PR #8, opened 2018-07-16
POC: Sketch of localfilesystem and commiting

martindurant:
@mariusvniekerk , suggesting here is where you might want to work something out. 
This contains minimal stuff for method aliasing (simplest possible), an actual example file-system for local work, and a hacky "transaction" that writes to temporary files, moving to final destination when the transaction is over.
As usual, everything I happened to code here is up for altering or criticism.

```
In [7]: with fs.transaction():
   ...:     with fs.open('./temp', 'wb') as f:
   ...:         f.write(b'hi')
   ...:     print(fs.ls('.'))
   ...: print(fs.ls('.'))
   ...:
   ...:
[]
['/Users/mdurant/temp/temp']
```

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -27,8 +27,8 @@ class AbstractFileSystem(object):
         A reasonable default should be provided if there are no arguments
         """
         self.autocommit = True
-        self.files = None
         self._intrans = False
+        self._transaction = Transaction(self)
         self._singleton[0] = self
         for new, old in aliases:
             setattr(self, new, getattr(self, old))
@@ -40,9 +40,18 @@ class AbstractFileSystem(object):
         If no instance has been created, then create one with defaults
         """
         if not cls._singleton[0]:
-            return AbstractFileSystem()
+            return cls()
         else:
             return cls._singleton[0]
+
+    @property
+    def transaction(self):
+        """A context within which files are committed together upon exit
+
+        Requires the file class to implement `.commit()` and `.discard()`
+        for the normal and exception cases.
+        """
+        return self._transaction
 
     def invalidate_cache(self, path=None):
         """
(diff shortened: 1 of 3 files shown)


martindurant:
(and a prod to make the aliases work with docstrings, please :) )

mariusvniekerk:
Yeah i think that would probably work as a minimal example. 

I think the interesting thing for distributed filesystems would be to consider making the actual transaction class itself a thing that can be moved around.  This is easily achievable by implementers by making fs.transaction() return an object with a __enter__ and __exit__.  
Not entirely sure if we want to be as strict by default?

fsspec/spec.py
```diff
@@ -220,12 +287,22 @@ def open(self, path, mode='rb', block_size=None, **kwargs):
         if 'b' not in mode:
             mode = mode.replace('t', '') + 'b'
             return io.TextIOWrapper(
-                self.open(self, path, mode, block_size, **kwargs))
+                self.open(path, mode, block_size, **kwargs))
+        else:
+            ac = kwargs.pop('autocommit', not self._intrans)
+            f = self._open(path, mode=mode, block_size=block_size,
+                           autocommit=ac, **kwargs)
+            if not ac:
```
martindurant: Should autocommit be necessarily True is `_intrans` isn't set, and should we avoid collecting up instances that might never be cleared?

mariusvniekerk: `autocommit` only applies to writing at `__exit__` so i think its a sane default in terms of user expectations

mariusvniekerk: Thats more explicit.  👍

martindurant:
OK, going to merge this now for further iteration. Of course, it does not yet explicitly deal with distributed transactions, since the file objects within a Transaction are not necessarily pickleable (but they should be, in this context) - however, you would be able to call `FileSystem.current().transaction.complete(True|False)` within any worker.

PR #16, opened 2018-10-09
Copy code from Dask

martindurant:


martindurant:
Note that this code eschews the compatibility cruft in Dask, and is therefore explicitly py3 only. It is to be considered as a replacement for code within Dask, only after it drops py2 support. Naturally, thorough testing and documentation must also be in place at that time.
 
Ref: https://github.com/dask/dask/issues/4047

PR #20, opened 2018-11-04
SFTP

martindurant:
New implementation for SSH-FTP via paramiko. Uses docker to test.


martindurant:
Write transactions could reasonably be accomplished by writing to /tmp/UUID and mv on commit?

PR #21, opened 2018-11-15
Ftp

martindurant:
Adding FTP FS

Also abstracted out buffered file reader/writer, since s3/gcs/adl essentially all do the same thing, and now FTP can use the same structure too.


martindurant:
Note that I am not adding SSL (FTPS?) until someone asks for it, since I don't want to create certs just for the purpose of testing.

PR #23, opened 2019-01-04
WIP: HDFS

martindurant:


martindurant:
cc @jcrist ; the webHDFS implementation is functional here for file listing, reading and writing (but no tests yet). Note the `data_proxy` kwarg, which makes this work with your test cluster by assuming the hostname mapping is known a priory - only to be used in testing, I think.

jcrist:
> Note the data_proxy kwarg, which makes this work with your test cluster by assuming the hostname mapping is known a priory - only to be used in testing, I think.

I'm not sure I follow - why is this necessary?

martindurant:
> why is this necessary

Reading and writing are done directly from the data node, but you get the appropriate address from the name node. It will refer to the data node by the address it knows internally, but to test from outside the cluster, we need to replace that with "localhost" for things to work.

jcrist:
Is this a problem that others would ever experience as well? Are there other cases where you'd want to alter the addresses (knox/other proxies? idk). Perhaps `rewrite_address` with a callable argument would be more flexible?

If you were to add `master.example.com`/`worker.example.com` to `etc/hosts` on your test machine that would also resolve things (not that that's a better solution).

martindurant:
@jcrist , the cluster is going into safe mode, any idea why? Is there any special config that makes it run OK within Travis?

jcrist:
How much data are you moving around? safemode can happen if you're storing too much data and the cluster is overloaded.

martindurant:
This is happening before any data transfer, unfortunately

PR #25, opened 2019-01-10
Fix setup.py and requirements.txt

gsakkis:
Fixes https://github.com/martindurant/filesystem_spec/issues/24


martindurant:
Thank you.

PR #27, opened 2019-01-11
Fixes: missing protocol attr, isfile, isdir 

gsakkis:
Two fixes:
- External filesystem classes may not have a `protocol` attribute (e.g. `S3FileSystem`)
- isfile and isdir should not raise `FileNotFoundError` is the path doesn't exist

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -59,7 +59,7 @@ def get_filesystem_class(protocol):
             raise RuntimeError(str(err))
         registry[protocol] = getattr(mod, name)
     cls = registry[protocol]
-    if cls.protocol == 'abstract' or cls.protocol is None:
+    if getattr(cls, 'protocol', None) in ('abstract', None):
         cls.protocol = protocol
 
     return cls
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -405,11 +405,17 @@ class AbstractFileSystem(object):
 
     def isdir(self, path):
         """Is this entry directory-like?"""
-        return self.info(path)['type'] == 'directory'
+        try:
+            return self.info(path)['type'] == 'directory'
+        except FileNotFoundError:
+            return False
 
     def isfile(self, path):
         """Is this entry file-like?"""
-        return self.info(path)['type'] == 'file'
+        try:
+            return self.info(path)['type'] == 'file'
+        except FileNotFoundError:
+            return False
 
     def cat(self, path):
         """ Get the content of a file """
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -184,6 +184,23 @@ def test_not_found():
             pass
 
 
+def test_isfile():
+    fs = LocalFileSystem()
+    with filetexts(files, mode='b'):
+        for f in files.keys():
+            assert fs.isfile(f)
+        assert not fs.isfile('not-a-file')
+
+
+def test_isdir():
+    fs = LocalFileSystem()
+    with filetexts(files, mode='b'):
+        for f in files.keys():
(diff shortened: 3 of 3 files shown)


fsspec/registry.py
```diff
@@ -59,7 +59,7 @@ def get_filesystem_class(protocol):
             raise RuntimeError(str(err))
         registry[protocol] = getattr(mod, name)
     cls = registry[protocol]
-    if cls.protocol == 'abstract' or cls.protocol is None:
+    if getattr(cls, 'protocol', None) in ('abstract', None):
```
martindurant: The class will always have a protocol attribute, so long as it derives from AbstractFileSystem (which has "abstract")

fsspec/spec.py
```diff
@@ -405,11 +405,17 @@ def size(self, path):
 
     def isdir(self, path):
         """Is this entry directory-like?"""
-        return self.info(path)['type'] == 'directory'
+        try:
```
martindurant: Quite right, this better reflects `os.path`.
Care to make a test of the two nonexistent cases, at least for the LocalFileSystem?

fsspec/registry.py
gsakkis: Well at least currently `S3FileSystem` is not derived from `AbstractFileSystem`, which is how I stumbled on this.

fsspec/spec.py
gsakkis: Will do

fsspec/registry.py
martindurant: https://github.com/dask/s3fs/pull/161

martindurant: (you should not use any of the code in fsspec with anything *not* derived or otherwise completely compatible with AbstractFileSystem)

gsakkis: As things stand, versions `0.1.2` and `0.1.3` are broken, at least for s3:
```
$ python -c "import fsspec; fsspec.filesystem('s3')"
...
AttributeError: type object 'S3FileSystem' has no attribute 'protocol'
```
The patch fixes it and is backwards compatible; even when dask/s3fs#161 lands it doesn't hurt to leave the fix in.

martindurant: The S3FileSystem that you get without the WIP PR, though, won't actually function as you expect given the fsspec spec. I guess your change can't hurt anything, it's true, but what are you wanting to do? I haven't checked, so I'm not sure that basic things like `open_files` would work with the pre-PR version.

martindurant: (but yes, you have persuaded me that this line can change as you suggest)

gsakkis: (btw I added `isfile` and `isdir` to `s3fs` in https://github.com/gsakkis/s3fs/commit/3b119ef041d655bbec15a5aca807e4163e12a9d5; I think that's needed even after https://github.com/dask/s3fs/pull/161 because the default `AbstractFileSystem` implementation cannot be used there)

martindurant: Why cannot it be used? 
You seem to have some code interacting with these implementations, is that public?

gsakkis: Because ` self.info(path)` raises an exception in S3 if path is a directory (and rightly so), so at least `isdir` needs to be implemented otherwise.

gsakkis: My code is a simple [Path](https://docs.python.org/3/library/pathlib.html#concrete-paths)-like implementation; it's not generic and mature enough to release it as a general library or a PR to fsspec (not yet at least) but I could post it as a gist if interested.

fsspec/spec.py
gsakkis: Done!

martindurant:
Thank you!

PR #30, opened 2019-02-19
fix kerberos auth

AlbertDeFusco:
I was able to get the following code to work from a Kerberized EMR cluster.

```python
from fsspec.implementations.webhdfs import WebHDFS

webhdfs = WebHDFS('ip-172-31-56-231.ec2.internal', kerberos=True)
files = webhdfs.ls('/user/instructor')
print(files)

data = webhdfs.cat('/user/instructor/autompg.csv')
print(data[:1024])
```

### DIFF
--- a/fsspec/implementations/webhdfs.py
+++ b/fsspec/implementations/webhdfs.py
@@ -85,12 +85,10 @@ class WebHDFS(AbstractFileSystem):
         self._connect()
 
     def _connect(self):
+        self.session = requests.Session()
         if self.kerb:
             from requests_kerberos import HTTPKerberosAuth
-            self.session = requests.Session(
-                auth=HTTPKerberosAuth(**self.kerb_kwargs))
-        else:
-            self.session = requests.Session()
+            self.session.auth = HTTPKerberosAuth(**self.kerb_kwargs)
 
     def _call(self, op, method='get', path=None, data=None,
               redirect=True, **kwargs):


fsspec/implementations/webhdfs.py
```diff
@@ -85,12 +85,10 @@ def __init__(self, host, port=50070, kerberos=False, token=None, user=None,
...
 
     def _connect(self):
+        self.session = requests.Session()
         if self.kerb:
             from requests_kerberos import HTTPKerberosAuth
-            self.session = requests.Session(
-                auth=HTTPKerberosAuth(**self.kerb_kwargs))
-        else:
-            self.session = requests.Session()
+            self.session.auth = HTTPKerberosAuth(**self.kerb_kwargs)
```
martindurant: Isn't this the wrong way around - the Kerberos auth now appears to be in the `kerb is False` path. Was there a specific error with the original code?

AlbertDeFusco: The original error was that `requests.Session()` does not accept `auth` as a keyword argument.

According to the change I'm adding the kerberos authentication when `self.kerb` is True.

martindurant: sorry, misread the diff colours :|

martindurant:
OK, this is good to see and nice that it's now tested with kerberos auth. 

Note that in general, fsspec is experimental, but pretty good for this kind of thing. It is a side-project of mine that I would like to see some effort put into to become the general file-systems-for-everything package it should be.

PR #34, opened 2019-02-28
Add experimental FUSE support

martindurant:
Note that FUSE always asks for paths starting with '/', so we strip this.
If you are mounting a real path, therefore, be sure to end it with '/' to
make the join work.

Use:

```python
import fsspec.fuse
fs = fsspec.filesystem('sftp', host='localhost')
fsspec.fuse.run(fs, '/Users/mdurant/temp/', '/Users/mdurant/mounting/')
```
(the destination must exist and be an empty directory)
Now the contents of the former directory are available in the latter location, for reading and writing.

This will not work for file implementations that do not support random access, and will have poor random-access performance for ones that only simulate random access with read-ahead (e.g., gcsfs, s3fs).

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -12,12 +12,14 @@ User Functions
    fsspec.filesystem
    fsspec.get_filesystem_class
    fsspec.get_mapper
+   fsspec.fuse.run
 
 .. autofunction:: fsspec.open_files
 .. autofunction:: fsspec.open
 .. autofunction:: fsspec.filesystem
 .. autofunction:: fsspec.get_filesystem_class
 .. autofunction:: fsspec.get_mapper
+.. autofunction:: fsspec.fuse.run
 
 Base Classes
 ------------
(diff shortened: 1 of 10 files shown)


martindurant:
(on OSX, the system automatically makes or reads files on any new device, so expect `.DS_Store` files to appear, and maybe not to be able to quit with ^C - you may need to call `umount` or `diskutil unmount` in another terminal)

martindurant:
Naturally, this requires fusepy, and will only work on posix.

pzwang:
For Windows, this seems to be one possible way to use FUSE on WSL:

> dokany currently provides a fuse API wrapper that uses the dokan kernel driver to produce a library that translates between the FUSE API and the Dokan API. This means that FUSE drivers built under cygwin and mingw are compatible.
> Now, this library is user space code that is linked at build time, but it kind of shows that a somewhat working 1:1 mapping between FUSE and Dokan is possible. Binary compatibility inside WSL would be quite cool, and handy -- especially for scripts relying on setting up sshfs mounts dynamically.

http://dokan-dev.github.io/

https://github.com/dokan-dev/dokany/tree/master/dokan_fuse

martindurant:
If someone on win would like to try, would be happy to see it

martindurant:
(will add documentation on this, and then merge - probably on the weekend)

PR #40, opened 2019-04-23
start to abstract out the read cache for files

martindurant:
allows for implementing mmap from https://github.com/dask/s3fs/pull/177


martindurant:
@blackmagic02881 , what do you think of this as a general solution?

martindurant:
(the tests don't actually use the mmap cache yet, might have made a mistake abstracting the bytes-based cache)

blackmagic02881:
looks good for me. sorry that could not reply promptly. deeply underwater recently in day job.

Issue #45, opened 2019-05-30
HTTPFileSystem hanging on open


rabernat:
I am trying to use fsspec.implementations.http.HTTPFileSystem to read a byte range from a remote url. Here's what I'm doing

```python
from fsspec.implementations.http import HTTPFileSystem
path = 'https://data.nas.nasa.gov/ecco/download_data.php?file=/eccodata/llc_4320/compressed/0000010368/Theta.0000010368.data.shrunk'
fs = HTTPFileSystem()
f_obj = fs.open(path)
# ... later would come the reading commands
```
This hangs forever. I have turned on urllib3 debugging to see what's happening. I get the following log output
```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (2): data.nas.nasa.gov:443
```
If I interrupt the running command, I get this stack trace
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/anaconda/envs/pangeo/lib/python3.6/site-packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, chunked, **httplib_request_kw)
    376             try:  # Python 2.7, use buffering of HTTP responses
--> 377                 httplib_response = conn.getresponse(buffering=True)
    378             except TypeError:  # Python 3

TypeError: getresponse() got an unexpected keyword argument 'buffering'

During handling of the above exception, another exception occurred:

KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-27-556e6da93e62> in <module>
----> 1 f_obj = fs.open(path)
      2 f_obj

~/anaconda/envs/pangeo/lib/python3.6/site-packages/fsspec/spec.py in open(self, path, mode, block_size, **kwargs)
    562             ac = kwargs.pop('autocommit', not self._intrans)
    563             f = self._open(path, mode=mode, block_size=block_size,
--> 564                            autocommit=ac, **kwargs)
    565             if not ac:
    566                 self.transaction.files.append(f)

~/anaconda/envs/pangeo/lib/python3.6/site-packages/fsspec/implementations/http.py in _open(self, url, mode, block_size, **kwargs)
    117         kw.pop('autocommit', None)
    118         if block_size:
--> 119             return HTTPFile(url, self.session, block_size, **kw)
    120         else:
    121             kw['stream'] = True

~/anaconda/envs/pangeo/lib/python3.6/site-packages/fsspec/implementations/http.py in __init__(self, url, session, block_size, **kwargs)
    165         try:
[...]

PR #46, opened 2019-05-30
Add size_policy to HTTPFile

martindurant:
Ref #45

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -116,7 +116,8 @@ class HTTPFileSystem(AbstractFileSystem):
         mode: string
             must be "rb"
         block_size: int or None
-            Bytes to download in one request; use instance value if None.
+            Bytes to download in one request; use instance value if None. If
+            zero, will return a streaming Requests file-like instance.
         kwargs: key-value
             Any other parameters, passed to requests calls
         """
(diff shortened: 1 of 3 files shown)


martindurant:
Thoughts on this, @rabernat ? There are quite a few things now that a user might need to know, which are pretty hard to explain, but flexibility that can be very useful. `size_policy` can be an arg to HTTPFileSystem (without the int option).

rabernat:
Have you tested this branch on my url?

```
https://data.nas.nasa.gov/ecco/download_data.php?file=/eccodata/llc_4320/compressed/0000010368/Theta.0000010368.data.shrunk
```

martindurant:
Yes, like
```
f_obj = fs.open(path, size_policy='get')
```
or give the exact file size, since you happen to know in this case.

rabernat:
Then it solves my problem. Thanks so much for this quick fix!

rabernat:
How would one hope to test for this?

Does the test suite actually spin up or mock an http server?

martindurant:
Yes, it spins up a python standard http file server, which has specific behaviours of its own. To get a full set of tests for all the different things servers might do... would take some pretty sophisticated mocking. Maybe could be done better with various known-behaviour docker images.

PR #57, opened 2019-06-27
Strip prefix when making a mapping

### LINKED ISSUE #56: Stripping of prefix for FSMap
`FSMap.__init__` should probably strip the protocol / prefix of the `root` so that `S3Map('s3://mybucket/file')` is treated the same as `S3Map('mybucket/file')`.

```python
In [15]: s3 = s3fs.S3FileSystem()

In [16]: map1 = s3fs.S3Map('s3://a/b', s3)

In [17]: map2 = s3fs.S3Map('a/b', s3)

In [18]: assert map1.root == map2.root
AssertionError                            Traceback (most recent call last)
<ipython-input-18-f02e96dd7165> in <module>
----> 1 assert map1.root == map2.root

AssertionError:
```

martindurant:
Fixes #56

### DIFF
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -33,7 +33,7 @@ class FSMap(MutableMapping):
 
     def __init__(self, root, fs, check=False, create=False):
         self.fs = fs
-        self.root = root.rstrip('/')  # we join on '/' in _key_to_str
+        self.root = fs._strip_protocol(root).rstrip('/')  # we join on '/' in _key_to_str
         if create:
             self.fs.mkdir(root)
         if check:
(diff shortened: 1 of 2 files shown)


martindurant:
Some things would work like that, but most operations on file-systems require complex infrastructure fixtures to work. The stuff in the base classes like the case here should not be overridden, so I don't know that there's anything to test. Having said that, it's good to have the s3fs mapping tests pass too as extra assurance (they do) and once I've thrown the switch to have s3fs and gcsfs depend on fsspec, I'll run the test suites of each in this repo too, alongside the tests that are only here.

PR #66, opened 2019-07-15
WIP Add experimental dask worker file system

martindurant:
(read-only files for now, although you can create directories, copy move
and delete stuff)


martindurant:
OK, merging this and iterating more following.

PR #73, opened 2019-07-23
When reading whole file, set size of HTTPFile

martindurant:
Fixes https://github.com/CityOfLosAngeles/intake-dcat/issues/14


ian-r-rose:
Thanks @martindurant!

PR #86, opened 2019-08-02
fix #85 local file get and put should support recursive=True

### LINKED ISSUE #85: put and get with file fs should support recursive=True
Currently just  implemented with `shutil.copyfile()`, i.e.

```
        """ Copy within two locations in the filesystem"""
        shutil.copyfile(path1, path2)

    get = copy
    put = copy
```

which only works for single files.

kalefranz:
fix #85


martindurant:
Thanks @kalefranz !

PR #93, opened 2019-08-05
Try more workarounds

martindurant:


fsspec/implementations/http.py
```diff
@@ -147,6 +147,17 @@ def size(self, url):
         """Size in bytes of the file at path"""
         return file_size(url, session=self.session, **self.kwargs)
 
+    def info(self, url, size_policy='head'):
+        try:
+            size = file_size(url, self.session, size_policy)
+            return {'name': url, 'size': size, 'type': 'file'}
+        except:
+            try:
+                self.session.get(url, stream=True)
```
jcrist: Any reason not to try `file_size(url, self.session, 'get')` if `head` fails, and then fallback to `None`? The `get` should also be a good backup check for file existence, with the benefit you may be able to get size from it?

I *think* the following may work:

```python
def info(self, url):
    for policy in ['head', 'get']:
        try:
            size = file_size(url, self.session, policy)
            break
        except ValueError as exc:
            size = None  # the ValueError from in `file_size`, would maybe be better just return `None` in `file_size`
            break
        except Exception:
            pass
    else:
        raise FileNotFoundError(url)
    return {"name": url, "size": size, "type": "file"}
```

martindurant: OK, I included it. Good to have extra eyes on this code...
The only potential problem with this is making extra requests for every file, but that should be better than exceptions that are hard to understand or remedy.

jcrist:
In dask we get the size from `fs.info`. What filesystems may not provide this? Is returning `None` here the best option (seems fine to me)?

jcrist:
Basically I'm wondering if "size may be none as reported from info" can be interpreted as part of the filesystem spec.

martindurant:
Right, this means that Dask can't chop the file into blocks, and will error if it tries. It may be good to catch that dask-side though

martindurant:
HTTP is the only FS which doesn't know size for sure right now. I suppose there could be others later.

jcrist:
> It may be good to catch that dask-side though

Agreed. Before I make this change, I want to ensure that "If the filesystem doesn't know the file size it returns None, otherwise it returns an integer" is part of the spec.

martindurant:
Here is the chance to define exactly that! :)

jcrist:
Perhaps add that to the `info` docstring in `spec.py`?

martindurant:
Yep, doing so now

martindurant:
(note that the hdfs cluster, used for webhdfs testing, has started to fail occasionally)

PR #94, opened 2019-08-06
fix glob regex escaping for some special characters

crepererum:


fsspec/implementations/tests/test_local.py
```diff
@@ -257,6 +258,21 @@ def test_abs_paths(tmpdir):
     # assert res == 'hi'
 
 
+@pytest.mark.parametrize('chars', ['+', '++', '(', ')', '|', '\\'])
+def test_glob_weird_characters(tmpdir, chars):
+    tmpdir = str(tmpdir)
+
+    subdir = os.path.join(tmpdir, 'test' + chars + 'x')
```
martindurant: I woudn't use `os.path` here, but `'/'` or `'\\'` explicitly

crepererum: why do you think so? We're testing the local FS here and the pathsep is one of the two options depending on the OS.

martindurant: but we only run CI on linux, so it would be good to test both explicitly. Internally, all paths which look like windows are converted to posix at the first opportunity, so it should work the same on both platforms.

martindurant:
Thanks, @crepererum

PR #100, opened 2019-08-08
Deal with symlinks

martindurant:
Should report the size of the target, since that will automatically
accesses upon open()

Fixes https://github.com/dask/dask/issues/5233

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -45,7 +45,7 @@ class LocalFileSystem(AbstractFileSystem):
         else:
             return paths
 
-    def glob(self, path):
+    def glob(self, path, **kargs):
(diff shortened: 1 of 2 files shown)


martindurant:
Note that I *want* `ls()` to report that a name is in fact a link, but reporting the size of the link itself is not useful.

martindurant:
@philippegr , this should solve your problem. However, I'll let this PR stay open a little while, because it isn't wholly clear to me that this is the right thing to do.

philippegr:
Thank you @martindurant, I am not too sure of the release process of fsspec, seems very frequent, so it should make it to the next release?

martindurant:
It certainly will, and it won't be long.

PR #101, opened 2019-08-08
Pickleable MMapCache & correctly handle empty file

### LINKED ISSUE #90: error when reading empty file using CachingFileSystem
example:
```
from fsspec.implementations.cached import CachingFileSystem
fs = CachingFileSystem(target_protocol='s3')
fs.open('s3://bucket/empty')

```
```
{'fn': '01c8b40c487e3c79a7549715f6fff04a15e6b0017bcb83cb98d9278b0b4ed4ec', 'blocks': set()}
  File "/fsspec_test.py", line 3, in <module>
    fs.open('s3://bucket/empty')
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/implementations/cached.py", line 157, in <lambda>
    self, *args, **kw
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/spec.py", line 669, in open
    autocommit=ac, **kwargs)
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/implementations/cached.py", line 157, in <lambda>
    self, *args, **kw
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/implementations/cached.py", line 132, in _open
    fn, blocks)
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/core.py", line 397, in __init__
    self.cache = self._makefile()
  File "/usr/local/miniconda3/envs//lib/python3.7/site-packages/fsspec/core.py", line 409, in _makefile
    fd.seek(self.size - 1)
OSError: [Errno 22] Invalid argument
```

mtrbean:
Alternative fix for @martindurant's PR #99 

Also fixes #90 for caching empty file

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -399,22 +399,24 @@ class MMapCache(BaseCache):
     def _makefile(self):
         import tempfile
         import mmap
-        from builtins import open
+
+        if self.size == 0:
+            return bytearray()
+
         # posix version
         if self.location is None or not os.path.exists(self.location):
             if self.location is None:
                 fd = tempfile.TemporaryFile()
+                self.blocks = set()
             else:
-                fd = open(self.location, 'wb+')
+                fd = io.open(self.location, 'wb+')
             fd.seek(self.size - 1)
             fd.write(b'1')
             fd.flush()
         else:
-            fd = open(self.location, 'rb+')
-        self._file = fd
+            fd = io.open(self.location, 'rb+')
 
-        f_no = fd.fileno()
-        return mmap.mmap(f_no, self.size)
+        return mmap.mmap(fd.fileno(), self.size)
 
     def _fetch(self, start, end):
         start_block = start // self.blocksize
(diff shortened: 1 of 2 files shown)


fsspec/core.py
```diff
@@ -432,6 +434,19 @@ def _fetch(self, start, end):
...
+        state = self.__dict__.copy()
+        # Remove the unpicklable entries.
+        del state['cache']
+        del state['_file']
+        return state
+
+    def __setstate__(self, state):
+        # Restore instance attributes
+        self.__dict__.update(state)
+        self.blocks = set()
```
martindurant: Why not serialise this too? Blocks can only be added, so it will never go stale. Else, the data on disc is not actually useful for the reconstituted instance. I wonder if there's a way to tell which parts of a file contain any data...

mtrbean: I actually just realized that intention is to persist the data on disk because `CachingFileSystem` does explicitly control the the underlying file. I think we should clear out blocks if the underlying file is a `TemporaryFile`, otherwise I agree with you that we can serialize `blocks` too.

martindurant: Probably we should be sure to update `self.location` to the created file, if it is temporary, and then we can safely reuse it and its blocks.

mtrbean: Since the temporary file could be gone when this unpickled, so I do prefer clearing out the cache if `location` was not explicitly supplied. What do you think?

martindurant: I would say, check for the file's existence and if it's there, use it. Maybe we should store the fact that the original call was for a temporary file, so if the file is gone, we know to make a new one rather than error.

martindurant:
(sorry, force pushes do not send notifications)

PR #103, opened 2019-08-11
Added adl protocol to registry.py, and added lowercasing to isdir() a…

hayesgb:
Wanted to get started with the Azure Datalake Gen1 implementation.  I've added adl to registry.py, with error instructions to install the adlfs library currently located in my repository.  I will create a pip installable over the next few days for that, and then begin adding tests.  The only changes implemented here are to the registry.py file, and lowercasing of the returns from the isfile and isdir method calls in core.py.  This is needed because the response of an info() call from Azure is all-caps, which returns an error with core.py as written.

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -30,6 +30,8 @@ known_implementations = {
                 'err': 'webHDFS access requires "requests" to be installed'},
     's3': {'class': 's3fs.S3FileSystem',
            'err': 'Install s3fs to access S3'},
+    'adl': {'class': 'adlfs.AzureDatalakeFileSystem',
+            'err': 'Install adlfs to access files stored on Azure Datalake Gen1 from https://github.com/hayesgb/adlfs'},
     'cached': {'class': 'fsspec.implementations.cached.CachingFileSystem'},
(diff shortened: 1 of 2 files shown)


martindurant:
Hi, what is the status of this, at the moment? With the lower() operation, dod you manage to get things working with adlfs? I know that I made comments on your fork, but it would be nice to come to both a long-term plan and a quicker solution that would enable dask and azure-datalake to work together.

hayesgb:
By adding the .lower() change, and adding adl to the registry, the adlfs (not dask-adlfs) is working with Gen1 Datalake on Azure.  I’m currently working on adding a second class to the adlfs package that will interface with Gen2.  Once I have those working, then I’ll start adding tests and improving usability.

martindurant:
Would it be possible to move the lower() call into the external code rather than having it here in fsspec itself?

hayesgb:
I can probably just reimplement the method calls in adlfs.  However, will fsspec pick up the package if it’s not listed in fsspec’s registry.py file?

martindurant:
> will fsspec pick up the package if it’s not listed in fsspec’s registry.py

No, this is exactly the right thing to do. If you are still experimental/testing, you can edit `fsspec.registry` directly at runtime.

PR #105, opened 2019-08-13
Remove fspath from AbstractFile since broke some pyarrow

martindurant:
Fixes https://github.com/intake/filesystem_spec/issues/104


martindurant:
Do you need a swift release here?

TomAugspurger:
Nope. I don't think there was a release of fsspec with fspath, right?

martindurant:
Indeed, you are right again!

PR #106, opened 2019-08-15
Have glob return matching dirs too

martindurant:
Fixes https://github.com/dask/s3fs/issues/219


martindurant:
Requesting comments here

PR #107, opened 2019-08-16
Attribute 'info' is now retrieved from pyarrow

AlexisMignon:
The attribute 'info' was added to the list of attributes to be retrieved from the underlying HadoopFileSystem object in PyArrowHDFS.


AlexisMignon:
This PR is related to the following issue in dask: https://github.com/dask/dask/issues/5285

PR #109, opened 2019-08-19
Strip protocol in paths returned by 'ls'

AlexisMignon:
The previous changes had as an effect to by pass the 'ls' method which is used in the base class AbstractFileSystem.

There seem to be a difference between what is returned by ls in the test docker in dask and what is returned using an actual HDFS server.

In my case, when doing:

```python
hdfs=pyarrow.hdfs.connect(host="myserver", port=xxxx)
hdfs.ls("path/to/file", detail=True)
```
... the protocol information is prepended to the path while it is not the case in the test docker.

This explains while nothing was detected. My corrections consists in calling `_strip_protocol` on the output as expected by the `info` method on the `AbstractFileSystem`.

this PR is related to the dask issue: https://github.com/dask/dask/issues/5285
and reverts and corrects the previous modification in the PR: https://github.com/intake/filesystem_spec/pull/107

### DIFF
--- a/fsspec/implementations/hdfs.py
+++ b/fsspec/implementations/hdfs.py
@@ -67,6 +67,9 @@ class PyArrowHDFS(AbstractFileSystem):
         if detail:
             for p in out:
                 p['type'] = p['kind']
+                p["name"] = self._strip_protocol(p["name"])
+        else:
+            out = [self._strip_protocol(p) for p in out]
         return out
 
     @staticmethod
@@ -107,7 +110,6 @@ class PyArrowHDFS(AbstractFileSystem):
             'get_space_used', 'df', 'chmod', 'chown', 'disk_usage',
             'download', 'upload', '_get_kwargs_from_urls',
             'read_parquet', 'rm', 'stat', 'upload',
-            'info',
         ]:
             return getattr(pahdfs, item)
         else:


TomAugspurger:
> and reverts and correct modifictation in the PR: #107

Should that be the **in**correct modification?

AlexisMignon:
I mean it reverts and corrects (i.e. fixes) the modification in PR...

TomAugspurger:
Thanks for the clarification. Based on [the docs](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.ls) this seems to be the correct change:

> full path to the entry (without protocol)

Can we add a test for this change, to ensure that we don't regress on this in the future?

martindurant:
Thank you for finding this. 
Testing with the Dask image shows that there is nothing special about the image: if you use the default server, you get paths without the protocol, and if you specify the server, you get "hdfs://...".

This must always have been the case, I'm not quite following why the move to fsspec exposed it. Nevertheless, this should go in, but I agree with @TomAugspurger that a test would be good. That test would have to be in Dask, though, which is the only place that runs the HDFS implementation right now. Since we run hadoop-test-cluster anyway for the webhdfs test, it would be reasonable to reuse that infrastructure here - @jcrist , would that be hard?

jcrist:
`libhdfs` (the library underneath `pyarrow`'s wrapper) has differences between versions/builds that sometimes result in prefixes in certain calls, and sometimes doesn't. This has to do with the version of hadoop/libhdfs, and not the version of pyarrow. This has bit us in other libraries (e.g. https://github.com/jcrist/hdfscm/issues/2). This may be why there's a difference between test environments.

> Since we run hadoop-test-cluster anyway for the webhdfs test, it would be reasonable to reuse that infrastructure here - @jcrist , would that be hard?

You'd have to run the tests inside the docker container, since you'd need access to libhdfs, and port access to the hdfs namenode. We do this for other libraries using `hadoop-test-cluster`, no reason it wouldn't work here.

PR #116, opened 2019-08-27
.walk() should respect "detail=True"

JohnEmhoff:
This PR addresses two issues: passing `detail` as a kwarg to
`.walk()` fails because it internally calls `.ls()`
with both `detail=True` and whatever kwargs were passed
originally, causing `detail` to be specified twice.

The second is that `.walk()` does not respect `detail=True`
and always returns bare file names. My fix here is to just
shove the dict from `.ls()` into the return list unmodified --
this feels slightly incongruous with the non-detail version
which strips prefixes from paths, but this feels like the
right thing to do?

Anywhoo, this is a bit messy so I'm happy to clean it up
if you have any suggestions. It could also use some tests:
is there any particular way you'd like it tested? A test in
`test_local`?

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -252,7 +252,7 @@ class AbstractFileSystem(up):
         """Remove a directory, if empty"""
         pass  # not necessary to implement, may not have directories
 
-    def ls(self, path, **kwargs):
+    def ls(self, path, detail=False, **kwargs):
         """List objects at path.
 
         This should include subdirectories and files at that location. The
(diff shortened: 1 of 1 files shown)


martindurant:
As I said elsewhere, I believe it may make sense to add this to `find()`, but not for `walk()`.

martindurant:
OK, ok - let's have this. Can you please resolve the conflict (sorry) ?

fsspec/spec.py
```diff
@@ -331,24 +331,40 @@ def walk(self, path, maxdepth=None, **kwargs):
...
         for info in listing:
             # each info name must be at least [path]/part , but here
             # we check also for names like [path]/part/
             name = info['name'].rstrip('/')
+            if detail:
+                to_append = info
+            else:
+                to_append = name.rsplit("/", 1)[-1]
+
             if info['type'] == 'directory' and name != path:
```
dhirschfeld: Is this a bug? At least in my info dict a directory is denoted with `type='dir'`:

image

martindurant: Should definitely be "directory" - is "dir" coming right from ftplib? Then the code in `ls` should rewrite it to "directory".

dhirschfeld: > *is "dir" coming right from ftplib?*

Yep - I haven't changed anything in the output. Maybe another quirk of my FTP server. I'll look to fix that up in `ls`

PR #120, opened 2019-08-28
Validate minversion for external packages.

TomAugspurger:
Closes #120


martindurant:
Are you still working here? Looks good and simple enough to me.

TomAugspurger:
I think this should be good.

PR #121, opened 2019-08-29
BUG: Fixed file fetching issue

TomAugspurger:
Closes https://github.com/dask/s3fs/issues/225

Which apparently was broken by 4749ab9c5f786a6ce9cdc3a098c12f205e8cf207.

Still working on a test.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -498,7 +498,7 @@ class BytesCache(BaseCache):
                 self.cache = self.fetcher(start, bend)
                 self.start = start
             else:
-                new = self.fetcher(self.end, bend)
+                new = self.fetcher(self.end, max(end, bend))
                 self.cache = self.cache + new
 
         self.end = self.start + len(self.cache)


TomAugspurger:
Note: https://github.com/dask/s3fs/pull/229 fixes the issuer for s3fs. I *think* there's still a bug in the bytes caching (which this is hopefully fixing) but I may be wrong.

TomAugspurger:
Note: https://github.com/dask/s3fs/pull/229 fixes the issuer for s3fs. I *think* there's still a bug in the bytes caching (which this is hopefully fixing) but I may be wrong. Will wait for @martindurant to chime in.

fsspec/core.py
```diff
@@ -498,7 +498,7 @@ def _fetch(self, start, end):
                 self.cache = self.fetcher(start, bend)
                 self.start = start
             else:
-                new = self.fetcher(self.end, bend)
+                new = self.fetcher(self.end, max(end, bend))
```
martindurant: Not sure how this helps, `bend` is either equal to `end`, equal to the file size, or always larger than `end`

PR #128, opened 2019-09-05
Add simpler read-ahead cache

martindurant:
Needs tests, but can confirm that the s3fs suite is happy with it, except for the one test which specifically depends on BytesCache behaviour.


martindurant:
Defer whether to make ReadAheadCache the default to the future, would like people to try this first.

Issue #129, opened 2019-09-09
`infer_compression` does not support zstd/lz4 compression.


asford:
The `fsspec.core.open` interface provides a `compression="infer"` option, which infers a file's compression via file extension. I would expect that this interface support all compression types, including the built-in lz4 and zstd compression tools, but it does not currently recognize the `lz4` and `zstd` compression types. Ideally, any dynamically registered compression method should support a set of file extensions for compatibility with `infer_compression`.

The current implementation is a *bit* tricky to fix, as a simple check against `fsspec.compression.compr` introduces an import cycle between `utils` and `compression`. As `utils.infer_compression` is used by dask, we can assume that this function must remain in it's current namespace.

asford:
Fixed in #130

PR #130, opened 2019-09-09
Add `infer_compression` support for zstd/lz4 compression.

asford:
Proof of concept fix for #129. Not suitable in current from, as "infer_compression" is moved to break the `utils`/`compression` import cycle.

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -35,19 +35,22 @@ def register_compression(name, callback, extensions, force=False):
         ValueError: If name or extensions already registered, and not force.
 
     """
-    if name in compr and not force:
-        raise ValueError("Duplicate compression registration: %s" % name)
-
-    compr[name] = callback
-
     if isinstance(extensions, str):
         extensions = [extensions]
+
+    # Validate registration
+    if name in compr and not force:
+        raise ValueError("Duplicate compression registration: %s" % name)
 
     for ext in extensions:
         if ext in fsspec.utils.compressions and not force:
             raise ValueError(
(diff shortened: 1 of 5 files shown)


martindurant:
What is WIP about this, do you intend to change things?

asford:
I realized, post implementation, that `infer_compression` is part of the fsspec interface exposed to dask. I'm going to rework the implementation to avoid a breaking change to the interface.

This could either (a) use an import cycle between utils and compression to create a shim for the older interface or (b) leave the file-extension-to-compression-type mapping in the utils layer and add a utility function in compression to register both the compression method and it's intended file extensions. Currently leaning toward (b), but do you have a preference?

martindurant:
I am not sure, happy to see your preferred solution.

asford:
Opted for B, adding a registration function for compression methods with explicit names and filesystem extensions.

This is now ready for review.

martindurant:
Thank you, this is going in.
I have never come across a .lz4 or .zst file, which is why I never knew to include those :)

asford:
Great, thanks.

We're currently using lz4 to handle compression of big-ish (25-50gb) S3 data partitions with text data accessed via dask and msgpack encoded data accessed by fsspec-backed pytorch data loaders. This has been a *great* solution solution so far, we're often seeing faster processing of lz4 compressed data than uncompressed data read from S3.

martindurant:
Raises the question of whether things like msgpack should be included as "encoding" methods; right now, text is the only encoding, which is pretty typical for files.

Issue #135, opened 2019-09-13
issue with ftp traversal, KeyError "name"


hardingnj:
I think this is a bug (version `0.4.4`), caused by some entries (`cdir` and `pdir`) on FTP not having the `name` attribute. Very happy to submit a PR if so, should be a straightforward fix.

```
import fsspec
host = fsspec.filesystem('ftp', host='ngs.sanger.ac.uk') 
host.ls(".")
```
gives:
```
[{'modify': '20180615104058',
  'perm': 'fle',
  'type': 'cdir',
  'unique': 'FC02U80',
  'unix.group': '0',
  'unix.mode': '0755',
  'unix.owner': '0'},
 {'modify': '20180615104058',
  'perm': 'fle',
  'type': 'pdir',
  'unique': 'FC02U80',
  'unix.group': '0',
  'unix.mode': '0755',
  'unix.owner': '0'},
 {'modify': '20190513162006',
  'perm': 'fle',
  'type': 'dir',
  'unique': 'FC02U83',
  'unix.group': '0',
  'unix.mode': '0755',
  'unix.owner': '0',
  'name': '/./production',
  'size': 0},
 {'modify': '20180615104110',
  'perm': 'fle',
  'type': 'dir',
  'unique': 'FC02U80000080',
  'unix.group': '0',
  'unix.mode': '0755',
  'unix.owner': '0',
  'name': '/./scratch',
  'size': 0},
 {'modify': '20180119093730',
  'perm': 'adfr',
  'size': 1099,
  'type': 'file',
  'unique': 'FC02U3937',
  'unix.group': '0',
  'unix.mode': '0644',
  'unix.owner': '0',
  'name': '/./README'}]
```

but
```
host.open("README")
```
gives:
```
...

~/git/vector-ops/binder/deps/conda/envs/vector-ops-c668a73/lib/python3.6/site-packages/fsspec/implementations/ftp.py in <listcomp>(.0)
    108         path = self._strip_protocol(path)
    109         files = self.ls(self._parent(path), True)
--> 110         return [f for f in files if f['name'] == path][0]
    111 
    112     def _open(self, path, mode='rb', block_size=None, autocommit=True,

KeyError: 'name'
```
I guess because when traversing `pdir` or `cdir`, it fails to find the attribute. Perhaps because line  `109` is not functioning as intended in allowing dirs as well?

martindurant:
Converting cdir, pdir -> dir fixes this? I must admit I don't know what those other labels mean.

[...]

PR #136, opened 2019-09-13
Add check for type == file before checking name == path

hardingnj:
- sometimes if entry does not have name eg "." or ".." causes KeyError.

- all entries of type file have a name

re #135

### DIFF
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -79,12 +79,10 @@ class FTPFileSystem(AbstractFileSystem):
         out = []
         if path not in self.dircache:
             try:
-                out = list(self.ftp.mlsd(path))
+                out = [(fn, details) for (fn, details) in self.ftp.mlsd(path) if fn not in ['.' '..']]
                 for fn, details in out:
                     if path == '/':
                         path = ''  # just for forming the names, below
-                    if fn in ['.', '..']:
-                        continue
                     details['name'] = '/'.join([path, fn.lstrip('/')])
                     if details['type'] == 'file':
                         details['size'] = int(details['size'])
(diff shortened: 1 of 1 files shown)


fsspec/implementations/ftp.py
```diff
@@ -106,7 +105,7 @@ def ls(self, path, detail=True):
     def info(self, path, **kwargs):
         # implement with direct method
         path = self._strip_protocol(path)
-        files = self.ls(self._parent(path), True)
+        files = [x for x in self.ls(self._parent(path), True) if x['type'] == 'file']
```
martindurant: So, I believe you should be able to get an `info()` of a directory (should return a dict where `type=="dir"`). What happens if instead of this, in L86 we exclude those pdir and cdir entries only?

hardingnj: Yes- that's better I think. I can't think of any case where cdir/pdir would be useful.

fsspec/implementations/ftp.py
```diff
@@ -79,12 +79,10 @@ def ls(self, path, detail=True):
         out = []
         if path not in self.dircache:
             try:
-                out = list(self.ftp.mlsd(path))
+                out = [(fn, details) for (fn, details) in self.ftp.mlsd(path) if fn not in ['.' '..']]
```
martindurant: The string here is "...", not what you intended; but I think the functionality is exactly the same as before, no? What about the "pdir", "cdir" that you initially wanted to guard against?

martindurant:
I took the liberty of pushing what I think you meant - I hope that's OK with you.

fsspec/implementations/ftp.py
hardingnj: Yes- sorry!

PR #139, opened 2019-09-16
Fix XZ compression name

jrbourbeau:
It looks like #130 changed the LZMA algorithm compression name from "xz" to "lzma" which is causing backward compatibility issues. For example, using `fsspec==0.4.4`:

```python
In [1]: import fsspec

In [2]: fsspec.__version__
Out[2]: '0.4.4'

In [3]: from fsspec.core import get_compression

In [4]: get_compression('some_file.xz', 'xz')
Out[4]: 'xz'
```

while with the new 0.4.5 `fsspec` release, we have:

```python
In [1]: import fsspec

In [2]: fsspec.__version__
Out[2]: '0.4.5'

In [3]: from fsspec.core import get_compression

In [4]: get_compression('some_file.xz', 'xz')
ValueError                                Traceback (most recent call last)
<ipython-input-4-31b6a6cda193> in <module>
----> 1 get_compression('some_file.xz', 'xz')

~/miniconda/envs/fsspec-test/lib/python3.7/site-packages/fsspec/core.py in get_compression(urlpath, compression)
    206         compression = infer_compression(urlpath)
    207     if compression is not None and compression not in compr:
--> 208         raise ValueError("Compression type %s not supported" % compression)
    209     return compression
    210

ValueError: Compression type xz not supported
```

This PR reverts the name change back to "xz". 

Came across this on a Dask CI build failure (ref https://travis-ci.org/dask/dask/jobs/585748693#L1057-L1088)

### DIFF
--- a/fsspec/tests/test_compression.py
+++ b/fsspec/tests/test_compression.py
@@ -11,6 +11,7 @@ def test_infer_custom_compression():
     """Inferred compression gets values from fsspec.compression.compr."""
     assert infer_compression("fn.zip") == "zip"
     assert infer_compression("fn.gz") == "gzip"
+    assert infer_compression("fn.xz") == "xz"
     assert infer_compression("fn.unknown") is None
     assert infer_compression("fn.test_custom") is None
     assert infer_compression("fn.tst") is None
--- a/fsspec/tests/test_compression.py
+++ b/fsspec/tests/test_compression.py
@@ -11,7 +11,6 @@ def test_infer_custom_compression():
     """Inferred compression gets values from fsspec.compression.compr."""
     assert infer_compression("fn.zip") == "zip"
     assert infer_compression("fn.gz") == "gzip"
-    assert infer_compression("fn.xz") == "xz"
     assert infer_compression("fn.unknown") is None
     assert infer_compression("fn.test_custom") is None
     assert infer_compression("fn.tst") is None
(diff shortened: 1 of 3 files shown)


jrbourbeau:
Note I also removed the `try`/`except` around the `lzma` import as `lzma` was added to the standard library in Python 3.3.

TomAugspurger:
> Note I also removed the try/except around the lzma import as lzma was
added to the standard library in Python 3.3.

Just a heads up: there's a decent number of people out there with broken
Python installations that lack the lzma module. Pandas
chose to revert a change doing an unconditional `import lzma` because we
got so many reports about it.

jrbourbeau:
Thanks for the heads up @TomAugspurger. I reverted the `import` changes (making `lzma` a conditional import again)

fsspec/tests/test_compression.py
```diff
@@ -48,6 +48,11 @@ def test_infer_custom_compression():
         del compressions["tst"]
 
 
+def test_lzma_compression_name():
+    pytest.importorskip("lzma")
+    assert infer_compression("fn.xz") == "xz"
```
martindurant: OK, I suppose I was confused. The file extension is still xz, and the compression should be called xz, even though the module/codec is lzma? Should we perhaps include "lzma" in the list of compressions also?

jrbourbeau: > The file extension is still xz, and the compression should be called xz, even though the module/codec is lzma?

That's my understanding based on the first paragraph of https://docs.python.org/3/library/lzma.html (although there's also a legacy `.lzma` extension which is also supported : / ). Naming the compressions "xz" is primarily to maintain compatibility with previous behavior. Adding an "lzma" compression seems reasonable too.

martindurant: Alright so should add it

martindurant:
Thanks for the fix.

PR #141, opened 2019-09-17
Make threadlocal data, especially for HTTP sessions

martindurant:


martindurant:
@mrocklin , how does this look? I am trialing it for HTTP, where we have *not* had any reported problems, but the same model would apply to gcsfs and s3fs.

martindurant:
ping @mariusvniekerk , @mrocklin or @TomAugspurger : would love feedback here before applying this to s3fs

TomAugspurger:
One question on the pickle stuff: it looks like your SerialisableLocal is to enable pickling `AbstractFileSystem`? What's the advantage to having the thread local on the instance rather than a (thread-local) global at the module level?

martindurant:
Different instances will, in general, have separate (configured) sessions *and* one per thread. To do it in a global, we'd need to have a lookup for which session belongs to which instance (but instances are hashable).

fsspec/implementations/tests/test_http.py
```diff
@@ -128,3 +127,23 @@ def test_random_access(server, headers):
...
+    out = []
+
+    def target():
+        out.append(h.session)
+
+    t = threading.Thread(target=target)
+    t.start()
+    t.join()
+
+    assert out[0] != id(h.session)
```
mrocklin: It looks like you're calling `id` on one but not the other here.

martindurant:
The session object is specifically not being serialised (it cannot, it holds a threadpool and sockets). The thing that's serialisable is the thread-local holder, which is always empty for pickling. threading.local() is not serialisable.

mrocklin:
Why are trying to serialize the thread-local holder?  Can we just avoid trying to serialize it in the first place, and so avoid making the serializeable-thread-local thing?

martindurant:
It's either make this little thing serialisable, or to have similar logic in the main file system class - that might be fragile, since derived classes might want to override it.

mrocklin:
Maybe the logic could be somewhat general.  If you're not doing custom serialization in any of the subclasses then something like the following might work:

```python
def __getstate__(self):
    d = self.__dict__.copy()
    del d['session']
    return d
```

martindurant:
> If you're not doing custom serialization in any of the subclasses

but we do need to allow this, many of them need to set up implementation-specific connections, some of which may need to be thread-local. How can "session" be thread-local if it is an attribute, except to have some instance-specific look-up?

martindurant:
Anyway, seems like single-threaded workers have problems with s3fs too, so putting this on hold.

PR #143, opened 2019-09-19
Remove universal wheel

TomAugspurger:
Py3 only packages aren't universal.

I already pushed the git tag. I'll manually override this (or do a 0.5.1 if needed).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[bdist_wheel]
-universal: 1
-
 [metadata]
 long_description: file: README.rst
 


TomAugspurger:
Hmm doesn't seem like there's an option for overriding universal. So I'll push 0.5.1 (and probably 0.5.0 as an sdist-only release).

martindurant:
I don't think we need wheels at all...

TomAugspurger:
Wheels are generally preferred to source distributions, even for pure-python packages.

TomAugspurger:
The most prominent reason is probably not having to execute `setup.py` when `pip install`ing a package.

PR #149, opened 2019-09-25
Fix bottleneck in AbstractFileSystem.find

jobh:
By changing the list to a set, my test case (writing a zarr array
through s3fs) got 50x higher throughput and went from CPU to IO bound.

This fix is minimal with no semantic changes. It is a question whether the `if name in out` on line 387 and `if path not in out` on line 389 should be dropped, but line 387 in particular is a change in behaviour since it's not necessarily `name` that is added to `out`.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -379,17 +379,17 @@ class AbstractFileSystem(up):
         kwargs are passed to ``ls``.
         """
         # TODO: allow equivalent of -name parameter
-        out = []
+        out = set()
         for path, dirs, files in self.walk(path, maxdepth, **kwargs):
             if withdirs:
                 files += dirs
             for name in files:
                 if name and name not in out:
-                    out.append("/".join([path.rstrip("/"), name]) if path else name)
+                    out.add("/".join([path.rstrip("/"), name]) if path else name)
         if self.isfile(path) and path not in out:
             # walk works on directories, but find should also return [path]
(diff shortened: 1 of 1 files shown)


martindurant:
This is definitely a positive change, thank you. Can you please check if it makes any difference to change the order of checks on L389?

jobh:
I'm just a tourist in this code, and my use case / test case probably doesn't trigger the test on line 389 at all. Do you mean difference in performance?

If I was more confident with the code, I would assume that `path` cannot be walked if it is a file, and that `name` cannot be empty, and reorder it to
```
        out = set()
        if self.isfile(path):
            # walk works on directories, but find should also return [path]
            # when path happens to be a file
            out.add(path)
        else:
            for path, dirs, files in self.walk(path, maxdepth, **kwargs):
                if withdirs:
                    files += dirs
                for name in files:
                    out.add("/".join([path.rstrip("/"), name]) if path else name)
        return sorted(out)
```
and at that point it wouldn't matter if `out` was a list or a set. But then I'd worry that I didn't think of some weird corner case.

martindurant:
Unfortunately, a path might sometimes be both a file and a directory :)
I like your change and will merge it, and may get around to seeing whether my suggestion makes any measurable difference.
Please feel free to be more than a tourist, if it interests you!

jobh:
I was afraid of something like that :)

Thanks!

PR #152, opened 2019-10-10
Add github FS implementation

martindurant:
(no tests, because this relies on online service)

(the github API may enforce rate limits, making this FS only usable in light traffic situations)


martindurant:
@jsignell , do you find this interesting/useful?

martindurant:
Putting this in because it's explicitly marked as experimental, and isn't yet referenced as a known FS or in the docs. I still think it can be useful, though.

PR #153, opened 2019-10-11
Remove [] from around open()...split('\n')

sscherfke:
`split()` already returns a list so you don't need to wrap it in another list.


martindurant:
I wonder why this wasn't a problem so far?

sscherfke:
I think pip is okay with that but I have a script that converts pypi packages to conda packages and that failed with this package (and previously with other packages, as well ;-)).

PR #158, opened 2019-10-17
Skip fuse tests correctly.

### LINKED ISSUE #151: Fuse is no longer optional in tests
The import of `fsspec.fuse` and the `importorskip` are in reverse, so it fails to import `fuse` before it can realize the file should be skipped. This worked correctly in 0.5.1.

QuLogic:
Fixes #151.


martindurant:
Thanks for this.

PR #162, opened 2019-10-21
FTPFileSystem Fall back to using `dir` when `mlsd`extension is not avail on server

acrosby:
Not sure if it necessary to try to match the format of the extra **details** from `mlsd` command, so I haven't done that here (unix users/groups, permissions, unique, etc.). `dir` output is dependent on the server platform, but I have assumed Linux `ls -l` format to be the most common and it solves the issues I ran into. 

Also, I had to tweak the `_parent` call in `info` to strip a double forward slash (`//`) that seemed to crop up.  Tests seem to continue to pass, but now the following things all seem to work.

```python
host = fsspec.filesystem('ftp', host='ftpprd.ncep.noaa.gov')  
host.ls("/")   

host = fsspec.filesystem('ftp', host='ngs.sanger.ac.uk')  
host.get("scratch/README", "test.readme.txt")

fs = fsspec.filesystem('ftp', host="ftpprd.ncep.noaa.gov")
with fs.open("ftp://ftpprd.ncep.noaa.gov/pub/data/nccf/com/gfs/prod/gfs.20191020/18/gfs.t18z.aircft.tm00.bufr_d.nr", "rb") as f:
    with open("temp", "wb") as nc:
        nc.write(f.read())
```

### DIFF
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -124,6 +124,7 @@ class FTPFileSystem(AbstractFileSystem):
             out = [f for f in files if f["name"] == path][0]
         except IndexError:
             raise FileNotFoundError(path)
+        return out
 
     def _open(self, path, mode="rb", block_size=None, autocommit=True, **kwargs):
         path = self._strip_protocol(path)
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -81,11 +81,15 @@ class FTPFileSystem(AbstractFileSystem):
         out = []
         if path not in self.dircache:
             try:
-                out = [
-                    (fn, details)
-                    for (fn, details) in self.ftp.mlsd(path)
-                    if fn not in [".", ".."] and details["type"] not in ["pdir", "cdir"]
-                ]
+                try:
+                    out = [
+                        (fn, details)
+                        for (fn, details) in self.ftp.mlsd(path)
+                        if fn not in [".", ".."]
+                        and details["type"] not in ["pdir", "cdir"]
+                    ]
+                except Error:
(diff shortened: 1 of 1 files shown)


martindurant:
Sounds good! Are you still working here?

> `host.get("scratch/README"`

I believe the paths should always start with "/" (i.e., it might be considered a path relative to some remote home directory otherwise), but I suppose it is added automatically?

acrosby:
@martindurant Yeah the behavior is a little weird, I think `FTPFileSystem` is sort of operating like you are always in the root-most directory of the ftp, so yeah in this case `scratch/README` is relative to the root (which is is). `/scratch/README`,  `./scratch/README`, and `/./scratch/README` all seem to work as well still. 

I'm not sure why `info`/`_parent` calls were adding `//` to the paths, but it was breaking the `ls` call, even for `mlsd` compliant servers so `get` and `open` were also failing.

Unless you want me to make some more changes, it could merged in if you are ready.

acrosby:
@martindurant actually, I think I'm going to add in a few more tweaks while I am at it

martindurant:
ok, thank you - let me know.

acrosby:
@martindurant Seems to be working reliably now if you are ready to merge it in, seems like CI failed from a miniconda.sh build.  Thanks!

NB: I made a possibly opinionated decision by setting the now-configurable timeout to 30s by default. The the timeout is required (and should be relatively short) in the case of avoiding hanging on `voidresp` for ftp servers that wait to send 2XX code after a binary transfer until connection close.

martindurant:
That will do nicely, thank you.
I expect we may in the future have to revisit the text mangling for servers which have different format output.

PR #163, opened 2019-10-23
Handle content-length of 0

TomAugspurger:
cc @martindurant.

This test is failing right now. Some layer is truncating the response when the `Content-Length` is 0, so the `assert f.read() == data` is failing. `f.read()` is returning `b''`. Any suggestions on an alternative test?

It is working for the original issue though

```python
In [1]: import dask.dataframe as dd
df
In [2]: df2 = dd.read_csv('https://fred.stlouisfed.org/graph/fredgraph.csv?id=SP500')

In [3]: df2.compute()
Out[3]:
            DATE    SP500
0     2009-10-23  1079.60
1     2009-10-26  1066.95
2     2009-10-27  1063.41
3     2009-10-28  1042.63
4     2009-10-29  1066.11
...          ...      ...
2603  2019-10-16  2989.69
2604  2019-10-17  2997.95
2605  2019-10-18   2986.2
2606  2019-10-21  3006.72
2607  2019-10-22  2995.99

[2608 rows x 2 columns]
```

Closes https://github.com/dask/dask/issues/5517

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -180,7 +180,7 @@ class HTTPFileSystem(AbstractFileSystem):
             # get failed, so conclude URL does not exist
             if size is False:
                 raise FileNotFoundError(url)
-        return {"name": url, "size": size, "type": "file"}
+        return {"name": url, "size": size or None, "type": "file"}
 
 
 class HTTPFile(AbstractBufferedFile):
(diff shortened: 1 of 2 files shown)


martindurant:
I am puzzled that the pandas load works, but the simple read() doesn't. Would be good to figure out why the two take different paths.

martindurant:
So this fixes the original case, in which HEAD was giving content length 0, but GET was giving right length. If the content length were 0 in the GET too, I think requests itself doesn't fetch the data.

Issue #164, opened 2019-10-29
infer_storage_options fails with on file paths with two leading slashes


samkaufman:
Calling `infer_storage_options("//tmp/f”)` returns:

```python
{'protocol': 'file', 'path': '/f', 'host': 'tmp’}
```

but I expected it to return:

```python
{'protocol': 'file', 'path': '//tmp/f’}
```

martindurant:
Works correctly on master

PR #165, opened 2019-10-29
Test leading/non-leading POSIX path slashes

samkaufman:
This adds a test which fails when the following is called:

```python
infer_storage_options("//mnt/datasets/test.csv”)
```

It returns:

```python
{'protocol': 'file', 'path’: '/datasets/test.csv', 'host': ‘mnt’}
```

This corresponds to issue #164.


samkaufman:
@martindurant Happy to. I’m not sure when I’ll have a second, so I wanted to get this in the tracker in case anybody else runs into it, but feel free to assign to me.

PR #168, opened 2019-10-31
Fixes #166

### LINKED ISSUE #166: AbstractFileSystem walk method calls ls with 3 parameters, but only 2 expected
Although `ls` is expected to be overwritten by child classes, it's still worth noting that the `walk` method passes 3 arguments (2 positional and `**kwargs`) 

https://github.com/intake/filesystem_spec/blob/d774f0ef92b5f7bd28ec0b1b4e4fdb62ee8bea60/fsspec/spec.py#L344

While `ls` only accepts 1 position and `**kwargs`
https://github.com/intake/filesystem_spec/blob/d774f0ef92b5f7bd28ec0b1b4e4fdb62ee8bea60/fsspec/spec.py#L264

This causes `TypeError` to be raised instead of the expected `NotImplementedError`. Looking at the docstring for `ls`, it looks like walk should be called with `detail=True` or `ls` changed to have "detail" explicitly declared in the function signature (or both).

jonathan-hourany:
Adds key word `detail` to `ls` signature and to `self.ls` call in `walk`

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -261,7 +261,7 @@ class AbstractFileSystem(up):
         """Remove a directory, if empty"""
         pass  # not necessary to implement, may not have directories
 
-    def ls(self, path, **kwargs):
+    def ls(self, path, detail=True, **kwargs):
         """List objects at path.
 
         This should include subdirectories and files at that location. The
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks!

PR #170, opened 2019-11-01
Fixes #169

### LINKED ISSUE #169: `glob` does not support range
Although stated in the docstring that ranges (e.g. `s3://test/date=2019-10-0[1-5]`) are supported it turns out they are not. Using the `s3fs.S3FileSystem` library, which relies on this implementation of `glob`, returns an empty list when supplying a path with a range in it.

This is because the opening brace (i.e. `[`) is not detected in glob's if-check
https://github.com/intake/filesystem_spec/blob/56fd518316a5ec05357d6e62c9a41a1b981802ee/fsspec/spec.py#L445
By adding `"[" not in path` or using the `glob.has_magic` function in the if-check this bug is solved

jonathan-hourany:
Uses builtin `glob.has_magic` to detect when special characters are used in the string
Separates if-checks for `indstar` and `indques` so that `path.find` is called twice instead of four times

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -436,13 +436,21 @@ class AbstractFileSystem(up):
         kwargs are passed to ``ls``.
         """
         import re
+        from glob import has_magic
 
         ends = path.endswith("/")
         path = self._strip_protocol(path)
-        indstar = path.find("*") if path.find("*") >= 0 else len(path)
-        indques = path.find("?") if path.find("?") >= 0 else len(path)
+        indstar = path.find("*")
+        indques = path.find("?")
+
+        if indstar < 0:
+            indstar = len(path)
+        if indques < 0:
+            indques = len(path)
+
         ind = min(indstar, indques)
-        if "*" not in path and "?" not in path:
+
+        if not has_magic(path):
             root = path
             depth = 1
             if ends:
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -440,15 +440,11 @@ class AbstractFileSystem(up):
 
         ends = path.endswith("/")
         path = self._strip_protocol(path)
-        indstar = path.find("*")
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -436,13 +436,21 @@ def glob(self, path, **kwargs):
...
+        from glob import has_magic
 
         ends = path.endswith("/")
         path = self._strip_protocol(path)
-        indstar = path.find("*") if path.find("*") >= 0 else len(path)
-        indques = path.find("?") if path.find("?") >= 0 else len(path)
+        indstar = path.find("*")
+        indques = path.find("?")
+
+        if indstar < 0:
```
martindurant: I unwinding these necessary? Seems the change in the `if` below is the correction you were after.

jonathan-hourany: I unwound them though because in the master branch if `if path.find("*") >= 0` returned `True` then `path.find("*")` would be called a second time. This is an extremely small optimization though, It's not necessary, and if the previous style is prefered I'm happy to revert

jonathan-hourany:
Actually, before merging I'll see if I can write test cases for this function. Should we close this PR for now or keep it up? I'll be out for a few days so I might be a little while in getting to it.

martindurant:
Happy to wait

jonathan-hourany:
@martindurant I got some tests written up! Please don't hesitate to let me know if you'd like anything changed about them or the rest of the code.

Also, I reverted the if-check behavior on the magic character indices. I figured optimizations or code clean up probably warrant their own separate PR. Let me know if that works for you :slightly_smiling_face:

martindurant:
Looks totally decent to me, thank you - I'll merge this now.
On cleaning up, well it's even more complicated I suppose, because of course those characters might always be escaped...

PR #177, opened 2019-11-07
REF: Alternative docstring handling

TomAugspurger:
Dynamically updating the class docstring while creating an
instance seemed a bit strange.

I'd prefer to provide a decorator to do this once at class definition
time. Downstream projects will need to opt into this by adding the decorator to their class.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -5,9 +5,10 @@ import posixpath
 import re
 import tempfile
 from fsspec import AbstractFileSystem
-from fsspec.utils import stringify_path
+from fsspec.utils import stringify_path, inherit_docstrings
 
 
+@inherit_docstrings
 class LocalFileSystem(AbstractFileSystem):
     """Interface to files on local storage
 
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -5,10 +5,9 @@ import posixpath
 import re
 import tempfile
 from fsspec import AbstractFileSystem
-from fsspec.utils import stringify_path, inherit_docstrings
+from fsspec.utils import stringify_path
 
 
-@inherit_docstrings
 class LocalFileSystem(AbstractFileSystem):
     """Interface to files on local storage
 
@@ -78,25 +77,21 @@ class LocalFileSystem(AbstractFileSystem):
         return result
 
     def copy(self, path1, path2, **kwargs):
-        """ Copy within two locations in the filesystem"""
         shutil.copyfile(path1, path2)
 
     def get(self, path1, path2, **kwargs):
-        """ Copy file to local """
         if kwargs.get("recursive"):
             return super(LocalFileSystem, self).get(path1, path2, **kwargs)
         else:
             return self.copy(path1, path2, **kwargs)
 
     def put(self, path1, path2, **kwargs):
-        """ Upload file from local """
         if kwargs.get("recursive"):
             return super(LocalFileSystem, self).put(path1, path2, **kwargs)
         else:
             return self.copy(path1, path2, **kwargs)
 
     def mv(self, path1, path2, **kwargs):
-        """ Move file from one location to another """
         os.rename(path1, path2)
 
     def rm(self, path, recursive=False, maxdepth=None):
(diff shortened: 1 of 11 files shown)


martindurant:
OK, fair enough, but then this should be applied to s3fs, gcsfs and adlfs/abfs

TomAugspurger:
Right. They'll need to add the decorator.

martindurant:
Also the gdrive implementation

fsspec/utils.py
```diff
@@ -302,3 +302,32 @@ def stringify_path(filepath):
...
+            continue
+
+        m = getattr(cls, method)
+        n = getattr(AbstractFileSystem, method, None).__doc__
+        if not callable(m) or not n or n in (m.__doc__ or ""):
+            # ignore if a) not a method, b) no superclass doc
+            # c) already includes docstring
+            continue
+        if m.__doc__ and hasattr(getattr(AbstractFileSystem, method), "__doc__"):
+            template = """\
```
jrbourbeau: Just a heads up, I tried this out with `s3fs` and am getting a `numpydoc` error from having multiple `Parameter` sections

<details>
<summary>Traceback:</summary>

```terminal
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v2.2.1
making output directory... done
WARNING: html_static_path entry '_static' does not exist
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 4 source files that are out of date
updating environment: [new config] 4 added, 0 changed, 0 removed
reading sources... [ 25%] api
Exception occurred:
  File "/Users/jbourbeau/miniconda/envs/s3fs-dev/lib/python3.7/site-packages/numpydoc/docscrape.py", line 422, in _error_location
    raise ValueError(msg)
ValueError: The section Parameters appears twice in the docstring of <function S3FileSystem.ls at 0x10a7e6378> in /Users/jbourbeau/github/dask/s3fs/s3fs/core.py.
The full traceback has been saved in /var/folders/m3/6sdjxnm164q0bf5pw5dj24t80000gp/T/sphinx-err-n0jao0x7.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

</details>

What if we don't include the upstream docstring if it's defined for the subclass?

TomAugspurger: Thanks for trying that out. Older versions of numpydoc may be OK with it (that's what we do on dask/dask)

FWIW, that's my preference. I'd rather downstream packages explicitly reference the upstream docs. Something like

```python
def copy(...):
    """
    My docstring.

    See Also
    ---------
    fsspec.AbstractFileSystem.copy
```

will produce a link to the fsspec docs in the HTML version, as long as intersphinx config is setup.

jrbourbeau: I think in dask/dask if the subclass defines it's own `Parameters` section, then the corresponding upstream section has it's title modified to be `Extra Parameters`. 

Your linking to `fsspec` docs proposal seems like a good way to go

TomAugspurger: Does OK by you @martindurant? If so, I'll close this PR and make PRs in downstream libraries to update their docs.

martindurant: Yes, that's OK. We can even reuse the function from dask.

TomAugspurger: Sorry, may not have been clear. My preference (and I think @jrbourbeau's) is to no implicitly inherit the parent docstrings. Rather, the children should either

1. Link to the parent docstrings when they overwite
2. Do the magic to combine docstrings if they so desire.

TomAugspurger:
Another thing we noticed: These "Upstream ..." docs weren't actually appearing in the sphinx docs, since the docstring was mutated until an instance was actually created (and sphinx wouldn't create one).

So I'll repurpose this PR to just remove the doc magic.

fsspec/implementations/local.py
```diff
@@ -77,25 +77,21 @@ def info(self, path, **kwargs):
         return result
 
     def copy(self, path1, path2, **kwargs):
-        """ Copy within two locations in the filesystem"""
```
TomAugspurger: These were identical to the parent class's. Sphinx will pick them up automatically.

martindurant:
+1 from me

PR #181, opened 2019-11-08
REF: Refactor AbstractFileSysem init

TomAugspurger:
This PR changes how init works for the base AbstractFileSystem. The
primary goal is to simplify things, which will hopefully result in more
predictable behavior downstream.

The primary most immediate change that the subclass's `__int__` is run *only* when
the instance is newly created. The `__init__` should be skipped when
loading a cached instance. This is accomplished with a metaclass.

There are two secondary changes here, which I can likely split out if desired

1. **Garbage collection of file systems**. On master, circular references in `_cache` and elsewhere mean that `del fs` won't actually garbage collect the object. It's still alive in the cache. Which means that newly created instances will be loaded from the cache. This may be a sensible behavior, I'm not sure. To support it, the only necessary change is to make `_cache` a dict rather than a `WeakValueDictionary`.
2. **Method aliases**. On master, we dynamically add these alias methods inside `__init__`. It'd be better to do this when the class is defined. I've done this in a backwards-compatible way, though subclasses will need to update to avoid a warning.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -77,25 +77,21 @@ class LocalFileSystem(AbstractFileSystem):
         return result
 
     def copy(self, path1, path2, **kwargs):
-        """ Copy within two locations in the filesystem"""
         shutil.copyfile(path1, path2)
 
     def get(self, path1, path2, **kwargs):
-        """ Copy file to local """
         if kwargs.get("recursive"):
             return super(LocalFileSystem, self).get(path1, path2, **kwargs)
         else:
             return self.copy(path1, path2, **kwargs)
 
     def put(self, path1, path2, **kwargs):
-        """ Upload file from local """
         if kwargs.get("recursive"):
             return super(LocalFileSystem, self).put(path1, path2, **kwargs)
         else:
             return self.copy(path1, path2, **kwargs)
 
     def mv(self, path1, path2, **kwargs):
-        """ Move file from one location to another """
         os.rename(path1, path2)
 
     def rm(self, path, recursive=False, maxdepth=None):
@@ -111,7 +107,6 @@ class LocalFileSystem(AbstractFileSystem):
         return LocalFileOpener(path, mode, fs=self, **kwargs)
 
     def touch(self, path, **kwargs):
-        """ Create empty file, or update timestamp """
         path = self._strip_protocol(path)
         if self.exists(path):
             os.utime(path, None)
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -4,11 +4,11 @@ import shutil
 import posixpath
 import re
 import tempfile
-from fsspec import AbstractFileSystem, AliasMixin
+from fsspec import AbstractFileSystem
 from fsspec.utils import stringify_path
 
 
-class LocalFileSystem(AbstractFileSystem, AliasMixin):
+class LocalFileSystem(AbstractFileSystem):
     """Interface to files on local storage
 
     Parameters
@@ -20,7 +20,6 @@ class LocalFileSystem(AbstractFileSystem, AliasMixin):
     """
 
     root_marker = "/"
-    _add_aliases = True
 
     def __init__(self, auto_mkdir=True, **kwargs):
         super().__init__(**kwargs)
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -4,11 +4,11 @@ import shutil
 import posixpath
 import re
 import tempfile
-from fsspec import AbstractFileSystem
+from fsspec import AbstractFileSystem, AliasMixin
 from fsspec.utils import stringify_path
 
 
-class LocalFileSystem(AbstractFileSystem):
+class LocalFileSystem(AbstractFileSystem, AliasMixin):
     """Interface to files on local storage
 
     Parameters
@@ -20,6 +20,7 @@ class LocalFileSystem(AbstractFileSystem):
     """
 
     root_marker = "/"
+    _add_aliases = True
 
     def __init__(self, auto_mkdir=True, **kwargs):
         super().__init__(**kwargs)
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -56,7 +56,11 @@ class SFTPFileSystem(AbstractFileSystem):
     def mkdir(self, path, mode=511):
         self.ftp.mkdir(path, mode)
 
-    def makedirs(self, path, mode=511):
+    def makedirs(self, path, exist_ok=False, mode=511):
+        if exist_ok:
+            raise ValueError(
+                "SFTPFileSystem.makedirs does not support the 'exist_ok' option."
+            )
         parts = path.split("/")
         path = ""
         for part in parts:
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -57,12 +57,12 @@ class SFTPFileSystem(AbstractFileSystem):
         self.ftp.mkdir(path, mode)
 
     def makedirs(self, path, exist_ok=False, mode=511):
-        if exist_ok:
-            raise ValueError(
-                "SFTPFileSystem.makedirs does not support the 'exist_ok' option."
-            )
+        if self.exists(path) and not exist_ok:
+            raise FileExistsError("File exists: {}".format(path))
+
         parts = path.split("/")
         path = ""
+
         for part in parts:
             path += "/" + part
             if not self.exists(path):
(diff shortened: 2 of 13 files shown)


TomAugspurger:
cc @martindurant. The primary design question I have is whether instances should be outlive the usual garbage collection rules. Either is easily doable. I don't have a good feeling about what is best here.

cc @jrbourbeau if you have thoughts. For context, on master

```python
fs = FileSystem()
del fs
```

will not actually garbage collect the instance because the `_cache` holds a reference to it. If we later come along and do

```python
fs2 = FileSystem()
```

`fs2` will actually be the same object as the original `fs`, and can reuse anything it cached.

TomAugspurger:
With these changes, there's a single failure in s3fs

```python
    def test_change_defaults_only_subsequent():
        """Test for Issue #135

        Ensure that changing the default block size doesn't affect existing file
        systems that were created using that default. It should only affect file
        systems created after the change.
        """
        try:
            S3FileSystem.cachable = False  # don't reuse instances with same pars

            fs_default = S3FileSystem()
            assert fs_default.default_block_size == 5 * (1024 ** 2)

            fs_overridden = S3FileSystem(default_block_size=64 * (1024 ** 2))
            assert fs_overridden.default_block_size == 64 * (1024 ** 2)

            # Suppose I want all subsequent file systems to have a block size of 1 GiB
            # instead of 5 MiB:
            S3FileSystem.default_block_size = 1024 ** 3

            fs_big = S3FileSystem()
>           assert fs_big.default_block_size == 1024 ** 3
E           assert 5242880 == (1024 ** 3)
E            +  where 5242880 = <s3fs.core.S3FileSystem object at 0x11f749a90>.default_block_size

```

It needs to be updated to include

```diff
@@ -134,6 +142,7 @@ class S3FileSystem(AbstractFileSystem):
     read_timeout = 15
     default_block_size = 5 * 2**20
     protocol = 's3'
+    _extra_tokenize_attributes = ('root_marker', 'connect_timeout', 'read_timeout', 'default_block_size', )

     def __init__(self, anon=False, key=None, secret=None, token=None,
                  use_ssl=True, client_kwargs=None, requester_pays=False,
```

else modifications to class attributes interact poorly with the cache. I think that's acceptable.

TomAugspurger:
In the interest of moving things along, I've reverted the changes to how these are garbage collected. I've made `_cache` a dict again, so that it keeps instances alive until the cache is cleared.

So now the only real change is to how often `__init__` is called. Now it'll be called exactly once, when the instance is first created.

fsspec/implementations/local.py
```diff
@@ -20,6 +20,7 @@ class LocalFileSystem(AbstractFileSystem):
     """
 
     root_marker = "/"
+    _add_aliases = True
```
martindurant: Could just be the default on AbstractFileSystem, not repeated for every derivative? 
Could simply remove and add aliases regardless?

fsspec/spec.py
```diff
@@ -8,75 +10,176 @@
...
+
+    In the interim, using an implicitly inherited alias will emit a warning. To silence the warning,
+    inherit from AliasMixin and set `_add_aliases=True` on the filesystem class.
+    """
+
+    _add_aliases = False
+    _alias_message = "Using implicitly added alias '{}'. Inherit from AliasMixin to continue using this method."
+
+    @_warn_implicit_alias
+    def makedir(self, path, create_parents=True, **kwargs):
```
martindurant: Might we not list the explicit signatures, since then we have to change them in multiple places in the future? 
Also the repetitive pattern might be explicit, but it is verbose and perhaps error-prone.

fsspec/spec.py
```diff
@@ -8,75 +10,176 @@
...
+        self._cache = {}
+
+    def __call__(self, *args, **kwargs):
+        cls = type(self)
+        extra_tokens = tuple(
+            getattr(self, attr, None) for attr in self._extra_tokenize_attributes
+        )
+        token = tokenize(cls, *args, *extra_tokens, **kwargs)
+        if cls.cacheable and token in self._cache:
+            return self._cache[token]
```
martindurant: Is it a single global cache for all fsspec instances?

fsspec/implementations/local.py
TomAugspurger: No. This is the only way to know that a subclass as *opted in* to having these aliases. Otherwise, we wouldn't be able to know if we need to warn or not.

fsspec/spec.py
TomAugspurger: This preserves the function signature. If you want to use you certainly could just have `*args, **kwargs`, but then you would need more magic to rewrite the signature.

fsspec/spec.py
TomAugspurger: No. The caching is per subclass type.

fsspec/implementations/local.py
martindurant: I am thinking we might just making mandatory and skip opting out or in, and no need for warnings. That would be equivalent to simply including them in spec

```
    cp = copy
```
etc

TomAugspurger: Fine by me. Do you recall the motivation for not doing that in the first place?

martindurant: Not as such. It is possible (should be checked) that the overrides in gcsfs and s3fs do not agree on which one of a pair of possibilities they implement.
Actually, thinking: if `cp = copy` in the superclass and the subclass defines its `copy`, does it get the subclass `cp`?

martindurant: No, it does not :|

TomAugspurger: I don't follow. Are you talking about master? Things seem OK on this branch.

```
In [2]: import fsspec

In [3]: class MyFS(fsspec.AbstractFileSystem):
   ...:     def copy(self, *args, **kwargs):
   ...:         print("my copy!")
   ...:         return super().copy(*args, **kwargs)
   ...:

In [4]: fs = MyFS()

In [6]: fs.cp('a', 'b')
my copy!
```

martindurant: I'm talking about the putative really simple way to do it, which it turns out doesn't work. Your version works correctly, so I'm happy to go with it

TomAugspurger:
@martindurant this is passing. It should be backwards compatible, aside from `add_aliases=False` no longer preventing those methods from being added. You get them no matter what.

s3fs and gcsfs tests pass for me locally.

TomAugspurger:
I don't think that `__init_subclass__` is an option. That's executed when a subclass is *defined*, rather than created.

```python
In [1]: class Base:
   ...:     def __init_subclass__(cls, *args, **kwargs):
   ...:         print("init subclass")
   ...:         super().__init_subclass__(*args, **kwargs)
   ...:

In [2]: class Bar(Base):
   ...:     def __init__(self, *args, **kwargs):
   ...:         print("init!")
   ...:         super().__init__(*args, **kwargs)
   ...:
init subclass
```

We need control over what happens with `fs = MyFS()`, which IIUC is only possible with metaclasses.

jrbourbeau:
Thanks for looking into that @TomAugspurger. +1 for the current changes then

fsspec/implementations/sftp.py
```diff
@@ -56,7 +56,11 @@ def _get_kwargs_from_urls(urlpath):
     def mkdir(self, path, mode=511):
         self.ftp.mkdir(path, mode)
 
-    def makedirs(self, path, mode=511):
+    def makedirs(self, path, exist_ok=False, mode=511):
+        if exist_ok:
+            raise ValueError(
+                "SFTPFileSystem.makedirs does not support the 'exist_ok' option."
```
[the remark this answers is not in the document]
TomAugspurger: Ah... I see. We unconditionally *do* `exist_ok=True`.

I'll see how hard it is to implement this properly

TomAugspurger:
The pickling tests may fail here.

TomAugspurger:
Should be good now.

Did you want to take another look @martindurant?

martindurant:
I had just the couple of comments above, nothing new. Otherwise happy.

TomAugspurger:
Fixing the downstream s3fs test failure in https://github.com/dask/s3fs/pull/261  (some extra state between tests now? Not sure.)

I needed to re-record the gcsfs tests locally to get them to pass. Presumably we're making different HTTP calls now, though it's not clear why (most likely different caching?). So I think we'll merge 

* https://github.com/dask/s3fs/pull/261
* this
* re-record gcfs tests

Then I'll kick of a build on master and everything should be passing again.

TomAugspurger:
OK s3fs passed on a rebuild after https://github.com/dask/s3fs/pull/261.

Merging this and re-recording gcsfs.

TomAugspurger:
Hmm having some trouble with gcs while replaying the tests (recording is fine, just testing isn't).

I'll fight with it a bit longer, and revert this if needed.

TomAugspurger:
Rerecorded in https://github.com/dask/gcsfs/pull/198.

Running against our CI in https://travis-ci.org/intake/filesystem_spec/jobs/610566689

PR #188, opened 2019-11-12
API: Kwargs handling for Caches

TomAugspurger:
This removes `**kwargs` from the signature of all the byte Cache
classes. The were not being used, and accepting kwargs can lead to
subtle errors.

It also changes how kwargs are forwarded from the opened File to
the cache. We use a dedicated `storage_options` dict, which is
more flexible than popping off specific keywords from `**kwargs`.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -441,7 +441,7 @@ class BaseCache(object):
         How big this file is
     """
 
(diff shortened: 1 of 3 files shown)


TomAugspurger:
(CI all passed. Not sure why GitHub isn't reporting that yet).

PR #191, opened 2019-11-12
Add BlockCache

TomAugspurger:
An alternative caching implementation that's intended as a replacement for BytesCache. The idea is to only ever make requests of size `blocksize`. At a high-level, the design is

* Cache data in a `Dict[int, bytes]`, a mapping from block number to the bytes for that chunk
* When serving a `fetch_range` request,
  * determine all the blocks touched by start, end, and in between
  * Make actual get requests for uncached blocks
  * Return the requested bytes by slicing the blocks and joining

---

I was kicking around this idea a couple months ago when we were hitting bugs in BytesCache, which has some surprisingly complicated logic.

This implementation tries to get similar characteristics (cache reads as in-memory bytestrings).

I think the main thing to decide is whether we should have some kind of `maxsize` parameter to control how large this cache will get. Right now, we'll end up caching the full contents of the file if a byte has been requested from every chunk. We could cap that somehow, and evict blocks based on some strategy (least recently used, used the least number of times, etc.)

cc @martindurant if you have a chance to look, though not a priority. I need to do some performance testing / profiling.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -578,17 +578,27 @@ class BlockCache(BaseCache):
     """
     Cache holding memory as a set of blocks.
 
-    Blocks are stored as a Dict[int, bytes]. Requests are only
-    ever made for `blocksize`.
+    Requests are only ever made `blocksize` at a time, and are
+    stored in an LRU cache. The least recently accessed block is
+    discarded when more than `maxblocks` are stored.
+
+    Parameters
+    ----------
+    blocksize : int
+        The number of bytes to store in each block.
+        Requests are only ever made for `blocksize`, so this
+        should balance the overhead of making a request against
+        the granularity of the blocks.
+    fetcher : Callable
+    size : int
+        The total size of the file being cached.
+    maxblocks : int
+        The maximum number of blocks to cache for. The maximum memory
+        use for this cache is then ``blocksize * maxblocks``.
     """
 
     def __init__(self, blocksize, fetcher, size, maxblocks=32):
         super().__init__(blocksize, fetcher, size)
-        # TODO: Consider an LRU-cache rather than a dict.
-        # This would avoid ever having the full file in memory if that
-        # becomes a problem.
-        self._count = 0
-        self._counts = {}
         self.nblocks = math.ceil(size / blocksize)
         self.maxblocks = maxblocks
         self._fetch_block_cached = functools.lru_cache(maxblocks)(self._fetch_block)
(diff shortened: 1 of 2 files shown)


martindurant:
Definitely I think this should be done as a LRU cache rather than storing everything; otherwise a good idea.
One possible improvement: when you are fetching more than one contiguous block, would be good to combine the calls, because often it's the total number of fetch calls which is the limiting factor. This is also true for for the CachingFileSystem (which has protocol "blockcache", which could be confusing).

I previously commented that ReadaheadCache might be the best default, what do you think?

TomAugspurger:
> I previously commented that ReadaheadCache might be the best default, what do you think?

Just glanced through it. It also seems like a reasonable alternative.

I suppose the main difference is BlockCache would be better for read patterns where you go back to an earlier block?

```python
._fetch(10, 20)
._fetch(20, 30)
._fetch(10, 15)  # in cache
```

I think third `._fetch` would have been a fresh request for ReadAheadCache, and not loaded from cache? So if we expect mostly-sequential reads, then ReadAheadCache is probably the way to go as it's simpler than this BlockCache.

martindurant:
Correct: block cache would be good when you repeatedly access some parts of a file in a random-access pattern; readahead cache for when you are reading mostly serially through the file.

For your example, the third fetch might also hit the cache for the right block-size. For a size of 5, say, you would indeed need to read from remote again.

TomAugspurger:
Refactored to use an LRU cache for getting / setting blocks. Things are more complicated now, and I don't see an easy way to do this LRU caching with the optimization that multi-block fetches be done in a single call.

---

So our choice is between

1. Use ReadAheadCache:
  * pros: simpler, better for large requests
  * cons: worse for repeated reads, when we're jumping around the file "a bit"
2. Use BlockCache:
  * pros: better for repeated reads with jumps
  * cons: probably more complicated, worse for multi-block requests. Choosing the right `maxblocks` isn't straightforward.

It's hard for me to know what is better in the abstract :/

martindurant:
Right, so of course it works out workload-dependent, and most users will not expect to have to read the docstrings to work out what's the best for them. I am marginally for read-ahead, because it's easy to understand. The current BytesCache is somewhere between these two, but too complex.

TomAugspurger:
See https://github.com/intake/filesystem_spec/pull/193 for changing the default to readahead. I think that's the right choice for now.

BlockCache may be useful, but isn't a priority right now.

martindurant:
So long as it makes it into the docs, so we can point people to it.

TomAugspurger:
OK, going to merge this in so that we can experiment easily on different workloads. I have a followup PR coming with docs and a reorg.

PR #193, opened 2019-11-12
API: Change the default caching to readahead

TomAugspurger:
xref https://github.com/intake/filesystem_spec/pull/191.

The BytesCache cache is complex and has been the source of bugs in the past. This switches over to the simpler readahead cache by default.

This cache is worse for workloads where you're likely to re-read an earlier part of the file, as those contents would have been dropped from the cache.

The previous behavior can be restored with `cache_type='bytes'` when creating a File (a config to switch this would be nice).

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -872,7 +872,7 @@ class AbstractBufferedFile(io.IOBase):
         mode="rb",
         block_size="default",
         autocommit=True,
-        cache_type="bytes",
+        cache_type="readahead",
         cache_options=None,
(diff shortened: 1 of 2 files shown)


martindurant:
"bytes" is also a nicely obscure name, so may want to think about changing it in the future (but not now when the default is changing anyway)

TomAugspurger:
It is obscure :)

Only failure is gcsfs, which will need to be re-recored. I've verified that I can rerecord locally, so I'll make a gcsfs PR following this.

PR #195, opened 2019-11-12
Reorganize caches

TomAugspurger:
Moves caches to a new module, `caching.py`, as `core.py` was getting a tad long.

We import the caches in `core.py` to stay backwards compat, but document the caches at their new location.

I'll point out a real change inline.


fsspec/spec.py
```diff
@@ -661,11 +661,27 @@ def _parent(cls, path):
...
             return cls.root_marker
 
-    def _open(self, path, mode="rb", block_size=None, autocommit=True, **kwargs):
+    def _open(
+        self,
+        path,
+        mode="rb",
+        block_size=None,
+        autocommit=True,
+        cache_options=None,
```
TomAugspurger: Passing through `cache_options` is a real change. I may have introduced a bug when I did my `cache_options` PR, though I'm not sure.

martindurant: This is not actually a change to any implementation, though, since they all override `_open`. Not all use `AbstractBufferedFile`, though, so the cache may not be relevant. I think it's fine to include here, and certainly in the docs of what to implement for new implementations.

PR #202, opened 2019-11-14
Add `cache_options` to AbstractFileSystem._open subclasses

TomAugspurger:
Confirmed that the dask HDFS tests are passing with this.

See https://github.com/intake/filesystem_spec/issues/201 for more work to be done along this line.

Closes https://github.com/intake/filesystem_spec/issues/198

cc @jrbourbeau

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -200,7 +200,14 @@ class CachingFileSystem(AbstractFileSystem):
         if not path.startswith(self.protocol):
             path = self.protocol + "://" + path
         if mode != "rb":
-            return self.fs._open(path, mode=mode, **kwargs)
+            return self.fs._open(
+                path,
+                mode=mode,
+                block_size=block_size,
+                autocommit=autocommit,
+                cache_options=cache_options,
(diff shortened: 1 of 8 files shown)


martindurant:
Can you also please check what happens for file types *not* based on AbstractBufferedFile? I can see that `SFTPFileSystem._open()` discards unused kwargs, for instance, so that's OK.

TomAugspurger:
I think HDFSFile is another example.

Is there a reason these *don't* return subclasess of AbstractBufferedFile? I think having `AbstractFileSystem.open` return a common type for all subclasess would be helpful.

martindurant:
You could, if you were prepared to write wrappers to pass on calls to the original file object. The implementation of [local files](https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/local.py#L159) does this partially and might be a decent guide. There are corner cases though: http returns a stream-only (no seek) requests file-like object in the case of block_size=0 - that would need to be wrapped too.

martindurant:
Short story: I agree it might be nice, but involves some work.

PR #203, opened 2019-11-14
Rename argument in _Cached

TomAugspurger:
Typically metaclasses use `cls` as the first argument.


martindurant:
Would a pickle test with multiple instances have picked this up?

TomAugspurger:
Added a test that was failing on master.

Issue #200, opened 2019-11-14
access to dropbox?


fkloosterman:
I was wondering if there is already a filesystem implementation that provides access to dropbox (I suppose using the dropbox API)? If not, what alternatives for dropbox access with fsspec exist?

martindurant:
There is no such implementation.
However, the implementations in this repo, as well as the new https://github.com/intake/gdrivefs (in development) should show that they are not hard to make, so long as the remote API is reasonably flexible. The [python library](https://dropbox-sdk-python.readthedocs.io/en/stable/) does look OK at a glance.

MarineChap:
Dear, 
I am working with @fkloosterman. We developed our (partial) implementation for dropbox. For the moment, I cloned the filesystem_spec and manually add in the "registry.py" my own implementation. It is working well but there is another way to add an implementation ? I would rather prefer to use your package instead of a local modified package. 
(I would also be happy to share with you the implementation if you are interested but it is really a partial implementation. We only want to use it for upload file in dropbox).
Thanks

martindurant:
Please post a PR to here, with a change to fsspec.registry.known_implementations specifying the import/class of the implementation, a protocol (I suggest "dropbox") and a meaningful message upon import failure, telling people how to install your package.
Happy to see it included!

martindurant:
ref #206

MarineChap:
Hello! Nice, my first open-source PR!! ;) Forgive me I am a newbie in open-source. As it is a short implementation, I was not thinking to create my own package for this. It is not possible to add it in the PR to be part of the available fsspec implementation ? Like this, eventually someone can add the missing part later, no?

martindurant:
Sorry, I had assumed that you already had a package.
In this case, you would need to not only do what I suggested above for the registry, but include your implementation in the PR too, in the directory fsspec/implementations/ . Obviously, if the implementation is incomplete, you should explicitly document how. Also, you should document how to install necessary dependencies (an abbreviated version would go in the error message in the registry too).
The best thing would be to get started, so comments can go alongside the code.

MarineChap:
#226

PR #207, opened 2019-11-21
Dropbox implementation + add existing implementation in the registry

MarineChap:
Dropbox (partial) implementation : Only download 1 file at the time (tested with jpeg image)
issue #200


martindurant:
Thank you for contributing!
Before proceeding, it seems to me that the [URL link](https://dropbox-sdk-python.readthedocs.io/en/latest/api/dropbox.html#dropbox.dropbox.Dropbox.files_get_temporary_link) capability would allow for fetching parts of a file, with RANGE header keys like a HTTPFile. That would be very useful for chunking and parallel access, which we generally care about in fsspec.

I note that there is no file listing here, which does seem fairly easy to do with the dropbox API.

Also, the API allows for uploading in chunks via a [session](https://dropbox-sdk-python.readthedocs.io/en/latest/api/dropbox.html#dropbox.dropbox.Dropbox.files_upload_session_start), so it would be nice to implement writing too. 

Aside from that, documentation will be important here, such as the `token` parameter in `__init__` - how is htis generated?

MarineChap:
The token is the dropbox key generated by each user by adding an app in their account. Where I should add this documentation ? In class comment or in an error message in case of fail due to a missing token ?

I had the file listing but for the fetch range, it is really confusing. I don't fully understand how it works ... So, seems difficult to implement. 

And finally, indeed, upload seems easy to implement and I already have a draft but I don't see how it is integrated in the filesystem class. I just need to overload the method _upload_chunk ?

martindurant:
> In class comment or in an error message

In the class docs first, since it should be a required parameter, `token=`; for the error, you should surface the failure in a way that makes sense. So don't swallow the HTTP exception for a `print`, but let it bubble up.

> the fetch range

From my brief reading, it seems that grabbing the URL and using the existing machinery of fsspec.implementations.http.HTTPFile is best. You would need to create a `Session` containing the credentials. The file size is already known from the metadata.

> I just need to overload the method _upload_chunk

`_upload_chunk` writes data to an existing upload session and commits it when `final is True`, `_initiate_upload` creates such a session.

MarineChap:
Hello, 
I have finally understood how to do the download file with fetch_range. It is working but I am unsure how to proceed. I have copy-past the fetch_range, fetch_all, read method from the http implementation and not import them. I was thinking it is better to keep the implementation independent from each other, seems good for you ? 
I have also implement the upload part.

martindurant:
@TomAugspurger, @rabernat , you may be interested by this, another very nice piece linking to broadly-available public storage.

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,244 @@
...
+        """ Initiate the upload session
+        """
+        session = self.dbx.files_upload_session_start(self.buffer.getvalue())
+        self.commit = dropbox.files.CommitInfo(path=self.path)
+        self.cursor = dropbox.files.UploadSessionCursor(
+            session_id=session.session_id, offset=self.offset
+        )
+
+
+class AllBytes(object):
```
TomAugspurger: Most likely should implement this in `fsspec.caching` (or move the one from fsspec.implementations.http there).

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,244 @@
...
+    """ Interface dropbox to connect, list and manage files
+    Parameters:
+    ----------
+    token : str
+          Generated key by adding a dropbox app in the user dropbox account. 
+          Needs to be done by the user
+
+    """
+
+    def __init__(self, **storage_options):
```
TomAugspurger: A bit clearer to have token as a required argument I think.

```suggestion
    def __init__(self, token, *args, **storage_options):
```

but still pass it through to `super()`.

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            return {"name": url, "size": None, "type": "unknow"}
+
+
+class DropboxDriveFile(AbstractBufferedFile):
+    """ fetch_all, fetch_range, and read method are based from the http implementation
+    """
+
+    def __init__(
+        self, fs, dbx, path, session=None, block_size=None, mode="rb", **kwargs
```
TomAugspurger: I'd recommend matching the signature of the base class. 
```suggestion
        self,
        fs,
        path,
        mode="rb",
        block_size="default",
        autocommit=True,
        cache_type="readahead",
        cache_options=None,
        *,
        dbx
        **kwargs
```

That uses a keyword only argument for dbx. Though I wonder: can you just get it from `fs` instead?

MarineChap:
Thank you for your review! I followed your advices. (I didn't know the possibility to directly commit suggestions sorry I will do it next time). 
For testing, I have a script which run main features but it is using my dropbox with my personal token. So I don't really know how to test it without using it.

martindurant:
Yeah, I'm not sure we have a good testing story we can bring to this one. It might have to be included as "use at your own risk". We have different testing strategies for different implementations:
- s3fs uses an emulator of the storage service
- ssh and some others use real servers deployed on docker
- gcsfs uses an HTTP request recording library (painful to use)
- gdrive uses travis environment secrets and tests against the real service (this requires PRs are only on in-repo branches, not forks)

MarineChap:
Do you think it could be merge soon ? Can I do something else to allow this ? 
And if it is merged, do you have a schedule for the next release, please ? 

I have one of my new feature based on this implementation + intake module and I am waiting to release it. I would appreciate to have just an approximat' deadline. Like this, I can decide, whether or not, I should do a monkey patch in waiting.  Thank you !!

fsspec/caching.py
```diff
@@ -370,6 +370,16 @@ def __len__(self):
         return len(self.cache)
 
 
+class AllBytes(object):
```
martindurant: This is copied from the http implementation, which seems unnecessary duplication. Indeed, if we can fetch ranges, then this is not required anymore at all

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,239 @@
...
+    """ Interface dropbox to connect, list and manage files
+    Parameters:
+    ----------
+    token : str
+          Generated key by adding a dropbox app in the user dropbox account.
+          Needs to be done by the user
+
+    """
+
+    def __init__(self, token, *args, **storage_options):
```
martindurant: storage_options become kwargs and are eventually passed to session.get I think - it should say that here. Are there any options that are worth mentioning?

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,239 @@
...
+        list_file = []
+        list_item = self.dbx.files_list_folder(
+            path, recursive=True, include_media_info=True
+        )
+        items = list_item.entries
+        while list_item.has_more:
+            list_item = self.dbx.files_list_folder_continue(list_item.cursor)
+            items = list_item.entries + items
+
+        if detail:
```
martindurant: Noting that this does not cache listings for future calls

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,239 @@
...
+        path : str
+            file path to inspect in dropbox
+        mode: str
+            Normal file modes.'rb' or 'wb'
+        block_size: int or None
+            The amount of read-ahead to do, in bytes. Default is 5MB, or the value
+            configured for the FileSystem creating this file
+        """
+        for key, value in kwargs.items():
+            print("{0} = {1}".format(key, value))
```
martindurant: This should be removed

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,239 @@
...
+        self.session = fs.session if fs.session is not None else requests.Session()
+        super().__init__(fs=fs, path=path, mode=mode, block_size=block_size, **kwargs)
+
+        self.path = path
+        self.dbx = self.fs.dbx
+        while "//" in path:
```
martindurant: `while` is not needed here, `replace` does the job in one go, unless you expect "//////"-like strings

fsspec/implementations/dropboxdrivefs.py
```diff
@@ -0,0 +1,239 @@
...
+
+        The expectation is that the server returns only the requested bytes,
+        with HTTP code 206. If this is not the case, we first check the headers,
+        and then stream the output - if the data size is bigger than we
+        requested, an exception is raised.
+        """
+        kwargs = self.kwargs.copy()
+        headers = kwargs.pop("headers", {})
+        headers["Range"] = "bytes=%i-%i" % (start, end - 1)
+        r = self.session.get(self.url, headers=headers, stream=True, **kwargs)
```
martindurant: Much of this seems to be copied from HTTPFile - which is OK, I guess, but less duplication would be good

MarineChap: In a previous comment, I asked you how to proceed with this "copy-past". I did not know if I should import http implementation to use this method and avoid redundancy or if copy-past was fine as the implementations should stay independent between them. 
Do you see another way to go ? and if not, which solution do you prefer? Thanks

martindurant: I believe you could have subclassed HTTPFile successfully to reduce the amount of code here, but without trying it, I cannot be sure - you may have had to override many of the methods anyway. Initially I suspected that you could just *use* HTTPFile directly, passing the known size in the constructor.

TomAugspurger:
Up to you I think, but my preference would be to have this in its own package.

Regradless of the decision here, we should define a policy for that.

fsspec/caching.py
MarineChap: see my comment below about "copy-past". 
Also, your colleague advises me to put this class in the caching file. If the http implementation is modified to used it from the catching module, it will remove the redundancy problem at least for this part of the code and it seems also more coherent. But as the pull request is not about the http implementation, I didn't wanted to modify it my-self here. 
Tell me what you prefer about this please. :)

martindurant: Agree that this is clearly a caching method and should be included in the caching module, although it is not particularly useful for most files, since it just keeps the whole thing in memory. Moving the def out of the http module now is fine - there are tests for that functionality that would fail if somehow moving the code went wrong.

fsspec/implementations/dropboxdrivefs.py
MarineChap: When I try to subclassed HTTPFile, I had two problems:
- the mode error raise: can be fixed with a workaround making the code very ugly
- the info method called with the temporary url path instead of the dropbox path => crashed the code and I couldn't find another workaround. 
Even if I find a workaround, we definitly cannot call this an improvement. The code is unreadable and very much complicated for nothing. 

When I try to directly used an HTTPFile, I add again the info method problem that I don't see how to fix.

MarineChap: Indeed, both methods are working. Except, that by subclassed HTTPFile, a little trick is needed to avoid error raise by the mode "wb" 
=> send "rb" in super and set again the mode "wb" after (as well as 4 bool variables) 
Like it makes the code little more dirty, I prefer the second solution (using the object HTTPFile object)

fsspec/implementations/dropboxdrivefs.py
MarineChap: Sorry it is probably because of my english but I am not sure to understand your point here...
And also, no, I don't think there are others options used in this implementation.

martindurant: I mean that any extra keyword arguments get stored in the instance, and passed to the `get` call below. This should be documented, as well as any specific options that this might be useful for.

fsspec/implementations/dropboxdrivefs.py
martindurant: OK, well we don't like dirty code, so no need to change it.

fsspec/implementations/dropboxdrivefs.py
MarineChap: I try to look how it is done in other implementations but I don't see any difference... Can you give an hint about what you were expecting here please ? 
Thanks :)

martindurant: Some implementations populate a `dirache` attribute, which is a dict of {path: [file, file]} storing the listings that have been seen so far. It has proven tricky to use, but potentially speeds up repeat listings of the same directory. You are fine to proceed without it, but may get users later complaining that calls with `glob` and `walk` are slow.

fsspec/implementations/dropboxdrivefs.py
MarineChap: The implementation is not using others options... I am just following the signature in spec.py. 
So, I am not sure to understand what to do here. Sorry :/

martindurant: Oh, I think I was getting confused between the kwargs here and the kwargs for the file class. You should just remove `self.kwargs = storage_options` in that case, since this is done by the superclass anyway.

MarineChap:
Hello guys, 
What do you think about the last version? Did you decide if you prefer a separate package or include the implementation in yours ? 
Thanks for the feedback. :)

martindurant:
@TomAugspurger would I believe prefer a separate package, and he is currently the main maintainer here, so I defer to him. That means more work for you, unfortunately: you will need to form your module into a package and release on pypi and ideally conda-forge too. At least, it does mean that you can change the code as fast as you need and release whenever you want. In that case, I would reduce this PR to include the move of the AllBytes class (if desired) and make the registry point to the new location of the implementation. Note that dropboxfs is already [taken](https://github.com/btimby/fs-dropbox/blob/master/dropboxfs.py) (that code may also be useful here).

MarineChap:
Ok. No problem for the additional work, I am always happy to try to do new stuffs!
For what I can see, dropboxfs is now maintained [here](https://github.com/smartfile/fs-dropbox). And the good point is that it implements some tests! 
But in any case, it won't be a problem as I called my module dropboxdrivefs and I didn't find any other occurrences in pypi and conda-forge. 
Thanks you for your help !

MarineChap:
Hello! Holiday are finished and I finally upload my package in pypi ([here](https://pypi.org/project/dropboxdrivefs)).
Do you think you will be able to merge this branch any time soon?

Also, I try to do a patch monkey but in fact I have no idea how to do it. 
Could you give some hints please?  
Thanks 
Happy new year ! :)

martindurant:
> I try to do a patch monkey

Sorry, what are you trying to patch?

MarineChap:
Sorry, I was not clear. In waiting to you, to eventually merge this branch and release  a new version of your package, I was trying to implement a monkey patch to make it already available for my colleagues in our environment. (basically, modify the registry file to add the dropbox implementation class). My manager advises me to it but it is the first time I am trying to do something like that. ^^

martindurant:
Oh I see. The following would be enough:
```
from fsspec.registry import registry
import dropboxfilefs
registry["dropbox"] = dropboxdrivefs.DropboxDriveFileSystem
```

...but I will merge this now. Releases are pretty frequent, but I can't promise a specific date. 

By the way, you may want to raise a new issue in this repo announcing the availability of your repo and pypi package, so that people can comment.

MarineChap:
Nice! Thank you for your help, I learn a lot! It is a good start for the year, my first PR merged ! :)

About the patch monkey, I have already tried that but as I am working directly with intake module (and not fsspec) it does not work for some reasons. But as you merged it, it is enough for me to develop the next step of my project on top of that in waiting for the release.
I will create the issue as you advice it.

PR #216, opened 2019-12-03
Add cache timeout and directory cache

TomAugspurger:
Closes https://github.com/intake/filesystem_spec/issues/215

WIP for now, I need to incorporate https://github.com/intake/filesystem_spec/issues/215#issuecomment-561357180.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -119,3 +119,11 @@ Read Buffering
 
 .. autoclass:: fsspec.caching.BlockCache
    :members:
+
+.. _dircache:
+
+Listing Cache
+-------------
+
+.. autoclass:: fsspec.DirCache
+   :members:
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Version 0.7.0
+-------------
+
+* Added ``cache_timeout`` option to :class:`fsspec.AbstractFileSystem` (:pr:``).
+* Changed :attr:`fsspec.AbstractFileSystem` to be an instance of :class:`fsspec.DirCache` rather
+  than a dictionary (:pr:``).
+
 Version 0.6.1
 -------------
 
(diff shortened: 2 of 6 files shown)


TomAugspurger:
@martindurant  Is there a standard schema for what entries in DirCache should look like? I see that `github` and `ftp` use them. In fsspec/spec.py, it's only used in `_ls_from_cache`, which seems to expect something like

```
{
    "path": {
        "name": name,
    }
} 
```

Standardizing the structure here would be good (maybe in a dataclass?), but I'm not sure what all to include yet.

TomAugspurger:
For `ftp`, elements of DirCache are `List[Tuple[str, Dict]]`, where the first item of the tuple is the path of the element, and the dict has a schema like `modify, perm, size, file, unique, name`.

```python
(Pdb) pp out[:2]
[('__init__.py',
  {'modify': '20190813183127',
   'name': '/__init__.py',
   'perm': 'r',
   'size': 0,
   'type': 'file',
   'unique': '1000004g2058413d7'}),
 ('__pycache__',
  {'modify': '20191127162327',
   'name': '/__pycache__',
   'perm': 'el',
   'size': 0,
   'type': 'dir',
   'unique': '1000004g206bad42b'})]
```

those are all the files under the path (the key).

For `github` we just have a `List[Dict]`, and the keys in the dict are `name, mode, type, size, sha`. 

These are inconsistent. At the moment, I'm leaning toward a namedtuple structure like

```python
CacheItem = namedtuple("CacheItem", ["name", "details"])
```

where `name` is a string, and `details` is a dict with anything. Hopefully that will suffice.

martindurant:
The canonical structure should be:

```
{'cached_path`: [
    {"name": 'file_path", 
     "size": 10,
     "type": "file"},
    ...
   ]
}
```
The FTP case is clearly based on the output of the client library, and ought to be processed into canonical form, as it done for s3, gcs...

martindurant:
i.e., the key is the path that we did a listing for

martindurant:
This still looks useful to me

martindurant:
Superceded by #243

PR #218, opened 2019-12-12
Files can be added to cache after first save

elephantum:


TomAugspurger:
@elephantum can you clarify the reason for this change?

elephantum:
@TomAugspurger I stumbled upon an issue, when it was impossible to add new files into the cache after it was initialised (i.e. file was created).

Our use case requires an ability to add new new files in consecutive runs of the same script.

If I understood everything correctly, then the current implementation ignores all the files that are absent in `cache` file, making it impossible to add information about new files to the `cache` after `cache` is written to the disk.

This PR fixes the issue for us.

elephantum:
To clarify it further, the issue is the following:

if `cache` file exists then the control goes into the branch https://github.com/intake/filesystem_spec/pull/218/files#diff-fd28cf28001d94bc7d51f2fe0d6f9823L131

There code iterates over all files that exist in current `cache`

Then everything else is ignored on the line: https://github.com/intake/filesystem_spec/pull/218/files#diff-fd28cf28001d94bc7d51f2fe0d6f9823L140

So if I have information about files A and B in memory, but there's a `cache` file with info about A only, then info about B will not be written into the cache.

martindurant:
I think you may have a point, though this takes some thinking about - it depends on exactly when a save happens versus accessing the new files, right? Probably it would be best to make a test showing the problem and then demonstrate that this fixes it.

elephantum:
@martindurant please take a look at the test I implemented. without my fix it fails

fsspec/implementations/tests/test_cached.py
```diff
@@ -312,3 +312,20 @@ def test_takes_fs_instance():
...
+    fs.save_cache()
+
+    fs.cat(original_file)
+    assert len(fs.cached_files[-1]) == 1
+
+    fs.save_cache()
+
+    fs2 = fsspec.filesystem(
+        'filecache', target_protocol="file", cache_storage=cache_location, 
+        do_not_use_cache_for_this_instance=True # using FS from cache is masking the issue
```
martindurant: That is ... inventive!

martindurant:
LGTM and thank you. 
Please fix the linting issue.

elephantum:
@martindurant I managed to take over lint, all's good )

martindurant:
Thanks!

PR #219, opened 2019-12-12
Fix for gcsfs protocol which is tuple

elephantum:


martindurant:
`_strip_protocol` should already deal with this, I don't suppose that the code is ever reached. You could check this with simple print/pdb.

elephantum:
@martindurant I did this change because of the actual problem I had while chaining FS's.

```
GCSFS = fsspec.filesystem('gcs')

FS = fsspec.filesystem(
    "filecache", 
    target_protocol=GCSFS, 
)

FS.open('somefile')
```

This code fails on the latest release of filespec and my PR fixes it.

martindurant:
OK, right got it - I thought this was the other way around.
The PR fails lint, but otherwise looks good.

elephantum:
@martindurant fixed formatting

martindurant:
Thank you!

PR #222, opened 2019-12-20
Deprecate auto_mkdir default value

### LINKED ISSUE #212: Deprecate auto_mkdir=True for LocalFileSystem
xref https://github.com/intake/filesystem_spec/pull/211 and https://github.com/dask/dask/issues/5526.

This is a bit of a strange API, that's no longer needed after #211. I'd like to change the default from True to `None`. Users will need to specify `auto_mkdir=True` to retain the old behavior, or `=False` to opt into the new behavior now.

jrbourbeau:
Currently we're still falling back to the current behavior of `auto_mkdir=True` if `auto_mkdir` is not specified. I'm not sure how quickly we can move to setting `auto_mkdir=False` by default. Is doing that in this PR too soon, or should we wait for a release with a `FutureWarning`?

Closes #212

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -4,6 +4,7 @@ import shutil
 import posixpath
 import re
 import tempfile
+import warnings
 from fsspec import AbstractFileSystem
 from fsspec.utils import stringify_path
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
I think let's do a release with a warning first.

TomAugspurger:
Thanks!

PR #224, opened 2019-12-31
tests: fix listdir sorting instability

Mic92:
The filesystem gives no guarantee how files in two directories are ordered.

        # the filenames in each cache are the same, but the data is different
>       assert os.listdir(cache1) == os.listdir(cache2)
E       AssertionError: assert ['d5ab5d3de20...2ba', 'cache'] == ['cache', 'd5...f00d04f152ba']
E         At index 0 diff: 'd5ab5d3de205324fbd07a48bc1f8cce33fffbb450198966e1524f00d04f152ba' != 'cache'
E         Use -v to get the full diff


martindurant:
Thanks!

Issue #229, opened 2020-01-20
PermissionError on Windows when using LocalFileSystem.transaction


dhirschfeld:
```python
from fsspec.implementations.local import LocalFileSystem


fs = LocalFileSystem()

with fs.open('data.txt', 'w') as f:
    f.write('ABC')

with fs.transaction:
    with fs.open('data.txt', 'w') as f:
        f.write('DEF')
```
```python-traceback
Traceback (most recent call last):

  File "<ipython-input-9-afc4aa1abe2e>", line 3, in <module>
    f.write('DEF')

  File "C:\Users\dhirschf\envs\dev\lib\site-packages\fsspec\transaction.py", line 24, in __exit__
    self.complete(commit=exc_type is None)

  File "C:\Users\dhirschf\envs\dev\lib\site-packages\fsspec\transaction.py", line 36, in complete
    f.commit()

  File "C:\Users\dhirschf\envs\dev\lib\site-packages\fsspec\implementations\local.py", line 219, in commit
    os.rename(self.temp, self.path)

PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\dhirschf\\AppData\\Local\\Temp\\2\\tmpa03bxzd_' -> 'C:/Users/dhirschf/code/sandbox/data.txt'
```

dhirschfeld:
***Note:*** This works fine without the transaction so it appears that the transaction is keeping an open handle to the file?

At the point of the exception the `self.f` attribute points to the temp file which appears to be closed:
```python
self
Out[2]: <fsspec.implementations.local.LocalFileOpener at 0x21e1be53c08>
self.fs
Out[3]: <fsspec.implementations.local.LocalFileSystem at 0x21e1a2be548>
self.f
Out[4]: <_io.BufferedWriter name='C:\\Users\\dhirschf\\AppData\\Local\\Temp\\2\\tmphtjujnel'>
self.f.closed
Out[5]: True
```

martindurant:
I have looked a little, and so far not found why the rename is blocked. Deleting the file is also blocked, in the case of an exception in the transaction context.

dhirschfeld:
Yeah, I haven't had any luck tracking it down so far. File locking on Windows is a nasty can of worms :/

martindurant:
If you are interested, `i, name = tempfile.mkstemp()` was the culprit - this uses `os.open` and the file does not get closed by the garbage collector.

dhirschfeld:
[...]

PR #230, opened 2020-01-20
Use atomicwrites if available

dhirschfeld:


dhirschfeld:
I ~thought~ hoped this might magically fix my problem over in #229 

Unfortunately, it didn't.

Nevertheless, it might still be worthwhile leveraging the more sophisticated replace implementation in [`atomicwrites`](https://python-atomicwrites.readthedocs.io/en/latest/#how-it-works)  if available

martindurant:
[`os.replace'](https://docs.python.org/3/library/os.html#os.replace) seems to do exactly what `atomic_replace` does (and the doc says it is atomic). I'd be happy to do `os.rename`->`os.replace`; silent overwriting seems like the write thing to do, since that is what `open(, mode='wb')` does.

dhirschfeld:
Yeah, it looks like `os.replace` does the same thing:
https://github.com/untitaker/python-atomicwrites/issues/25#issuecomment-358050598

https://github.com/python/cpython/blob/7514f4f6254f4a2d13ea8e5632a8e5f22b637e0b/Modules/posixmodule.c#L4319

It might be worthwhile to use that then instad of `os.rename`.

No great urgency but I'll try to find time to update the PR...

PR #233, opened 2020-01-23
Update LocalFileSystem to expand user home directory

jrbourbeau:
This PR updates `LocalFileSystem._strip_protocol` to expand `"~"` with the full user home directory path. This allows us to handle local paths like `"~/foo/bar/baz.csv"`

Following up on the conversation in https://github.com/dask/dask/issues/5819

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -6,6 +6,7 @@ import os.path
 import sys
 from contextlib import contextmanager
 import tempfile
+import getpass
 
 import pytest
 import fsspec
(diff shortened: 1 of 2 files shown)


fsspec/implementations/tests/test_local.py
```diff
@@ -439,3 +440,11 @@ def test_auto_mkdir_warns():
     LocalFileSystem.clear_instance_cache()
     with pytest.warns(FutureWarning, match="auto_mkdir=True has been deprecated"):
         LocalFileSystem()
+
+
+def test_strip_protocol_expanduser():
+    path = "file:///C:\\foo\\bar" if sys.platform == "win32" else "file://~/foo/bar"
```
jrbourbeau: I don't have access to a windows machine, so I haven't been able to test that `"file:///C:\\foo\\bar"` passes locally

martindurant: Shouldn't "~" work on windows too?
Also, I notice there are three "/" characters  in the win path, not two.

jrbourbeau: Oops my mistake, I forgot to include the `"~"` there. I'll add it now.

I added the extra "/" based on 

https://github.com/intake/filesystem_spec/blob/c325a40e5c3bb73100c0f922bccf2e0d4e8bca3a/fsspec/implementations/tests/test_local.py#L84

Should the protocol always be `file://` (two slashes)?

martindurant: I don't know why there was three there. I hope two is right... Let's get windows tests on this. I can do it sometime not today.

PR #236, opened 2020-01-28
Allow connecting to the Name-node via https

yamrzou:
This fixes an error encountered when using WebHDFS : `ConnectionError: ('Connection aborted.', BadStatusLine('\x15\x03\x03\x00\x02\x02P'))`, when the Name-node only uses HTTPS.

### DIFF
--- a/fsspec/implementations/webhdfs.py
+++ b/fsspec/implementations/webhdfs.py
@@ -83,7 +83,7 @@ class WebHDFS(AbstractFileSystem):
             return
         super().__init__(**kwargs)
         self.url = "{protocol}://{host}:{port}/webhdfs/v1".format(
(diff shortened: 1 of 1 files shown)


martindurant:
This PR is failing linting, please run `black` on your code. Otherwise it looks good! Do you know if it's easy to set the docker server that we test against to run HTTPS, so that we can test if this works?

yamrzou:
I kept waiting for a response to the hadoop-test-cluster build issue, but it seems to take some time.. Anyways I tested this manually in our production cluster, along with Kerberos enabled and it worked successfully.

martindurant:
I don't think hadoop-test-cluster can be set up for HTTPS, that would require the generation of certs; I wouldn't be surprised if some other docker image out there has done it, though.

PR #238, opened 2020-02-10
fix https://github.com/dask/s3fs/issues/190

xuchen:
Tested my use case of saving scipy npz and it worked

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1036,7 +1036,7 @@ class AbstractBufferedFile(io.IOBase):
         """
         loc = int(loc)
         if not self.mode == "rb":
-            raise ValueError("Seek only available in read mode")
+            raise OSError("Seek only available in read mode")
         if whence == 0:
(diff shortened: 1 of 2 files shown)


martindurant:
Assuming this doesn't break any of the existing tests, it would be nice to add a test to `test_file`, using FTP to read/write a ZipFile with a small binary blob like `b"hello zip"`.

martindurant:
Spoke too soon: `test_file_write_attributes` (also in `test_file`) apparently checks for `ValueError` (the previous behaviour) explicitly, so the new `OSError` sneaks out and breaks the test.

martindurant:
I think:
```
z = zipfile.ZipFile(f, mode='w')
with z.open('innerfile', mode='w') as zz:
       zz.write(b'hello zip')
```
(because zipfile is not a single compressed file, but an archive of files)

martindurant:
Perfect, thank you.

PR #243, opened 2020-03-04
instance- and dir-caches

martindurant:
@TomAugspurger what do you think, is this a simple way to solve things? I'm thinking every fsspec instance will have a `dircache` instance attached, but they are not required to use it (e.g., local never would). What exactly the cache contains is up to the implementation, but the structure in the docstring covers s3fs and gcsfs . When I plug it in, `invalidate_cache()` should maybe be moved to the new class, with an optional flag on whether directory parents should also be invalidated.

### DIFF
--- a/fsspec/dircache.py
+++ b/fsspec/dircache.py
@@ -24,7 +24,13 @@ class DirCache(MutableMapping):
     caching off
     """
 
-    def __init__(self, use_listings_cache=True, listings_expiry_time=None, max_paths=None, **kwargs):
+    def __init__(
+        self,
+        use_listings_cache=True,
+        listings_expiry_time=None,
+        max_paths=None,
+        **kwargs
+    ):
         """
 
         Parameters
@@ -84,4 +90,7 @@ class DirCache(MutableMapping):
         return (k for k in self._cache if k in self)
 
     def __reduce__(self):
-        return DirCache, (self.use_listings_cache, self.listings_expiry_time, self.max_paths)
+        return (
+            DirCache,
+            (self.use_listings_cache, self.listings_expiry_time, self.max_paths),
+        )
--- a/fsspec/dircache.py
+++ b/fsspec/dircache.py
@@ -1,8 +1,9 @@
 from functools import lru_cache
 import time
+from collections.abc import MutableMapping
 
 
-class DirCache(dict):
+class DirCache(MutableMapping):
     """
     Caching of directory listings, in a structure like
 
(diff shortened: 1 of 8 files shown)


fsspec/dircache.py
```diff
@@ -0,0 +1,50 @@
+from functools import lru_cache
+import time
+
+
+class DirCache(dict):
```
TomAugspurger: Since you're changing the `__init__` signature, I think it'd be better to compose this class with a `._cache` method rather than inheriting from dict. You might inherit from MutableMapping instead.

fsspec/dircache.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __getitem__(self, item):
+        if self.exp:
+            if self._times.get(item, 0) - time.time() < self.exp:
+                del self._cache[item]
+        if self.max:
+            self.q(item)
+        return self._cache[item]  # maybe raises KeyError
+
+    def __setitem__(self, key, value):
```
TomAugspurger: Should this validate the structure of `value`?

fsspec/dircache.py
```diff
@@ -0,0 +1,50 @@
...
+        },
+        ...
+        ],
+     "path1": [...]
+    }
+
+    Parameters to this class control listing expiry or indeed turn
+    caching off
+    """
+    def __init__(self, usecache=True, expiry_time=False, max_size=None):
```
TomAugspurger: Is this max_size in terms of elements or bytes?

Evicting based on the number of elements doesn't seem especially useful to me, at least no compared to the size of the values. But perhaps s3fs or gcsfs already does that?

fsspec/dircache.py
```diff
@@ -0,0 +1,50 @@
...
+        self._cache = {}
+        self._times = {}
+        if max_size:
+            self.q = lru_cache(max_size + 1)(lambda key: self._cache.pop(key, None))
+        self.use = usecache
```
TomAugspurger: These names should match the names of the arguments to init.

fsspec/dircache.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    Parameters to this class control listing expiry or indeed turn
+    caching off
+    """
+    def __init__(self, usecache=True, expiry_time=False, max_size=None):
+        self._cache = {}
+        self._times = {}
+        if max_size:
+            self.q = lru_cache(max_size + 1)(lambda key: self._cache.pop(key, None))
```
TomAugspurger: It feels a bit strange to have the max_size parameter affect the attribute where we actually store things. That leads to stuff like `if self.max` later on. Can we just always store it in the same place?

fsspec/dircache.py
martindurant: I'll probably inherit from nothing in the end, since I don't expect calls to `keys()`, etc.

fsspec/dircache.py
martindurant: Not certain, I don't know if we want to force implementations here. Might make things simpler if we do.

fsspec/dircache.py
martindurant: It is number of elements as it stands, because it's far easier to implement using LRU. 
There is no such expiry anywhere yet; gcsfs has time-based expiry.

fsspec/dircache.py
martindurant: ok

fsspec/dircache.py
martindurant: `._cache` stored the listings. The q and times dicts hold only information on what to automatically evict, and are not accessed if eviction is not desired.

martindurant:
@TomAugspurger , this can be considered working, although the docs should be updated to explain this (and changes propagated to s3fs, gcsfs).

Did you have an opinion if `invalidate_cache` should live in the `DirCache` implementation?

martindurant:
> if invalidate_cache should live in the DirCache implementation?

I'm going against this in the end, because there can be different requirements in the implementations (e.g., on s3, the creation of a key may affect all parent dirs, but on ssh you must explicitly make those dirs first)

PR #248, opened 2020-03-17
Small fixes for a general zarr store

martindurant:
ref: https://github.com/zarr-developers/zarr-python/issues/540


martindurant:
(gcsfs needs rerecording of the one test - it does pass)

PR #257, opened 2020-03-24
Implement chained URLs

martindurant:
@TomAugspurger : the API for calling s3 via a dask worker is like

```
dd.read_csv("dask://localhost:8786::s3://bucket/path/*.csv", storage_options={'s3': {'anon': True}})
```

Obviously needs more testing...

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -197,6 +197,21 @@ def open_files(
     -------
     List of ``OpenFile`` objects.
     """
+    chain = _un_chain(urlpath, kwargs)
+    if len(chain) > 1:
+        kwargs = chain[0][2]
+        inkwargs = kwargs
+        for ch in chain[1:]:
+            urls, protocol, kw = ch
+            inkwargs['target_protocol'] = protocol
+            inkwargs['target_options'] = kw.copy()
+            inkwargs = inkwargs['target_options']
+        protocol = chain[0][1]
+        if isinstance(urlpath, str):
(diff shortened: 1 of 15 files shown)


martindurant:
Ref: https://github.com/dask/dask/pull/6011

martindurant:
@TomAugspurger , thoughts on auto_mkdir here, as requires by pyarrow? It would be nice to avoid the warning showing up, but I'm not certain what the default value should be; we certainly don't want to break arrow usage of fsspec.

TomAugspurger:
I think auto_mkdir is unnecessary after https://github.com/intake/filesystem_spec/pull/211. So it should be fine to make that change, but I'd recommend doing it in its own PR.

---

Did you think at all about making the protocol `dask-s3fs://` rather than `dask:...:s3fs:://``? I haven't looked closely at the implementation, but I'm wondering if that would simplify things. Ideally this would be just another backend, and would require changes in core and spec?

martindurant:
`dask-s3fs://` wouldn't allow you to encode extra path parts into the URL, like the example with including the dask client address, or where you want to access some file(s) within a zip (there is a test for "simplecache::zip://*::file://filepath", which extracts files to the cache from the target file, which could have been remote)

TomAugspurger:
Do you think that encoding things like the scheduler address in the URL is a good idea? My initial reaction is that would would fall more naturally under a `storage_args` (with a default of getting a global client, so it shouldn't be necessary).

martindurant:
both work

PR #259, opened 2020-03-26
Only catch FileNotFound in mapping get

### LINKED ISSUE #255: Potential bad interaction with zarr and missing chunks
We've encountered a situation using zarr on GCS via gcsfs where an attempt to retrieve a chunk object fails for some transient reason, but the fsspec gcsfs mapping raises a KeyError, which zarr interprets as meaning the object doesn't exist. Zarr's default behaviour in that case is to fill the chunk entirely with the fill value set for the array.

What this means is that any transient errors in GCS can result in chunks of an array being returned as filled with the fill value, rather than the actual data, without any error being raised. This is obviously a serious issue.

Currently the fsspec `FSMap` class catches any exception within `__getitem__` and reraises as a KeyError: 

https://github.com/intake/filesystem_spec/blob/cc7cbc0cc58c96c427708c1e9a98b0d6c99bd1f0/fsspec/mapping.py#L77

I think this should be modified in some way, such that `FSMap` is able to distinguish between a genuine "this file/object does not exist" error and any other kind of error. In the case where a "this file/object does not exist" error is encountered, it should be safe to reraise that from `__getitem__` as a `KeyError`, but any other type of error I suggest would be safer to raise either by propagating the underlying error as-is, or by reraising as a `RuntimeError`. This would mean that any transient errors would get propagated all the way up through zarr, preventing any silent replacement of data with fill values.

For the time being we have worked around this with a wrapper class that transforms all `KeyError` into `RuntimeError` as [described here](https://nbviewer.jupyter.org/gist/alimanfoo/66d7760440eba981214d483f6d917c79). Note that this only works for the case where we know ahead of time that all chunks of an array should exist, and therefore we never expect to experience a `KeyError` for any reason.

martindurant:
For discussion whether this is a good idea. Sees to solve zarr's problem, but would be good to test on large datasets in the cloud.

Fixes #255

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -15,7 +15,7 @@ class MemoryFileSystem(AbstractFileSystem):
     protocol = "memory"
     root_marker = ""
 
-    def ls(self, path, detail=False):
+    def ls(self, path, detail=False, **kwargs):
         if path in self.store:
             # there is a key with this exact name, but could also be directory
             out = [
@@ -127,7 +127,10 @@ class MemoryFileSystem(AbstractFileSystem):
         self.store[path2] = MemoryFile(self, path2, self.store[path1].getbuffer())
 
     def cat(self, path):
-        return self.store[path].getvalue()
+        try:
+            self.store[path].getvalue()
+        except KeyError:
+            raise FileNotFoundError(path)
 
     def _rm(self, path):
         del self.store[path]
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -128,7 +128,7 @@ class MemoryFileSystem(AbstractFileSystem):
 
     def cat(self, path):
         try:
-            self.store[path].getvalue()
+            return self.store[path].getvalue()
         except KeyError:
             raise FileNotFoundError(path)
 
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -71,9 +71,9 @@ class FSMap(MutableMapping):
 
     def __getitem__(self, key, default=None):
         """Retrieve data"""
-        key = self._key_to_str(key)
+        k = self._key_to_str(key)
         try:
-            result = self.fs.cat(key)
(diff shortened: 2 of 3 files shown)


martindurant:
Actually, one concrete case for confound would be `IsADirectoryError`, i.e., we tried to read a path that does exist, but is not a file. It may be reasonable to put a `fs.isfile(path)` into `__getitem__`, which also leads to `KeyError` if False.

alimanfoo:
Thanks Martin.

> Actually, one concrete case for confound would be `IsADirectoryError`, i.e., we tried to read a path that does exist, but is not a file. It may be reasonable to put a `fs.isfile(path)` into `__getitem__`, which also leads to `KeyError` if False.

FWIW that is what the zarr DirectoryStore [currently does](https://github.com/zarr-developers/zarr-python/blob/e7d5f39b4488a2d230a7ef5df7e69ef848f6ae78/zarr/storage.py#L785). But for the more general case that could include cloud stores, it might hurt performance to add an `isfile` check if it involved an extra service call. I'm happy with the PR as it stands.

openSourcerer9000:
I'm so confused about this. I'm trying to run[ this example notebook](https://nbviewer.org/github/NOAA-OWP/AORC-jupyter-notebooks/blob/master/jupyter_notebooks/AORC_Zarr_notebook.ipynb), sometimes it returns data, and sometimes NaNs, even when querying the same spatiotemporal bounds on different calls! I'm not sure where to ask this, since it may be an S3 thing, but if someone knows the answer: why is this not consistent? How do I ensure my calls receive data?

I'm a Zarr-head, and believe this should be the easiest way to access this CF-compliant data, but how is this usable if you're playing Russian roulette trying to access data and coming up with blanks most of the time?

I'm not sure I'd rather see errors cancelling an entire operation, especially if some data is coming through. But I would like to be able to differentiate between missing data and understand what's hanging up S3 calls.

fsspec v
'2024.3.1'
zarr v
'2.17.1'

PR #263, opened 2020-03-31
WebHDFS - Support some intricacies of HttpFS

airosa:
Attempting to add support in webhdfs to some intricacies of [HttpFS]( https://hadoop.apache.org/docs/current/hadoop-hdfs-httpfs/index.html).

[Create and write to a file]( https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Create_and_Write_to_a_File) and [append to a File]( https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Append_to_a_File) both require request header "content-type" with value "application/octet-stream". You will get the following error message if you do not set it: “_HTTP Status 400 - Data upload requests must have content-type set to application/octet-stream_”. Set the content-type always to application/octet-stream for the above requests.

[Open and read a file]( https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Open_and_Read_a_File) and [get file checksum]( https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Get_File_Checksum) “usually” redirect the request.  However HttpFS does not seem to redirect (even if noredirect flag is not set) meaning there is no Location response header. For these requests check if the Location response header is actually returned and use the Location value only if it is returned.

### DIFF
--- a/fsspec/implementations/webhdfs.py
+++ b/fsspec/implementations/webhdfs.py
@@ -226,10 +226,14 @@ class WebHDFS(AbstractFileSystem):
     def ukey(self, path):
         """Checksum info of file, giving method and result"""
         out = self._call("GETFILECHECKSUM", path=path, redirect=False)
-        location = self._apply_proxy(out.headers["Location"])
-        out2 = self.session.get(location)
-        out2.raise_for_status()
-        return out2.json()["FileChecksum"]
+        if "Location" in out.headers:
+            location = self._apply_proxy(out.headers["Location"])
+            out2 = self.session.get(location)
+            out2.raise_for_status()
+            return out2.json()["FileChecksum"]
+        else:
+            out.raise_for_status()
+            return out.json()["FileChecksum"]
 
     def home_directory(self):
         """Get user's home directory"""
@@ -352,7 +356,11 @@ class WebHDFile(AbstractBufferedFile):
             This is the last block, so should complete file, if
             self.autocommit is True.
         """
-        out = self.fs.session.post(self.location, data=self.buffer.getvalue())
+        out = self.fs.session.post(
+            self.location,
+            data=self.buffer.getvalue(),
+            headers={"content-type": "application/octet-stream"},
+        )
         out.raise_for_status()
         return True
 
@@ -369,7 +377,9 @@ class WebHDFile(AbstractBufferedFile):
         location = self.fs._apply_proxy(out.headers["Location"])
         if "w" in self.mode:
             # create empty file to append to
-            out2 = self.fs.session.put(location)
+            out2 = self.fs.session.put(
+                location, headers={"content-type": "application/octet-stream"}
+            )
             out2.raise_for_status()
         self.location = location.replace("CREATE", "APPEND")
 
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks for this!

Do you suppose there's any way to test these unorthodox situations?

> You will get the following error message if you do not set it

I am glad to see such a specific error message, although it makes we wonder why they don't just assume the default.

martindurant:
btw: would be nice to add to the docstring that this class implements webhdfs and httpfs, right?

airosa:
No problem and thank you for this super package! 

Forgot to mention that I was running against CDH [5.15.1](https://docs.cloudera.com/documentation/enterprise/release-notes/topics/cdh_vd_cdh_package_tarball_515.html#packaging_5151) cluster with hadoop-2.6.0+cdh5.15.1+2822.

> btw: would be nice to add to the docstring that this class implements webhdfs and httpfs, right?

Mentioning support for HttpFS would indeed be nice as many people are probably accessing HttpFS gateways rather than WebHDFS directly on namenodes/datanodes - particularly if they have access to secured Hadoop clusters.  

> Do you suppose there's any way to test these unorthodox situations?

Did you mean creating test cases for these?

martindurant:
> Did you mean creating test cases for these?

exactly.

airosa:
> Did you mean creating test cases for these? exactly.

This should be possible but it looks like the underlying [htcluster](https://github.com/jcrist/hadoop-test-cluster) does not spin up a HttpFS server in addition to the Namenode and Datanode. This would be a bit of work in that project. I did modify and manually run the tests in the test_webhdfs.py against a HttpFS server and indeed there was one more anomaly which broke  the test_simple test case. 

Looks like HttpFS returns HTTP response code 500 instead of 403 for a permission error. Fortunately the [error responses](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Error_Responses) do follow the proper [RemoteException](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#RemoteException_JSON_Schema) json schema. 

I attempted to fix the error responses by mapping the exception names to Python exceptions. I also  included the returned exception message in the raised exception. These messages are very useful in troubleshooting as for example AccessControlException contains both the actual user name and the remote path. With this fix all the existing webhdfs tests passed.

Finally, also adjusted the docstring. Thanks for your patience!

BTW, noticed that _initiate_upload does a rm instead of setting the overwrite flag to true but that is another story.

martindurant:
> _initiate_upload does a rm instead of setting the overwrite flag to true 

Does that require correction?
If not, I'll merge this now, looks good.

It would indeed be great to get a test cluster that does all the things (including a backend for ssh). htcluster is quite old now and not maintained, but this is pretty niche.

airosa:
> Does that require correction?
 It is not related to HttpFS support so I would keep it separate from this pull reg.

 > If not, I'll merge this now, looks good.
Excellent, thanks a lot!

PR #268, opened 2020-04-04
Feature/modified

limx0:
- Add `created` and `modified` methods to `AbstractFileSystem`
- Add implementation for `LocalFileSystem`
- Some place holders in conftest.py for testing further FS additions (can remove if preferred)


martindurant:
The created and modified time seem to be exactly equal, failing the test. I don't know what the FS time resolution is, and of course the OS will cache file IO. You probably need a `sleep(0.05)`.

On the conftest addition: for anyone that doesn't have all the required packages of all of the backends, this will cause all the tests to skip, which would be unfortunate. I would go with simple here.

Please run `black` on your code in order to pass the lint run.

limx0:
Thanks @martindurant, I've removed the unused filesystems from the conftest and blacked the code. Ready for another look when you have the time - thanks.

fsspec/implementations/tests/test_common.py
```diff
@@ -0,0 +1,33 @@
+import datetime
+import time
+import pytest
+
+from fsspec import AbstractFileSystem
+from fsspec.implementations.tests.conftest import READ_ONLY_FILESYSTEMS
+
+TEST_FILE = "file"
```
martindurant, line 8: Will this create a (temporary) file in the local directory? May be better using `tempfile.mktemp`

fsspec/spec.py
```diff
@@ -976,6 +976,12 @@ def download(self, rpath, lpath, recursive=False, **kwargs):
         """Alias of :ref:`FilesystemSpec.get`."""
         return self.get(rpath, lpath, recursive=recursive, **kwargs)
 
+    def created(self, path):
```
martindurant, line 979: They sound pretty self-explanatory, but would be good to include docstrings anyway.

limx0:
Added the fixes - thanks for the feedback @martindurant

PR #275, opened 2020-04-16
DVC, git interface; and more

martindurant:
cc @jsignell


martindurant:
needs tests for:
- [ ] compression in cache
- [ ] local git
- [ ] dvc (needs a lot of deps)

Issue #277, opened 2020-04-17
numpy slice assignment failing on `readinto`


JacksonMaxfield:
`readinto` failing to perform slice assignment:

```
~/active/cell/filesystem_spec/fsspec/spec.py in readinto(self, b)
   1249         # for i in range(len(data)):
   1250         #     b[i] = data[i]
-> 1251         b[: len(data)] = data
   1252         return len(data)
   1253 

ValueError: invalid literal for int() with base 10: ''
```

**`fsspec` version:** 0.7.2

**Context:**
I am one of the developers for [`aicsimageio`](https://github.com/AllenCellModeling/aicsimageio), relevant to this issue is the fact that we use delayed dask arrays to read microscopy images, but as a part of that we support very large files (usually this means >200GB). We natively support handing in `str` and `Path` to the `AICSImage` object as one would expect and I was in the process of extending support to reading from `S3File` or really, _any_ `fsspec` file like object.

Bear with me...
**Commands for reproduction setup:**
```bash
git clone git@github.com:AllenCellModeling/aicsimageio.git
cd aicsimageio
git checkout feature/support-s3-file
pip install -e .
python
```

(The above was done in a conda environment on python 3.7)

**Commands to reproduce:**
```python
from aicsimageio import AICSImage
from s3fs import S3File, S3FileSystem
fs = S3FileSystem(anon=True)
f = S3File(fs, "allencell/aics/pipeline_integrated_single_cell/cell_images_3d/00001bb5_16872_35317_reg.tiff")
img = AICSImage(f)
print(img.shape)
```

**Traceback:**
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/maxfield/active/cell/aicsimageio/aicsimageio/aics_image.py", line 207, in shape
    return self.size()
  File "/home/maxfield/active/cell/aicsimageio/aicsimageio/aics_image.py", line 197, in size
    return tuple([self.dask_data.shape[self.dims.index(dim)] for dim in dims])
  File "/home/maxfield/active/cell/aicsimageio/aicsimageio/aics_image.py", line 197, in <listcomp>
    return tuple([self.dask_data.shape[self.dims.index(dim)] for dim in dims])
  File "/home/maxfield/active/cell/aicsimageio/aicsimageio/aics_image.py", line 152, in dask_data
    reader_data = self.reader.dask_data
[...]

PR #278, opened 2020-04-17
Fix slice assignment during readinto for numpy reads

JacksonMaxfield:
In short, the slice assignment is failing on `readinto`, replacing with a simple loop and individual item assignment works.

See #277 for more details

Happy to optimize this and writes tests, would love feedback on how to do that.


martindurant:
This ought to be *much* slower than the assignment. Have you any idea why it's breaking? There should be a test showing that this works in at least some case.

martindurant:
closes #277

martindurant:
Perfect, thank you.

PR #279, opened 2020-04-17
On recursive get, skip directories when writing files.

journeymansix:
When using get with recursive=True, I was getting errors on the directory nodes because it was trying to write to them as a files. This fixes the problem.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -607,11 +607,12 @@ class AbstractFileSystem(up, metaclass=_Cached):
             lpaths = [lpath]
         for lpath, rpath in zip(lpaths, rpaths):
             with self.open(rpath, "rb", **kwargs) as f1:
-                with open(lpath, "wb") as f2:
-                    data = True
-                    while data:
-                        data = f1.read(self.blocksize)
-                        f2.write(data)
+                if not os.path.isdir(lpath):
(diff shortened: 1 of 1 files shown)


fsspec/spec.py
```diff
@@ -607,11 +607,12 @@ def get(self, rpath, lpath, recursive=False, **kwargs):
             lpaths = [lpath]
         for lpath, rpath in zip(lpaths, rpaths):
             with self.open(rpath, "rb", **kwargs) as f1:
-                with open(lpath, "wb") as f2:
-                    data = True
-                    while data:
-                        data = f1.read(self.blocksize)
-                        f2.write(data)
+                if not os.path.isdir(lpath):
```
journeymansix, line 610: Directory was created in os.makedirs above.

martindurant:
Fixes https://github.com/dask/gcsfs/issues/248 ?

journeymansix:
Yes, it should. As https://github.com/vincentlepage said, it appears it
would happen on any recursive copy.

journeymansix:
Actually this probably needs a more comprehensive look. While it fixes the recursive get, I'm not sure what the behavior should be for all the other the combinations of lpath and rpath: dir or file, top-level lpath exists or doesn't exist. This may break other behavior.

PR #282, opened 2020-04-21
Allow chained URL syntax for get_mapper

martindurant:
cc @raphaeldussin , would allow your example to become
```yaml
  temp_zip:
    description: zarr zipstore
    driver: zarr
    args:
      urlpath: 'zip::temp_zip.zip'
      consolidated: False
```


martindurant:
(although using CATALOG_DIR or absolute path to the zip file would be better

PR #284, opened 2020-04-23
extent put to support source paths ending with a slash

dynamix:
As explained in https://github.com/intake/filesystem_spec/issues/281

### DIFF
--- a/fsspec/tests/test_api.py
+++ b/fsspec/tests/test_api.py
@@ -57,6 +57,19 @@ def test_get_put(tmpdir):
     fs.put(tmpdir, "/more", recursive=True)
     assert fs.find("/more") == ["/more/dir/two", "/more/one", "/more/three"]
 
+    entries = ["/dir/two", "/one", "/three"]
+
+    fs.mkdir("/exists")
+    fs.put(tmpdir, "/exists", recursive=True)
+    tmpbase = os.path.basename(tmpdir)
+
+    assert fs.find("/exists") == ["/exists/" + tmpbase + p for p in entries]
+
+    fs.mkdir("/existstwo")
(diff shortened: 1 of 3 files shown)


martindurant:
The docstring should be updated to show, that if the target path exists, then the behaviour you'll see depends on whether you have that final "/", to make it clear.
We shoulc make clear that we are copying bash-style copying, rather than [copytree](https://docs.python.org/2/library/shutil.html#shutil.copytree) which requires the target must not exists.

fsspec/tests/test_api.py
```diff
@@ -57,6 +57,19 @@ def test_get_put(tmpdir):
     fs.put(tmpdir, "/more", recursive=True)
     assert fs.find("/more") == ["/more/dir/two", "/more/one", "/more/three"]
 
+    entries = ["/dir/two", "/one", "/three"]
```
martindurant, line 60: I agree with fixing the memory FS;
but perhaps the best way to show that we indeed do exactly what is expected, it might be best to use another temporary directory and make a number of tests of (inpath, outpath) and actually call "cp -r {inpath} {outpath}" (with the shell, on posix only) to show that the right thing happens.

martindurant:
@dynamix , do you intend to work on this in the near future?

dynamix:
@martindurant I missed the GH notification. I can update the docstring.  Regarding your other comment: the idea would be to compare if the outcome of calling `cp -r {inpath} {outpath}` is identical to the `fs.put`?

martindurant:
Exactly - on posix systems at least

martindurant:
I believe that get/put does now work this way, but worth checking for edge cases still

Issue #285, opened 2020-04-23
Cache chaining fails with recent s3fs


ian-r-rose:
I ran into this problem with using the `WholeFileCacheFileSystem` chained with s3. The more recent versions s3fs have a list of protocols: `["s3", "s3a"]`, but `fsspec` expects the protocol attribute to be a string.

So using `s3fs` 3.5, the following succeeds:
```python
import fsspec
from fsspec.implementations.cached import WholeFileCacheFileSystem

target = fsspec.filesystem("s3", anon=True)
fs = WholeFileCacheFileSystem(fs=target)
fs.open("s3://landsat-pds/scene_list.gz")
```
But using `s3fs` version 0.4.2 it produces the following error:
```python
Traceback (most recent call last):
  File "test_s3_chain.py", line 6, in <module>
    fs.open("s3://landsat-pds/scene_list.gz")
  File "/home/ian/miniconda3/lib/python3.7/site-packages/fsspec/implementations/cached.py", line 299, in <lambda>
    return lambda *args, **kw: getattr(type(self), item)(self, *args, **kw)
  File "/home/ian/miniconda3/lib/python3.7/site-packages/fsspec/spec.py", line 775, in open
    **kwargs
  File "/home/ian/miniconda3/lib/python3.7/site-packages/fsspec/implementations/cached.py", line 299, in <lambda>
    return lambda *args, **kw: getattr(type(self), item)(self, *args, **kw)
  File "/home/ian/miniconda3/lib/python3.7/site-packages/fsspec/implementations/cached.py", line 351, in _open
    if not path.startswith(self.target_protocol):
TypeError: startswith first arg must be str or a tuple of str, not list
```

martindurant:
Sorry, that's annoying - the cost of not testing every feature with every combination of backends. The following fix would do it
```diff
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -105,7 +105,8 @@ class CachingFileSystem(AbstractFileSystem):
         self.compression = compression
         self.same_names = same_names
         self.target_protocol = (
-            target_protocol if isinstance(target_protocol, str) else fs.protocol
+            target_protocol if isinstance(target_protocol, str) else
+            (fs.protocol if isinstance(fs.protocol, str) else fs.protocol[0])
         )
         self.load_cache()
         self.fs = fs if fs is not None else filesystem(target_protocol, **self.kwargs)
```

martindurant:
[...]

PR #287, opened 2020-04-24
Fix multi protocol cache

ian-r-rose:
Implements the fix suggested in #285


ian-r-rose:
CI Failure looks like a network hiccup

PR #291, opened 2020-05-05
Add panel-based GUI

martindurant:
(using my sigslot mechanism)


martindurant:
@jsignell , thought you'd appreciate this new version

martindurant:
Screen Shot 2020-05-05 at 3 24 16 PM

martindurant:
Some next-level questions:
- some backends can derive kwargs from the url; should they be interpreted and fill in the kwargs box automatically, or just used as-is?
- since we now support chaining, how would you do something like "browse the contents of the currently selected zip file" or pick a protocol (e.g., dask::s3) other than the simple list?
- it would be good to optionally show the docstring and/or signature of the currently-selected protocol's backend somewhere

PR #298, opened 2020-05-15
Provide explicit way to register implementations

### LINKED ISSUE #293: Provide an explicit way to register a protocol
I'm working on a [v3io driver](https://github.com/v3io/v3io-fs) for fsspec.
I can't find a way to register the `v3io` protocol, currently I'm doing
```python
from fsspec.registry import known_implementations

known_implementations[V3ioFS.protocol] = {
     'class': 'v3iofs.V3ioFS',
}
```

Which is  seems like working with internal implementaion detilas.

I suggest adding `register_protocol(protocol: str, cls: type)` function that will allow implementations to register their protocol. I should probably raise an error if the protocol is already registered.

martindurant:
Fixes #293 

It would still be good to implement an entrypoints method (althoug that would be even better without the dependency on `entrypoints`, to keep this package clean!)

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -38,6 +38,8 @@ Base Classes
    fsspec.core.BaseCache
    fsspec.core.get_fs_token_paths
    fsspec.dircache.DirCache
+   fsspec.registry.ReadOnlyRegistry
+   fsspec.registry.register_implementation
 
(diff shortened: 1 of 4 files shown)


martindurant:
@tebeka , does this meet your expectations?

tebeka:
> @tebeka , does this meet your expectations?

Yes, thanks!

PR #299, opened 2020-05-19
Allow OpenFile.open to survive del of OpenFile

martindurant:


martindurant:
This is required, because pandas' `get_filepath_or_buffer` expects to get a file-like object from an arbitrary URL (with optional compression, etc) and we would rather use `fsspec.open` to achieve this, rather than replicate extracting the protocol and instantiating the filesystem. That latter option is still possible, however.

PR #310, opened 2020-06-09
Async HTTP implementation

martindurant:
Experiment with applying async methods to a backend implementation.

The idea is to not change the base API, except maybe offering extra "batching" kwargs where appropriate. Thus, any backend that isn't async (can't be or the work hasn't been done), nothing will change at all, and where the backend is async, the existing methods will still be sync and call async things only internally. (The coroutines will be available for other devs on a "know what you're doing" basis)

The following functionality stands to benefit:
- down/uploading multiple files
- down/uploading single large files in chunks
- batch delete/copy/move operations

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -32,7 +32,7 @@ def sync(loop, func, *args, callback_timeout=None, **kwargs):
             if callback_timeout is not None:
                 future = asyncio.wait_for(future, callback_timeout)
             result[0] = await future
-        except Exception as exc:
+        except Exception:
             error[0] = sys.exc_info()
         finally:
             thread_state.asynchronous = False
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -1,0 +1,194 @@
+import asyncio
+import functools
+import inspect
+import re
+import sys
+import threading
+
+from .utils import other_paths
+
+# this global variable holds whether this thread is running async or not
+thread_state = threading.local()
+private = re.compile("_[^_]")
+
+
+def sync(loop, func, *args, callback_timeout=None, **kwargs):
+    """
+    Run coroutine in loop running in separate thread.
+    """
+
+    e = threading.Event()
+    main_tid = threading.get_ident()
+    result = [None]
+    error = [False]
+
+    async def f():
+        try:
+            if main_tid == threading.get_ident():
+                raise RuntimeError("sync() called from thread of running loop")
+            await asyncio.sleep(0)
+            thread_state.asynchronous = True
+            future = func(*args, **kwargs)
+            if callback_timeout is not None:
+                future = asyncio.wait_for(future, callback_timeout)
+            result[0] = await future
+        except Exception as exc:
+            error[0] = sys.exc_info()
+        finally:
+            thread_state.asynchronous = False
+            e.set()
+
+    asyncio.run_coroutine_threadsafe(f(), loop=loop)
+    if callback_timeout is not None:
+        if not e.wait(callback_timeout):
+            raise TimeoutError("timed out after %s s." % (callback_timeout,))
+    else:
+        while not e.is_set():
+            e.wait(10)
+    if error[0]:
+        typ, exc, tb = error[0]
+        raise exc.with_traceback(tb)
+    else:
+        return result[0]
+
+
+async def _run_as_coroutine(func, *args, **kwargs):
+    return func(*args, **kwargs)
+
+
+def sync_wrapper(func, obj=None):
+    """Given a function, make so can be called in async or bocking contexts
+
+    Leave obj=None if defining within a class. Pass the instance if attaching
+    as an attribute of the instance.
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        self = obj or args[0]
+        loop = self.loop
+        # second condition below triggers if this is running in the thread of the
+        # event loop *during* the call to sync(), i.e., while running
+        # asynchronously
+        if self.asynchronous or getattr(thread_state, "asynchronous", False):
+            if inspect.iscoroutinefunction(func):
+                # directly make awaitable and return is
+                return func(*args, **kwargs)
+            else:
+                # make awaitable which then calls the blocking function
+                return _run_as_coroutine(func, *args, **kwargs)
+        else:
+            if inspect.iscoroutinefunction(func):
+                # run the awaitable on the loop
+                return sync(loop, func, *args, **kwargs)
+            else:
+                # just call the blocking function
+                return func(*args, **kwargs)
+
+    return wrapper
+
+
+def async_wrapper(func):
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def get_loop():
+    """Create a running loop in another thread"""
+    loop = asyncio.new_event_loop()
+    t = threading.Thread(target=loop.run_forever)
+    t.daemon = True
+    t.start()
+    return loop
+
+
+# these methods should be implemented as async by any async-able backend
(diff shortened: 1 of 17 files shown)


martindurant:
I should comment that this PR has already become gnarly, since some refactoring is included here to allow the appearance of batch methods like `get` with multiple inputs, globs or recursive trees. That accounts for the removal of code in spec, which has been consolidated in utils, and tests added in test_utils.

It is the changed to the http implementation that are of most interest here, particularly the async chaining stuff. Note, for example, the [horrible hack](https://github.com/intake/filesystem_spec/pull/310/files?file-filters%5B%5D=.py#diff-01721d9c9819199817bbb92aff4dfe16R21) to determine if a call is happening in an async context or not - this is required because I want to reuse code like expand_paths without making *everything* async and rewriting or duplicating the whole API.

martindurant:
@TomAugspurger @mrocklin : this is a working implementation where tests pass and there are no warnings for HTTP. Unfortunately, it doesn't seem any faster than a sequential get of URLs from independent servers. I also tried having the event loop in a different thread, but that just moves the waiting from select to wait for thread-locks...

fsspec/implementations/http.py
```diff
@@ -110,25 +151,84 @@ def ls(self, url, detail=True, **kwargs):
...
         kwargs = self.kwargs.copy()
-        kwargs["stream"] = True
         try:
-            r = self.session.get(path, **kwargs)
-            r.close()
-            return r.ok
+            r = await self.session.get(path, **kwargs)
+            async with r:
+                return r.status < 400
         except requests.HTTPError:
```
TomAugspurger, line 227: Does the exception caught here need to change?

fsspec/implementations/http.py
```diff
@@ -1,18 +1,54 @@
...
 
 
+def is_running_async():
+    for frame in inspect.stack(context=0):
+        if frame.function == "run_until_complete":
+            return True
+    return False
+
+
+def sync_wrapper(func):
```
TomAugspurger, line 28: To make sure I understand the general idea: the sync version always calls the async implementation? So if I do something like

```python
fs = S3FSFileSystem()
fs.ls("test-bucket/")
```

That will eventually run code on the event loop?

martindurant: The only guarantee is that the original API is sync, i.e., returns concrete results; but it *may* be implemented on the event loop. Some backends (HDFS) won't work well with async, and it will take some time to add async to all those backends that will. 
You could instead say that it always runs something on the event loop, but some functions may block for a while, but I had not imagined it that way.

fsspec/implementations/http.py
martindurant: probably; I should check what aiohttp does when a connection fails completely.

martindurant: probably `aiohttp.ClientError`

fsspec/implementations/http.py
TomAugspurger: What will happen if I'm doing some aysncio stuff on the event loop, and then I call some fsspec stuff from the main thread. Is it possible that some blocking IO ends up happening on the asyncio event loop, which will block the unrelated asyncio stuff I was doing?

martindurant: I *believe* that if you call something in fsspec, then coroutines will still process on the event loop while the caller waits for result; things get more complicated if any of the fsspec coroutines themselves call sync (e.g., via a method defined only on the superclass, which right now has nothing async in it).
Perhaps this could be tested by `cat`ing a large download while a periodic coroutine appnds timestamps to a list?

martindurant: (but you are getting to the root of my async knowledge, exactly why I want other people to look at this code!)

mrocklin: I recommend stealing the approach taken by dask.distributed.  It handles this and a variety of other situations smoothly.

martindurant: You mean https://github.com/dask/distributed/blob/master/distributed/utils.py#L292 ? I don't know what's going on there, but it's essentially what I had in mind. However, shouldn't running in a separate thread be unnecessary? The given loop would, by default, be in the current thread.

martindurant:
@mrocklin , this is more distributed-like, with the loop in a thread, and in limited manual testing, seems to run slower. I don't know what would happen in this version is a coroutine tried to call sync somewhere in the stack.

fsspec/implementations/http.py
```diff
@@ -110,25 +175,86 @@ def ls(self, url, detail=True, **kwargs):
         else:
             return list(sorted(out))
 
-    def cat(self, url):
-        r = self.session.get(url, **self.kwargs)
-        r.raise_for_status()
-        return r.content
+    ls = sync_wrapper(_ls)
```
Carreau, line 178: I got caught a couple of time when implementing a class with both sync and async method, blokgin myself by mistake. Personally I prefer to have a superclass that have no sync method and a superclass (or a utility function) that "syncify" them. That way I'm sure when working on  the superclass that I'm not going to call sync method in sync context.

fsspec/implementations/http.py
Carreau: Just as a note, NJS recently pointed me to https://github.com/oremanj/greenback/ in a [thread](https://trio.discourse.group/t/concurrency-and-trio-as-implmentation-detail-in-blocking-sync-method/302/2) about reentrancy. It might be useful. 

Be careful if your main eventloop is running and the method you implement as "sync" are actually letting the event loop running you may actually break some assumptions.


When needing to "nest" eventloop, if it is a possibility, then you can also use a different runner (for example trio from within asyncio). That will really make your trio-syncified-method really sync (and blocking the outer eventloop) without the need to use threads.

mrocklin: The Dask solution was designed to support simultaneous use in both async and sync situations, which I think was the situation that @TomAugspurger was getting at.

martindurant:
@mrocklin , the new module `asyn` contains much of what we discussed. The evil is mostly rooted [here](https://github.com/intake/filesystem_spec/pull/310/files#diff-dc33a6f87f2ae7e9d1e2361e99f61491R192), which adds sync and async versions of methods as needed, but only for file-systems that have a class attribute `async_impl=True`, only HTTP for now.
I would be interested if you have any suggestions on how to *test* async methods - I assume I can copy from the likes of gen.test in distributed? Currently all of the test functions call the sync methods which might call async internally.

Anyone else watching this thread, this would be a good time to settle the user API. Currently, the batch methods are: rm, cp, mv, get (download), put (upload), mcat (contents of files). Suggests perhaps the latter should just be cat(recursive=), and corresponding cat_file. Get/put/cat could be concurrent per file, and/or chunks of single file(s) - which all need kwarg names.

Opening a file for read or write is complicated; read-a-chunk could be async the parts of writing could be async (init upload, send chunk, finalise), but not so obviously the read()write() methods, since obviously not all operations can be done in arbitrary order, and there is state (file location, buffer). This problem can all be deferred, if get/put doesn't follow the default code, which does an open().

fsspec/asyn.py
```diff
@@ -0,0 +1,206 @@
...
+        lpaths = other_paths(rpaths, lpath)
+        await asyncio.gather(
+            *[
+                self._get_file(rpath, lpath, **kwargs)
+                for lpath, rpath in zip(lpaths, rpaths)
+            ]
+        )
+
+
+def make_sync_methods(obj):
```
Carreau, line 192: So this _also_ create async methods right ? 

1) Do you think a better like "mirror" methods would be better; 
and
2) do you think some `obj` would implement both the sync and async in which case IIUC this would overwrite the custom impl ?

martindurant: Agreed, the name could be better (obviously the func first did only one of the jobs...).
Yes, you would squash the sync implementation; it could apply only if the method being overwritten is missing of the same as the AsyncFileSystem class.

PR #319, opened 2020-06-15
OpenFile newline in reduce

### LINKED ISSUE #318: Pickling OpenFile object ignores newline attribute
When pickling `OpenFile` objects, the `newline` attribute is not persisted. This causes issues on Windows, where a value of `None` inserts an additional `\r` when using the file with a stream already containing correct line termination.

```python
import fsspec
import pickle


def main():
    test = fsspec.open(__file__, newline=b"")
    
    pickled = pickle.dumps(test)
    restored = pickle.loads(pickled)
    
    assert test.newline == restored.newline
    return


if __name__ == '__main__':
    main()
```

Is there an underlying reason as to why this is not pickled, or was it simply missed during the addition of the `newline` attribute?

martindurant:
Fixes #318

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -84,7 +84,7 @@ class OpenFile(object):
                 self.compression,
                 self.encoding,
                 self.errors,
-                self.newline
+                self.newline,
             ),
         )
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks!

PR #320, opened 2020-06-16
allow Path object in open

martindurant:
Ref: https://github.com/dask/dask/issues/6327


TomAugspurger:
Thanks for the quick fix.

PR #327, opened 2020-06-18
Fix local file ops

### LINKED ISSUE #326: local filesystem isn't consistent
- If you set `auto_mkdir` to True for a local filesystem, it won't actually listen to this, for example when copying files, it just invokes `shutil.copyfile`
- if you pass an uri with the scheme included in it to certain functions, the local filesystem will give an error.

martindurant:
Always strip protocol, and respect auto_mkdir when copying

Fixes #326


pvanderlinden:
The function `touch` also needs to listen to auto_makedir

martindurant:
Hm, CI not running. All is good locally, so merging.

Issue #325, opened 2020-06-17
Clearing the cache / Question 2/2


gmaze:
Hi folks !

This is 2/2 questions wrt to fsspec cache handling when we want to clear it.

## Context:
I develop a python library to work with ocean data collected using a specific type of instruments (Argo floats): https://github.com/euroargodev/argopy
Data can be fetched from local files or from an erddap server online.

I recently tried to use fsspec to implement file/resource management and ease handling of cached data.
I ended up wrapping ``file``, ``memory`` and ``http`` fsspec file systems into [classes](https://github.com/euroargodev/argopy/blob/master/argopy/stores/fsspec_wrappers.py) that managed custom methods like open_dataset or open_dataframe.

## Question 2:
Since the fsspec method wasn't working the way I expected (see issue #324), I implemented my own method in the [custom file system wrapper](https://github.com/euroargodev/argopy/blob/master/argopy/stores/fsspec_wrappers.py). It works like this:
- Every time a call is made to ``open``, I register the opened path in an internal list (``self.cache_registry``)
- When users call on the ``clear_cache()`` method, I load the cache and remove the corresponding entry from the pickle ``cache`` file and I also delete the corresponding hash file.

This goes basically like this:
```python
import pickle
import tempfile
import shutil
import fsspec
import os

class argo_store():
    protocol = 'file'
    
    def __init__(self, cache: bool = False, cachedir: str = "", **kw):
        """ Create a file storage system for argopy

            Parameters
            ----------
            cache : bool (False)
            cachedir : str (from OPTIONS)

        """
        self.cache = cache
        self.cachedir = OPTIONS['cachedir'] if cachedir == '' else cachedir
        self.cache_registry = []  # Will hold uri cached by this store instance
        if not self.cache:
            self.fs = fsspec.filesystem(self.protocol, **kw)
        else:
            self.fs = fsspec.filesystem("filecache",
                                        target_protocol=self.protocol,
                                        target_options={'simple_links': True},
                                        cache_storage=self.cachedir,
                                        expiry_time=86400, cache_check=10, **kw)

    def store_path(self, uri):
        if not uri.startswith(self.fs.target_protocol):
            path = self.fs.target_protocol + "://" + uri
        else:
            path = uri
        return path

    def register(self, uri):
        """ Keep track of files open with this instance """
        if self.cache:
            self.cache_registry.append(self.store_path(uri))
        
    def _clear_cache_item(self, uri):
[...]

Issue #322, opened 2020-06-17
OSError: [Errno 18] Invalid cross-device link error on readthedocs


gmaze:
Hi folks,
I develop a python library to work with ocean data collected using a specific type of instruments (Argo floats): https://github.com/euroargodev/argopy
Data can be fetched from local files or from an erddap server online.

I recently tried to use fsspec to implement file/resource management and ease handling of cached data.
I ended up wrapping ``file``, ``memory`` and ``http`` fsspec file systems into [classes](https://github.com/euroargodev/argopy/blob/master/argopy/stores/fsspec_wrappers.py) that managed custom methods like open_dataset or open_dataframe.

I worked nicely up to the point where the new master branch was pushed to readthedocs, which failed because of a:
```bash
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmp1twhthrp' -> 'mycache_folder/cache'
```

The error is raised when readthedocs executes [a notebook where I demonstrate the use of the caching system of **argopy**](https://github.com/euroargodev/argopy/blob/master/docs/performances.ipynb).
The error is raised by the sspec/implementations/cached.py ``save_cache(self)`` method in the line where the new cache replace the old one: ``os.replace(fn2, fn)``.

This is apparently raised because ``os`` expect files to be from the same file system.

Is this only related to readthedocs ? and if you have any ideas about how to fix this, that would be great ! One possibility is by using ``shutil.move`` instead of ``os.rename``.

[The RTD log build with the error can be seen here](https://readthedocs.org/projects/argopy/builds/11273728/).
The interesting peace is here (sorry for the RTD character mess):
```bash
[0;32m~/checkouts/readthedocs.org/user_builds/argopy/envs/latest/lib/python3.6/site-packages/fsspec/implementations/cached.py[0m in [0;36msave_cache[0;34m(self)[0m
[1;32m    157[0m         [0;32mwith[0m [0mopen[0m[0;34m([0m[0mfn2[0m[0;34m,[0m [0;34m"wb"[0m[0;34m)[0m [0;32mas[0m [0mf[0m[0;34m:[0m[0;34m[0m[0;34m[0m[0m
[1;32m    158[0m             [0mpickle[0m[0;34m.[0m[0mdump[0m[0;34m([0m[0mcache[0m[0;34m,[0m [0mf[0m[0;34m)[0m[0;34m[0m[0;34m[0m[0m
[...]

PR #328, opened 2020-06-18
start cache clear ops

martindurant:
POC for #325 
@gmaze , this is not quite what you need, but I though would be useful for the general user. What do you think?

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -3,7 +3,11 @@ import pickle
 import logging
 import os
 import hashlib
+<<<<<<< Updated upstream
 from shutil import move
+=======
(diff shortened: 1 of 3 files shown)


gmaze:
awesome !
This should provide a minimal way of clearing the cache

Is the ``os.remove(fn)`` ok and won't trigger an error like #322 ?

martindurant:
> Is the os.remove(fn) ok and won't trigger an error like #322 ?

If the cache dir is writable, we should be able to do this.

gmaze:
Thanks a lot @martindurant for being so fast with this !
I'll be able to use this in https://github.com/euroargodev/argopy right away

PR #331, opened 2020-06-22
WIP: zip `ls("")` output is different from ls("/")

sodre:
The same is true for any path that is a directory as well.

My question, is that the expected behavior?

### DIFF
--- a/fsspec/implementations/tests/test_zip.py
+++ b/fsspec/implementations/tests/test_zip.py
@@ -23,7 +23,7 @@ def tempzip(data={}):
             pass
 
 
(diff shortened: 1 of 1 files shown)


sodre:
@martindurant, would you be able to confirm that the expected behavior of ZipFileSytem "ls" command is correct? I wrote a small unit test to capture the `ls('/') vs ls('')`

martindurant:
The zip implementation defines `root_marker = ""`, meaning that paths are *not* expected to have the initial "/", and `_strip_protocol` includes `.lstrip("/")`, which ought to deal with it. The method `ls` should call this.

sodre:
closing in favor of #332

Issue #297, opened 2020-05-14
tar backend?


martindurant:
A TAR file is like a file-system in that it contains directories and files. These are simply concatenated end-to-end, with short headers, but after a single pass of the file, the whole set of files and their offsets can be known (and maybe stored somewhere). Indeed, since each header contains the size of the file, you can skip the data bits altogether on this pass.

*However*, tar files are often stored compressed (gz, bz2), so to seek to a particular location to read a contained file amounts to scanning through all the compressed file up to that point. Any tar backend would only work well with uncompressed files.

cc @kmpaul

martindurant:
This is as opposed to ZIP files, which are compressed file-wise and contain an index, so that each file can be accessed immediately.

kmpaul:
@martindurant Thanks for starting this!  I think this would be interesting to explore.  I think you are correct that a compressed tarball would not work efficiently; you would need to uncompress it before you could use it as a filesystem.

martindurant:
> you would need to uncompress it before you could use it as a filesystem

ideally uncompress it as you are receiving it and store the expanded version!

kmpaul:
> ideally uncompress it as you are receiving it and store the expanded version!

Yes!  Indeed that would be nicer.  That might be a good default, especially since most compressed tarballs have "double compression" (i.e., compression of the tarball containing compressed files), meaning that the expansion of the compressed tarball shouldn't take up too much more space than the original.

martindurant:
Indexing:
```
tf = tarfile.TarFile('file.tar')
out = {ti.get_info()['name']:  (ti.get_info(), ti.offset_data) for ti in tf}
```
(I assume python's tar is smart enough here to seek past the data segments).

kmpaul:
@martindurant Should I just start putting together a draft PR for this?

martindurant:
If you have the time, that would be great. It is up to you whether you make a way to persist the file contents index.
Note that the input to `TarFile` can be a file-like object, so this can work on remote TAR files too, except that random-access would be slow (but maybe good enough).

kmpaul:
Ok.  I may have to start this next week, but I'll put together a draft PR.

kmpaul:
Oooh!  Thanks, @martindurant!

zklaus:
I think this issue has been addressed and can be closed, no?

martindurant:
Yes, closing.
[...]

PR #332, opened 2020-06-23
Use dir_cache to return ZipFilesystem info()

sodre:
This new implementation gives us a O(1) lookup for any calls to zip.info(path).
The previous version was O(n), even after dir_cache was computed.

For example, one can now open a large zipfile stored in s3 very efficiently using the following
code:

```python
In [1]: import fsspec                                                                                                                                                                                              

In [2]: %%time  
   ...: of = fsspec.open_files("zip://**.jpg::blockcache://fast-ai-coco/unlabeled2017.zip::s3://", s3={"anon":True}) 
   ...: len(of) 
   ...:  
   ...:                                                                                                                                                                                                            
CPU times: user 2.65 s, sys: 117 ms, total: 2.77 s
Wall time: 4.69 s
Out[2]: 123403
```

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -49,7 +49,7 @@ class MMapCache(BaseCache):
 
     def __init__(self, blocksize, fetcher, size, location=None, blocks=None):
         super().__init__(blocksize, fetcher, size)
-        self.blocks = set() if blocks is None else blocks
+        self.blocks = set() if blocks is None else set(blocks)
         self.location = location
         self.cache = self._makefile()
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -49,7 +49,7 @@ class MMapCache(BaseCache):
 
     def __init__(self, blocksize, fetcher, size, location=None, blocks=None):
         super().__init__(blocksize, fetcher, size)
-        self.blocks = set() if blocks is None else set(blocks)
+        self.blocks = set() if blocks is None else blocks
         self.location = location
         self.cache = self._makefile()
 
--- a/fsspec/implementations/tests/test_zip.py
+++ b/fsspec/implementations/tests/test_zip.py
@@ -47,3 +47,25 @@ def test_pickle():
         fs = fsspec.get_filesystem_class("zip")(fo=z)
         fs2 = pickle.loads(pickle.dumps(fs))
         assert fs2.cat("b") == b"hello"
+
+
+def test_info():
+    with tempzip(data) as z:
+        fs = fsspec.get_filesystem_class("zip")(fo=z)
+
+        with pytest.raises(FileNotFoundError):
+            fs.info("i-do-not-exist")
+
+        # Iterate over all directories
+        for d in {os.path.dirname(f) for f in data.keys()}:
+            d_info = fs.info(d)
+            assert d_info["type"] == "directory"
+            assert d_info["size"] == 0
+            assert d_info["name"] == (f"{d}/" if d != fs.root_marker else fs.root_marker)
+
+        # Iterate over all files
+        for f, v in data.items():
+            f_info = fs.info(f)
+            assert f_info["type"] == "file"
+            assert f_info["size"] == len(v)
+            assert f_info["name"] == f
--- a/fsspec/implementations/tests/test_zip.py
+++ b/fsspec/implementations/tests/test_zip.py
@@ -49,6 +49,23 @@ def test_pickle():
         assert fs2.cat("b") == b"hello"
 
 
+def test_all_dirnames():
+    with tempzip() as z:
+        fs = fsspec.get_filesystem_class("zip")(fo=z)
+
+        # fx are files, dx are a directories
+        assert fs._all_dirnames(["f1"]) == {""}
+        assert fs._all_dirnames(["f1", "f2"]) == {""}
+        assert fs._all_dirnames(["f1", "f2", "d1/f1"]) == {"", "d1/"}
+        assert fs._all_dirnames(["f1", "f2", "d1/f1", "d1/f2"]) == {"", "d1/"}
+        assert fs._all_dirnames(["f1", "f2", "d1/f1", "d1/f2", "d2/f1"]) == {
+            "",
+            "d1/",
+            "d2/",
+        }
+        assert fs._all_dirnames(["d1/d1/d1/f1"]) == {"", "d1/", "d1/d1/", "d1/d1/d1/"}
+
+
 @pytest.mark.parametrize("implementation", ["super", "cache"])
 def test_info(implementation):
     with tempzip(data) as z:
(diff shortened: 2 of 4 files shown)


martindurant:
Thanks for this! Goes to show that the dev's assumptions are not necessarily reflective of reality. 123k files in a single 19GB zip??

I am surprised that directory listing is slow, even though your benchmarks prove your case. I would have thought that s3's reading cache and/or the local copy would be fast enough. Do you know of any documentation describing the directory lookup for zip?

I will give the code a thorough look soon. The addition of tests and passing of existing tests is encouraging.

> of = fsspec.open_files("zip://**.jpg::blockcache://fast-ai-coco/unlabeled2017.zip::s3://", s3={"anon":True}) 

Shouldn't this be

> of = fsspec.open_files("zip://**.jpg::blockcache::s3://fast-ai-coco/unlabeled2017.zip", s3={"anon":True})

fsspec/implementations/tests/test_zip.py
```diff
@@ -30,6 +30,8 @@ def test_empty():
     with tempzip() as z:
         fs = fsspec.get_filesystem_class("zip")(fo=z)
         assert fs.find("") == []
+        with pytest.raises(FileNotFoundError):
+            fs.info("")
```
martindurant, line 34: One could argue that the root always exists and is a directory-like thing, since you can list it.

fsspec/implementations/tests/test_zip.py
```diff
@@ -47,3 +49,100 @@ def test_pickle():
         fs = fsspec.get_filesystem_class("zip")(fo=z)
         fs2 = pickle.loads(pickle.dumps(fs))
         assert fs2.cat("b") == b"hello"
+
+
+def test_all_dirnames():
+    with tempzip() as z:
+        fs = fsspec.get_filesystem_class("zip")(fo=z)
```
martindurant, line 56: same as `fsspec.filesystem("zip", fo=z)`

fsspec/implementations/zip.py
```diff
@@ -55,6 +56,9 @@ def __init__(
         self.block_size = block_size
         self.dir_cache = None
 
+        if kwargs.get("_info_implementation") == "base":
```
martindurant, line 59: The plan is to remove this, as soon as the case is made for the new functionality, right?

fsspec/implementations/zip.py
```diff
@@ -75,6 +82,16 @@ def _get_dirs(self):
                 )
                 self.dir_cache[f["name"]] = f
 
+    def info(self, path, **kwargs):
+        self._get_dirs()
+        path = self._strip_protocol(path)
+        if path in self.dir_cache:
+            return self.dir_cache[path]
+        elif path + "/" in self.dir_cache:
```
martindurant, line 90: These are directories?

fsspec/implementations/zip.py
```diff
@@ -133,3 +141,16 @@ def _open(
...
+        """ Returns *all* directory names for each path in paths, including intermediate ones.
+
+        Parameters
+        ----------
+        paths: Iterable of path strings
+        """
+        if len(paths) == 0:
+            return set()
+
+        dirnames = {os.path.dirname(path) for path in paths} - {f"{self.root_marker}"}
```
martindurant, line 155: `self.root_marker` is a string anyway, no need for format.

fsspec/implementations/zip.py
```diff
@@ -133,3 +141,16 @@ def _open(
 
     def ukey(self, path):
         return tokenize(path, self.fo, self.protocol)
+
+    def _all_dirnames(self, paths):
```
martindurant, line 145: Might want to extract this to a separate function, so as not to pollute the FS namespace. Since it's marked private, maybe that's unimportant.

fsspec/implementations/zip.py
sodre: Yes. The `dir_cache` stores directories with `/` appended to them.

fsspec/implementations/zip.py
sodre: I originally had this function as part of test_zip.py, because I needed a way to generate all the intermediate directories. Then I realized that we needed the same behavior for the class as well. 

I am happy to move it where you prefer. Here are the options I can think:
   1. Private at ZipFileSystem (as it is)
   1. Keep it private but move it up to AbstractFileSystem, and make the "root_maker" a function parameter.
   1. Make it public and place it on utils.py

fsspec/implementations/zip.py
sodre: I got a little too carried away on `f"{}"-syntax`. It is fixed...

fsspec/caching.py
```diff
@@ -49,7 +49,7 @@ class MMapCache(BaseCache):
 
     def __init__(self, blocksize, fetcher, size, location=None, blocks=None):
         super().__init__(blocksize, fetcher, size)
-        self.blocks = set() if blocks is None else blocks
+        self.blocks = set() if blocks is None else set(blocks)
```
sodre, line 52: This fixes #333 on my local tests, but I did not have time to write a proper unit test for it.

The test needs to:
   - Open a block cache
   - Partially read the remote content.
   - Close the block cache
   - Reopen the block cache.
   - Ensure no exceptions are raised.

Not sure how to write that test yet though :(

sodre:
> Thanks for this! Goes to show that the dev's assumptions are not necessarily reflective of reality. 123k files in a single 19GB zip??
>

I know! That is also how they release the files directly in the COCO website. In a sense, I am glad they are  Zips and not "tgz" as you pointed out the issues in #297. 
 
> I am surprised that directory listing is slow, even though your benchmarks prove your case. I would have thought that s3's reading cache and/or the local copy would be fast enough. Do you know of any documentation describing the directory lookup for zip?
> 

The reason looks ups are slow is because it relies on calling "ls", which is quite complicated in the ZipFileSystem case.

> I will give the code a thorough look soon. The addition of tests and passing of existing tests is encouraging.
> 
> > of = fsspec.open_files("zip://**.jpg::blockcache://fast-ai-coco/unlabeled2017.zip::s3://", s3={"anon":True})
> 
> Shouldn't this be
> 
> > of = fsspec.open_files("zip://**.jpg::blockcache::s3://fast-ai-coco/unlabeled2017.zip", s3={"anon":True})

I think there is still a bug in the URL unpacking. I looked at the code but could not quite figure out why it was not working.

fsspec/implementations/tests/test_zip.py
martindurant: Hm, I'd say
- yes, ls("") should of course work
- therefore, info("") should, I suppose, return a directory - but after some thought, I am less certain. One would expect isdir("") == True, for example, but we don't have any permissions or other info for this entry as we do for any other dir in the archive
- find("") should *not* include "" itself, even for withdirs=True, you only get children

fsspec/implementations/tests/test_zip.py
```diff
@@ -47,3 +49,100 @@ def test_pickle():
...
+        assert fs._all_dirnames(["f1", "d1/f1", "d2/f1"]) == {"d1", "d2"}
+        assert fs._all_dirnames(["d1/d1/d1/f1"]) == {"d1", "d1/d1", "d1/d1/d1"}
+
+
+def test_ls():
+    with tempzip(data) as z:
+        lhs = fsspec.get_filesystem_class("zip")(fo=z)
+
+        assert lhs.ls("") == ["a", "b", "deeply/"]
+        assert lhs.ls("/") == lhs.ls("")
```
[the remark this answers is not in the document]
martindurant: (OK, but the handling of paths in s3fs has been a constant headache, since it is backed by a simulated filesystem, and may not be the best example to follow!)

fsspec/implementations/zip.py
martindurant: In the end, I say leave it here. We can always move it later if need be.

martindurant:
OK, so please remove the flag and make the new method the only method.
We can also take this opportunity fo fix the URL bug you mention above, or postpone it for a new issue.

fsspec/implementations/tests/test_zip.py
sodre: @martindurant, I left the behavior as it was before. I think if we need that to change we can open a separate issue/PR.

sodre:
@martindurant, this PR should be ready for a final review and merge.

For my use-case I still need to figure out why #333 is happening. I  will post additional comments on that PR.

martindurant:
Why "Revert "Force MMapCache blocks to "set" type" ?

sodre:
> Why "Revert "Force MMapCache blocks to "set" type" ?

That commit leaked into this PR when I was working on #333. With my latest commit to #333, we can actually replicate the bug, so I'll add the fix there.

Please feel free to squash the commit if you prefer.

martindurant:
OK, I don't mind either way.
This is going in.

PR #333, opened 2020-06-25
Error reopening a blockcache

sodre:
I am having a little bit of a problem reopening a url that has a block cache in the middle.
Here is the exception I am getting:

It looks like a simple change from "add" to "append" fixes the problem, but I am not sure if it creates issues in other places.

The code that generate the exception is [here](https://github.com/sodre/intake-coco2017/blob/master/main.ipynb). To reproduce, run the notebook once, terminate the kernel, and start it back again. The exception will show up the second time.

<details>

```python-traceback
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/.conda/envs/intake-coco2017/lib/python3.8/site-packages/ipywidgets/widgets/interaction.py in update(self, *args)
    254                     value = widget.get_interact_value()
    255                     self.kwargs[widget._kwarg] = value
--> 256                 self.result = self.f(**self.kwargs)
    257                 show_inline_matplotlib_plots()
    258                 if self.auto_display and self.result is not None:

<ipython-input-3-6a5c22e1a584> in show(dataset, id)
     13 @interact(dataset=dataset_widget, id=partition_widget)
     14 def show(dataset, id):
---> 15     img = dataset.read_partition((id,))[0]
     16     io.imshow(img)

~/.conda/envs/intake-coco2017/lib/python3.8/site-packages/intake_xarray/base.py in read_partition(self, i)
     49         """
     50         import numpy as np
---> 51         self._load_metadata()
     52         if not isinstance(i, (tuple, list)):
     53             raise TypeError('For Xarray sources, must specify partition as '

~/.conda/envs/intake-coco2017/lib/python3.8/site-packages/intake/source/base.py in _load_metadata(self)
    124         """load metadata only if needed"""
    125         if self._schema is None:
--> 126             self._schema = self._get_schema()
    127             self.datashape = self._schema.datashape
    128             self.dtype = self._schema.dtype

~/.conda/envs/intake-coco2017/lib/python3.8/site-packages/intake_xarray/image.py in _get_schema(self)
    352 
    353         if self._ds is None:
--> 354             self._open_dataset()
    355 
    356             # convert to dataset for serialization

~/.conda/envs/intake-coco2017/lib/python3.8/site-packages/intake_xarray/image.py in _open_dataset(self)
    335         from dask.bytes import open_files
    336 
--> 337         files = open_files(self.urlpath, **self.storage_options)
    338         if len(files) == 0:
    339             raise Exception("No files found at {}".format(self.urlpath))

[...]

PR #335, opened 2020-06-25
Reopen the blockcache issue

sodre:
PR #333 was accidentally closed when #332 was merged. Please look at #333 for the comments.

### DIFF
--- a/fsspec/implementations/tests/test_cached.py
+++ b/fsspec/implementations/tests/test_cached.py
@@ -77,8 +77,7 @@ def test_blockcache_workflow(ftp_writable, tmp_path, disable_fs_caching):
     with fs.open("/out", block_size=5) as f:
         assert f.read(5) == b"test\n"
         f.seek(30)
-        with pytest.raises(AttributeError):
-            assert f.read(5) == b"test\n"
+        assert f.read(5) == b"test\n"
 
 
 @pytest.mark.parametrize("impl", ["filecache", "blockcache"])
--- a/fsspec/implementations/tests/test_cached.py
+++ b/fsspec/implementations/tests/test_cached.py
@@ -37,7 +37,15 @@ def test_idempotent():
     assert fs3.storage == fs.storage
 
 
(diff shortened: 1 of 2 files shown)


martindurant:
OK, now I see it fails. Does `set()` fix it, as you had before?
This is separate from the chaining URL syntax issue, right?

sodre:
> OK, now I see it fails. Does `set()` fix it, as you had before?
> This is separate from the chaining URL syntax issue, right?

This is separate from the chaining url, I don't know yet how to to fix that problem.

The underlying root cause are these problems:
https://github.com/intake/filesystem_spec/blob/63c421861005644b802c9a5ce9b891a683220c70/fsspec/implementations/cached.py#L155-L156

There are two possible fixes:
   - It can go into MMap constructor like I did before.
   - It can go into CachingFileSystem during load_cache. I prefer to put the fix here because save_cache is the one choosing to serialize `blocks` using list.

sodre:
@martindurant, I did some git blame digging and I found the [commit code.py:L50-L54](https://github.com/intake/filesystem_spec/commit/d56fc22298b34d60bd3fddcbb822688a4b9b12df) that broke the reopening of the blockcache.

It looks like there was a plan to store a list of block ranges instead of individual blocks, do you remember?

sodre:
@martindurant, this PR is ready for review.

fsspec/implementations/tests/test_cached.py
```diff
@@ -37,6 +37,49 @@ def test_idempotent():
     assert fs3.storage == fs.storage
 
 
+@pytest.fixture()
+def disable_fs_caching():
```
martindurant, line 41: You probably just wanted `fs = FTPFileSystem(host, port, user, pw, skip_instance_cache=True)` instead of this fixture

sodre: Sorry, I had not seen that option, I needed to use it only for the blockcache. If the tests pass, it should be ready to merge.

Issue #334, opened 2020-06-24
URL Unpacking does not work as expected when chaining 3(+) FileSystems


sodre:
Currently URL unpacking expects the following syntax if we want to open a zip file stored in S3 with a blockcache in the middle. The use case in question is a 19GB file for the Coco dataset.

This is the syntax that works:

> of = fsspec.open_files("zip://**.jpg::blockcache://fast-ai-coco/unlabeled2017.zip::s3://", s3={"anon":True}) 

Shouldn't this be:

> of = fsspec.open_files("zip://**.jpg::blockcache::s3://fast-ai-coco/unlabeled2017.zip", s3={"anon":True})

_Originally posted by @martindurant in https://github.com/intake/filesystem_spec/pull/332#issuecomment-648167709_

martindurant:
I wonder if you could make an example with fewer stages and only local files (so we can include in tests). 
Do you see the problem with a urls like
- "zip://*::simplecache::file://localfile.zip"
- "zip://inner_file::simplecache::file://localfile.zip"
- "zip://*::simplecache://::file://localfile.zip"

sodre:
I will take a look at creating the test case later tonight.

martindurant:
This can be closed?

PR #336, opened 2020-06-26
Reproduce #334 in a test case

sodre:
This PR should reproduce issue #334 without having to rely on an s3 connection to a large 19GB file.

### DIFF
--- a/fsspec/tests/test_api.py
+++ b/fsspec/tests/test_api.py
@@ -3,6 +3,7 @@
 import contextlib
 import os
 import pickle
+import sys
 import tempfile
 
 import pytest
@@ -197,6 +198,7 @@ def test_multilevel_chained_fs():
             assert f.read().decode("utf-8") == f.name
 
 
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="no seek in old zipfile")
 def test_multilevel_chained_fs_zip_zip_file():
     """This test reproduces intake/filesystem_spec#334"""
     import zipfile
--- a/fsspec/tests/test_api.py
+++ b/fsspec/tests/test_api.py
@@ -211,6 +211,7 @@ def test_multilevel_chained_fs_zip_zip_file():
         with open_file as f:
             assert f.read().decode("utf-8") == f.name
 
+
 def test_chained_equivalent():
     d1 = tempfile.mkdtemp()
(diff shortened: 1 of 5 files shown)


fsspec/tests/test_api.py
```diff
@@ -158,6 +161,57 @@ def test_chained_fs():
...
+        # filename, content
+        z.writestr("foo.txt", "foo.txt")
+        z.writestr("bar.txt", "bar.txt")
+
+    with zipfile.ZipFile(f2, mode="w") as z:
+        with open(f1, "rb") as f:
+            z.writestr("f1.zip", f.read())
+
+    # We expected this to be the correct syntax
+    of = fsspec.open_files(f"zip://*.txt::zip://f1.zip::file://{f2}")
```
sodre, line 207: This example works without any issues. This indicates to me that the issue is in the blockcache code base and not in the url parsing/unchaining.

martindurant: In this example you have some txt files in f1.zip, which in turn is within f2.zip? That is how I read that URL.

sodre: Yes, that is correct.

fsspec/tests/test_api.py
```diff
@@ -158,6 +161,57 @@ def test_chained_fs():
...
+
+    assert len(of) == 2
+    for open_file in of:
+        with open_file as f:
+            assert f.read().decode("utf-8") == f.name
+
+
```
sodre, line 214:
```suggestion

```

fsspec/tests/test_api.py
```diff
@@ -158,6 +161,56 @@ def test_chained_fs():
...
+            z.writestr("f1.zip", f.read())
+
+    # We expected this to be the correct syntax
+    of = fsspec.open_files(f"zip://*.txt::zip://f1.zip::file://{f2}")
+
+    assert len(of) == 2
+    for open_file in of:
+        with open_file as f:
+            assert f.read().decode("utf-8") == f.name
+
```
sodre, line 213:
```suggestion

```

martindurant:
OK with you, @sodre ?

sodre:
Hi @martindurant, it looks like it works now. I am not necessarily proud of the solution, but it is as much as I can do at this point.

Issue #306, opened 2020-05-28
samba backend


martindurant:
https://pysmb.readthedocs.io/en/latest/api/smb_SMBConnection.html appears to give a very thorough understandable interface to Samba for us to wrap, with random access. We can easily test these using one of the docker-based samba servers.

telamonian:
I support this effort 100%. There's a dearth of vfs backends in general that support smb, and AFAIK there's only one [existing python smb vfs](https://github.com/althonos/fs.smbfs). A new solid python implementation would be very welcome.

WIth [jupyter-fs](https://github.com/jpmorganchase/jupyter-fs), I'm aiming to build a tool that sits at least one level above vfs backends in the stack (and so is backend agnostic). So far, all of the integrated backends are pyfilesystem-based, including https://github.com/althonos/fs.smbfs. I'll share what I've learned in the effort about smb + vfs.

The smb use case that people care about is the native Windows share client/server. Windows share is historically why the CIFS standard exists in the first place (and it seems like `MS-CIFS` is generally treated as the "real" standard). Windows is the source of all problems:

- In terms of getting started, it's very tricky to initialize a working client/server and test share on the Windows containers available on the public CI services. You

- `pysmb` doesn't cover all needed functionality. In turn, this is because the patchwork of standards on which `pysmb` is based don't cover eveything. The biggest pain point here that I've run into is that there is no specification/support for translating posix <=> windows file permissions.

- Focusing on the permissions issue, every vfs backend for smb has to roll their own "universal enough" posix/windows permissions handling. Instead, there's a [13 year old unaccepted draft spec](https://tools.ietf.org/id/draft-ietf-nfsv4-acl-mapping-05.txt) and (theoretically) a reference implementation somewhere in the code of the OSS samba server (see notes under "Interoperability Strategies" [here](https://wiki.linux-nfs.org/wiki/index.php/ACLs)).

juarezr:
The package pysmb only handles SMB protocol v2.

The PR #340 has a implementation using the [smbprotocol](https://github.com/jborean93/smbprotocol) package.

PR #340, opened 2020-06-29
handle smb:// protocol using smbprotocol package

juarezr:
This PR add support for reading and writing files and folders in Windows and Samba shares using the [smbprotocol](https://github.com/jborean93/smbprotocol) package.

Included testing using a docker image with samba installed.

### DIFF
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -6,13 +6,14 @@ Windows Samba network shares by using package smbprotocol
 
 from stat import S_ISDIR, S_ISLNK
 import datetime
-import types
 import uuid
 
 import smbclient
 
 from .. import AbstractFileSystem
 from ..utils import infer_storage_options
+
+# ! pylint: disable=bad-continuation
 
 
 class SMBFileSystem(AbstractFileSystem):
@@ -188,26 +189,14 @@ class SMBFileSystem(AbstractFileSystem):
     ):
         """
         block_size: int or None
-            If 0, no buffering, if 1, line buffering, if >1, buffer that many
-            bytes, if None use default from paramiko.
+            If 0, no buffering, 1, line buffering, >1, buffer that many bytes
         """
-        buffering = block_size if block_size is not None and block_size >= 0 else -1
+        bls = block_size if block_size is not None and block_size >= 0 else -1
         wpath = _as_unc_path(self.host, path)
-        if autocommit is False:
-            # writes to temporary file, move on commit
-            # TODO: use transaction support in SMB protocol
-            share = path.split("/")[1]
-            path2 = "/{}{}/{}".format(share, self.temppath, uuid.uuid4())
-            wpath2 = _as_unc_path(self.host, path2)
-            smbf = smbclient.open_file(wpath2, mode, buffering=buffering, **kwargs)
-            smbf.temppath = path2
-            smbf.targetpath = path
-            smbf.fs = self
-            smbf.commit = types.MethodType(_commit_a_file, smbf)
-            smbf.discard = types.MethodType(_discard_a_file, smbf)
-        else:
-            smbf = smbclient.open_file(wpath, mode, buffering=buffering, **kwargs)
-        return smbf
+        if "w" in mode and autocommit is False:
+            temp = _as_temp_path(self.host, path, self.temppath)
+            return SMBFileOpener(wpath, temp, mode, block_size=bls, **kwargs)
+        return smbclient.open_file(wpath, mode, buffering=bls, **kwargs)
 
     def copy(self, path1, path2, **kwargs):
         """ Copy within two locations in the same filesystem"""
@@ -230,17 +219,16 @@ class SMBFileSystem(AbstractFileSystem):
         smbclient.rename(wpath1, wpath2, **kwargs)
 
 
-def _commit_a_file(self):
-    self.fs.mv(self.temppath, self.targetpath)
-
-
-def _discard_a_file(self):
-    self.fs.rm(self.temppath)
-
-
 def _as_unc_path(host, path):
     rpath = path.replace("/", "\\")
     unc = "\\\\{}{}".format(host, rpath)
+    return unc
+
+
+def _as_temp_path(host, path, temppath):
+    share = path.split("/")[1]
+    temp_file = "/{}{}/{}".format(share, temppath, uuid.uuid4())
+    unc = _as_unc_path(host, temp_file)
     return unc
 
 
@@ -249,3 +237,49 @@ def _share_has_path(path):
     if path.endswith("/"):
         return parts > 2
     return parts > 1
+
+
+class SMBFileOpener(object):
+    """writes to remote temporary file, move on commit"""
+
+    def __init__(self, path, temp, mode, block_size=-1, **kwargs):
+        self.path = path
+        self.temp = temp
+        self.mode = mode
+        self.block_size = block_size
+        self.kwargs = kwargs
+        self.smbfile = None
+        self._incontext = False
+        self._open()
+
+    def _open(self):
+        if self.smbfile is None or self.smbfile.closed:
+            self.smbfile = smbclient.open_file(
+                self.temp, self.mode, buffering=self.block_size, **self.kwargs
+            )
+
+    def commit(self):
+        """Move temp file to definitive on success."""
+        # TODO: use transaction support in SMB protocol
+        smbclient.replace(self.temp, self.path)
+
+    def discard(self):
+        """Remove the temp file on failure."""
+        smbclient.remove(self.temp)
+
+    def __fspath__(self):
+        return self.path
+
+    def __iter__(self):
+        return self.smbfile.__iter__()
+
+    def __getattr__(self, item):
+        return getattr(self.smbfile, item)
+
+    def __enter__(self):
+        self._incontext = True
+        return self.smbfile.__enter__()
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._incontext = False
+        self.smbfile.__exit__(exc_type, exc_value, traceback)
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -100,18 +100,6 @@ class SMBFileSystem(AbstractFileSystem):
             connection_timeout=self.timeout,
         )
 
-    def _as_unc_path(self, path):
-        rpath = path.replace("/", "\\")
-        unc = "\\\\{}{}".format(self.host, rpath)
-        return unc
-
-    @classmethod
-    def _share_has_path(cls, path):
-        parts = path.count("/")
-        if path.endswith("/"):
-            return parts > 2
-        return parts > 1
-
     @classmethod
     def _strip_protocol(cls, path):
         return infer_storage_options(path)["path"]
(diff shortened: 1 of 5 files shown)


juarezr:
Ops! Failing on installing smbprotocol package install.
It's installed with `pip`.
Any recommendations here?

tox.ini
```diff
@@ -22,6 +22,7 @@ conda_deps=
     s3fs
     pyftpdlib
     cloudpickle
+    smbprotocol
```
telamonian, line 25: I think this just needs to be moved from the `conda_deps` list here to the `deps` list below

telamonian:
@juarezr I figured out the issues with the CI. I can't push directly to this PR, so I opened a new PR with your code + my fixes over at #341. Feel free to build on #341, or to just steal my code and add it to this PR, whatever floats your boat.

fsspec/registry.py
```diff
@@ -154,6 +154,10 @@ def register_implementation(name, cls, clobber=False, errtxt=None):
         "class": "fsspec.implementations.git.GitFileSystem",
         "err": "Install pygit2 to browse local git repos",
     },
+    "smb": {
+        "class": "fsspec.implementations.smb.SMBFileSystem",
+        "err": 'SMBFileSystem requires "smbprotocol[kerberos]" to be installed',
```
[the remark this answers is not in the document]
juarezr: It depends on server configuration.
It's better to make it clear for the dev.

tox.ini
juarezr: I'll fix and push a new commit.

fsspec/implementations/smb.py
```diff
@@ -0,0 +1,225 @@
...
+        elif S_ISLNK(s.st_mode):
+            t = "link"
+        else:
+            t = "file"
+        res = {
+            "name": path + "/" if t == "directory" else path,
+            "size": s.st_size,
+            "type": t,
+            "uid": s.st_uid,
+            "gui": s.st_gid,
```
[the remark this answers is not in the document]
juarezr: Good catch.

martindurant:
Thanks for the implementation! I will try to look through it soon.

martindurant:
Fixes #306

fsspec/implementations/smb.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        buffering = block_size if block_size is not None and block_size >= 0 else -1
+        wpath = self._as_unc_path(path)
+        if autocommit is False:
+            # writes to temporary file, move on commit
+            # TODO: use transaction support in SMB protocol
+            share = path.split("/")[1]
+            path2 = "/{}{}/{}".format(share, self.temppath, uuid.uuid4())
+            wpath2 = self._as_unc_path(path2)
+            f = smbclient.open_file(wpath2, mode, buffering=buffering, **kwargs)
```
martindurant, line 189: Instead of patching the instance, you could make a wrapper class or subclass containing the extra methods

fsspec/registry.py
```diff
@@ -154,6 +154,10 @@ def register_implementation(name, cls, clobber=False, errtxt=None):
         "class": "fsspec.implementations.git.GitFileSystem",
         "err": "Install pygit2 to browse local git repos",
     },
+    "smb": {
```
martindurant, line 157: Is there a distinction to be made between SMB protocols here?

fsspec/registry.py
martindurant, line 159: I agree that it would be better not to require kerberos generally, and catch the auth error when connecting if the particular connection does need it.

fsspec/implementations/smb.py
```diff
@@ -0,0 +1,225 @@
...
+import uuid
+from .. import AbstractFileSystem
+from ..utils import infer_storage_options
+
+
+class SMBFileSystem(AbstractFileSystem):
+    """Downloads or uploads to Windows and Samba network drives.
+
+    The argument `path` (str) must have a URI with format:
+    `smb://workgroup;user:password@server:port/share/folder/file.csv`.
```
martindurant, line 13: Note that most implementations *accept* complex URLs, and create kwargs from them - as you do - but also allow simple paths, in this case "/share/folder/file.csv", so long as the init kwargs are given.

juarezr: Any suggestion for this?

fsspec/registry.py
juarezr: I hope that the message got better after the last commits.
Can check the changed `err`, please?

fsspec/registry.py
juarezr: I have found only `smb://`  in [SMB Filesharing URL Scheme](https://tools.ietf.org/html/draft-crhertel-smb-url-00#section-2).

Do you know something else?

fsspec/implementations/smb.py
juarezr: Oh! I missed that.

I'll fix and commit next.

juarezr: Adressed in d8f676b

fsspec/registry.py
juarezr: Adressed in 5391607

fsspec/implementations/smb.py
martindurant: My suggestion is that, in general, "path" arguments can be of the form "/share/folder.file.csv".

fsspec/implementations/smb.py
martindurant: Sorry I wasn't clear, I mean attaching methods to `f`, the file-like instance. It may be overkill, but you could do something like [this](https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/local.py#L189) for local files.

fsspec/registry.py
martindurant: I have no idea, samba is alien to me :|
If your implementation tries to connect to a service of the other protocol, you just fail, right?
Perhaps @telamonian has an opinion.

fsspec/implementations/tests/test_smb.py
```diff
@@ -0,0 +1,94 @@
...
+        import smbclient
+
+        smbclient.reset_connection_cache()
+        stop_docker(container)
+
+
+def test_simple(smb_params):
+    adir = "/home/adir"
+    adir2 = "/home/adir/otherdir/"
+    afile = "/home/adir/otherdir/afile"
```
martindurant, line 48: Please forget my earlier comment, you clearly show that ordinary paths work as expected, *and* that whole URLs are accepted by `open`, just as it should be.

fsspec/registry.py
telamonian: I need to look up some specs/how `pysmb` does it, but I think it's essentially correct to have the single protocol. You can then have kwargs that limit you to eg no SMB1. I'll look up the details

fsspec/implementations/smb.py
juarezr: This is ok according to [this comment bellow](https://github.com/intake/filesystem_spec/pull/340#pullrequestreview-440215111)

fsspec/implementations/smb.py
```diff
@@ -0,0 +1,251 @@
...
+class SMBFileSystem(AbstractFileSystem):
+    """Allow reading and writing to Windows and Samba network shares.
+
+    When using `fsspec.open()` for getting a file-like object the URI
+    should be specified as this format:
+    `smb://workgroup;user:password@server:port/share/folder/file.csv`.
+
+    Example::
+        >>> import fsspec
+        >>> with fsspec.open('smb://myuser:mypassword@myserver.com/'
```
martindurant, line 27: but you *could* also do
```
with fsspec.open('/share/folder/file.csv',
        host="myserver.com",
        username="myuser",
        password="mypassword"):
```

fsspec/registry.py
telamonian: Okay, so this isn't covered by the standard (it instead describes [a flexible negotiation process](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/80850595-e301-4464-9745-58e4945eb99b) for which features a connection will support; this is why smb has a reputation of being security swiss-cheese).

On the other hand, `smbprotocol` itself provides a nice and simple way of [specifying a minimum SMB_X protocol for your connection](https://github.com/jborean93/smbprotocol#additional-info):

>By default, the negotiation process will use the latest dialect that is supported by the server but this can be overridden if required. When this is done by the following code
>
>```
>import uuid
>
>from smbprotocol.connection import Connection, Dialects
>
>connection = Connection(uuid.uuid4(), "server", 445)
>connection.connect(Dialects.SMB_3_0_2)
>```
>
>While you shouldn't want to downgrade to an earlier version, this does allow you to set a minimum dialect version if required.

Ideally we should support a `dialect` kwarg for `SMBFileSystem`, which would take a [`Dialects`](https://github.com/jborean93/smbprotocol/blob/40161522e3add7ec0382c08042b343d012c1b0ec/smbprotocol/__init__.py#L49) arg. Possibly it would be simpler for `dialect` to instead take a `str` arg, which would then get parsed to a `Dialects`, since this would prevent endusers from having to do `from smbprotocol.connection import Dialects` in their own code.

martindurant: Does that imply that smbprotocol does, in theory, support all protocols? Then this really is an ideal solution, better than I had original thought.
Agree that being to set parameters on the `connect` call is useful. If they are few, they could have specific entries in `__init__`, or we could have a `connect_kwargs=` if there are many possibilities.

telamonian: So, roughly speaking it looks like `pysmb` supports SMB1 and SMB2, while `smbprotocol` supports SMB2 and SMB3 (with `smbprotocol` having more granular support for specific SMB_X_Y_Z versions).

I'm personally more excited by the fact that `smbprotocol` actually has proper Windows CI set up for their repo (my enthusiasm is slightly dampened by the fact [said CI is currently failing](https://ci.appveyor.com/project/jborean93/smbprotocol/builds/33450600); though it looks like it's specifically a python 3.5 issue. Which I don't care about, and nor should anyone else).

martindurant: We *do not* care about py35 :)

martindurant: How does SMB1/2/3 map to usage in the wild?

telamonian: If you're on Windows 7 or higher, then you can use SMB2. So there's allllmost no extant use case for SMB1, which is good, because it is super insecure

martindurant: Win7 is past end-of-life, I say we ignore it; "smb" will mean this implementation only. In the unlikely case that someone makes an SMB1 implementation (like the pyfilesystems one), then we can give it a different protocol marker.

>  September 17, 2019 | 4.11 | SMB1 is disabled by default.

telamonian: > Win7 is past end-of-life

LOL you wish. You might be shocked at the millions upon millions of remaining corporate and government users. I know I was.

But it shouldn't be an issue here, since Win7 does support SMB2. So I agree with this being the implementation for `smb://`.

martindurant: OK :)

fsspec/implementations/smb.py
juarezr: This is missing smb:// somewhere, right?

fsspec/implementations/smb.py
juarezr: I got it now!
Unfortunately my city was hit by strong winds and rain, and I don't have electricity to explore it right now.

fsspec/implementations/smb.py
martindurant: Sorry, 
```
with fsspec.open('smb://share/folder/file.csv',
        host="myserver.com",
        username="myuser",
        password="mypassword"):
```
(although you can also give `protocol="smb"` instead)

martindurant:
Please do update docs/source/api.rst under "Built-in Implementations" to include this implementation, when things are ready.

fsspec/implementations/smb.py
juarezr: What exactly should change here? 
Should we keep the doc example? Change it? Extend it?

fsspec/implementations/smb.py
juarezr: I hope this is addressed in e4d9988. 
Could you review again, please?

juarezr:
> Please do update docs/source/api.rst under "Built-in Implementations" to include this implementation, when things are ready.

Done.

martindurant:
Is anything more required here? I am happy to merge if you say you are done.
Thanks for the work!

juarezr:
@martindurant ,

LGTM.

If there is no issue left, please merge.
Also further details can be handled in future PRs.

martindurant:
@telamonian : merging this now. Please do keep testing and post issues/fixes as they come.
Thank you @juarezr

juarezr:
Other PR related to `fsspec`: [petl](https://github.com/petl-developers/petl) remote file [support](https://github.com/petl-developers/petl/pull/494).

PR #343, opened 2020-06-30
fixed typo "gui" => "gid"

telamonian:
While reviewing #340, I noticed the same typo there and in `fsspec/implementations/sftp.py`:

https://github.com/intake/filesystem_spec/blob/f29eb56d0b8960cf2b4245afc8016c9aec3bb6bf/fsspec/implementations/sftp.py#L91

This is supposed to be `"gid": ...` right? I assume the type in #340 was copied from here

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,8 +91,6 @@ celerybeat-schedule
 .venv
 venv/
 ENV/
-.idea/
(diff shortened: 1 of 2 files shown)


martindurant:
Oops :)

PR #344, opened 2020-07-02
Pass glob kwargs to super class from local fs

masda70:
Proposed solution to https://github.com/intake/filesystem_spec/issues/283


TomAugspurger:
Thanks @masda70!

PR #348, opened 2020-07-07
Avoid async hell

martindurant:
@TomAugspurger , this is the "sidestep" I talked about, can take you through it if you like.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -162,16 +162,22 @@ class AsyncFileSystem(AbstractFileSystem):
         self.loop = loop or get_loop()
         super().__init__(*args, **kwargs)
 
-    async def _rm(self, path, recursive=False):
-        path = await self._expand_path(path, recursive=recursive)
-        await asyncio.gather(*[self._rm_file(p) for p in path])
+    async def _rm(self, path, recursive=False, **kwargs):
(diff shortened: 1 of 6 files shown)


fsspec/asyn.py
```diff
@@ -162,38 +162,55 @@ def __init__(self, *args, asynchronous=False, loop=None, **kwargs):
...
+        await asyncio.gather(*[self._rm_file(p, **kwargs) for p in path])
 
-    async def _copy(self, path1, path2, recursive=False, **kwargs):
-        paths = await self.expand_path(path1, recursive=recursive)
-        path2 = other_paths(paths, path2)
+    def rm(self, path, recursive=False, **kwargs):
+        path = self.expand_path(path, recursive=recursive)
+        sync(self.loop, self._rm, path, **kwargs)
+
+    async def _copy(self, paths, path2, **kwargs):
```
TomAugspurger, line 172: Add to the `defualt_async_methods` list above?

martindurant: Yep, the doc and some of the lists will need updating.
I still might make the expand_path tree of functions async too (i.e., multiple concurrent calls to ls), in the future, but I'm happy to have the batch ops working for now.

PR #354, opened 2020-07-20
Fix race condition of lzma import when using threads

### LINKED ISSUE #353: lzma.LZMAFile AttributeError in Python without LZMA
I sometimes get a `module 'lzma' has no attribute 'LZMAFile' ` error in a threaded environment (which I don't have any control over, unfortunately 😟 - it's an company internal "cloud" engine). It's not reproducible and I was unable so far to create a local example for this. 

The stack trace is:
```
  File "/root/.local/lib/python3.6/site-packages/adlfs/__init__.py", line 1, in <module>
    from .core import AzureDatalakeFileSystem
  File "/root/.local/lib/python3.6/site-packages/adlfs/core.py", line 12, in <module>
    from fsspec import AbstractFileSystem
  File "/root/.local/lib/python3.6/site-packages/fsspec/__init__.py", line 5, in <module>
    from .mapping import FSMap, get_mapper
  File "/root/.local/lib/python3.6/site-packages/fsspec/mapping.py", line 3, in <module>
    from .core import split_protocol
  File "/root/.local/lib/python3.6/site-packages/fsspec/core.py", line 6, in <module>
    from .compression import compr
  File "/root/.local/lib/python3.6/site-packages/fsspec/compression.py", line 77, in <module>
    register_compression("lzma", lzma.LZMAFile, "xz")
AttributeError: module 'lzma' has no attribute 'LZMAFile'
```

After digging some time I found [Python issue 39430: tarfile.open(mode="r") race condition when importing lzma](https://bugs.python.org/issue39430) which also suffered from the same AttributeError. It looks like there is a race condition when using `import lzma` compared to `from lzma import xyz` in a threaded environment.

Would it be ok to change the lzma import in [fsspec/compression.py](https://github.com/intake/filesystem_spec/blob/master/fsspec/compression.py#L76-L79)? What do you think? I did a local modification of those lines and so far the AttributeError does not longer occur.

Happy to get your feedback on this and opening a PR if we agree on this.

dfigus:
Fix a race condition of the import lzma that occurs when this is executed inside threads. Refer to similar issue in the python tar library [Python issue 39430: tarfile.open(mode="r") race condition when importing lzma](https://bugs.python.org/issue39430).

This fixes #353.

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -73,10 +73,10 @@ register_compression("bz2", BZ2File, "bz2")
 register_compression("gzip", lambda f, **kwargs: GzipFile(fileobj=f, **kwargs), "gz")
 
 try:
-    import lzma
+    from lzma import LZMAFile
 
-    register_compression("lzma", lzma.LZMAFile, "xz")
-    register_compression("xz", lzma.LZMAFile, "xz", force=True)
+    register_compression("lzma", LZMAFile, "xz")
+    register_compression("xz", LZMAFile, "xz", force=True)
 except ImportError:
     pass
 


martindurant:
Thank you!

PR #364, opened 2020-07-28
test_local: fix username not in home path (#253)

### LINKED ISSUE #253: test_strip_protocol_expanduser assumes home paths
`test_strip_protocol_expanduser` assumes that `$USER` is part of the path of `$HOME`, but there's no requirement for that to be so.

ArchangeGabriel:
It’s not required that the username must be part of the path to that user home. Instead, check whether the call to expand the `~` on fsspec side was correctly made.

Fixes #253.

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -8,7 +8,6 @@ import sys
 from contextlib import contextmanager
 import posixpath
 import tempfile
-import getpass
 
 import pytest
 import fsspec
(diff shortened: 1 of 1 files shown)


martindurant:
Can you please check the lint with `black` and `flake8`

```
fsspec/implementations/tests/test_local.py:11:1: F401 'getpass' imported but unused
```

ArchangeGabriel:
@martindurant Sorry for missing that, fixed. They are still CI failures, but they seem unrelated.

martindurant:
OK, thank you!

PR #369, opened 2020-08-04
HTTPFileSystem requires requests and aiohttp

manics:
If `requests` but not `aiohttp` is installed the current error message is misleading since it only mentions requests.


martindurant:
Thanks for this.

Issue #366, opened 2020-08-01
dependency on aiohttp should be in the requirements.txt


bmccandless:
the update to 0.8.0 brings in a dependency to aiohttp, which is not listed in the requirements.   
I therefore get an import error when using fsspec.

juarezr:
I have the same error while [building](https://github.com/petl-developers/petl/pull/507) for [python3.6](https://travis-ci.org/github/petl-developers/petl/jobs/714909092#L1221) and [python3.7](https://travis-ci.org/github/petl-developers/petl/jobs/714909093#L1363).

Any chance of including `aiohttp` as dependecy or maybe a import check as workaround?

Thanks

martindurant:
fsspec does *not* depend on aiohttp, only the HTTP backend does. Attempting to use the HTTP backend without it should display `'HTTPFileSystem requires "requests" and "aiohttp" to be installed'` (on master).

juarezr:
Understood.

Although I fixed my build, other environments relying in `fsspec` + `requests` installed but without `aiohttp` could break.

This configures a serious issue that needs fixing here?

mivade:
I actually get the following error when running tests of a project with 0.8.0:

```
ImportError: HTTPFileSystem requires "requests" to be installed
```

However looking at the full traceback what is actually happening is this:

```
ModuleNotFoundError: No module named 'aiohttp'
```

so something is clearly off and it looks like `aiohttp` *is* actually required for HTTP support but we get an error message that hides this fact.

martindurant:
The message is fixed on master https://github.com/intake/filesystem_spec/blob/master/fsspec/registry.py#L100

mivade:
Glancing at `setup.py`, it wouldn't be a bad idea to use "extras" to make it easier to get these dependencies by doing something like `pip install fsspec[http]`. I can put together a PR for this shortly.

martindurant:
Good idea. Obviously there are quite a few... I wonder if it could be done in a way not to repeat the information in registry.py . 
(Indeed, it would be good to implement #206 , which is tangentially related)

PR #371, opened 2020-08-05
Use extras_require in setup.py for optional dependencies

mivade:
Include `requests` and `aiohttp` as both appear to be required for HTTP support now.

This relates to comments in #366 

Are there any other optional dependencies we should add here? Once those are added I would also update the documentation to include hints about this, so I'm marking this as a draft PR for now.


martindurant:
Yes, quite a few! Any of the items in registry that are imported from within the fsspec package and have error messages, e.g., distributed for dask, paramiko for ssh, pyarrow for hdfs.

mivade:
Thanks @martindurant, I'll take a look there.

setup.py
```diff
@@ -32,5 +32,19 @@
     packages=["fsspec", "fsspec.implementations"],
     python_requires=">3.6",
     install_requires=open("requirements.txt").read().strip().split("\n"),
+    extras_require={
+        "aws": ["s3fs"],
```
martindurant, line 36: I would probably use the same names as the protocols, so "s3" for this one and "gcs" instead of "google"

mivade: I wasn't sure what the best approach was there. I think matching it is fine even if that means there are some redundancies.

docs/source/index.rst
```diff
@@ -31,6 +31,11 @@ Installation
 
    pip install fsspec
 
+Not all included filesystems are usable by default without installing extra
+dependencies. For example to be able to access data in S3::
+
+   pip install fsspec[s3]
```
martindurant, line 37: Missing s3 in the list :)

martindurant: Please also check that this works - because s3fs depends on fsspec, fsspec[s3] would depend on itself.

martindurant:
This PR is still marked as draft. Ready to go?
I have not personally checked that something like `pip install fsspec[s3]` would do the right thing.

mivade:
Not yet. I still want to add automated testing that the extras work. I'll mark it as ready for review once I've done so.

martindurant:
ping @mivade , do you think you'll have time to finish this PR off?

mivade:
Yeah, I should be able to get to the testing later this week. Sorry for the delay!

mivade:
@martindurant I've added some pretty basic testing here. I took the existing S3 and GCSFS tox environments and removed the conda dependencies installed there to instead use the `extras=` line. This tells tox to install the package with the provided extras (e.g., something like `pip install .[gcs]`). Of course this doesn't test every single extra defined in this PR; doing so would require some considerable rework of how the test environments are setup.

martindurant:
Looks good!

Issue #368, opened 2020-08-04
Version 0.8.0 marked as compatible with Python 3.5 but uses f-strings


dlindelof:
Version 0.8.0 is marked on PyPI as requiring Python >= 3.5, but there's an f-string in core.py. Python 3.6 is required for f-strings.

martindurant:
Gahh, the *selectors* say 3.6+... I suppose that's not enough.
Would you like to make the correction as a PR?

dlindelof:
I'll be happy to. Should we replace the use of f-strings with normal strings to keep compatibility with Python 3.5, or stop supporting Python 3.5 altogether?

martindurant:
No py35 seems to have been dropped in lots of places (including support libraries for some backends), so lets drop it.

mivade:
I'm noticing that `tox.ini` is currently testing against Python 3.5, 3.6, and 3.7 but not 3.8.

dlindelof:
Seems to have been fixed by 21c4316?

martindurant:
Yep, might have slipped in :)

tsibley:
@dlindelof Can you clarify where you saw/see Python >=3.5 for fsspec 0.8.0 in PyPI? On the [0.8.0 release page](https://pypi.org/project/fsspec/0.8.0/), I see "Requires: Python >3.5".

I want to adjust fsspec setup.py's `python_requires` to address #402 but don't want to revert something affecting this issue.

tsibley:
@dlindelof Ah ha! Nevermind. I see now that `>3.5` means `>3.5.0` and is not equivalent to `>=3.6`.

PR #372, opened 2020-08-06
Test against Python 3.8, drop 3.5 testing

mivade:
This aligns with the classifiers and discussion in #368.


martindurant:
Thanks!

PR #379, opened 2020-08-20
fix foreground parameter not being used issue

xyb:


martindurant:
Hm, interesting!
Did you have cause to use this? Thanks for the fix.

xyb:
@martindurant Haha, just reading the code and saw this error.

PR #406, opened 2020-09-13
Fix missing kwargs in call to _copy in asyn

ysgit:


martindurant:
Thanks, merging. Did you experience a particular problem from this?

ysgit:
@martindurant thanks. Yes, when using s3fs and trying to update the timestamp on (touch) a file by copying it onto itself, one must pass the kwarg `MetadataDirective='REPLACE'`

PR #408, opened 2020-09-15
Added support for hat dollar lbrace rbrace regex character escapes.

shipmints:
Fixes https://github.com/intake/filesystem_spec/issues/390


martindurant:
This is good with me.
let's see if other more knowledgable people comment.
I do wonder what happens in the case of already quoted special characters, but we can leave that for a follow-up, if it ever arises.

martindurant:
You are failing on flake8 style:

fsspec/spec.py:451:89: E501 line too long (166 > 88 characters)
fsspec/spec.py:491:89: E501 line too long (103 > 88 characters)
fsspec/spec.py:492:89: E501 line too long (114 > 88 characters)

PR #409, opened 2020-09-15
Cat error options

martindurant:
To allow async usage in zarr, provide options of what happens on missing items when doing `cat` (or mapping `getitems`).


martindurant:
Marking this for review.
We should extend the on_error parameter to other bulk operations, ideally without much code repetition. In the meantime, this is enough to get async reading on zarr working. So, since the API won't change, this can be merged.

PR #412, opened 2020-09-17
Make hash_name a method on CachingFileSystem to make it easier to change.

owal-mike:
Right now it's hard to change the hashing method without overriding `_close`, this PR makes it a method so a subclass can implement a different scheme.


martindurant:
To get this to work, the method name needs to be included in the list of known methods in getattribute. This is obviously not a very nice way to do things, happy to hear of an alternative.

owal-mike:
Pretty sure a metaclass could be used to collect all of the methods but all of the extra magic would probably make it even more confusing.

martindurant:
OK, this is good, thank you.
The getattribute thing can wait until someone teaches me a better way.

PR #414, opened 2020-09-18
handle multibyte dtypes

heeler:
First off thank you so much for fsspec and all the work it entailed! 
We think we have a fix for a small bug.
We ran into the following problem.
When we execute this script: 
``` python
from s3fs import S3FileSystem
from tifffile import imread
fs = S3FileSystem(anon=True)
with fs.open("s3://aics-modeling-packages-test-resources/aicsimageio/test_resources/resources/s_1_t_1_c_1_z_1.tiff") as open_resource:
    print(type(open_resource))
    print(imread(open_resource).shape)
```
we encounter the following exception:
```
<class 's3fs.core.S3File'>
  File "/Users/jamies/Library/Application Support/JetBrains/PyCharm2020.1/scratches/scratch_13.py", line 14, in <module>
    print(imread(open_resource).shape)
  File "/Users/jamies/Sandbox/Scratch/Jackson/tifffile/tifffile/tifffile.py", line 721, in imread
    return tif.asarray(**kwargs)
  File "/Users/jamies/Sandbox/Scratch/Jackson/tifffile/tifffile/tifffile.py", line 2804, in asarray
    result = self.filehandle.read_array(
  File "/Users/jamies/Sandbox/Scratch/Jackson/tifffile/tifffile/tifffile.py", line 7498, in read_array
    raise ValueError(f'failed to read {size} bytes')
ValueError: failed to read 308750 bytes
Process finished with exit code 1
```
When we traced it back it seems to be the read() in readinto() doesn't seem to allow for array types of more than 1 byte so this fixes that by simply casting to a bytearray prior to getting length.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1364,7 +1364,7 @@ class AbstractBufferedFile(io.IOBase):
 
         https://docs.python.org/3/library/io.html#io.RawIOBase.readinto
         """
-        data = self.read(len(b))
+        data = self.read(len(bytearray(b)))
         memoryview(b).cast("B")[: len(data)] = data
         return len(data)
 
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1364,8 +1364,9 @@ class AbstractBufferedFile(io.IOBase):
 
         https://docs.python.org/3/library/io.html#io.RawIOBase.readinto
         """
-        data = self.read(len(bytearray(b)))
-        memoryview(b).cast("B")[: len(data)] = data
+        out = memoryview(b).cast("B")
+        data = self.read(out.nbytes)
+        out[: len(data)] = data
         return len(data)
 
     def readuntil(self, char=b"\n", blocks=None):
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -1,11 +1,15 @@
 import pickle
 import json
 
+import os
 import pytest
 from fsspec.spec import AbstractFileSystem, AbstractBufferedFile
+from fsspec.implementations.ftp import FTPFile, FTPFileSystem
+from fsspec.implementations.local import LocalFileSystem
 import fsspec
 
 import numpy as np
+from pathlib import Path
 
 
 class DummyTestFS(AbstractFileSystem):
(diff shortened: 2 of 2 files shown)


martindurant:
Thanks for tracking this down.

I believe `bytearray()` makes a copy. Does the following work?
```diff
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1364,8 +1364,9 @@ class AbstractBufferedFile(io.IOBase):

         https://docs.python.org/3/library/io.html#io.RawIOBase.readinto
         """
-        data = self.read(len(b))
-        memoryview(b).cast("B")[: len(data)] = data
+        out = memoryview(b).cast("B")
+        data = self.read(out.nbytes)
+        out[: len(data)] = data
         return len(data)
```

It would be great to have a test for readinto with a non-byte type; would you mind creating this test? It would have to be one of the derivative classes, like FTPFile.

heeler:
You're right, avoiding the copy is the way to go and your suggestion makes sense. I will take another shot at writing a test. I just have to choose something that is a descendant of AbstractBufferedFile.  Hopefully, I'll have something for you shortly. Thanks.

heeler:
I've added `test_readinto_with_multibyte`. It's not particularly pretty but it fails on 7 of the 10 types with the original implementation and it succeeds on all of them with the new implementation. If you have suggestions on how to clean up the test I'm all ears.

heeler:
I'll look into why this is failing on Travis. I ran the tests successfully locally with tox but obviously there's some kind of problem. I'll get to this as soon as I can.

martindurant:
Wow, this is rather more complicated than what I had in mind - can you use the `ftp_writable` fixture and interact directly? There are some tests that do this in test_core and test_file. I would say this test belongs in the latter (because it tests the file interface, rather than the filesystem spec, even though both of those are defined in spec.py).

heeler:
I had a feeling I was on the wrong path. Thanks for the suggestion. I'll give it a go later today.

heeler:
Thanks for the suggestion, ftp_writable was definitely a cleaner way to go.

martindurant:
Perfect, thanks.

PR #419, opened 2020-09-23
Uses get_event_loop for < 3.7 compatibility

pschork:
Resolves https://github.com/intake/filesystem_spec/issues/410


fsspec/asyn.py
```diff
@@ -16,7 +16,7 @@
 
 def _run_until_done(coro):
     """execute coroutine, when already in the event loop"""
-    loop = asyncio.get_running_loop()
+    loop = asyncio.get_event_loop()
```
martindurant, line 19: Perhaps add `assert loop.is_running()` ?

PR #421, opened 2020-09-24
Fix the dircache is not correctly expired after transaction is completed.

xyb:


fsspec/spec.py
```diff
@@ -217,6 +218,13 @@ def end_transaction(self):
         """Finish write transaction, non-context version"""
         self.transaction.complete()
         self._transaction = None
+        # The invalid cache must be cleared after the transcation is completed.
+        for path in self._invalidated_caches_in_txn:
+            self.invalidate_cache(path)
```
martindurant, line 223: And then
```
self._invalidated_caches_in_txn.clear()
```

fsspec/implementations/ftp.py
```diff
@@ -166,6 +168,13 @@ def mv(self, path1, path2, **kwargs):
     def __del__(self):
         self.ftp.close()
 
+    def invalidate_cache(self, path=None):
+        if path and path in self.dircache:
+            del self.dircache[path]
+        else:
+            self.dircache.clear()
```
martindurant, line 175: It seems that this gets called if the path passed happens not to be int he cache. It should only be called if path is None
```
if path is None:
    self.dircache.clear()
else:
    self.dircache.pop(path, None)'
```

fsspec/spec.py
```diff
@@ -217,6 +218,13 @@ def end_transaction(self):
...
+
+    def _record_invalidated_cache(self, path):
```
martindurant, line 225: Being only two lines, this could go right into `invalidate_cache` rather than make a new method

martindurant (commented):
It took me a moment to understand what you were doing, but I have it now.

I left a couple of comments.

"txn" stands for "transaction"?

fsspec/implementations/ftp.py
xyb: Yes, good catch! I have fixed this bug and another one in the `_parent` method by the way. Please take a look again.

fsspec/spec.py
xyb: Fixed.

fsspec/spec.py
xyb: Merged.

xyb:
@martindurant Fixed and replaced the abbreviation "txn".

martindurant (commented):
Sorry, a couple more comments.

fsspec/spec.py
```diff
@@ -808,7 +818,7 @@ def rm(self, path, recursive=False, maxdepth=None):
     def _parent(cls, path):
         path = cls._strip_protocol(path.rstrip("/"))
         if "/" in path:
-            return cls.root_marker + path.rsplit("/", 1)[0]
+            return os.path.join(cls.root_marker, path.rsplit("/", 1)[0])
```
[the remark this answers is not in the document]
xyb: > Also, I suspect that this would make "//path/..."

Exactly, that's why I fix the `_parent` method.

> If there was a specific error that you saw, it would be good to get that as a separate test.

In theory, it's correct. But
1. The `_parent` method is private, I think the private method is unstable, testing it is not a good idea;
2. In fact, the `_parent` method only causes errors in the `zip.py` and `ftp.py` files in which `root_marker` is not empty. It works well in `spec.py`, so you can't test the actual problem in `spec.py`, or it's not easy to write a straightforward test.
So I think it's ok to indirectly test the _parent method in ftp.py for now.

But in the long run, all the logic about directories included `_parent` should be refactored, may be extracted into a utils file, and tested well. For example, we have too much code like this:
```
            elif (
                path
                and len(path) < len(p.strip("/"))
                and all(
                    (a == b) for a, b in zip(path.split("/"), p.strip("/").split("/"))
                )
            ):
```

fsspec/spec.py
```diff
@@ -228,7 +233,12 @@ def invalidate_cache(self, path=None):
             If None, clear all listings cached else listings at or under given
             path.
         """
-        pass  # not necessary to implement, may have no cache
+        # Not necessary to implement invalidation mechanism, may have no cache.
+        # But if have, you should call this method of parent class from your
+        # subclass to ensure expiring caches after transacations correctly.
+        # See the implementaion of FTPFileSystem in ftp.py
+        if bool(self._transaction):
```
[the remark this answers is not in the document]
xyb: Modified.

fsspec/spec.py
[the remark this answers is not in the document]
xyb: Oh, and if we officially support windows and mac osx, we should add them to our CI environment. I may post a pull request for that later.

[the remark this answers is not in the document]
martindurant: That would be appreciated

PR #431, opened 2020-10-02
Deep async

martindurant:
@jhnnsrs , can you check how this works with your real-world example? I do not think this is good code, but makes things pass, when I try the s3fs test in https://github.com/dask/s3fs/pull/374 with `--count=10`.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -5,6 +5,7 @@ import re
 import os
 import sys
 import threading
+import time
 
 from .utils import other_paths, is_exception
 from .spec import AbstractFileSystem
@@ -12,12 +13,10 @@ from .spec import AbstractFileSystem
 # this global variable holds whether this thread is running async or not
 thread_state = threading.local()
 private = re.compile("_[^_]")
-depth = [0]
 
 
 def _run_until_done(coro):
     """execute coroutine, when already in the event loop"""
-    depth[0] += 1
     if sys.version_info < (3, 7):  # pragma: no cover
         raise RuntimeError(
             "async file systems do not work completely on py<37. "
@@ -31,14 +30,12 @@ def _run_until_done(coro):
     del asyncio.tasks._current_tasks[loop]
     runner = loop.create_task(coro)
     while not runner.done():
-        print(depth[0])
         try:
             loop._run_once()
         except IndexError as e:
-            pass
+            time.sleep(0.001)
     asyncio.tasks._register_task(task)
     asyncio.tasks._current_tasks[loop] = task
-    depth[0] -= 1
     return runner.result()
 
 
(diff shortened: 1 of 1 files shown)


jhnnsrs:
sorry for coming back so late. 
Fix works great for datasets smaller than (1024,1024,50,2,1) but falls short on anything beyond, same 

```
  File "/usr/local/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 1844, in _run_once
    handle = self._ready.popleft()
IndexError: pop from an empty deque
````

Requirements:
```
aiobotocore==1.1.1
aiohttp==3.6.2
aioitertools==0.7.0
asciitree==0.3.3
async-timeout==3.0.1
attrs==20.2.0
botocore==1.17.44
chardet==3.0.4
dask==2.28.0
docutils==0.15.2
fasteners==0.15
fsspec @ file:///code/filesystem_spec/dist/fsspec-0.7.4%2B309.g77564f5-py3-none-any.whl
idna==2.10
jmespath==0.10.0
monotonic==1.5
multidict==4.7.6
numcodecs==0.7.2
numpy==1.19.2
pandas==1.1.2
python-dateutil==2.8.1
pytz==2020.1
PyYAML==5.3.1
s3fs==0.5.1
six==1.15.0
toolz==0.11.1
typing-extensions==3.7.4.3
urllib3==1.25.10
wrapt==1.12.1
xarray==0.16.1
yarl==1.6.0
zarr==2.4.0
```
Thanks!

martindurant:
I was running with a size `(1024, 1024, 10, 9, 1)` and repeats `--count=10` and passing, but if I up the time even further, it breaks for me too :|. Increasing the sleep time only converts the exceptions into deadlocks.

I am puzzled that
- you can get an IndexError inside a block explicitly protected by `except IndexError`
- that a sleep makes any difference (this is the only thread handling events, other threads can only add events; but maybe it allows them to drop references and let garbage get collected?)
- zarr should not be calling mkdirs on every single chunk write. Even for a local file-system, I find that it hurts performance for many-small-writes quite a lot; we know that all the chunks live in the same dir 

--EDIT--

I have found the one-line fix (does not need change to fsspec) :). However, I am still not understanding exactly why the multiply-nested calls caused the race, and what to do about it. This will pobably turn up elsewhere.

martindurant:
Since the fix is in s3fs, this need not be merged (it is questionable code anyway, and you should see my local version); I'll leave this open here for now.

PR #442, opened 2020-10-08
DOC: hyperlink registry

raybellwaves:
I test locally later.

You can add a rtd build check using the site (click button at botton of image)
Screenshot from 2020-10-08 08-00-05

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -55,7 +55,7 @@ The current list of known implementations can be found as follows
     known_implementations
 
 These are only imported on request, which may fail if a required dependency is missing. The dictionary
-``fsspec.registry`` contains all imported implementations, and can be mutated by user code, if necessary.
+:py:class:`fsspec.registry` contains all imported implementations, and can be mutated by user code, if necessary.
 
 
 .. toctree::
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -55,7 +55,7 @@ The current list of known implementations can be found as follows
     known_implementations
 
 These are only imported on request, which may fail if a required dependency is missing. The dictionary
-:py:class:`fsspec.registry` contains all imported implementations, and can be mutated by user code, if necessary.
+:py:class:`fsspec.registry.known_implementations` contains all imported implementations, and can be mutated by user code, if necessary.
 
 
 .. toctree::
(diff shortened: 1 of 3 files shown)


raybellwaves:
couldn't get the hyperlink to work. I belive because https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/registry.html isn't part of the API.

martindurant:
Feel free to add it. The registry could probably do with more work to include searching for entrypoints anyway.

raybellwaves:
bit unsure where to put it. Dropped it in Built-in Implementations but not sure if `autofunction` is appropriate. The hyperlink works but it's not very useful. Ideally want to show the dict when you click on the hyperlink from the home page
```
"file": {"class": "fsspec.implementations.local.LocalFileSystem"},
    "memory": {"class": "fsspec.implementations.memory.MemoryFileSystem"},
    "dropbox": {
        "class": "dropboxdrivefs.DropboxDriveFileSystem",
        "err": (
            'DropboxFileSystem requires "dropboxdrivefs",'
            '"requests" and "dropbox" to be installed'
        ),
    },
...
```

Screenshot from 2020-10-08 21-05-36

martindurant:
A couple of possible solutions for directly including the dictionary: https://stackoverflow.com/questions/27875455/displaying-dictionary-data-in-sphinx-documentation/29789910

martindurant:
(ping)

docs/source/index.rst
```diff
@@ -55,7 +55,7 @@ The current list of known implementations can be found as follows
     known_implementations
 
 These are only imported on request, which may fail if a required dependency is missing. The dictionary
-``fsspec.registry`` contains all imported implementations, and can be mutated by user code, if necessary.
+:py:class:`fsspec.registry.known_implementations` contains all imported implementations, and can be mutated by user code, if necessary.
```
martindurant, line 58: Actually, on second reading, this is wrong. `fsspec.registry` is indeed the set of classes that have been used so far: the values are python class objects. `known_implementations` is a dictionary where the values are dictionaries of import details, location of the class and reasonable error messages - it serves as the source of the classes for `registry` via `fsspec.get_filesystem_class`, but you could edit either one at runtime.

martindurant:
Sorry, this seems to have got lost. Still interested?

Issue #339, opened 2020-06-29
Default filesystem


martindurant:
Idea: to be able to register an instance as the "default" for a given protocol, so that `filesystem(protocol)`, without additional kwargs, gets that particular instance. Will mean ability for pandas (for example) to do IO without figuring out a way to pass storage_options or other arguments through.

martindurant:
@jcrist : this is the idea that I mentioned. What the "default" for a given protocol is could be defined as part of configuration options.

PR #453, opened 2020-10-15
Config system

martindurant:
Allows configuring of any kwarg parameter for backends.
Precedence is (latest wins):
- .ini and .json files in the config directory, in listdir order
- environment variables of the form `FSSPEC_<protocol>_<kwarg>` (will be lower-cased, so cannot use for kwargs with upper case characters in the name)
- any dynamic changes to the `fsspec.config.conf` dictionary
- parameters passed at instantiation time

JSON file can look like:
```json
{"gcs":
  {"requester_pays": true}
}
```
INI file:
```ini
[gsc]
requester_pays=true
```
(note that JSON files will have correct types, but ini files only support strings)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+Version 0.8.4
+-------------
+
+Features:
+
+- function ``can_be_local`` to see whether URL is compatible with ``open_local``
+- concurrent cat with filecaches, if backend supports it
+
+Fixes:
+
+- dircache expiry after transaction
+- blockcache garbage collection
+- close for HDFS
+- windows tests
+- glob depth with "**"
+
 Version 0.8.3
 -------------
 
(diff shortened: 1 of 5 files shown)


martindurant:
Just like Dask, we could introduce a context to allow temporarily setting the config. I don't think that's a priority.

nbren12:
Looks great! Would it be better to adhere to the [XDG standard](https://unix.stackexchange.com/questions/24347/why-do-some-applications-use-config-appname-for-their-config-data-while-other) for the config directory location. Basically using `~/.config/fsspec`. Many applications (e.g. dask, matplotlib) adhere to this standard.

martindurant:
Yes, that's a good point. It will still be a surprise, maybe, for Windows users, but I would rather not add `appdirs` as a dependency.

martindurant:
Fixes #339 (but not in the way suggested there)

nbren12:
This looks great! Thanks so much for putting it together. I'm sure it will come in handy.

PR #454, opened 2020-10-15
Fix MemoryFileSystem recursive rm bug

namanjain:
This change fixes an existing bug that throws an error when removing a
directory recursively which contains non-psuedo sub-dirs.

As MemoryFileSystem allows you to create files without creating the
parent directories first, those parent directories are not added to
`pseudo_dirs`. We should ignore deleting such directories in recursive
remove operation since once the underneath files are deleted, the
directory would be automatically deleted.

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -164,6 +164,17 @@ class MemoryFileSystem(AbstractFileSystem):
         else:
             raise FileNotFoundError
 
+    def rm(self, path, recursive=False, maxdepth=None):
+        paths = self.expand_path(path, recursive=recursive, maxdepth=maxdepth)
+        for p in reversed(paths):
+            # If the expanded path doesn't exist, it is only because the expanded
+            # path was a directory that does not exist in self.pseudo_dirs. This
+            # is possible if you directly create files without making the
+            # directories first.
+            if not self.exists(p):
+                continue
+            self.rm_file(p)
+
     def size(self, path):
         """Size in bytes of the file at path"""
         if path not in self.store:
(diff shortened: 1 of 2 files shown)


fsspec/implementations/memory.py
```diff
@@ -164,6 +164,20 @@ def _rm(self, path):
...
             raise FileNotFoundError
 
+    def rm(self, path, recursive=False, maxdepth=None):
+        paths = self.expand_path(path, recursive=recursive, maxdepth=maxdepth)
+        for p in reversed(paths):
+            # If the expanded path doesn't exist, it is only because the expanded
+            # path was a directory that does not exist in self.pseudo_dirs. This
+            # is possible if you directly create files without making the
+            # directories first.
+            if not self.exists(p):
```
martindurant, line 174: If it cannot happen, we probably shouldn't test for it?

fsspec/implementations/memory.py
```diff
@@ -164,6 +164,20 @@ def _rm(self, path):
         else:
             raise FileNotFoundError
 
+    def rm(self, path, recursive=False, maxdepth=None):
```
martindurant, line 167: Aside from the branch that should never get called, I don't see how this is different from `AbstractFileSystem.rm`.

martindurant (commented):
You have demonstrated the problem well with the test, but the putative fix doesn't seem to help.

fsspec/implementations/memory.py
namanjain: Maybe my comment was confusing. I have my comments on the line above, not below. So this if condition can happen, but what cannot happen is checked on line 177 (that `p` is `path`). I can remove that assert if it makes it confusing, I was just adding some defense in case something changes in the future.

martindurant: Please do remove the assert and comment

namanjain: Done!

martindurant:
Thank you!

PR #459, opened 2020-10-28
Fix and test expand_path recursive

### LINKED ISSUE #456: expand_path fails with recursive magic
```python
fs = LocalFileSystem()
fs.expand_path("./*", recursive=True)
# TypeError: unsupported operand type(s) for +=: 'set' and 'list'
```
[This](https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L793) should probably be `out |= set(self.expand_path(p))`

ahirner:
Should eventually fix #456 
Before that, needs clarification what `expand_path` is supposed to return: directories (including root), files, both? See failing test.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -785,10 +785,13 @@ class AbstractFileSystem(up, metaclass=_Cached):
             self.cp_file(p1, p2, **kwargs)
 
     def expand_path(self, path, recursive=False, maxdepth=None):
-        """Turn one or more globs or directories into a list of all matching files"""
+        """Turn one or more globs or directories into a list of all matching paths
+        to files or directories."""
         if isinstance(path, str):
             out = self.expand_path([path], recursive, maxdepth)
         else:
+            # reduce depth on each recursion level unless None or 0
+            maxdepth = maxdepth if not maxdepth else maxdepth - 1
             out = set()
             path = [self._strip_protocol(p) for p in path]
             for p in path:
@@ -796,9 +799,14 @@ class AbstractFileSystem(up, metaclass=_Cached):
                     bit = set(self.glob(p))
                     out |= bit
                     if recursive:
-                        out += self.expand_path(p)
+                        # We add all the globbed results under the assumption
+                        # .glob() expands paths (only files are documented but
+                        # in test they do) and files paths are OK with this function
+                        out |= set(self.expand_path(list(bit), recursive=recursive,
+                                                    maxdepth=maxdepth))
                     continue
                 elif recursive:
+                    # TODO: maxdepth is not used and probably should
                     rec = set(self.find(p, withdirs=True))
                     out |= rec
                 if p not in out and (recursive is False or self.exists(p)):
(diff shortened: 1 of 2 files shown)


martindurant:
I suppose the docstring should have said "paths".

Note that the purpose here is to enable bulk operations. Any calling code could infer a set of directories from files alone, or, conversely, could infer which of the paths are meant to be directories. What we *don't* want, is to call isfile/isdir/exists within the function, and the recursion would be at best slow, at worst hang (when called in many items within an async context).

ahirner:
@martindurant Thanks for clarification. To get a better understanding of `expand_paths` use case, would you then suggest that `s3fs` uses `find()` instead to prepare the bulk `get()` operation? (right now doesn't filter expand_paths and only works with file paths, re https://github.com/dask/s3fs/issues/376).

martindurant:
expand_path does use find. Yes, you could override functionality in s3fs to use find directly for the purpose of cp/get/cat, but you'd loose code reusability; and you still need to infer directories at some point for the case of get (because you need to create directories locally).

ahirner:
Udpated PR with two reasonable and one failing but commented out test (for CI).

Could you comment on the two `TODO` s @martindurant and/or think of another test case? Thank you very much.

fsspec/spec.py
```diff
@@ -785,20 +785,31 @@ def copy(self, path1, path2, recursive=False, **kwargs):
...
+                        # .glob() expands paths (only files are documented but
+                        # in test they do) and files paths are OK with this function
+                        out |= set(
+                            self.expand_path(
+                                list(bit), recursive=recursive, maxdepth=maxdepth
+                            )
+                        )
                     continue
                 elif recursive:
+                    # TODO: maxdepth is not used and probably should
```
martindurant, line 812: Agreed, `find` does accept `maxdepth`, so add it here.

fsspec/tests/test_spec.py
```diff
@@ -124,6 +134,49 @@ def test_glob(test_path, expected):
...
+                "top_level/second_level/date=2019-10-02",
+                "top_level/second_level/date=2019-10-02/a.parquet",
+                "top_level/second_level/date=2019-10-04",
+                "top_level/second_level/date=2019-10-04/a.parquet",
+            ],
+        ),
+        (("misc/foo.txt", "misc/*.txt"), ["misc/foo.txt"]),
+        # Note: fails because 'glob_test' paths are missing (intentionally?) from ls,
+        #  remove comments and add id to fail it
+        # Todo: figure out what test_fs.ls("") ought to return
```
martindurant, line 156: I would think `["glob_test", "misc", "top_level"]`

We assume "" is not itself a directory, but it could be thought of as one.

martindurant:
Note: this is *only* failing on lint now, but I would appreciate seeing how the commented case fails. 

If you merge from master, tests will run with github actions and be much faster.

martindurant:
(sorry for being slow to comment here, @ahirner )

ahirner:
Thanks @martindurant for finishing it off. As other work on fsspec, this is much appreciated and helpful to us.

PR #465, opened 2020-11-02
Add reference FS to registry and add docs

martindurant:


martindurant:
Adds one simple test. Can add more.

PR #469, opened 2020-11-11
add kwargs to exists

arzt:
add kwargs to `exists(...)` function and passes them to call of `info(...)`

Why is this meaningful?

When calling exists function in s3fs implementation the result is cached. So a second call will return `True` again, even if the file was deleted between the calls. The cache can be bypassed by providing parameter `refresh=True` to the call of `info(...)`.

This change solves this issue by simply forwarding all `kwargs` to `info(...)`

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -552,10 +552,10 @@ class AbstractFileSystem(up, metaclass=_Cached):
         else:
             return list(out)
 
-    def exists(self, path):
+    def exists(self, path, **kwargs):
         """Is there a file at the given path"""
         try:
-            self.info(path)
+            self.info(path, kwargs)
             return True
         except:  # noqa: E722
             # any exception allowed bar FileNotFoundError?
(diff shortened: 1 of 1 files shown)


martindurant:
Did you push another commit and then revert? TravisCI reports:
"This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#1137) for pull request #469 came in while this job was waiting to be processed."
 but I can't find any trace of that more recent build.

arzt:
Yeah, sorry. I pushed another commit, but no revert. this build seems to have worked: https://travis-ci.org/github/intake/filesystem_spec/builds/742930201

martindurant:
Weird... OK, I'll put this in. Note that some downstream implementations may have overrides of exists() and don't expect kwargs - but they will have their own docstrings in that case.

Specifically for s3fs, I assume you were deleting the files from some other process. You can also use invalidate_cache() to control when cached listings are expired.

PR #477, opened 2020-11-18
Http serialising files

martindurant:


martindurant:
Fixes https://github.com/pydata/xarray/issues/4591

martindurant:
Note: file position will be implicitly reset to start; which maybe OK or maybe not. Really, the open file ought not to be passed around via serialisation.

rabernat:
Does this have any relationship to https://github.com/intake/filesystem_spec/issues/476?

martindurant:
Yes and no.
When dask handles files for its various IO operations, it passes file system or OpenFile objects around, and things are only opened while the given function is running. The netCDF version seems to keep hold of the open file instead of the URL, and the HTTP variant of this was not serializable until that PR. Still, it should fix your issue.

PR #482, opened 2020-11-19
Add option to ignore file not found errors in copy

baxen:
Ignores by default in recursive copy but is not on by default for
non-recursive.

Fix first half of dask/gcsfs#307

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -196,14 +196,25 @@ class AsyncFileSystem(AbstractFileSystem):
         maybe_sync(self._rm, self, path, **kwargs)
 
     async def _copy(self, paths, path2, **kwargs):
-        await asyncio.gather(
-            *[self._cp_file(p1, p2, **kwargs) for p1, p2 in zip(paths, path2)]
+        return await asyncio.gather(
+            *[self._cp_file(p1, p2, **kwargs) for p1, p2 in zip(paths, path2)],
+            return_exceptions=True
         )
 
-    def copy(self, path1, path2, recursive=False, **kwargs):
+    def copy(self, path1, path2, recursive=False, on_error=None, **kwargs):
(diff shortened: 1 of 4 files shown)


fsspec/asyn.py
```diff
@@ -196,14 +196,25 @@ def rm(self, path, recursive=False, **kwargs):
...
+        elif on_error is None:
+            on_error = "raise"
+
         paths = self.expand_path(path1, recursive=recursive)
         path2 = other_paths(paths, path2)
-        maybe_sync(self._copy, self, paths, path2, **kwargs)
+        result = maybe_sync(self._copy, self, paths, path2, **kwargs)
+
+        for ex in filter(is_exception, result):
+            if on_error == "ignore" and isinstance(ex, FileNotFoundError):
```
martindurant, line 215: This is OK for now, but I think we may want to make the exceptions to ignore configurable in the future.

martindurant:
The PR is failing on lint - please merge from master, which has updated the version of `black` used in the CI.

This could use a test! You would want to copy with a list of files, which should raise unless "ignore" is given explicitly, but a cp-recursive will ignore unless "raise" is given (to trigger this, you will need a new fs instance to remove a file in a directory). I *think* you can only test the blocking versions here, and the async versions will need to be tested in gcsfs/s3fs. This is a prod to me to make more FSs async, to better test the interface. sftp or webhdfs would be good candidates.

baxen:
I added a test for sync in local filesystem here, and then a test for asyn in gcsfs in dask/gcsfs#309. That will not pass of course until this is merged and the tests on gcsfs include the udpates. But they do pass with an editable install:

image

martindurant:
I'll just wait for linting to return green, then merge. 
Thanks for doing this! If you have enthusiasm, you may wish to try `rm` next.

PR #492, opened 2020-12-04
Fix typo in code example

cthoyt:


cthoyt:
@martindurant that's a personal record for speed of merging a pull request 🥂

martindurant:
Still only average if you're measuring per character :)

Issue #495, opened 2020-12-08
feature request: dbfs


raybellwaves:
dbfs as in Databricks File System (https://docs.databricks.com/data/databricks-file-system.html)

I can write the pyspark.sql.dataframe.Dataframe to a parquet file and use data.dataframe.read_parquet to read it in. However, I tend to use tables (dbfs) for efficient usage on databricks. I also use delta tables which are essentially a way of adding version numbers to tables (https://docs.delta.io/latest/index.html).

It would be nice to bring this data into the pydata ecosystem for creating dashboards, EDA, machine learning etc.

I started taking a look at this over at dask-sql using pyhive https://github.com/nils-braun/dask-sql/issues/83

martindurant:
Looks very doable given the API at https://docs.databricks.com/dev-tools/api/latest/dbfs.html 

However, the docs seems to suggest that intended usage is to *mount* the FS, so you can interact with it as local files. Is this method lacking? Are you looking for something that you can access from outside the cluster?

raybellwaves:
Thanks. Wasn't familiar with that API. I'll check it out.

Yes. Ideally outside of the cluster. e.g. (for dask sql) instead of
`c.create_table("taxi", "az://nyctlc/green/puYear=2019/puMonth=1/*.parquet", storage_options=storage_options)`
it could be
`c.create_table("taxi", "dbfs://nyctlc.green", storage_options=storage_options)`

martindurant:
Setting it up looks pretty simple, but I have no idea how you would go about testing it

raybellwaves:
Likewise. Would be nice to get help from databricks folks. Or if there is something like azurite which is used for testing adlfs.

martindurant:
> if there is something like azurite

I'm not aware of it, but someone should at least search around.

nils-braun:
I have unfortunately also not found any test/mock software (in general it seems, the ecosystem for client software to be used outside of the cluster is quite rare for dbfs). I think one would need to mock the API ourselves...
However, as we only need the `dbfs` part, it is only around 10 endpoints. I am happy to help here if you like!

martindurant:
I suppose we could always record with vcrpy, so long as *some* developer has access to a real cluster; but my experience with gcsfs suggests this is often more pain than it's worth.

> However, as we only need the dbfs part, it is only around 10 endpoints. I am happy to help here if you like!

That would be great! Perhaps we could make our own mini "brickulite".

nils-braun:
[...]

PR #504, opened 2020-12-18
Add DBFS scratch

martindurant:
Start for #495 (completely untested, just for fun)


martindurant:
@raybellwaves , this is done in sync mode - could be made to match the async stuff in the http module.

martindurant:
cc @nils-braun

nils-braun:
Nice start! Definitely something we can build on :-)

nils-braun:
Thanks for making me a collaborator on this branch, @martindurant. I pushed some linting fixes and added tests using vcr (as we discussed, this is not the best solution but should at least serve as a first iteration). I have also described how to record the test casettes again. 
I have also added some more documentation.

martindurant:
Looks good!
I'd be happy to merge this and continue to iterate.

I was unaware of `@pytest.mark.vcr()`! Perhaps things in gcsfs can be simplified.

A couple of thoughts:
- the test only runs on windows, I didn't see changes in the other envs?
- we should change MANIFEST not to include the recordings in the pypi package, which we want to keep small; indeed, perhaps all test files should be pruned.

nils-braun:
Thanks for the review!
* I added the "pytest-vcr" to the `tox.ini`, which is used for testing on linux (you can see e.g. [here](https://github.com/intake/filesystem_spec/pull/504/checks?check_run_id=1617314228#step:4:89) that the test file is used). In general, the environment for non-windows only contains python and tox, the rest is controlled via the tox file.
* Very good point! Will do so.

nils-braun:
Small update: I just checked the [docu](https://packaging.python.org/guides/using-manifest-in/) and what actually happens when I call `python setup.py sdist`  (or `python setup.py bdist_wheel`). In both cases, the tests and everything in it (including the record files) are already excluded. So I think we do not need to do anything here.

martindurant:
Really? I opened https://github.com/intake/filesystem_spec/pull/512, and the package size before/after seem to differ quite a lot (much of that is docs). 
Maybe I am not surprised by the yaml files not being picked up... so perhaps you are right and this is not an issue.

martindurant:
> seem to differ quite a lot

correction: only the sdist; so maybe it was indeed due to docs alone.

nils-braun:
@martindurant - did you had the time to check again the sdist size issue?
If not, from my side we would be ready to merge as a first iteration. Is there some documentation that I should add somewhere?

martindurant:
> did you had the time to check again the sdist size issue?

I'm not bothered in the end, specially since you are definitely *not* adding the yamls here. I was waiting for you to convert from draft PR. Looking again now.

martindurant:
To do on the next iteration: add to API docs under Built-in Implementations, and add to changelog.

nils-braun:
> > did you had the time to check again the sdist size issue?
> 
> I'm not bothered in the end, specially since you are definitely *not* adding the yamls here. I was waiting for you to convert from draft PR. Looking again now.

Ah sorry, yes that is correct. Did not know I can also do this. Thanks!

raybellwaves:
Thanks for this @martindurant and @nils-braun !

raybellwaves:
Just learned about this as an FYI: https://github.com/DataThirstLtd/DBFS-Explorer

martindurant:
Interesting - two years old and untouched since. I wonder if anyone uses fsspec's GUI?

raybellwaves:
TIL! https://github.com/intake/filesystem_spec/blob/master/fsspec/gui.py

martindurant:
https://filesystem-spec.readthedocs.io/en/latest/features.html#file-selector

QuLogic:
The VCR cassettes here don't appear to be correct, as these tests attempt to connect somewhere and fail when there's no network.

QuLogic:
Ah, that was my mistake, I forgot to install `pytest-vcr`. Sorry for the noise.

Issue #506, opened 2020-12-20
HTTP backend has flaky behavior when enumerating the directory index


amotl:
Dear @martindurant, @TomAugspurger and contributors,

first things first: Thanks a stack for conceiving and maintaining this excellent library.

@gutzbenj is currently migrating the data acquisition backend of [wetterdienst](https://github.com/earthobservations/wetterdienst) to use `fsspec` for downloading data from the HTTP server at https://opendata.dwd.de/. For ingesting the directory index, we are currently using a hand-rolled implementation [1] based on BeautifulSoup and lxml.

Now that we are switching to fsspec, we found that it would **randomly** include a **single folder** within its results list when enumerating a remote HTTP filesystem using `fs.find(url)`. It is expected that it would only return **files** as results, right? We added a concise repro at [2].

The respective folder is not always the same, so the behavior is flaky. It can be revealed by invoking the examples [1] vs. [2] by using:
```
# Run current implementation.
python list-remote-files-dwd.py | grep -v zip

# Run implementation based on fsspec.
python fsspec-dwd.py | grep -v zip
```

Alternatively, this produces a deterministic yet wrong result when counting the number of enumerated files:
```
python list-remote-files-dwd.py | wc -l
  194595
python fsspec-dwd.py | wc -l
  194596
```
The result list count of the fsspec-based implementation is off by one.

Thank you in advance for taking the time to look into this.

Keep up the spirit and with kind regards,
Andreas.

[1] https://gist.github.com/gutzbenj/88e69d10447698d099f7227a731add9b
[2] https://gist.github.com/amotl/9fc67b696cbab9f0667be60de4dcf2be

P.S.: We are using Python 3.8 and fsspec 0.8.5.

amotl:
We thought it would only happen on large trees, but actually it doesn't. So we amended the repro programs to run on arbitrary URLs and wanted to take the chance to provide an alternative DWD_URL because the runtime on it is much lower.

### Setup
```
wget https://gist.githubusercontent.com/gutzbenj/88e69d10447698d099f7227a731add9b/raw/list-remote-files-dwd.py
wget https://gist.githubusercontent.com/amotl/9fc67b696cbab9f0667be60de4dcf2be/raw/fsspec-dwd.py
```

### Configure
```
export DWD_URL=https://opendata.dwd.de/climate_environment/CDC/observations_germany/climate/daily
```

### Invoke
And then run the examples like:
```
# Run current implementation.
[...]

PR #507, opened 2020-12-21
Fix appending of random directory within find() method

gutzbenj:
At #506 we've seen that when using the HTTPFileSystem's ``find`` method, it appears that the method would randomly add one of the found directories to the returned files listing. This is due to conflicting variable naming here:
https://github.com/intake/filesystem_spec/blob/a5bb95155b450ec2443872a45a4a58ce90748297/fsspec/spec.py#L430-L437

To fix this I renamed resulting **path** from ``self.walk()`` to **path_**. See here:
https://github.com/earthobservations/filesystem_spec/blob/e3956827181393b30a57ab6d11ede52931a4dec8/fsspec/spec.py#L430-L437

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -427,7 +427,7 @@ class AbstractFileSystem(up, metaclass=_Cached):
         path = self._strip_protocol(path)
         out = dict()
         detail = kwargs.pop("detail", False)
-        for path_, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
+        for _, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
             if withdirs:
                 files.update(dirs)
             out.update({info["name"]: info for name, info in files.items()})
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -427,7 +427,7 @@ class AbstractFileSystem(up, metaclass=_Cached):
         path = self._strip_protocol(path)
         out = dict()
         detail = kwargs.pop("detail", False)
-        for path, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
+        for path_, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
             if withdirs:
                 files.update(dirs)
             out.update({info["name"]: info for name, info in files.items()})
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Dev
+-------------
+
+Fixes:
+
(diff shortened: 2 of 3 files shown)


martindurant:
Thanks for finding this! 

I expect this didn't show up with existing tests of `find` because, unlike the HTTP case, we can filter out directories easily, right? It would be nice to have a test for HTTP that would have shown the failure, but is now fixed after this change.

martindurant:
Fixes #506

fsspec/spec.py
```diff
@@ -427,7 +427,7 @@ def find(self, path, maxdepth=None, withdirs=False, **kwargs):
         path = self._strip_protocol(path)
         out = dict()
         detail = kwargs.pop("detail", False)
-        for path, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
+        for path_, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
```
amotl, line 430: Maybe we can say just this here:
```suggestion
        for _, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
```

gutzbenj:
Dear Martin,

I added two tests for the FTP and HTTP filesystem querying the Debian repo at [1]. As you can see from the test
- the url which was queried is included in HTTPs result while for FTP this one is missing
- for the case of HTTP this one has found the additional filter hrefs at this directory
- the FTP method returns the relative paths

I propose that the FTP method is aligned with the one from HTTP by adding the base url to the found files thus returning the full filepath. Would you agree on this point?

[1] https://ftp.fau.de/debian-cd/current/amd64/log/success/

martindurant:
> I propose that the FTP method is aligned with the one from HTTP by adding the base url to the found files thus returning the full filepath. Would you agree on this point?

I don't think I do. In the case of FTP, the host/port are arguments to the class, so each set of arguments is an independent file-system with a set of paths within it (which may depend on the specific user info given) and a single persistent connection. Conversely, HTTP sees the internet as a single filesystem, and makes new connections for every access. Although it's true that you can phrase a unique FTP path like "ftp://user:pass@host:port/path", this is for convenience only. Obviously for a case like this, where you have FTP-on-HTTP, the distinction is blurred.

> for the case of HTTP this one has found the additional filter hrefs at this directory

I suppose this is OK, though one wonders what these things are, and whether they are likely to stay the same in the future.

martindurant:
Ah, those hrefs are new sorted views of the same directory. There is no way to know that they are not new content.

Ready to merge now?

martindurant:
Fixes https://github.com/dask/s3fs/issues/393

gutzbenj:
Yes, ready.

PR #519, opened 2021-01-19
Fix mapping tests on big-endian machines.

QuLogic:


martindurant:
Thank you, this seems very reasonable.
I wonder, what is the big-endian system in question?

QuLogic:
It is Fedora on s390x.

PR #522, opened 2021-01-22
Mark 3.6.0 as supported as well?

mariusvniekerk:
Technically we didn't support 3.6.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     license="BSD",
     keywords="file",
     packages=["fsspec", "fsspec.implementations"],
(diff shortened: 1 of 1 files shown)


martindurant:
3.6 is *poorly* supported by the async, if any nested calls are made (which is quite likely)

mariusvniekerk:
we should probably then move it to `>=3.7.0`

mariusvniekerk:
since atm we support `3.6.1` but not `3.6.0`

martindurant:
In that case, I suggest >=3.7. I've probably made this mistake elsewhere...

PR #533, opened 2021-02-09
Fix write when using transaction context

### LINKED ISSUE #531: AttributeError: '_io.BufferedReader' object has no attribute 'temp'
Hello,

I try to use `LocalFileSystem` with Transaction context manager and I get the following error:

```
AttributeError: '_io.BufferedReader' object has no attribute 'temp'
```

An example:

```python

def main():
    fs = LocalFileSystem()

    with fs.transaction:
        with fs.open('test.txt', 'w') as fp:
            fp.write('hello')

        with fs.open('test.txt', 'r') as fp:
            s = fp.read()
        
        print(s)

if __name__ == '__main__':
    main()
```

An exception:

```
  File "scratches\scratch_94.py", line 18, in <module>
    main()
  File "scratches\scratch_94.py", line 14, in main
    print(s)
  File "lib\site-packages\fsspec\transaction.py", line 24, in __exit__
    self.complete(commit=exc_type is None)
  File "lib\site-packages\fsspec\transaction.py", line 37, in complete
    f.commit()
  File "lib\site-packages\fsspec\implementations\local.py", line 246, in commit
    os.replace(self.temp, self.path)
  File "lib\site-packages\fsspec\implementations\local.py", line 261, in __getattr__
    return getattr(self.f, item)
AttributeError: '_io.BufferedReader' object has no attribute 'temp'
```

If I only use write operations within transaction context it works but read operation leads to the exception.

andersy005:
Fixes #531

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -594,3 +594,18 @@ def test_copy_errors(tmpdir):
             make_path_posix(os.path.join(dest2, "afile1")),
             make_path_posix(os.path.join(dest2, "afile2")),
         ]
+
+
+def test_transaction(tmpdir):
+    file = (tmpdir / "test.txt").as_posix()
+    fs = LocalFileSystem()
+
+    with fs.transaction:
+        content = "hello world"
+        with fs.open(file, "w") as fp:
+            fp.write(content)
+
+    with fs.open(file, "r") as fp:
+        read_content = fp.read()
+
+    assert content == read_content
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -597,7 +597,7 @@ def test_copy_errors(tmpdir):
 
 
 def test_transaction(tmpdir):
-    file = (tmpdir / "test.txt").as_posix()
+    file = str(tmpdir / "test.txt")
     fs = LocalFileSystem()
 
     with fs.transaction:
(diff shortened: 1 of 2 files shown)


martindurant:
This seems to be failing lint, but the gha details box is bizarrely empty

andersy005:
> but the gha details box is bizarrely empty 

GitHub is experiencing [some issues](https://www.githubstatus.com/) 

> This seems to be failing lint,

I used the pre-commit hooks locally. I am noticing that CI runs black and flake8 directly. Will fix the linting issues shortly

martindurant:
It would be reasonable to make an effort to have the CI and commit hooks be identical. I have found that new versions of `black` often have small, subtle, annoying changes.

andersy005:
@martindurant, unless I am missing something, the failing tests seem unrelated to changes in this PR

fsspec/implementations/tests/test_local.py
```diff
@@ -597,7 +597,7 @@ def test_copy_errors(tmpdir):
 
 
 def test_transaction(tmpdir):
-    file = (tmpdir / "test.txt").as_posix()
+    file = str(tmpdir / "test.txt")
```
martindurant, line 600: Would this do the right thing on Windows?

espdev: I can confirm that on windows we need to use as_posix path representation. I just recently fixed such issue in my code.

martindurant:
Test passes also on windows, so I will merge. This is actually a good demonstration that fsspec is doing the right thing with windows, at least for local paths.

PR #534, opened 2021-02-09
Make CI linting identical to pre-commit hooks linting

andersy005:
Addresses @martindurant's comment in https://github.com/intake/filesystem_spec/pull/533#issuecomment-776070058


martindurant:
That's a lot of changes!

andersy005:
> That's a lot of changes! 

:) Should be fixed now.

martindurant:
The failing test is touching an external network endpoint for ftp find, where the contents seem to have changed. The test should be removed, or changed to a set of `in` asserts rather than exact output equality.

andersy005:
> The test should be removed, or changed to a set of in asserts rather than exact output equality. 

Do you want these fixed in this PR?

martindurant:
Yes please, if you have the time

PR #535, opened 2021-02-09
DOC: cell blocks for installs

raybellwaves:
Can't say i've built the docs but the current page is missing some cell blocks

https://filesystem-spec.readthedocs.io/en/latest/index.html

Screen Shot 2021-02-09 at 4 09 17 PM


martindurant:
(failures I hope will be fixed in #534 )

PR #537, opened 2021-02-10
Fix find test

martindurant:


QuLogic:
This test should be VCR'd?

martindurant:
This uses FTP with a real server

PR #547, opened 2021-02-23
Fix invalid cross-device link when moving files for local fs

### LINKED ISSUE #544: LocalFileSystem: OSError: [Errno 18] Invalid cross-device link
Hello,

I get the following error when I try to use `LocalFileSystem` on WSL and write a file to Windows directory (the directory on host):

```
  File "... /fsspec/transaction.py", line 24, in __exit__
    self.complete(commit=exc_type is None)
  File "... /fsspec/transaction.py", line 37, in complete
    f.commit()
  File "... /fsspec/implementations/local.py", line 246, in commit
    os.replace(self.temp, self.path)
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmpursl97uu' -> '/mnt/c/Temp/file_name'
``` 

I think that when we use `os.replace` we will get cross-device link errors whenever the devices of the source and the destination paths are not same. Most likely I will get the same error when I will try to write to path on any mounted FS from another device. I found a similar issue: https://github.com/intake/filesystem_spec/issues/322

espdev:
Fixes #544

We can use `shutil.move` instead of `os.rename` to move files between block devices (file-systems). If the destination is on the current file-system, then `os.rename()` is used inside `shutil.move`.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -102,7 +102,7 @@ class LocalFileSystem(AbstractFileSystem):
     def mv_file(self, path1, path2, **kwargs):
         path1 = self._strip_protocol(path1).rstrip("/")
         path2 = self._strip_protocol(path2).rstrip("/")
-        os.rename(path1, path2)
+        shutil.move(path1, path2)
 
     def rm(self, path, recursive=False, maxdepth=None):
         path = self._strip_protocol(path).rstrip("/")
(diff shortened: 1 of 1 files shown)


fsspec/implementations/local.py
```diff
@@ -102,7 +102,7 @@ def put_file(self, path1, path2, **kwargs):
     def mv_file(self, path1, path2, **kwargs):
         path1 = self._strip_protocol(path1).rstrip("/")
         path2 = self._strip_protocol(path2).rstrip("/")
-        os.rename(path1, path2)
```
espdev, line 105: `mv_file` method does not inherit the interface of `AbstractFileSystem` class. It looks a little strange to me.

martindurant: In the base, we have cp which calls cp_file and rm which calls rm_file. The base mv uses these two, but here (and some other cases) we have a single call that will do it.

PR #548, opened 2021-02-27
document the use of entry_points to register new backends

d70-t:
I just found out that registering new backends is now possible using `entry_points`, thanks to #515. I think that this feature needs at least some basic documentation, so this is what this PR is for.

One think to **note** is that it actually works a little different from what is written in #515:

This is from the [initial message in 515](https://github.com/intake/filesystem_spec/pull/515#issue-548432617)
>```python
>    entry_points=[
>        'fsspec.specs': [
>            'myfs=myfs:MyFs',
>        ]
>    ],
>```

and this is what actually works:

```python
    entry_points={
        'fsspec.specs': [
            'myfs=myfs.MyFS',
        ],  
    },
```

The change of `[]` to `{}` seems to be a typo, but I am not sure if that is the case for `:` -> `.`. According to the [setuptools manual](https://setuptools.readthedocs.io/en/latest/userguide/entry_point.html), the syntax for defining entry points is:

```
<name> = [<package>.[<subpackage>.]]<module>[:<object>.<object>]
```

and thus the original comment in #515 would be correct in this regard, but the code is maybe wrong.

---

This PR documents what currently works. But probably we should change the code and the docs such that the filesystem **object** will be separated by a `:` in the `entry_points`.

### DIFF
--- a/docs/source/developer.rst
+++ b/docs/source/developer.rst
@@ -66,7 +66,27 @@ We may eventually refactor ``AbstractFileSystem`` to split the default implement
 the set of methods that you might implement in a new backend, and the
 documented end-user API.
 
-For now, new backends must register themselves on import
+In order to register a new backend with fsspec, new backends should register
+themselves using the `entry_points <https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html#entry-points-and-automatic-script-creation>`_
+facility from setuptools. In particular, if you want to register a new
+filesystem protocol ``myfs`` which is provided by the ``MyFS`` class in
+the ``myfs`` package, add the following to your ``setup.py``:
+
+.. code-block:: python
+
+    setuptools.setup(
+        ...
+        entry_points={
+            'fsspec.specs': [
+                'myfs=myfs.MyFS',
+            ],
+        },
+        ...
+    )
+
+
+Alternatively, the previous method of registering a new backend can be used.
+That is, new backends must register themselves on import
 (``register_implementation``) or post a PR to the ``fsspec`` repo
 asking to be included in ``fsspec.registry.known_implementations``.
 
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -42,4 +42,4 @@ if entry_points is not None:
     entry_points = entry_points()
     for spec in entry_points.get("fsspec.specs", []):
         err_msg = f"Unable to load filesystem from {spec}"
-        register_implementation(spec.name, spec.value.replace(':', '.'), errtxt=err_msg)
+        register_implementation(spec.name, spec.value.replace(":", "."), errtxt=err_msg)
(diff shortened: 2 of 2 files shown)


d70-t:
While implementing this for an own backend (d70-t/ipfsspec#2), I discovered that this is not working in Python 3.8 as in that version [`importlib.metadata`](https://docs.python.org/3.8/library/importlib.metadata.html#entry-points) is already available, but its `EntryPoint` object does not yet have a `module` attribute. As this concerns exactly [the line](https://github.com/intake/filesystem_spec/blob/ef1001a66b03155d5411aa0d5f8d830a3587851e/fsspec/__init__.py#L45) which probably must be changed regarding the `:` issue, I'd change this PR into a draft for now and ask how me might proceed?

d70-t:
Currently I think of two possible ways to solve this:

* Replace `spec.module` by `spec.value.replace(':', '.')` in [`__init__.py`](https://github.com/intake/filesystem_spec/blob/ef1001a66b03155d5411aa0d5f8d830a3587851e/fsspec/__init__.py#L45): seems to be a bit hackish, but should work for both, the currently implemented variant as well as if `:` is used in `entry_points`. Note that `.value` is available in Python 3.8.
* Use the [entrypoints](https://pypi.org/project/entrypoints/) library as `intake` does: seems to be more proper, but pulls in an additional dependency and is more complicated

martindurant:
Given the current situation, we should probably support both "package.module.class" and "package.module:class", sorry. Should be simple
```python
if ":" in spec:
     mod, class = spec.split(":", 1)
else:
    mod, class = spec.rsplit(".", 1)

**Comment by d70-t** [comment]
If I get it right, at this place the only thing required would be a dotted string, without splitting it into `mod` and `class`. I've added this changed to the PR for illustration. I think it would work like this.

There's however one more thing which bothers me. If any backend library which uses the `fsspec.specs` entry point is installed in a python 3.8 environment and a version of `fsspec` which does not include this PR is installed, then a simple
```python
import fsspec
```
will crash. We could either not bother and hope that prior to a wider adoption of the entry point, users will have upgrade the `fsspec` version or change the name of the entry point.

martindurant:
I don't see a problem with the current implementation here. If someone follows the cods, it will work, and if someone uses the entrypoints convention, it will also work. Is there any case where the part after the ":" might be more complex? I don't think so - or at least, we don't need to support it.

d70-t:
I also don't see a problem with this implementation. According to the setuptools manual, there will only be `.`-separated objects after the `:`.

Should I update the docs to recommend `:` then, to be in accordance with setuptools?

martindurant:
> Should I update the docs to recommend : then, to be in accordance with setuptools?

I don't think so, because the code doesn't actually handle it - it will attempt to `import` the whole string before the final ".". i.e., just because you can do `import mod; mod.classy.objecty`, doesn't mean you can do `import mod.classy`. 

It would be reasonable to respect the exact intention of ":" in the given string, if it's there.

d70-t:
Ok, that's reasonable. I'll leave it as is.

PR #557, opened 2021-03-04
spec: normalize the key for _ls_from_cache

### LINKED ISSUE #554: _ls_from_cache: weird behavior regarding comparisons with trailing slash
In this example, the path is stripped only for the contain check. When dircache contains the normalized version (`a/b`) but not `a/b/` it will throw a `KeyError` though it probably shouldn't since this is what that check appears to be stand for. Another case is right below that code that compares the normalized version with the obtained one from the cache, which might not be normalized. So I was wondering whether this is intentional or should we use `.rstrip()` in both of the cases (`self.dircache[path.rstrip('/')]`, `f["name"].rstrip("/") == path.rstrip("/")`) 
https://github.com/intake/filesystem_spec/blob/6ee47aa6c10868c383be5cb13837f2a26a099f5d/fsspec/spec.py#L342-L343
https://github.com/intake/filesystem_spec/blob/6ee47aa6c10868c383be5cb13837f2a26a099f5d/fsspec/spec.py#L349

isidentical:
Resolves #554

### DIFF
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -310,7 +310,7 @@ def test_ls_from_cache():
     # to actually verify that our stripping in the client
     # function works.
     assert (
-        fs.ls("top_level/second_level/", refresh=False, strip_proto=True)
+        fs.ls("top_level/second_level/", refresh=False, strip_proto=False)
         == uncached_results
     )
 
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -57,19 +57,21 @@ class DummyTestFS(AbstractFileSystem):
                 return item
         raise IndexError("{name} not found!".format(name=name))
 
-    def ls(self, path, detail=True, **kwargs):
-        path = self._strip_protocol(path)
+    def ls(self, path, detail=True, refresh=True, **kwargs):
+        if kwargs.pop("strip_proto", True):
+            path = self._strip_protocol(path)
 
-        files = {
-            file["name"]: file
-            for file in self._fs_contents
-            if path == self._parent(file["name"])
-        }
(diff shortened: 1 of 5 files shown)


isidentical:
@martindurant do you have access to a databricks instance? It appears to be that some operations doesn't invalidate the cache as of now (like .rm()/.mv() etc) which is uncovered by this PR though since I don't have access to such instance I am unable to record VCRs and test them in the CI. It would be much appreciated if you could push the latest recordings to my branch (or help me to setup in another way).

martindurant:
No, I'm afraid I don't

cc @nils-braun @andersy005

fsspec/tests/test_spec.py
```diff
@@ -57,19 +57,21 @@ def __getitem__(self, name):
...
-    def ls(self, path, detail=True, **kwargs):
-        path = self._strip_protocol(path)
-
-        files = {
-            file["name"]: file
-            for file in self._fs_contents
-            if path == self._parent(file["name"])
-        }
+    def ls(self, path, detail=True, refresh=True, **kwargs):
+        if kwargs.pop("strip_proto", True):
```
martindurant, line 61: Where is this ever `False`?

nils-braun:
I have created a PR with fixed cassettes. Thanks for the mention!

isidentical:
Thanks for regenerating cassettes @nils-braun

martindurant:
Do you have time to run the s3fs tests and figure out why two tests are failing? It may not be a coincidence that moto 2.0 was just released (i.e., setting the install req to <2 might fix this)

martindurant:
rerunning

martindurant:
Oh dear, I'm sorry: it needs changing in *this* repo in tox.ini under `[testenv:s3fs]`/`conda_deps`

PR #558, opened 2021-03-04
RF/BF: http - do not _fetch_all if file is smaller than blocksize/requested

yarikoptic:
This should prevent `_fetch_all` to overload the .cache with AllBytes
within async_fetch_all and thus use originally provided .cache and
store file on disk.

It is just a partial remedy/workaround to
https://github.com/intake/filesystem_spec/issues/553
and a proper solution is likely needing RFing so async_fetch_all could
feed the original .cache with the data for all blocks.

I just was looking around and decided not to stay silent and propose a PR -- feel free to drop it, but it seems to resolve the issue in my sample case at least and might give food for thought ;-)

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -431,6 +431,8 @@ class HTTPFile(AbstractBufferedFile):
         """
         if (
             (length < 0 and self.loc == 0)  # explicit read all
+            # but not when the size is known and fits into a block anyways
+            and not (self.size is not None and self.size <= self.blocksize)
         ):
             self._fetch_all()
         if self.size is None:
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -430,11 +430,7 @@ class HTTPFile(AbstractBufferedFile):
             read only part of the data will raise a ValueError.
         """
         if (
-            (length < 0 and self.loc == 0)
-            or (length > (self.size or length))  # explicit read all
-            or (  # read more than there is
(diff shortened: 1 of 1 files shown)


martindurant:
I think is `self.size` is known (not None) and less than a blocksize, then we want to pass through to the super() line, as the standard cache will take care of it. That is *unless* the cache is set to no-op-cache, but the user would have had to explicitly choose that, so maybe we should allow it.

yarikoptic:
> I think is `self.size` is known (not None) and less than a blocksize,  then we want to pass through to the super() line...

do you mean "even if `length=-1`"? if not - isn't that what would happen now with this patch?

martindurant:
Correct: since the whole file would be contained in a single block, why bother intercepting the normal caches?

yarikoptic:
something like a61f910babddace5059af80c7e25d5f5ca45e80f ?

martindurant:
Yes I believe so. This might result in a partial read f that block if a seek has already happened, but should behave OK.

yarikoptic:
looking at it again though -- wouldn't we have a side-effect from `.loc` somehow?  could I "seek" to the middle of the block somehow, get that tail, and then request full -- wouldn't `.loc` cause not fetching the leading part then? (I just don't know how `.loc` could be adjusted besides sequential growth there in `HTTPStreamFile._read`)

martindurant:
You would probably get the whole of the one and only block, but in two reads instead of one. Worth testing.

PR #560, opened 2021-03-06
Enable listings cache for HTTP filesystem

amotl:
Dear Martin,

we recognized your activity on the code base again and feared we would have missed some merge window already. So, we tried to hurry to bring in this patch I had slumbering on my workstation for some time already.

The patch will enable listings caching for the HTTP filesystem implementation. It is accompanied by appropriate tests and I will be happy if you could take the time to look into it and consider it for the next release if you like it. Please let us know about any amendments you would like to see.

With kind regards,
Andreas.

P.S.: After @gutzbenj submitted #507 the other day, this patch would be the last thing we would need for finally getting rid of the dbmfile-based cache we still use within Wetterdienst, see https://github.com/earthobservations/wetterdienst/issues/243. Because of this, we haven't migrated the code base to `filesystem_spec` yet, but would be very keen doing so.

On the CI side, as the old cache implementation is apparently not thread-safe (https://github.com/earthobservations/wetterdienst/pull/254#issuecomment-733946897), this would hopefully enable us to run all tests in parallel using [pytest-xdist](https://pypi.org/project/pytest-xdist), see https://github.com/earthobservations/wetterdienst/issues/253. At the same time, it would be a tremendous improvement for our users and also for ourselves, as we would be able to remove some code from the code base and stand on the shoulders of `filesystem_spec`.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import re
 import weakref
+from copy import copy
 from urllib.parse import urlparse
 
 import aiohttp
@@ -82,6 +83,18 @@ class HTTPFileSystem(AsyncFileSystem):
         self.client_kwargs = client_kwargs or {}
         self.kwargs = storage_options
         self._session = None
+
+        # Clean caching-related parameters from `storage_options`
+        # before propagating them as `request_options` through `self.kwargs`.
+        # TODO: Maybe rename `self.kwargs` to `self.request_options` to make
+        #       it clearer.
+        request_options = copy(storage_options)
+        self.use_listings_cache = request_options.pop("use_listings_cache", False)
+        request_options.pop("listings_expiry_time", None)
+        request_options.pop("max_paths", None)
+        request_options.pop("skip_instance_cache", None)
+        self.kwargs = request_options
+
         if not asynchronous:
             sync(self.loop, self.set_session)
 
@@ -113,7 +126,7 @@ class HTTPFileSystem(AsyncFileSystem):
             return par
         return ""
 
-    async def _ls(self, url, detail=True, **kwargs):
+    async def _ls_real(self, url, detail=True, **kwargs):
         # ignoring URL-encoded arguments
         kw = self.kwargs.copy()
         kw.update(kwargs)
@@ -147,7 +160,7 @@ class HTTPFileSystem(AsyncFileSystem):
                     # Ignore FTP-like "parent"
                     out.add("/".join([url.rstrip("/"), l.lstrip("/")]))
         if not out and url.endswith("/"):
-            out = await self._ls(url.rstrip("/"), detail=False)
+            out = await self._ls_real(url.rstrip("/"), detail=False)
         if detail:
             return [
                 {
@@ -159,6 +172,16 @@ class HTTPFileSystem(AsyncFileSystem):
             ]
         else:
             return list(sorted(out))
+        return out
+
+    async def _ls(self, url, detail=True, **kwargs):
+
+        if self.use_listings_cache and url in self.dircache:
+            out = self.dircache[url]
+        else:
+            out = await self._ls_real(url, detail=detail, **kwargs)
+            self.dircache[url] = out
+        return out
 
     ls = sync_wrapper(_ls)
 
--- a/fsspec/implementations/tests/test_http.py
+++ b/fsspec/implementations/tests/test_http.py
@@ -3,6 +3,7 @@ import contextlib
 import os
 import sys
 import threading
+import time
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
 import pytest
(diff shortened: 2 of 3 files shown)


fsspec/implementations/http.py
```diff
@@ -142,7 +154,18 @@ async def _ls(self, url, detail=True, **kwargs):
...
                     out.add("/".join([url.rstrip("/"), l.lstrip("/")]))
         if not out and url.endswith("/"):
-            return await self._ls(url.rstrip("/"), detail=True)
+            return await self._ls_real(url.rstrip("/"), detail=True)
+
+        return out
+
+    async def _ls(self, url, detail=True, **kwargs):
+
+        if self.use_listings_cache and url in self.dircache:
```
martindurant, line 163: When self.use_listings_cache is False, url will never be in dircache. Perhaps you don't need to store use_listings_cache at all.

martindurant: Indeed, have you seen [`_ls_from_cache`](https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L335) ?

martindurant (commented):
Thanks for adding! Just a small question.

fsspec/implementations/http.py
amotl: Hi Martin. Thanks for spotting this. Indeed, I haven't seen `_ls_from_cache` before and will make some appropriate amendments. I will probably be able to figure it out by looking at other examples how `_ls_from_cache` is used elsewhere.

amotl: However, I am not sure about the exception raised here?

https://github.com/intake/filesystem_spec/blob/767388a2b87d5281d41d7031618fd733d70d903b/fsspec/spec.py#L351-L353

So, it will trigger on empty directories? Is this intended?

amotl: Thank you for your response and the explanation. If it will just work (TM), I won't argue.

martindurant: Probably if you try _ls_from_cache with tests on a mock server, you'll find enough edges that, in hindsight, I wouldn't use it here :)

fsspec/implementations/tests/test_http.py
```diff
@@ -97,6 +97,36 @@ def test_list(server):
...
+
+
+def test_list_cache(server):
+    h = fsspec.filesystem("http", use_listings_cache=True)
+    out = h.glob(server + "/index/*")
+    assert out == [server + "/index/realfile"]
+
+
+def test_list_cache_with_expiry_time(server):
+    h = fsspec.filesystem("http", use_listings_cache=True, listings_expiry_time=30)
```
martindurant, line 113: You could test these three parametrised cases to actually trigger the condition: have a short listing expiry time and sleep past it; list more than 5 paths; make a new instance with/without the instance cache (one should already have its dircache populated, the other not).

martindurant:
(this PR should be easy to finish off, if you have a little time)

martindurant:
- friendly nudge -

fsspec/implementations/http.py
amotl: Hi Martin,

as you've resolved the conversation, I believe you don't expect me to do anything on this? Please let me know otherwise.

With kind regards,
Andreas.

fsspec/implementations/tests/test_http.py
amotl: Hi Martin,

thank you very much for your suggestions, I tried to implement some ideas from them as good as possible. However, maybe I am still missing some details. If you see the tests should be improved once more, I am humbly asking for your guidance and to elaborate a bit more about the details I might have missed.

> have a short listing expiry time and sleep past it

I just added b1f4cad55 in order to bring in your idea of verifying the actual contents of the directory cache beyond the expiration time.

> make a new instance with/without the instance cache (one should already have its dircache populated, the other not)

Do you mean something along the lines of ca0a61e6?

> list more than 5 paths

I am not sure I am following here. The test is currently using the some response mock as the other tests are also using. Is there some mock which would yield more than 5 paths? Would it actually make any difference?

With kind regards,
Andreas.

amotl:
Hi Martin,

> friendly nudge

Thank you. Finally, I have been able to catch up on this. Please let me know about further suggestions for this patch.

On my workstation, I observed different hiccups where the test suite would stall completely when invoking `pytest -vvvv -k test_http`. On the other hand I, can't see the outcome on CI, because [1] tells us that it would have to be approved beforehand? Did you recently change something on the GHA CI settings on this topic?

With kind regards,
Andreas.

[1] https://github.com/intake/filesystem_spec/actions/runs/780684958

martindurant:
You seem to have managed to hit an async deadlock in this code. I don't see why, but this may be good news - since deadlocks have occasionally occurred elsewhere, and this will give us the ability to investigate.

martindurant:
I cannot push to your branch, but https://github.com/earthobservations/filesystem_spec/pull/1 appears to fix the problem.

PR #305, opened 2020-05-27
WIP: TAR FS

martindurant:
cc #297 
@kmpaul , please have a go!

PR #561, opened 2021-03-07
Improve TarFileSystem and streamline with ZipFileSystem and LibArchiveFileSystem

amotl:
Dear Martin,

this patch builds upon your work at #305 and tries to improve the situation. Thank you very much for the great infrastructure routines already provided by `fsspec` and especially for starting this filesystem adapter the other day already. We need this functionality for Wetterdienst because there are also some tar files around at [1]. A demo program using `TarFileSystem` can be found at [2].

We tried to utilize the "libarchive" adapter first, but apparently it does not work well together with the HTTP filesystem, we will report about it in a separate issue.

So, this patch fills in some gaps of the tar filesystem implementation and accompanies it by respective tests derived from the zip filesystem adapter. Besides filling some gaps to make the tests work, it also adds a snippet to infer the compression type from the file name.

You can see the patch is not finished yet because `test_isdir_isfile` is currently deactivated.

However, we decided to submit it in order to get early feedback from your side. Specifically, we are asking for your opinion about introducing the `AbstractArchiveFileSystem` class. We did that because we found the relevant code to be duplicated between the `libarchive` and `zip` implementations already and didn't want to duplicate it for another time into the `tar` implementation. What is probably also missing is to introduce proper use of `self.dir_cache` into the tarfile implementation, as seen within the others.

With kind regards,
Andreas.

[1] https://opendata.dwd.de/climate_environment/CDC/grids_germany/daily/radolan/historical/bin/2006/*
[2] https://gist.github.com/amotl/f7739878b63df8fec2760a9c3f571533

cc @gutzbenj


amotl:
Hi Martin,

with 54c8e4241, we have been able to make the `test_isdir_isfile` test work. Along the lines, we have been able to identify more redundant code between `ZipFileSystem` and `LibArchiveFileSystem` and refactored it into `AbstractArchiveFileSystem`, so it can also be reused by `TarFileSystem`.

From our perspective, this worked out perfectly well. Please let us know what you think about it.

With kind regards,
Andreas.

P.S.: We added the other commit on purpose to this PR to make it easier for you to review the changes. However, before eventually merging it, we will be happy to squash those commits into two single ones. One for your baseline implementation and another one reflecting our improvements.

amotl:
Hi again,

> Still, we would like to add more tests reflecting some tar compression modes, the patch currently only uses an uncompressed synthesized tarfile for running them on.

3cd5ee6 improves the situation by validating all available tar file compression variants. Also, tests are now additionally run through a real fsspec filesystem implementation, the `LocalFileSystem`.

With kind regards,
Andreas.

martindurant:
Thanks for working on this! I'll have a look on Monday.

amotl:
Hello once more,

we just added 83a1afa, which streamlines archive file testing. It bundles all tests against archive file filesystems (zip, tar, libarchive) into a single test suite in order to make absolutely sure all of those will adhere to the same specification.

We found some anomalies between the test suites against those three different archive implementations and - after bundling together common code into `AbstractArchiveFileSystem` already - we thought it would also be a good idea to refactor the respective test code in order to improve the overall quality of those implementations.

With kind regards,
Andreas.

fsspec/implementations/tar.py
```diff
@@ -0,0 +1,146 @@
...
+from fsspec.spec import AbstractArchiveFileSystem
+from fsspec.utils import infer_compression
+
+typemap = {b"0": "file", b"5": "directory"}
+
+logger = logging.getLogger("tar")
+
+
+class TarFileSystem(AbstractArchiveFileSystem):
+    def __init__(
```
martindurant, line 17: Could use documentation

fsspec/implementations/tar.py
```diff
@@ -0,0 +1,146 @@
...
+            newfo = self.fo
+        else:
+            newfo = copy.copy(self.fo)
+        newfo.seek(offset)
+
+        return TarContainedFile(newfo, self.info(path))
+
+
+class TarContainedFile(object):
+    def __init__(self, of, info):
```
martindurant, line 103: Perhaps needs a repr, so that people know what their object is

fsspec/spec.py
```diff
@@ -1555,3 +1555,57 @@ def __enter__(self):
 
     def __exit__(self, *args):
         self.close()
+
+
+class AbstractArchiveFileSystem(AbstractFileSystem):
```
martindurant, line 1560: I think this should go into a module of its own, since it isn't part of the "spec", but a useful shared piece of code for the three implementations. Also, it should probably not be in the top-level `__all__` or imported in `fsspec.__init__` at all, to keep the namespace clean. No other file systems are imported there, except the "spec", in order to keep import times short and expose the canonical ways to make filesystems, like `fsspec.filesystem`.

martindurant (commented):
Looks like a good improvement, and the refactor certainly seems warranted and clean. I would prefer a separate module for the common superclass, however.

Please add the new implementation to the API docs.

It may be a good idea to add additional tests with file caching before or after dearchive. Where should be write about the concept of "indexing" a tar archive and possibly saving the offsets?

martindurant:
You may be interested by [fsspec-reference-maker](https://github.com/intake/fsspec-reference-maker) which provides a convenient way to view a set of binary chunks within one or more other files as a filesystem of its own. You could certainly apply this to an uncompressed tar. Indeed, if I make headway on unpacking the internal blocking scheme of compression algorithms (bzip and xz seem the only easy ones), then it might work for compressed tar too.

martindurant:
Do you plan on coming back to this in the near future, @amotl ?

amotl:
> Do you plan on coming back to this in the near future, @amotl?

Definitively, but I was a bit swamped since initially submitting this patch. Apologies. Otherwise, would you be keen on taking it over?

amotl:
> It may be a good idea to add additional tests with file caching before or after dearchive.

You hit a very good point here. After trying the new implementation with our demo program [1] in "cached" mode, the flaws immediately became obvious. Currently, because the bookkeeping of the `cached.py` implementation doesn't propagate the original filename (suffix) to the respective filesystem implementation, there is no way for the TarFilesystem to deduce the correct compression and will obviously fail on that.

I was already on my way into `cached.py` in order to evaluate potential improvements before I had to leave the code base due to other obligations.

I will be happy to hear back about your thoughts on this matter.

[1] https://gist.github.com/amotl/f7739878b63df8fec2760a9c3f571533

martindurant:
Sorry, I missed that you asked me a specific question here. 

> I was thinking about extending the detail bookkeeping data structure by an original attribute

This is reasonable detail to include in the file info, yes. Note that you *can* have the cacher use the original file-names too, but this is optional (not the default, because you might have the same name across different file-systems).

martindurant:
- friendly nudge -

fsspec/implementations/tar.py
amotl: I've address this with 21894bf6.

fsspec/implementations/tar.py
amotl: Implemented with e2e4e24c and tested with f2f6b0027.

fsspec/spec.py
amotl: Sure. e2e4e24cb now puts the common superclass into `fsspec.archive`.

amotl:
Dear Martin,

I tried to address more details of this patch in order to make the `TarFileSystem` work together with caching filesystems.

1. Before making any significant changes, 8ae20089b41 improves the test coverage in general.
   - For the generic `test_archive` tests, it adds `.tar.{gz,bz2,xz}` to the mix.
   - It also walks through some scenarios when using `url_to_fs()` to access a `TarFileSystem`.

2. Add support for caching
   > > It may be a good idea to add additional tests with file caching before or after dearchive.
   >
   > Currently, because the bookkeeping of the `cached.py` implementation doesn't propagate the _original_ filename (suffix) to the respective filesystem implementation, there is no way for the `TarFileSystem` to deduce the correct compression and will obviously fail on that.

   2a7133a4 improves the situation. I am happy that the adjustments to make the cached filesystems propagate the original file name to the specific filesystem implementation in order to support inferring the compression type from the file name were less invasive than expected. Please let me know what you think about this.

With kind regards,
Andreas.

PR #564, opened 2021-03-10
add importlib-metadata req

endremborza:
Hello

`import fsspec` breaks with earlier versions of importlib-metadata (obvs for python <3.8), as entry_points takes a positional argument prior to this commit:

https://github.com/python/importlib_metadata/commit/fa54a01a282da90c8b7445f84af648bbcf124e14#diff-a7bcfa00b0d753e504f430fc19095ce5409e080d5eb942d52969f7b7c5c409f6

adding might spare others the time


martindurant:
As you will have noticed, fsspec has no dependencies, and I'd like to keep it this way. Instead,  recommend catching this exception if it happens and moving on. The *only* use for importlib-metadata is to check the version of pyarrow, and there is a fallback option of just importing it.

endremborza:
thanks for the quick, response. 

how about this way?

(sorry about the linting commits, I tried reloving it on the github editor without cloning it and made a mess of it...)

martindurant:
OK, thank you!

PR #566, opened 2021-03-12
Libarchive implementation enhaced to support any filesystem with seekable files

juliospain:
The libarchive implementation was until now only supporting local files (with a file descriptor).

By implementing some low level code (not currently implemented in the python wrapper around the libarchive c library), we can read any file-like object which supports readinto seek (and tell), which covers the vast majority.

Note: libarchive was designed specifically for streaming data with no seek capability (tape recorders, sockets and so on). This is supported only for certain archive types (the most simpler?) and while extracting ALL the data. In any case, since the file system implementation allows for any file access we read the complete archive entries skipping the data, and this requires "seeking" capabilities.

### DIFF
--- a/fsspec/implementations/libarchive.py
+++ b/fsspec/implementations/libarchive.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
 from contextlib import contextmanager
-
 from ctypes import (
     CFUNCTYPE,
     POINTER,
--- a/fsspec/implementations/libarchive.py
+++ b/fsspec/implementations/libarchive.py
@@ -2,17 +2,18 @@ from __future__ import absolute_import, division, print_function
 
 from contextlib import contextmanager
 
-import libarchive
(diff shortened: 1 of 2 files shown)


martindurant:
This looks very useful, and I will look soon. I wonder, do you think libarchive might be interested in upstreaming the added functionality? That would make it available to a wider audience. On the other hand, if there's no place in their API to expose it, or it only makes sense int he context of fsspec, I'm happy for it to stay here.

juliospain:
Good point. I will check with them. Although the python wrapper seems to be "less maintained" (last commit 6 months ago), that's why I included this here.

martindurant:
As I said, fine either way - here is good, if they don't actively respond saying they would like to see the code included there.

martindurant:
If you didn't hear otherwise, this should now be merged here?

juliospain:
No answer so far.

Could be merged here and later simplified if the change is incorporated in the libarchive wrapper.

PR #569, opened 2021-03-14
Chained url fix

juliospain:
I think I refactored correctly the code in url_to_fs and get_fs_token_paths, and added a 'complex' chain test case (different ordering, 3 file systems involved)

I still think url_to_fs and get_fs_token_paths could be merged somehow. Probably url_to_fs is a simpler use of get_fs_token_paths and could be calling the other one. In any case, I leave that refactoring for another pull request.


juliospain:
Ok, I have no idea why it fails. In my pc it is working, so what's the difference? :(

martindurant:
Something went wrong here, core.py and test_core.py seem to have been completely rewritten.

Are you coding on windows, by any chance? It might be a difference of line-ending and that would maybe also explain a different behaviour in FTP. The failing test is at a call to `self.ftp.mlsd`, where I would have thought EOF is the same as an empty listing (but not sure).

juliospain:
Thanks for the hint about the EOL!
After some tries, it seems the ftp archive does NOT like to be wrapped with a simple cache in the test environment. I don't know how to debug this efficiently so I am changing the test to use a zip wrapper instead (the purpose of the test is the way to handle kwargs in open/open_files/etc.

fsspec/core.py
```diff
@@ -332,48 +332,48 @@ def _un_chain(path, kwargs):
...
             protocol in {"blockcache", "filecache", "simplecache"}
             and "target_protocol" not in kw
         ):
-            bit = previous_bit.replace(previous_protocol, protocol, 1)
+            bit = previous_bit
         out.append((bit, protocol, kw))
         previous_bit = bit
-        previous_protocol = protocol
     out = list(reversed(out))
-    # We should only do the url rewrite if the cache is in the middle of the chain
```
juliospain, line 351: In get_fs_token_paths it is easier to expand just the first fs paths. For that it is convenient to bring the target fs path to the cache fs.
Since this is an internal function only used by url_to_fs and get_fs_token_paths, no negative impact is expected.

fsspec/tests/test_core.py
```diff
@@ -178,3 +195,53 @@ def test_multi_context(tmpdir):
...
+    cls = fsspec.get_filesystem_class("ftp")
+    fs = cls(host=host, port=port, username=username, password=password)
+    with tempzip(data) as lfile:
+        fs.put_file(lfile, "archive.zip")
+
+    urls = [
+        "zip://afile",
+        "zip://afile::simplecache",
+        "simplecache::zip://afile",
+        "simplecache::zip://afile::simplecache",
```
martindurant, line 242: What happens with multiple "simplecache"s in the URL? In particular, I don't think there's a way to pass kwargs to a particular one of the instances.

martindurant:
Aside from the souble-simplecache issue, which is not a consequence of this PR, there is not anything needs doing here, right? So I can merge?

Issue #570, opened 2021-03-15
SFTP ls retrieves always detailed listing


juliospain:
The ls implementation for SFTP always retrieves the detailed info of each file or directory, just to throw it out if detailed was not needed.

```
    def ls(self, path, detail=False):
        out = ["/".join([path.rstrip("/"), p]) for p in self.ftp.listdir(path)]
        out = [self.info(o) for o in out]
        if detail:
            return out
        return sorted([p["name"] for p in out])
```

I have not tried it but something like this would make more sense:

```
    def ls(self, path, detail=False):
        paths = ["/".join([path.rstrip("/"), p]) for p in self.ftp.listdir(path)]
        if detail:
            return [self.info(p) for p in paths]
        else:
            return sorted(paths)
```

martindurant:
Sounds reasonable! The only reason to catch the info every time, is if we cache it - but we currently don'e

juliospain:
Funny thing, looking at the paramiko implementation and details, listdir is retrieving ALL the entries attributes and then returning only the names.

The entry details can be obtained using "listdir_attr" in a very efficient way. This takes in my use case around 370ms as opposed to the current implementation entry by entry 14seconds....

I'll work a PR out of this.

juliospain:
Closed with #571

Issue #526, opened 2021-02-02
Add ctime/mtime to list of expected values in info


martindurant:
Created and/or modified time is returned in the file info of most backends. We should endeavour to surface these in the file info dict with a common format (`datetime.datetime`? unix timestamp?) and key names.

e.g., 
```diff
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -78,6 +78,8 @@ class LocalFileSystem(AbstractFileSystem):
                 result["size"] = out2.st_size
             except IOError:
                 result["size"] = 0
+        result['created'] = datetime.datetime.utcfromtimestamp(result["created"])
+        result['modified'] = datetime.datetime.utcfromtimestamp(result["mtime"])
         return result
```

martindurant:
Marked as "good first issue" because this should be simple per implementation, but there are quite a few implementations to go through.

ap--:
## A list of filesystems and their info keys

I collected some about the `.info()` dicts of the different filesystems.
Posting it here in case it might be useful:

### AbstractFileSystem

"name", "size", "type"  

https://github.com/fsspec/filesystem_spec/blob/2a8e0ee2b9e8fc4f24e9fa2b257c1599a9e4711a/fsspec/spec.py#L669-L670


### arrow

"name", "size", "type", "mtime" (datetime | float | None)

https://github.com/fsspec/filesystem_spec/blob/2a8e0ee2b9e8fc4f24e9fa2b257c1599a9e4711a/fsspec/implementations/arrow.py#L101-L118

https://arrow.apache.org/docs/python/generated/pyarrow.fs.FileInfo.html#pyarrow.fs.FileInfo


### dask

returns whatever the remote fs returns.

https://github.com/fsspec/filesystem_spec/blob/2a8e0ee2b9e8fc4f24e9fa2b257c1599a9e4711a/fsspec/implementations/dask.py#L93-L97

### data

"name", "size", "type", "mimetype"

https://github.com/fsspec/filesystem_spec/blob/2a8e0ee2b9e8fc4f24e9fa2b257c1599a9e4711a/fsspec/implementations/data.py#L31-L35


### dbfs

"name", "size", "type"

https://github.com/fsspec/filesystem_spec/blob/2a8e0ee2b9e8fc4f24e9fa2b257c1599a9e4711a/fsspec/implementations/dbfs.py#L84-L90


### dirfs

returns whatever the remote fs returns.

[...]

PR #571, opened 2021-03-16
Faster implementation of SFTP listdir

juliospain:
As discussed in issue #570, a faster implementation of ls for SFTP using the already available paramiko functions.

### DIFF
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -1,3 +1,4 @@
+import logging
 import datetime
 import types
 import uuid
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -1,3 +1,4 @@
+import datetime
 import types
 import uuid
 from stat import S_ISDIR, S_ISLNK
@@ -6,6 +7,8 @@ import paramiko
 
 from .. import AbstractFileSystem
 from ..utils import infer_storage_options
+
+logger = logging.getLogger("fsspec.sftp")
 
 
 class SFTPFileSystem(AbstractFileSystem):
@@ -43,6 +46,7 @@ class SFTPFileSystem(AbstractFileSystem):
         self._connect()
 
     def _connect(self):
+        logger.debug("Connecting to SFTP server %s" % self.host)
         self.client = paramiko.SSHClient()
         self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         self.client.connect(self.host, **self.ssh_kwargs)
@@ -60,6 +64,7 @@ class SFTPFileSystem(AbstractFileSystem):
         return out
 
     def mkdir(self, path, mode=511):
+        logger.debug("Creating folder %s" % path)
         self.ftp.mkdir(path, mode)
 
     def makedirs(self, path, exist_ok=False, mode=511):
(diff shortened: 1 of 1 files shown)


fsspec/implementations/sftp.py
```diff
@@ -78,34 +78,46 @@ def rmdir(self, path):
...
+            "name": "",
+            "size": stat.st_size,
             "type": t,
-            "uid": s.st_uid,
-            "gid": s.st_gid,
-            "time": s.st_atime,
-            "mtime": s.st_mtime,
+            "uid": stat.st_uid,
+            "gid": stat.st_gid,
+            "time": stat.st_atime,
```
martindurant, line 99: Maybe a good time to settle on standard names for these fields, when available - see https://github.com/intake/filesystem_spec/issues/526

fsspec/implementations/sftp.py
```diff
@@ -78,34 +78,46 @@ def rmdir(self, path):
         self.ftp.rmdir(path)
 
     def info(self, path):
-        s = self.ftp.stat(path)
-        if S_ISDIR(s.st_mode):
+        stat = self._decode_stat(self.ftp.stat(path))
+        stat["name"] = path + "/" if stat["type"] == "directory" else path
```
martindurant, line 82: Isn't "/" also added in `_decode_stat`?

fsspec/implementations/sftp.py
```diff
@@ -78,34 +78,46 @@ def rmdir(self, path):
...
         }
+        if parent_path:
+            out["name"] = "/".join([parent_path.rstrip("/"), stat.filename])
+            out["name"] += "/" if t == "directory" else ""
+        return out
 
     def ls(self, path, detail=False):
-        out = ["/".join([path.rstrip("/"), p]) for p in self.ftp.listdir(path)]
-        out = [self.info(o) for o in out]
+        stats = [self._decode_stat(stat, path) for stat in self.ftp.listdir_iter(path)]
```
martindurant, line 108: +1

fsspec/implementations/sftp.py
```diff
@@ -78,34 +78,46 @@ def rmdir(self, path):
...
+            return stats
+        else:
+            paths = [stat["name"] for stat in stats]
+            return sorted(paths)
 
     def put(self, lpath, rpath):
         self.ftp.put(lpath, rpath)
 
     def get(self, rpath, lpath):
+        print(rpath)
```
martindurant, line 119: debug?

martindurant: NB: would be very reasonable to add a logger "fsspec.sftp" and add logging statements to the various methods.

martindurant (commented):
Thanks for working on this, looks like it's near.

fsspec/implementations/sftp.py
juliospain: Argh. Damn it. This was added several days ago and forgot to remove it.

fsspec/implementations/sftp.py
juliospain: The thing here is that I didn't want to duplicate all the `_decode_stat` functionality. However, the path name in one case (`info`) is only retrieved from the `info` argument, and in the case of `ls` is retrieved from the `stat` object (only the filename, the complete path needs to be added with the `ls` path argument)

The way to handle is a little bit dirty, I pass the "parent_path" to construct the name there, otherwise it is done "outside" (for info)

fsspec/implementations/sftp.py
juliospain: Good point the one raised in #526. I can do the conversion here already if it is decided.

fsspec/implementations/sftp.py
martindurant: So long as we don't end up with "//".

juliospain: By the way, is there a convention that directories end with "/"? (I just followed the existing code)

I am just asking because of other modules do explicitly omit the final '/' (os.path.normpath or pathlib)

martindurant: Now you mention it, since it already says "directory", it probably makes sense to drop it, and that also makes the code here simpler.

PR #575, opened 2021-03-18
ci: run py36 on the github actions

isidentical:


isidentical:
Oh, I thought 3.6 was already supported (judging by the classifiers) but apparently it is not; https://github.com/intake/filesystem_spec/pull/522. This might be a dead cause

martindurant:
You would need to change `python_requires=` in setup.py to get the CI to run. The classifiers have not been updates since the last release.

isidentical:
> You would need to change python_requires= in setup.py to get the CI to run.

Yes, I am aware. I thought it was the `3.6>=` before I made the PR, and I am a bit confused now regarding the official status of support.

isidentical:
Apparently all tests pass, though pyarrow seems to be a bit problematic when installed in <1, is there anything I should be aware when bumping it to 1>=?

martindurant:
Do you mean pyarrow -> v>1.0? That should be OK. The HDFS implementation is slated to be dropped, but not yet.

isidentical:
yep

martindurant:
OK, thanks for doing this - let's keep py36 alive a little longer!

isidentical:
Thanks @martindurant!

PR #582, opened 2021-03-24
Allow for URL ref without offset/size

manzt:
I took a look at the Reference V1 implementation, and noticed a `[url]` reference (no offset/size) wasn't currently supported. It is my understanding this is part of the spec and I have attempted to add support in this PR.


martindurant:
Perfect!

martindurant:
cc @DPeterK - this is the first part, exposing an arbitrary URL to look like a file in a filesystem or a key in a mapper. There's no "decoder" here. Note that the spec allows for url templating and generative keys and locations.

PR #587, opened 2021-03-29
Propagate the errors that happens on the initalizer when closing

### LINKED ISSUE #573: Handling errors that happened on the _initiate_upload separately
When opening a subclass of the AbstractBufferedFile, there might be some problems happening in the `_initiate_upload` function. Which would result in the absence of the necessary logic to finalize the upload. Here is a short example from s3fs;
```
with fs.open('test/foo/bar', 'wb') as stream:
    stream.write(b'x' * stream.blocksize)
```
If the `test` is a bucket that actually doesn't exist, we get a really weird and cryptic error
```
  File "/home/isidentical/s3fs/s3fs/core.py", line 235, in _call_s3
    return await method(**additional_kwargs)
  File "/home/isidentical/.venv38/lib/python3.8/site-packages/aiobotocore/client.py", line 154, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the CreateMultipartUpload operation: The specified bucket does not exist

The above exception was the direct cause of the following exception:

  File "t1.py", line 8, in <module>
    stream.write(b'x' * stream.blocksize)
  File "/home/isidentical/filesystem_spec/fsspec/spec.py", line 1360, in write
    self.flush()
  File "/home/isidentical/filesystem_spec/fsspec/spec.py", line 1395, in flush
    self._initiate_upload()
  File "/home/isidentical/s3fs/s3fs/core.py", line 1600, in _initiate_upload
    self.mpu = self._call_s3(
  File "/home/isidentical/s3fs/s3fs/core.py", line 1592, in _call_s3
    return self.fs.call_s3(method, self.s3_additional_kwargs, *kwarglist, **kwargs)
  File "/home/isidentical/filesystem_spec/fsspec/asyn.py", line 121, in wrapper
    return maybe_sync(func, self, *args, **kwargs)
  File "/home/isidentical/filesystem_spec/fsspec/asyn.py", line 100, in maybe_sync
    return sync(loop, func, *args, **kwargs)
  File "/home/isidentical/filesystem_spec/fsspec/asyn.py", line 71, in sync
    raise exc.with_traceback(tb)
  File "/home/isidentical/filesystem_spec/fsspec/asyn.py", line 55, in f
    result[0] = await future
  File "/home/isidentical/s3fs/s3fs/core.py", line 253, in _call_s3
    raise translate_boto_error(err)
FileNotFoundError: The specified bucket does not exist

During handling of the above exception, another exception occurred:

  File "t1.py", line 8, in <module>
    stream.write(b'x' * stream.blocksize)
  File "/home/isidentical/filesystem_spec/fsspec/spec.py", line 1557, in __exit__
    self.close()
  File "/home/isidentical/filesystem_spec/fsspec/spec.py", line 1524, in close
    self.flush(force=True)
  File "/home/isidentical/filesystem_spec/fsspec/spec.py", line 1397, in flush
    if self._upload_chunk(final=force) is not False:
  File "/home/isidentical/s3fs/s3fs/core.py", line 1717, in _upload_chunk
    UploadId=self.mpu["UploadId"],
TypeError: 'NoneType' object is not subscriptable
```

The reason seems to be that, no matter what kind of exception happened `__exit__` tries to close the file but since `_initate_upload` failed previously, the expectations of closing logic doesn't get satisfiedied;
https://github.com/intake/filesystem_spec/blob/3c6f9eae0d1516308b35ef9bec4548cf04ce164e/fsspec/spec.py#L1559-L1560

This could be overriden in the implementation by just holding a state like `initalized: bool = False` and checking this before `_upload_chunk` calls. Though it might be eaiser for all implementations if this  would be hosted in the fsspec itself, perhaps by guarding the functions in the AbstractBufferedFile level;
```py
    def initiate_upload(self, *args, **kwargs):
        self._initiate_upload(self, *args, **kwargs)
        self.initiated = True
    
    def upload_chunk(self, *args, **kwargs):
        if not self.initiated:
            return None
        self._upload_chunk(self, *args, **kwargs)
```

isidentical:
Resolves #573

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -17,7 +17,7 @@ Fixes:
 - get_mapper for caching filesystems (#559)
 - fix cross-device file move (#547)
 - store paths without trailing "/" for DBFS (#557)
-- errors that happen on ``_initatiate_upload`` when closing the
+- errors that happen on ``_initiate_upload`` when closing the
   ``AbstractBufferedFile`` will now be propagated (#587)
 
 Version 0.8.7
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -17,6 +17,8 @@ Fixes:
 - get_mapper for caching filesystems (#559)
 - fix cross-device file move (#547)
 - store paths without trailing "/" for DBFS (#557)
+- errors that happen on ``_initatiate_upload`` when closing the
+  ``AbstractBufferedFile`` will now be propagated (#587)
 
 Version 0.8.7
 -------------
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1395,7 +1395,11 @@ class AbstractBufferedFile(io.IOBase):
         if self.offset is None:
             # Initialize a multipart upload
             self.offset = 0
-            self._initiate_upload()
+            try:
+                self._initiate_upload()
+            except:  # noqa: E722
+                self.closed = True
+                raise
 
         if self._upload_chunk(final=force) is not False:
             self.offset += self.buffer.seek(0, 2)
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -303,9 +303,9 @@ def tests_file_open_error(monkeypatch):
         with fs.open("misc/foo.txt", "wb") as stream:
             stream.write(b"hello" * stream.blocksize * 2)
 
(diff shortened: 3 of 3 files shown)


martindurant:
Thanks, looks good

Issue #594, opened 2021-04-02
`infer_compression` doesn't work with CAPITAL file extensions.


itsbenweeks:
E.g. `fs.open('test.ZIP')` will not choose the `zip` interpreter. While choosing to `fs.open('test.ZIP', compression='zip')` would work, I'm dealing with an external call for fsspec that does not pass kwargs. I'd prefer to `.lower()` the string at https://github.com/intake/filesystem_spec/blob/master/fsspec/utils.py#L120 to address this.

martindurant:
Certainly, sounds reasonable a fix. Will you be submitting a PR?

itsbenweeks:
I will :)

PR #595, opened 2021-04-02
Let infer_compression work for UPPERCASE extensions.

itsbenweeks:
Addresses #594.

### DIFF
--- a/fsspec/tests/test_compression.py
+++ b/fsspec/tests/test_compression.py
@@ -48,6 +48,14 @@ def test_infer_custom_compression():
         del compressions["tst"]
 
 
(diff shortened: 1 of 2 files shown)


itsbenweeks:
Happy to write an appropriate test, but would appreciate a kind contributor pointing me to where it should be added.

martindurant:
See fsspec.tests.test_compression . test_infer_custom_compression has some infer tests, but maybe just make a new function test_uppercase_compression with one or two of the standard suffixes.

itsbenweeks:
Added. Thanks for your guidance, @martindurant!

PR #602, opened 2021-04-06
Allow optional offset/length for "gen" entry

manzt:
Accounts for: https://github.com/intake/fsspec-reference-maker/pull/20

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -190,7 +190,7 @@ class ReferenceFileSystem(AsyncFileSystem):
             for pr in products:
                 key = jinja2.Template(gen["key"]).render(**pr, **templates)
                 url = jinja2.Template(gen["url"]).render(**pr, **templates)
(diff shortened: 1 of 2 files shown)


fsspec/implementations/reference.py
```diff
@@ -190,10 +190,16 @@ def _process_references1(self, references):
             for pr in products:
                 key = jinja2.Template(gen["key"]).render(**pr, **templates)
                 url = jinja2.Template(gen["url"]).render(**pr, **templates)
-                offset = int(jinja2.Template(gen["offset"]).render(**pr, **templates))
-                length = int(jinja2.Template(gen["length"]).render(**pr, **templates))
-
-                self.references[key] = [url, offset, length]
+                if "offset" in gen and "length" in gen:
```
martindurant, line 193: Raise error if only one or the other is present `("offset" in gen) ^ ("length" in gen)` ?

martindurant:
Great. We should, in a future PR, actually start having some tests in here.

PR #604, opened 2021-04-08
Allow local, non-async FSs for references

martindurant:
Also introduces `template_overrides` kwargs, which will allow for easier testing


martindurant:
This opens the door to thorough testing with local files or MemoryFileSystem, but it means we should be sure to test an async backend too (i.e., HTTP, the only choice in this repo).

PR #606, opened 2021-04-09
Add tests for optional offset/length "gen" entry & optional "refs"

manzt:
Follow up to #602. Good idea to add tests. Optional `"refs"` wasn't implemented and is now added.


martindurant:
Right, `refs` was not originally optional, but `gen` might cover all keys.

PR #608, opened 2021-04-11
Fix support for symbolic links in `LocalFileSystem`

### LINKED ISSUE #591: `LocalFileSystem.isdir` inconsistent with `os.path.isdir` with symbolic links
We had a bit of a subtle bug here: https://github.com/PyTorchLightning/pytorch-lightning/pull/6730

Because `LocalFileSystem.info` returns `t: 'link'` for sym linked files / directories, `isdir` will always be false. This is inconsistent with the behaviour of `os.path.isdir` which follows sym links. We've been able to fix this by overriding `isdir` to just return `os.path.isdir`. I guess that this behaviour isn't intended? Not sure what the best fix is, overriding `isdir` works but is not the prettiest solution :smiley:

ethanwharris:
Fixes #591

- By returning `info['type'] = 'link'` for symbolic links, `LocalFileSystem.info` actually breaks more than just `isdir`, e.g. `isfile`, `walk` etc. are all affected too.
- This changes `LocalFileSystem.info['type']` to follow symlinks correctly and always give either `directory` or `file`.
- Adds an additional `LocalFileSystem.info['islink']` flag to enable checking for links without breaking the rest.

Let me know if this fix is acceptable, happy to implement an alternative if preferred :smiley:

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -59,21 +59,17 @@ class LocalFileSystem(AbstractFileSystem):
     def info(self, path, **kwargs):
         path = self._strip_protocol(path)
         out = os.stat(path, follow_symlinks=False)
-        dest = False
-        if os.path.islink(path):
-            t = "link"
-            dest = os.readlink(path)
-        elif os.path.isdir(path):
+        if os.path.isdir(path):
             t = "directory"
         elif os.path.isfile(path):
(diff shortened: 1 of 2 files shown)


martindurant:
Yes, I think this is fine. You are right that it probably best to keep close to what `os` does.

PR #611, opened 2021-04-13
Speed up reference loading.

martindurant:


martindurant:
Changes of note:
- use ujson to parse, if it is installed. Actually, that's not even necessarily the fastest, but it seems very common
- option "simple templates" which, if True, uses replace instead of jinja
- LRU cache for the templating (is simple_templates is False)
- quicker exit from building the directory cache

One obvious thing not implemented here is to have the directory listing be lazy, rather than computed up front. That will become more important when the size of the reference set itself becomes significant.

martindurant:
Folded into #640

PR #617, opened 2021-04-21
Add ability to interrupt all fsspec IO tasks

martindurant:
The idea is, that if for some reason fsspec IO has hanged, we can kick out the stalled tasks and/or get debug information about them. This could be called with `client.run()` in the dask distributed context


martindurant:
@mrocklin , I thought you might be interested in this possible companion PR. **IF** we see hangs from fsspec in workers, this might help to diagnose or work around them. Note that it ensures (hope!) that the future waiting on IO in another thread also finishes.

@shalin07 , if you are breve, you could try calling this on a cluster once you see problems.
ref: https://github.com/dask/dask/issues/7547

PR #627, opened 2021-04-30
Add exists as core method to fsspec

evinism:
Several times, I've used fsspec and been disappointed that there hasn't been an "exists" function on the top level, in the exact same way that there's an "open" function on the top level. 

This pull adds one!

Let me know if you don't think this is a good addition for whatever reason.

### DIFF
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -9,7 +9,7 @@ except ImportError:  # python < 3.8
 
 from . import caching
 from ._version import get_versions
-from .core import get_fs_token_paths, open, open_files, open_local
+from .core import get_fs_token_paths, open, open_files, open_local, exists
 from .mapping import FSMap, get_mapper
 from .registry import (
     filesystem,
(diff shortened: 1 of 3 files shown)


martindurant:
I appreciate the desire

Ref: https://github.com/intake/filesystem_spec/issues/41

There are *many* methods that we wish would be top-level usable, and I think that rather than add them one--by-one, we would be better off making a general class for it. How would you imagine your function working in the presence of chained URLs, e.g., with file caching? Would you pass through kwargs every time (and how do you do that if you include multiple filesystems?)?

Note that `open_file`/`open` is special, since it returns `OpenFile`/`OpenFiles` objects (unlike the file-like object of the implementation `open` methods) that are guaranteed serialisation safe, and support extra features such as transparent compression and text mode.

evinism:
Thank you for the feedback! If this difficulty continues to be an issue for us, I'll submit a proposal for such a universal file system.

PR #628, opened 2021-04-30
Add s3a:// as an alias for s3://

evinism:
s3fs operates on both `s3://` and `s3a://`-style protocols. We might consider doing the same for fsspec.

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -135,6 +135,7 @@ known_implementations = {
         "err": 'webHDFS access requires "requests" to be installed',
     },
     "s3": {"class": "s3fs.S3FileSystem", "err": "Install s3fs to access S3"},
+    "s3a": {"class": "s3fs.S3FileSystem", "err": "Install s3fs to access S3"},
     "adl": {
         "class": "adlfs.AzureDatalakeFileSystem",
         "err": "Install adlfs to access Azure Datalake Gen1",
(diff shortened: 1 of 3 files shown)


amotl:
Hi Martin,

this topic sparked my interest and I was pleasured to find a nice overview at [1]. Do you believe it would make sense to also explicitly add `s3n` here?

With kind regards,
Andreas.

[1] https://stackoverflow.com/questions/33356041/technically-what-is-the-difference-between-s3n-s3a-and-s3

---

### S3 Native FileSystem (URI scheme: s3n)
> A native filesystem for reading and writing regular files on S3. The advantage of this filesystem is that you can access files on S3 that were written with other tools. Conversely, other tools can access files written using Hadoop. The disadvantage is the 5GB limit on file size imposed by S3.

### S3A (URI scheme: s3a)
> A successor to the S3 Native, s3n fs, the S3a: system uses Amazon's libraries to interact with S3. This allows S3a to support larger files (no more 5GB limit), higher performance operations and more. The filesystem is intended to be a replacement for/successor to S3 Native: all objects accessible from s3n:// URLs should also be accessible from s3a simply by replacing the URL schema.

### S3 Block FileSystem (URI scheme: s3)
> A block-based filesystem backed by S3. Files are stored as blocks, just like they are in HDFS. This permits efficient implementation of renames. This filesystem requires you to dedicate a bucket for the filesystem - you should not use an existing bucket containing files, or write other files to the same bucket. The files stored by this filesystem can be larger than 5GB, but they are not interoperable with other S3 tools.

martindurant:
Probably no - "n" for "native" implies Hadoop implementation, which is not true for s3fs. In practice, I have not had people expect "s3n" to work here.

amotl:
> Probably no - "n" for "native" implies Hadoop implementation, which is not true for s3fs.

I understand. Thanks for educating me on this topic!

PR #629, opened 2021-05-03
asyn: use throttling for the get/put of AsyncFileSystem

### LINKED ISSUE #618: async: use some sort of chunking when calling put()/get()
The current implementation of async put runs all `_put_file` calls under the hood concurrently, which results with python opening all files at the same time and possibly hitting the open file limit. 
https://github.com/intake/filesystem_spec/blob/39f5137a66040ea70ede0f4a0e81d12ab41fd80d/fsspec/asyn.py#L205-L219

The limit that the user hits can be configured (ulimit) though I think it makes sense for fsspec to adjust the defaults (maybe spawn 200 coroutines at most, and then put these into a loop of `running, done = wait(bunch)` + `coros.update(create_coros(200 - len(running)))`.

isidentical:
Fixes #618

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -104,6 +104,10 @@ _DEFAULT_BATCH_SIZE = 128
 
 
 def _get_batch_size():
+    from fsspec.config import conf
+
+    if "gather_batch_size" in conf:
+        return conf["gather_batch_size"]
     if resource is None:
         return _DEFAULT_BATCH_SIZE
 
@@ -263,8 +267,9 @@ class AsyncFileSystem(AbstractFileSystem):
         The put_file method will be called concurrently on a batch of files. The
         batch_size option can configure the amount of futures that can be executed
         at the same time. If it is -1, then all the files will be uploaded concurrently.
-        If it is None, then the 1/8th of the system limit will be used. If it any other
-        positive number, then it will be used as the batch size.
+        The default can be set for this instance by passing "batch_size" in the
+        constructor, or for all instances by setting the "gather_batch_size" key
+        in ``fsspec.config.conf``, falling back to 1/8th of the system limit .
         """
         from .implementations.local import LocalFileSystem, make_path_posix
 
@@ -297,8 +302,9 @@ class AsyncFileSystem(AbstractFileSystem):
         The get_file method will be called concurrently on a batch of files. The
         batch_size option can configure the amount of futures that can be executed
         at the same time. If it is -1, then all the files will be uploaded concurrently.
-        If it is None, then the 1/8th of the system limit will be used. If it any other
-        positive number, then it will be used as the batch size.
+        The default can be set for this instance by passing "batch_size" in the
+        constructor, or for all instances by setting the "gather_batch_size" key
+        in ``fsspec.config.conf``, falling back to 1/8th of the system limit .
         """
         from fsspec.implementations.local import make_path_posix
 
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -100,38 +100,43 @@ except ImportError:
 else:
     ResourceEror = resource.error
 
-_DEFAULT_SOFT_LIMIT = 1024
+_DEFAULT_BATCH_SIZE = 128
 
 
-def _get_soft_limit():
+def _get_batch_size():
     if resource is None:
-        return _DEFAULT_SOFT_LIMIT
+        return _DEFAULT_BATCH_SIZE
 
     try:
         soft_limit, _ = resource.getrlimit(resource.RLIMIT_NOFILE)
     except (ImportError, ValueError, ResourceError):
-        return _DEFAULT_SOFT_LIMIT
+        return _DEFAULT_BATCH_SIZE
 
     if soft_limit == resource.RLIM_INFINITY:
-        return None
+        return -1
     else:
-        return soft_limit
+        return soft_limit // 8
 
 
-async def _throttled_gather(coros, disable=False, **gather_kwargs):
+async def _throttled_gather(coros, batch_size=None, **gather_kwargs):
     """Run the given coroutines in smaller chunks to
-    not crossing the file descriptor limit"""
-    soft_limit = _get_soft_limit()
+    not crossing the file descriptor limit.
 
-    if disable or soft_limit is None:
+    If batch_size parameter is -1, then it will not be any throttling. If
+    it is none, it will be inferred from the process resources (soft limit divided
+    by 8) and fallback to 128 if the system doesn't support it."""
+
+    if batch_size is None:
+        batch_size = _get_batch_size()
+
+    if batch_size == -1:
         return await asyncio.gather(*coros, **gather_kwargs)
 
-    chunk_size = soft_limit // 8
-    assert chunk_size > 0
+    assert batch_size > 0
 
     results = []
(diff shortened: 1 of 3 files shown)


fsspec/asyn.py
```diff
@@ -92,6 +92,50 @@ def get_loop():
     return loop[0]
 
 
+try:
+    import resource
```
martindurant, line 96: `resource` has been around a long time, why would the import fail?

fsspec/asyn.py
```diff
@@ -92,6 +92,50 @@ def get_loop():
...
+
+async def _throttled_gather(coros, disable=False, **gather_kwargs):
+    """Run the given coroutines in smaller chunks to
+    not crossing the file descriptor limit"""
+    soft_limit = _get_soft_limit()
+
+    if disable or soft_limit is None:
+        return await asyncio.gather(*coros, **gather_kwargs)
+
+    chunk_size = soft_limit // 8
```
martindurant, line 129: Why the `// 8` here? I would expect the limit to be called the limit. 

Also, the limit is not actually settable (unless it falls back to 1024), but I would expect it to be, probably have an option in the config too.

fsspec/asyn.py
```diff
@@ -128,6 +172,7 @@ class AsyncFileSystem(AbstractFileSystem):
     # for _* methods and inferred for overridden methods.
 
     async_impl = True
+    disable_throttling = False
```
martindurant, line 175: Seems to me not a good instance- or class-wise setting (and it's not documented). Would prefer to set a global variable and/or config.

fsspec/asyn.py
```diff
@@ -92,6 +92,50 @@ def get_loop():
...
+
+_DEFAULT_SOFT_LIMIT = 1024
+
+
+def _get_soft_limit():
+    if resource is None:
+        return _DEFAULT_SOFT_LIMIT
+
+    try:
+        soft_limit, _ = resource.getrlimit(resource.RLIMIT_NOFILE)
```
martindurant, line 111: Docs imply this can be called `RLIMIT_OFILE` on BSD (including macos?)

martindurant (commented):
Cool, thanks for putting this together.

I just have notes about the naming and how the chunksize value can be changed by the user.

Obviously some documentation of this would be appreciated, but I'm not really sure where it would go.

fsspec/asyn.py
isidentical: It doesn't work on windows (AFAIK)

fsspec/asyn.py
isidentical: > Why the // 8 here? I would expect the limit to be called the limit.

we might increase it to `// 4` though I don't think it would be safe to assume all file descripts of this process can be allocated in this procedure alone. Then it would fail if even a couple other files are open

> Also, the limit is not actually settable (unless it falls back to 1024), but I would expect it to be, probably have an option in the config too.

It can be set using `uname` though I guess we could expose it with `put()`.

fsspec/asyn.py
isidentical: The reason for implementations that do some other stuff in their `_put_file`/`_get_file` calls beside opening files on the local system.

fsspec/asyn.py
isidentical: Aren't they are aliases (i am not on mac, so can't test :().

fsspec/asyn.py
martindurant: Sorry, you misunderstood what I wanted to say

> we might increase it to // 4

I mean, that we should not have `//` by any factor here, and call the variable "gather batch size" from the start. In the function above where it is determined, we can have `// 8`, but the batch size can be changed by the user.

> It can be set using uname

I don't mean changing the actual OS limit, that seem not like something we should ver do. Rather, that we can guess a reasonable batchsize, but the user can change this batch size if they want.

isidentical: > I don't mean changing the actual OS limit, that seem not like something we should ver do. Rather, that we can guess a reasonable batchsize, but the user can change this batch size if they want.

Makes sense.

fsspec/asyn.py
martindurant: get and put should always use a local file. Indeed, all comms use sockets too, so there is reason to want to use batched gather everywhere. (In practice, the comms library like asyncio has a connection pool that does this throttling job too)

fsspec/asyn.py
martindurant: I don't know. I just checked from mac, and it's NOFILE for osx too! I guess darwin is not real BSD, but maybe we need to worry about it still.

isidentical: Tested on freebsd, openbsd, netbest but can't find any `RLIMIT_OFILE`, weird hmm.
```
gcc220$ python3
Python 3.8.6 (default, Oct 13 2020, 09:04:17) 
[Clang 10.0.1 ] on openbsd6
>>> import resource
>>> resource.RLIMIT_OFILE
AttributeError: module 'resource' has no attribute 'RLIMIT_OFILE'
>>> resource.RLIMIT_NOFILE
8
```
```
gcc300$ python3.7
Python 3.7.7 (default, May 13 2020, 14:31:34) 
[GCC 7.4.0] on netbsd9
>>> import resource
>>> resource.RLIMIT_OFILE
AttributeError: module 'resource' has no attribute 'RLIMIT_OFILE'
>>> resource.RLIMIT_NOFILE
8
```
```
Python 3.6.7 (default, Feb  7 2019, 01:15:51) 
[GCC 4.2.1 Compatible FreeBSD Clang 6.0.1 (tags/RELEASE_601/final 335540)] on freebsd12
>>> import resource
>>> resource.RLIMIT_OFILE
AttributeError: module 'resource' has no attribute 'RLIMIT_OFILE'
>>> resource.RLIMIT_NOFILE
8
>>> resource.getrlimit(resource.RLIMIT_NOFILE)
(87363, 87363)
```

martindurant: That's a big limit! :) If only that were true everywhere, we would never have seen this problem. I think macos's default is 256, but you can configure it.

OK, we can assume, then, that this some relic.

isidentical: > That's a big limit! :)

Not so great for a public machine with 2 cores / 2 gb of ram TBH :) In my local system (ubuntu) soft limit is 1024, which is quite easy to hit.

fsspec/asyn.py
```diff
@@ -261,13 +266,13 @@ async def _put(self, lpath, rpath, recursive=False, **kwargs):
...
             ],
-            disable=self.disable_throttling,
+            batch_size=batch_size,
         )
 
     async def _get_file(self, rpath, lpath, **kwargs):
         raise NotImplementedError
 
-    async def _get(self, rpath, lpath, recursive=False, **kwargs):
+    async def _get(self, rpath, lpath, recursive=False, batch_size=None, **kwargs):
```
martindurant, line 287: Sorry, I meant in the actual configuration, `fsspec.config`, a key like "async_gather_batch_size". If defined, it would be used instead of  _DEFAULT_BATCH_SIZE.

The version here would need to have the signature changed in any subclass that overrides this method.

martindurant:
I hope you don't mind the push - sometimes it's easier to describe in code.

isidentical:
I wasn't even aware of the config capabilities, so it's nice to learn!

Issue #295, opened 2020-05-12
Compatibility with the new Arrow FileSystem implementations


jorisvandenbossche:
**Background**

@martindurant as you know, last year we started developing new FileSystem implementations in the Apache Arrow project (https://issues.apache.org/jira/browse/ARROW-767, https://github.com/apache/arrow/pull/4225 is the PR with the initial abstract API, on which you gave feedback). Those developments have some implications for users using fsspec-compatible filesystems, and so as promised, with some delay, opening an issue here to discuss how to handle those implications (and since fsspec currently holds the pyarrow-compatibiliy layer, opening an issue here seems appropriate).

To summarize:

- The "old" filesystems are available under `pyarrow.filesystems` ([docs](https://arrow.apache.org/docs/python/filesystems_deprecated.html)). We basically only have a `LocalFileSystem` and `pa.hdfs.HadoopFileSystem` as concrete implementations. And in addition, there is the `DaskFileSystem` which is used by fsspec as base class, see more on that below.
- The "new" filesystems are available in the `pyarrow.fs` submodule ([docs](https://arrow.apache.org/docs/python/filesystems.html)). Those are python wrappers for the C++ implementations, and currently there are already concrete implementations for local, Hadoop and S3.

So an important difference is that the new filesystems are actual implementations in C++, and `pyarrow.fs` is only providing wrappers for those. This is done for good reasons: those filesystem are a shared implementation and are used by many different users of the Arrow project (and from C, C++, Python, R, Ruby, ..). Further, those filesystems are for example used in the Arrow Datasets project, which enables a bunch of new features in the ParquetDataset reading (and also enabled that you can now actually query a Parquet dataset from R). Those new filesystems have been an important part in moving the Arrow project forward.

But this also means that the filesystem that pyarrow functions expect is no longer an "interface" you can implement, but it actually needs a filesystem that wraps a C++ filesystem. 
(to be clear: all functionality that already existed before is right now still accepting the old filesystems, only the new `pyarrow.dataset` module already requires the new filesystems. But long term, we want to move consistently to the new filesystems). 

Concretely, this means that the feature of `fsspec` to automatically provide compatibility with payrrow will no longer work in the future:

> if installed, all file-system classes also subclass from `pyarrow.filesystem.FileSystem`, so can work with any arrow function expecting such an instance

This current compatibility means that eg pyarrow's `parquet.read_table/ParquetDataset` work with any fsspec filesystem.

---

**Concrete issues** 

[...]

PR #634, opened 2021-05-12
Update documentation to better reflect how PyArrow interoperability works

AndreiBarsan:
In the context of #295, it seems it will be non-trivial to seamlessly support pyarrow interoperability moving forward.

~It took me a bit of digging to realize why I couldn't easily interop between pyarrow and fsspec in my code base (I'm using pyarrow 4.0 but fsspec only supports pyarrow < 2.0). To this end, I added a small warning banner to the documentation to highlight this.~

fsspec no longer adds pyarrow superclasses to its fs files for pyarrow >= 2.0 since this is not necessary.

Made two other minor cosmetic fixes near my main change in `features.rst`.

**Testing plan:** I built the Sphinx docs on my machine and inspected my formatting changes visually, making sure that the new link works.

Let me know what you think!

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -94,20 +94,20 @@ object.
 PyArrow integration
 -------------------
 
-.. warning::
-    ``fsspec`` is only compatible with PyArrow's classic ``FileSystem``. Starting with version
-    2.0 (released in October 2020), `pyarrow`_ switched to a C++ filesystem implementation that
-    is no longer compatible with ``fsspec``. Please refer to 
-    `this GitHub issue <https://github.com/intake/filesystem_spec/issues/295>`_ for additional details.
-
-
-`pyarrow`_ has its own internal idea of what a file-system is (``pyarrow.filesystem.FileSystem``),
+`pyarrow`_ has its own internal idea of what a file-system is (``pyarrow.fs.FileSystem``),
 and some functions, particularly the loading of parquet, require that the target be compatible.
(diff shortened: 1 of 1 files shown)


martindurant:
Could you please clarify what doesn't work? We'd rather fix that.

AndreiBarsan:
Oh, I believe it's because they changed the API a lot and (I think) it's no longer in pure Python.

Currently `fsspec` tries to detect `pyarrow` and make its filesystem class a base class for `AbstractFileSystem`, but disables this behavior on L85 of ``spec.py``: https://github.com/intake/filesystem_spec/blob/e734622e2b837625d5c8f27477d6968d837f68b8/fsspec/spec.py#L85

Based on #295 it seems because actually inheriting from ``pyarrow.fs.FileSystem`` (the new-school 2.0+ pyarrow FS class) does not work.

martindurant:
As far as I understand, it is no longer necessary to inherit from pyarrow's filesystem class, we expect full functionality without it.

AndreiBarsan:
Good point. Upon closer inspection of the pyarrow docs, it seems that this is actually now still supported, just from the pyarrow side. So the user would just need to pass `fsspec` filesystems to, e.g., `pyarrow.dataset.dataset` and things should "just work".

I was originally confused because I was porting some custom code which used pyarrow's FileSystem to fsspec and was expecting the interfaces to be mostly compatible. They are in principle, except that pyarrow.fs.FileSystem just uses some different naming schemes.

I will rephrase the docs in the PR to remove the claim that things are not incompatible. Thank you!

PR #643, opened 2021-05-19
Treat 0 as None for async timeout

### LINKED ISSUE #642: Consider to handle the timeout=0 as None in asyn.py
After upgrade fsspec from 0.8.7 to 2021.5.0, I found my application not able to download from Azure Datalake. I found out the reason is I call fs.get(...) with extra argument timeout=0, which in Azure's API means infinite timeout, and with the new implementation for using asyncio, the timeout=0 not work any more, the equivalent usecase is timeout=None.

I would suggest treating timeout=0 as timeout=None. I think for have 0 second timeout is kind of not make sense at all in an async situation (no waiting also means you can't finish anything), and on the other hand, use 0 second timeout to indicate infinite waiting is quite natural for me, not sure how the others think.

Thanks!

scott-zhou:
In module asyn.py, when it get input argument timeout=0, it will wait for 0 second on async calls. To have 0 second timeout is kind of not make sense at all in an async situation (no waiting also means you can't finish anything), and on the other hand, use 0 second timeout to indicate infinite waiting is quite natural implementation.

This PR is for treat timeout=0 situation as timeout=None, which means for infinite timeout. This also align to some of the other storage client API (e.g. Azure Datalake).

Closes #642

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -14,6 +14,7 @@ private = re.compile("_[^_]")
 
 
 async def _runner(event, coro, result, timeout=None):
+    timeout = timeout if timeout else None  # convert 0 or 0.0 to None
     if timeout is not None:
         coro = asyncio.wait_for(coro, timeout=timeout)
     try:
@@ -34,6 +35,7 @@ def sync(loop, func, *args, timeout=None, **kwargs):
     """
     Make loop run coroutine until it returns. Runs in other thread
     """
+    timeout = timeout if timeout else None  # convert 0 or 0.0 to None
     # NB: if the loop is not running *yet*, it is OK to submit work
     # and we will wait for it
     if loop is None or loop.is_closed():
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -19,6 +19,8 @@ async def _runner(event, coro, result, timeout=None):
         coro = asyncio.wait_for(coro, timeout=timeout)
     try:
         result[0] = await coro
+    except asyncio.exceptions.TimeoutError:
+        result[0] = None
     except Exception as ex:
         result[0] = ex
     finally:
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -19,8 +19,6 @@ async def _runner(event, coro, result, timeout=None):
         coro = asyncio.wait_for(coro, timeout=timeout)
     try:
         result[0] = await coro
-    except asyncio.exceptions.TimeoutError:
-        result[0] = None
     except Exception as ex:
         result[0] = ex
     finally:
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -97,11 +97,11 @@ PyArrow integration
 `pyarrow`_ has its own internal idea of what a file-system is (``pyarrow.fs.FileSystem``),
 and some functions, particularly the loading of parquet, require that the target be compatible.
 As it happens, the design of the file-system interface in ``pyarrow`` *is* compatible with ``fsspec``
-(this is not by accident). 
+(this is not by accident).
 
-At import time, ``fsspec`` checks for the existence of ``pyarrow``, and, if ``pyarrow < 2.0`` is 
+At import time, ``fsspec`` checks for the existence of ``pyarrow``, and, if ``pyarrow < 2.0`` is
 found, adds its base filesystem to the superclasses of the spec base-class.
-For ``pyarrow >= 2.0``, ``fsspec`` file systems can simply be passed to ``pyarrow`` functions 
+For ``pyarrow >= 2.0``, ``fsspec`` file systems can simply be passed to ``pyarrow`` functions
 that expect ``pyarrow`` filesystems, and ``pyarrow`` `will automatically wrap them
 <https://arrow.apache.org/docs/python/filesystems.html#using-fsspec-compatible-filesystems>`_.
 
--- a/fsspec/tests/test_async.py
+++ b/fsspec/tests/test_async.py
@@ -16,6 +16,42 @@ def test_sync_methods():
     assert not inspect.iscoroutinefunction(inst.info)
 
 
+class _DummyAsyncKlass:
+    def __init__(self):
+        self.loop = fsspec.asyn.get_loop()
+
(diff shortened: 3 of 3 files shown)


martindurant:
Would you like to add a test, which is the same as another async test with timeout=0, showing that the call still succeeds?

scott-zhou:
> Would you like to add a test, which is the same as another async test with timeout=0, showing that the call still succeeds?

Yes, sure. I did not see any test case existed for fsspec.asyn.sync_wrapper, so I added 4 testcases for test timeout:
 - less than expected time, won't finish function
 - more than expected time, will finish function
 - None, will finish function
 - 0, as None, will finish function

And for test the fsspec.asyn.sync_wrapper, I created a dummy function for test, which will sleep for 1 second.

martindurant:
I am happy with this; if you wanted the tests to run on py3.6 too, you would need to use `loop.run_until_complete` - but I don't insist. I am happy to merge this tomorrow after the tests have passed.

scott-zhou:
> I am happy with this; if you wanted the tests to run on py3.6 too, you would need to use `loop.run_until_complete` - but I don't insist. I am happy to merge this tomorrow after the tests have passed.

It failed on windows. I don't have a windows machine right now, but I can find one and fix it later.

scott-zhou:
@martindurant I have fixed the failed test case on Windows, please double check. Thanks!

fsspec/asyn.py
```diff
@@ -14,10 +14,13 @@
 
 
 async def _runner(event, coro, result, timeout=None):
+    timeout = timeout if timeout else None  # convert 0 or 0.0 to None
     if timeout is not None:
         coro = asyncio.wait_for(coro, timeout=timeout)
     try:
         result[0] = await coro
+    except asyncio.exceptions.TimeoutError:
+        result[0] = None
```
martindurant, line 23: This doesn't seem like a good idea - it would make it appear that the coroutine succeeded, but had no result. Are you relying on the timeout being raised in `sync()` before this None-result can be collected?

martindurant: I'm pretty sure you are facing a race condition. The purpose of `_runner` is to run the coroutine and put the result *OR* exception into the output. `sync()`, if it doesn't timeout, examines this value to see if it's an exception or not. So in the current for you have, in the case of the coroutine timing out, the following could happen
- `_runner` times out first, puts `None` into the output, and then `sync()` thinks `_runner` completed successfully and then returns `None`
- `sync()` times out first and raises `TimeOut`, in which case it doesn't matter what `_runner` returns.

I would argue that both these scenarios should raise Timeout

> I observed the behavior can be different on Linux/OSX and Windows.

It is true that both the low-level "event loop poll"  and threading systems are different, which might account for what you are seeing.

scott-zhou: I agree with you, to set result[0] to be None on asyncio.exceptions.TimeoutError in `_runner` will cause `sync` think `_runner` completed successfully and returns None.

And I also agreed that this is actually a race condition about `_runner` timeout first or `sync` timeout first.

Just for clarification, here `_runner` timeout means `asyncio.wait_for()` timeout, and `sync` timeout means for `event.wait()` timeout.

Here is a different between this two timeout. `asyncio.wait_for` will raise asyncio.exceptions.TimeoutError exception, but `threading.Event.wait` timeout will finish gracefully without any exception.

My argument is I think `_runner`, `sync` and `sync_wrapper` want to turn an async method into a sync method, so the asyncio.exceptions.TimeoutError is the exception we want to hide it inside in `sync`. So I think we still need to suppress this exception.

I agree with you that suppress the TimeoutError in `_runner` will cause `sync` think `_runner` succeeds. How about handling it in `sync`? like:
```
    event.wait(timeout)
    if isinstance(result[0], asyncio.exceptions.TimeoutError):
        #  To suppress the TimeoutError exception and returns None on timeout occurs
        return None
    elif isinstance(result[0], BaseException):
        raise result[0]
    return result[0]
```

martindurant: >  threading.Event.wait timeout will finish gracefully without any exception.

We should raise a Timeout for this case too. 

> To suppress the TimeoutError exception and returns None on timeout occurs

but this *is* an error condition, and we should raise so that the use knows it. While it is true that the underlying coroutine *might* have succeeded, the whole purpose of a timeout is that, if things take longer than stated, we should get an exception.

scott-zhou: Aha, okay, that's my misunderstanding. I thought the `sync_wrapper` will turn async method to sync, and hide the timeout situation.

Then user of `sync_wrapper` should expect to get a timeout exception when they call it with a timeout, right?

Then in this case, here we might should consider to raise a timeout exception when `threading.Event.wait()` timeout. And it should be the same exception as `asyncio.wait_for()` timeout, because user don't really care which timeout it is.

In this case, maybe we should not re-raise the `asyncio.exceptions.TimeoutError`. Instead of that, we can create something like a `fsspec.TimeoutError` exception class and raise it on either timeout situation.

I will change my PR to catch the timeout exception in test case and change it back in `_runner`. If you agree with my previous idea (raise `fsspec.TimeoutError` on both timeout situations, not re-use `asyncio.exceptions.TimeoutError`), please let me know and I can do that in another PR.

martindurant: Yes, I think you have it right.

> Then user of sync_wrapper should expect to get a timeout exception when they call it with a timeout, right?

If you can figure out how the user might pass a timeout via `sync_wrapper`, sure. However, we don't expect the user to know, in general, that the call they are making is backed by an coroutine for some implementations and not others. Neither can we realistically set a global default timeout, since some operations (downloading thousands of files) might be expected to take very long.

scott-zhou: I have changed it back, and skip the timeout occur test case for now until the `sync_wrapper` has a consistent behavior on timeout. I can solve it later in another PR. It's actually out of scope of this PR, which is just want to treat timeout=0 as timeout=None.

Please have a double check, thanks!

martindurant:
OK, will merge this when it goes green again, and we can discuss the right behaviour of timeouts in general in a follow-up.

martindurant:
Thanks, @scott-zhou

PR #645, opened 2021-05-20
Raise FSTimeoutError on asyn timeout occurs

### LINKED ISSUE #644: Have consistent behavior when asyn.sync_wrapper decorated method have timeout
asyn.sync_wrapper can convert an async method to a sync method. When the user calls the decorated sync method, it accepts a timeout argument to specify the maximum execution time.

When the timeout happen, it can be two different behaviors:
 - timeout happen in `asyncio.wait_for()`, an asyncio.TimeoutError exception raised
 - timeout happen in `threading.Eveng.wait()`, the function will exit gracefully with None return value.

This is a bit confusing and ambiguous result. For the first case, the user should not care about the fsspec's implementation detail, so it should not expect an asyncio.TimeoutError. And for the second case, the user doesn't know if the function actually succeeds but returns a None result, or it is timeout.

I think fsspec should have it's own user-defined exception FSTimeoutError and use that exception on either case of timeout.

scott-zhou:
Create fsspec's own user-defined exception type FSTimeoutError, raise FSTimeoutError in `asyn.sync` function when timeout occurs.
 
Closes #644

### DIFF
--- a/fsspec/exceptions.py
+++ b/fsspec/exceptions.py
@@ -1,17 +1,10 @@
 """
 fsspec user-defined exception classes
 """
+import asyncio
 
 
-class FSBaseException(Exception):
-    """
-    Base exception for fsspec user-defined exceptions
-    """
-
-    ...
-
-
-class FSTimeoutError(FSBaseException):
+class FSTimeoutError(asyncio.TimeoutError):
     """
     Raised when a fsspec function timed out occurs
     """
--- a/fsspec/exceptions.py
+++ b/fsspec/exceptions.py
@@ -1,0 +1,19 @@
+"""
+fsspec user-defined exception classes
+"""
+
+
+class FSBaseException(Exception):
+    """
+    Base exception for fsspec user-defined exceptions
+    """
+
+    ...
+
+
+class FSTimeoutError(FSBaseException):
+    """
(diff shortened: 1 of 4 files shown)


fsspec/exceptions.py
```diff
@@ -0,0 +1,19 @@
+"""
+fsspec user-defined exception classes
+"""
+
+
+class FSBaseException(Exception):
```
martindurant, line 6: This isn't actually used anywhere

fsspec/exceptions.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Base exception for fsspec user-defined exceptions
+    """
+
+    ...
+
+
+class FSTimeoutError(FSBaseException):
```
martindurant, line 14: Should derive from the asyncio or concurrent.futures timeout. Note that there is also a `builtins.TimeoutError`

scott-zhou: Yes, right, and the BaseException can be ignored if we derive from asyncio.TimeoutError.

fsspec/exceptions.py
scott-zhou: Yep, fixed.

martindurant:
Thanks again

PR #647, opened 2021-05-25
optimisations

martindurant:
- remove abspath from linux/mac make_path_posix and move fast-path to start of long function
- logging in fuse
- give option to don't have instance cache keep instances alive (but still share) using config "weakref_instance_cache"
- allow min/max to cat_file to be negative, like a slice

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -243,6 +243,9 @@ class AsyncFileSystem(AbstractFileSystem):
             *[self._pipe_file(k, v, **kwargs) for k, v in path.items()]
         )
 
+    async def _cat_file(self, path, start=None, end=None, **kwargs):
+        raise NotImplementedError
+
     async def _cat(self, path, recursive=False, on_error="raise", **kwargs):
         paths = await self._expand_path(path, recursive=recursive)
         out = await asyncio.gather(
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -174,12 +174,12 @@ def make_path_posix(path, sep=os.sep):
     """ Make path generic """
     if isinstance(path, (list, set, tuple)):
         return type(path)(make_path_posix(p) for p in path)
+    if "~" in path:
+        path = os.path.expanduser(path)
     if sep == "/":
         # most common fast case for posix
         if path.startswith("/"):
             return path
-        elif "~" in path:
-            return os.path.expanduser(path)
         return os.getcwd() + "/" + path
     if (
         (sep not in path and "/" not in path)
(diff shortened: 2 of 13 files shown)


fsspec/implementations/local.py
```diff
@@ -46,31 +46,43 @@ def rmdir(self, path):
...
+            if os.path.isdir(path):
+                t = "directory"
+            elif os.path.isfile(path):
+                t = "file"
+            else:
+                t = "other"
         else:
-            t = "other"
+            # scandir DirEntry
+            out = path.stat(follow_symlinks=False)
```
ananthsub, line 71: is this a breaking change in the API? in pytorch lightning tests we're seeing this stacktrace now: 
```
kwargs = {}
        if isinstance(path, str):
            out = os.stat(path, follow_symlinks=False)
            link = os.path.islink(path)
            if os.path.isdir(path):
                t = "directory"
            elif os.path.isfile(path):
                t = "file"
            else:
                t = "other"
        else:
            # scandir DirEntry
>           out = path.stat(follow_symlinks=False)
E           TypeError: stat() got an unexpected keyword argument 'follow_symlinks'
```
example: https://github.com/PyTorchLightning/pytorch-lightning/runs/2766424283?check_suite_focus=true

martindurant: No the API did not change, this path is only meant to be called by internal functions. The test is whether the path is a str, and I suspect that the tmpdir fixture used for the test is not actually a str, but something str-like. `str(tempdir)` in the test would solve it. However, fsspec maybe should have coped with this in the more user-facing `isdir/info` functions (if my guess about the cause is correct).

ananthsub: yes, you're right. `tmpdir` is a py.path.local object. while we can update our tests to handle with this, does it make sense to update the instance check as well?

fsspec/asyn.py
```diff
@@ -243,6 +243,9 @@ async def _pipe(self, path, value=None, **kwargs):
             *[self._pipe_file(k, v, **kwargs) for k, v in path.items()]
         )
 
+    async def _cat_file(self, path, start=None, end=None, **kwargs):
```
TomAugspurger, line 246: @martindurant what's the purpose of `**kwargs` here? Should subclasses check that additional kwargs aren't passed?

martindurant: It's so that subclasses can pass any extra arguments (like headers) without changing the signature. If the subclass doesn't have any such thing it can do, then kwargs can be ignored.

martindurant: I guess its an indication to downstream that this is a reasonable place to add extra args, and it enables fsspec to add arguments (like start/end) without immediately braking other implementations for the common case that they are left with default values such as `None`.

Issue #650, opened 2021-06-01
Provide base test suite for implementators


TomAugspurger:
I think the fsspec ecosystem would benefit from a set of "base" tests defined in fsspec that can be used by downstream projects like s3fs and adlfs. This would help iron out the inconsistencies between these libraries, especially around behaviors with trailing `/`, exceptions, etc.

Pandas has used this with extension types, which are expected to implement the extension array interface. See https://pandas.pydata.org/docs/development/extending.html#testing-extension-arrays for more, but the basic idea is to use a form of property-based testing. fsspec defines the expected behavior as a test, in a way that only uses the public fsspec interface. Then implementors would define pytest fixtures to provide the filesystem with some known properties (e.g. certain files exist).

```python
@pytest.fixture
def fs():
    """Fixture defining the filesystem with these files:"""
    ...

class TestFilesystem:
    def test_ls_raises_filenotfound(self, fs):
        with pytest.raises(FileNotFoundError):
            fs.ls("/not/a/key")
```

https://github.com/pandas-dev/pandas/blob/master/pandas/tests/extension/conftest.py and https://github.com/pandas-dev/pandas/blob/master/pandas/tests/extension/base/__init__.py provide some examples.

martindurant:
Sounds very audible, but quite a lot of work! For sure, it would probably lead to the eventual solution of things like the path suffix issue; but there *are* some concrete differences between implementations. The most obvious of these is read-only versus rw.

TomAugspurger:
I don't think it's *too* much additional work. At least not on the margin once things are set up. For example, the test in https://github.com/intake/filesystem_spec/pull/649 could be adapted to a base test pretty easily. And in theory this lowers the burden on downstream libraries, if they can confidently rely on the base tests then they can skip writing some tests in their package.

There is some upfront work in defining the fixtures and base tests. If your open to it, I can do a small version, with just the `ls` raises test from that PR along with the scaffolding.

> but there are some concrete differences between implementations

Yep, in pandas we recommend extension arrays skip those tests that don't apply Either by not inheriting from that base class, or by skipping: https://github.com/pandas-dev/pandas/blob/9231b495aadac6e93a0385c24066d84a1fb9eac6/pandas/tests/extension/test_numpy.py#L211-L214

Your read-only vs. read-write example is perfect. There'd be something like `BaseReadOnlyTests`, `BaseReadTests` and `BaseWriteTests`.

isidentical:
[...]

PR #651, opened 2021-06-01
POC of base tests for fsspec behavior.

TomAugspurger:
xref https://github.com/intake/filesystem_spec/issues/650

### DIFF
--- a/fsspec/tests/base/__init__.py
+++ b/fsspec/tests/base/__init__.py
@@ -1,0 +1,3 @@
+from .spec import BaseFSTests, BaseReadTests
+
+__all__ = ["BaseFSTests", "BaseReadTests"]
--- a/fsspec/tests/base/spec.py
+++ b/fsspec/tests/base/spec.py
@@ -1,0 +1,27 @@
+import pytest
+
+
+class BaseSpecTests:
+    """Base class for all specification tests."""
+
+
+class BaseFSTests(BaseSpecTests):
+    """
+    Tests that the fixture object provided meets expectations. Validate this first.
+    """
+
+    def test_files_exist(self, fs, prefix):
+        assert fs.exists(f"{prefix}/exists")
+
+
+class BaseReadTests(BaseSpecTests):
+    """
+    Tests that apply to read-only or read-write filesystems.
+    """
+
+    def test_ls_raises_filenotfound(self, fs, prefix):
+        with pytest.raises(FileNotFoundError):
+            fs.ls(f"{prefix}/not-a-key")
+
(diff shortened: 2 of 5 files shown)


TomAugspurger:
Interestingly, `MemoryFileSystem` is failing:

```python
fsspec/tests/spec/test_memory.py::TestRead::test_ls_raises_filenotfound FAILED                                                                           [100%]

=========================================================================== FAILURES ===========================================================================
_____________________________________________________________ TestRead.test_ls_raises_filenotfound _____________________________________________________________

self = <test_memory.TestRead object at 0x7f3a3edfb760>, fs = <fsspec.implementations.memory.MemoryFileSystem object at 0x7f3a3ee0d340>, prefix = '/root/'

    def test_ls_raises_filenotfound(self, fs, prefix):
        with pytest.raises(FileNotFoundError):
>           fs.ls(f"{prefix}/not-a-key")
E           Failed: DID NOT RAISE <class 'FileNotFoundError'>

fsspec/tests/base/spec.py:25: Failed
=================================================================== short test summary info ====================================================================
FAILED fsspec/tests/spec/test_memory.py::TestRead::test_ls_raises_filenotfound - Failed: DID NOT RAISE <class 'FileNotFoundError'>
```

Is this deliberate?

```python
In [1]: import fsspec

In [2]: fs = fsspec.get_filesystem_class("memory")()

In [3]: fs.ls("/a")
Out[3]: []
```

martindurant:
Yep, MemoryFileSystem should probably raise. It has "pseudo directories" explicitly holding references that don't contain anything (i.e., empty directories), so we should be able to tell the difference between non-existent and empty directory.

martindurant:
`fsspec.implementations.memory.MemoryFileSystem.ls` is pretty ugly!

TomAugspurger:
OK, I've xfailed that MemoryFileSystem test for now: https://github.com/intake/filesystem_spec/issues/652.

Some general thoughts / questions:

1. Are we OK with the naming `BaseSpecTests`, `tests/base/spec.py`?
2. Are we OK with the layout of `tests/base/spec.py` defining the base tests, and `implementations/tests/spec/test*.py` for each implementation?
3. We should consider parametrizing `fs` by some common arguments to `AbstractFileSystem`, including `use_listings_cache` and `asynchronous`.

cc @isidentical from #650 if you have thoughts on the implementation. It is a bit awkward to have this `prefix` fixture that has to be defined. AFAICT, I needed that for HTTPFileSystem to work with this setup since its requests like `ls` include the full URL, rather than the URL relative to some root.

PR #655, opened 2021-06-04
Move Range helper to asyn, to reuse code in s3fs and gcsfs

martindurant:


martindurant:
@hayesgb @TomAugspurger , this makes a helper method available, which can translate start and end arguments to cat_file if those values are negative; i.e., `_cat_file(url, start, end)` should return the same as `_cat_file(url)[start:end]`.

martindurant:
This is useful for applications where you need just the end of a file, e.g., the parquet footer region.

PR #660, opened 2021-06-07
Add compression to LocalFileOpener

limx0:
This PR copies the compression implementation from `OpenFile` to `LocalFileOpener` to allow the `compression` kwarg to work for local files again. 

I've added a simple test also. This should fix https://github.com/intake/filesystem_spec/issues/659

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -7,6 +7,8 @@ import shutil
 import tempfile
 
 from fsspec import AbstractFileSystem
+from fsspec.compression import compr
+from fsspec.core import get_compression
 from fsspec.utils import stringify_path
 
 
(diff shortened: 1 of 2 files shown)


martindurant:
OK, I think this works fine. Note that `fsspec.open` does already handle this, and also explicitly keeps references to the intermediate file objects so that they can be cleanly closed. I am supposing that garbage collection is doing the right thing here, as the compressed file keeps a reference to the raw file.

limx0:
Thanks @martindurant - I'm doing some other operations with an open filesystem so this is just how I ended up opening the file (as opposed to just using fsspec.open). Should I add some extra testing to ensure these files are cleaned up?

martindurant:
I think it's OK. The extra tests would need some mocking or other inner hacks that might be unpleasant.

PR #662, opened 2021-06-08
Feedback after release

martindurant:
- registry loose versions dropped

- current directory cannot be deleted

- local _strip_protocol to remove trailing "/"


isidentical:
Apparently, some stuff is also broken on DVC related to the local filesystem, though didn't have time to debug it. Perhaps it was the `info()` with path-likes. Will look into it in depth tomorrow!

martindurant:
Thanks, @isidentical . Do you want to add DVC into the test runs here yet?

isidentical:
We have too many fs-unbound runs, so until they are all marked and grouped I don't think that it will worth it. CC: @efiop

PR #667, opened 2021-06-09
local: explicitly check for os.DirEntry

### LINKED ISSUE #666: 2021.06.0 regression: support for path-like objects are broken
Reproducer;
```
import os
import fsspec
from pathlib import Path

path = Path('/tmp/c.txt')
path.write_text('kitten')

print(fs.exists(path))
print(fs.exists(os.fspath(path)))
```

2021.05.0:
```
True
True
```

master:
```
False
True
```

isidentical:
`pathlib.Path()` and some other variants also has `stat()` method though they shouldn't use that codepath (since it is specifically designated for `os.DirEntry`). I don't have the context on why it doesn't checking for this in the first place though I can verify it makes related tests pass on our side (downstream). Resolves #666

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -58,7 +58,7 @@ class LocalFileSystem(AbstractFileSystem):
         return super().glob(path, **kwargs)
 
     def info(self, path, **kwargs):
-        if hasattr(path, "stat"):
+        if isinstance(path, os.DirEntry):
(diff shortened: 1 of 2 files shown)


martindurant:
The condition was meant to be an improvement on `not isinstance(str)`, but you are, of course, quite correct. Thanks for fixing.

PR #671, opened 2021-06-14
 asyn: introduce fsspec.asyn.fsspec_loop 

### LINKED ISSUE #670: Introduce a way of temporarily accessing the fsspec loop
Some file systems allow constructing structs from their original libraries (e.g in case of adlfs, this is azure sdk) outside of the filesystem's `__init__`. But if they are not properly bound to the fsspec's loop, we might get some errors (`Task is created under a different event loop...`), so I believe it would be really useful to have some sort of way to temporarily set the fsspec's event loop as the running one and then revert back. This is even useful the filesystems themselves, since if you are dealing with some asyncio constructs (e.g `Queue`) then you have to initialize it under the fsspec's loop otherwise you get the same error. I had to defer the initialization of those constructs to the connect logic, so such a functionality might also make things a little bit easier on that front too.

isidentical:
Resolves #670

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,7 +4,9 @@ Changelog
 Dev
 ---
 
+Enhancements
 
+- Introduce ``fsspec.asyn.fsspec_loop`` to temporarily switch to the fsspec loop.
 
 
 2021.06.0
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,7 +6,7 @@ Dev
 
 Enhancements
 
-- Introduce ``fsspec.asyn.fsspec_loop`` to temporarily switch to the fsspec loop.
+- Introduce ``fsspec.asyn.fsspec_loop`` to temporarily switch to the fsspec loop. (#671)
 
 
 2021.06.0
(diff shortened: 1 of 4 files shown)


docs/source/changelog.rst
```diff
@@ -4,7 +4,9 @@ Changelog
 Dev
 ---
 
+Enhancements
 
+- Introduce ``fsspec.asyn.fsspec_loop`` to temporarily switch to the fsspec loop.
```
isidentical, line 9:
```suggestion
- Introduce ``fsspec.asyn.fsspec_loop`` to temporarily switch to the fsspec loop. (#671)
```

martindurant:
I think this is a decent thing to include - anyone have a reason why this shouldn't happen or is somehow dangerous?

martindurant:
I am waiting for a decision on https://github.com/dask/s3fs/pull/495 (I asked for more feedback) and then I can do a new round of releases.

isidentical:
@martindurant it would be also really nice if you could please take a look at the callbacks, and maybe include them in the next release (if you see them as fit). Being able to use `get_file`/`put_file` APIs directly without making UI worse (no progress bars for total bytes transferred) would be a huge win for certain filesystems (e.g [sshfs](github.com/iterative/sshfs)).

PR #675, opened 2021-06-18
callbacks: Implement fsspec.callbacks

### LINKED ISSUE #668: [IDEA] Implement progress callbacks
It is a very good aspect of the UI to deliver users with updates regarding the process, if the operation is something that takes a while. There are certain operations in fsspec that are like blackboxes. You don't know anything until those are completed. I'd propose we add a callback system for certain APIs, so that client code can show progress bars, take action depending on those etc. etc. 

Here is the list of APIs that might be helpful with progress bars;
- `get_file`/`put_file` => A progress bar showing the full size of the remote file, and the current download/update status. 
- `get`/`put`/`cat` => Total number of files that are going to be transferred, and the current cursor (like `2 of 16 files are transffered`)

There are also possibilities for other APIs (e.g `ls`/`find`) to implement such callbacks, if they take a long time. 

Callbacks are totally optional, and no file system is expected to implement all of them for their all APIs. Also the callback system should be abstracted to level where they are totally separate from the actual logic in the client code. 

Here is the implementation that I've drafted (though I know it is a huge change, so it can drastically evolve): https://github.com/intake/filesystem_spec/compare/master...isidentical:progress?expand=1

Comments would be appreciated.

isidentical:
Resolves #668

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -10,6 +10,7 @@ except ImportError:
     import json
 
 from ..asyn import AsyncFileSystem, sync
+from ..callbacks import as_callback
 from ..core import filesystem, open
 from ..mapping import get_mapper
 from ..spec import AbstractFileSystem
@@ -199,9 +200,12 @@ class ReferenceFileSystem(AsyncFileSystem):
             f.write(data)
 
     def get_file(self, rpath, lpath, **kwargs):
+        callback = as_callback(kwargs.pop("callback", None))
         data = self.cat_file(rpath, **kwargs)
+        callback.lazy_call("set_size", len, data)
         with open(lpath, "wb") as f:
             f.write(data)
+        callback.lazy_call("absolute_update", len, data)
 
     def get(self, rpath, lpath, recursive=False, **kwargs):
         if self.fs.async_impl:
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -7,6 +7,7 @@ from stat import S_ISDIR, S_ISLNK
 import paramiko
 
 from .. import AbstractFileSystem
+from ..callbacks import as_callback
 from ..utils import infer_storage_options
 
 logger = logging.getLogger("fsspec.sftp")
@@ -119,12 +120,14 @@ class SFTPFileSystem(AbstractFileSystem):
             paths = [stat["name"] for stat in stats]
             return sorted(paths)
 
-    def put(self, lpath, rpath):
+    def put(self, lpath, rpath, **kwargs):
         logger.debug("Put file %s into %s" % (lpath, rpath))
+        callback = as_callback(kwargs.pop("callback", None))  # noqa: F841
         self.ftp.put(lpath, rpath)
 
-    def get(self, rpath, lpath):
+    def get(self, rpath, lpath, **kwargs):
         logger.debug("Get file %s into %s" % (rpath, lpath))
+        callback = as_callback(kwargs.pop("callback", None))  # noqa: F841
         self.ftp.get(rpath, lpath)
 
     def _open(self, path, mode="rb", block_size=None, **kwargs):
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -7,7 +7,6 @@ from stat import S_ISDIR, S_ISLNK
 import paramiko
 
 from .. import AbstractFileSystem
-from ..callbacks import as_callback
 from ..utils import infer_storage_options
 
 logger = logging.getLogger("fsspec.sftp")
@@ -120,14 +119,12 @@ class SFTPFileSystem(AbstractFileSystem):
             paths = [stat["name"] for stat in stats]
             return sorted(paths)
 
-    def put(self, lpath, rpath, **kwargs):
+    def put(self, lpath, rpath, callback=None, **kwargs):
         logger.debug("Put file %s into %s" % (lpath, rpath))
-        callback = as_callback(kwargs.pop("callback", None))  # noqa: F841
         self.ftp.put(lpath, rpath)
 
-    def get(self, rpath, lpath, **kwargs):
+    def get(self, rpath, lpath, callback=None, **kwargs):
         logger.debug("Get file %s into %s" % (rpath, lpath))
-        callback = as_callback(kwargs.pop("callback", None))  # noqa: F841
         self.ftp.get(rpath, lpath)
 
     def _open(self, path, mode="rb", block_size=None, **kwargs):
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -9,7 +9,7 @@ from errno import ESPIPE
 from glob import has_magic
 from hashlib import sha256
 
-from .callbacks import as_callback
+from .callbacks import as_callback, branch
 from .config import apply_config, conf
 from .dircache import DirCache
 from .transaction import Transaction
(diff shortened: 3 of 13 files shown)


fsspec/spec.py
```diff
@@ -730,13 +731,17 @@ def get_file(self, rpath, lpath, **kwargs):
...
+            return None
+
+        callback = as_callback(kwargs.pop("callback", None))
+        with self.open(rpath, "rb", **kwargs) as f1:
+            callback.lazy_call("set_size", getattr, f1, "size", None)
+            with open(lpath, "wb") as f2:
+                data = True
+                while data:
+                    data = f1.read(self.blocksize)
+                    f2.write(data)
```
martindurant, line 743:
```
len = f2.write(data)
```
so the call doesn't need to be lazy

fsspec/spec.py
```diff
@@ -750,25 +755,35 @@ def get(self, rpath, lpath, recursive=False, **kwargs):
...
-                    data = True
-                    while data:
-                        data = f1.read(self.blocksize)
-                        f2.write(data)
+            return None
+
+        callback = as_callback(kwargs.pop("callback", None))
+        with open(lpath, "rb") as f1:
+            callback.lazy_call(
+                "set_size", lambda: os.stat(lpath, follow_symlinks=False).st_size
```
martindurant, line 778:
```
size = f1.seek(0, 2)
f1.seek(0)
```
seems marginally faster (time was already spent opening the file).

fsspec/tests/test_async.py
```diff
@@ -72,7 +72,7 @@ def test_sync_wrapper_treat_timeout_0_as_none():
 
 
 @pytest.mark.skipif(sys.version_info < (3, 7), reason="no asyncio.run in <3.7")
-def test_throttled_gather(monkeypatch):
+def test_run_coros_in_chunks(monkeypatch):
```
martindurant, line 75: Why the rename?

martindurant (commented):
OK, a real review this time, so that we can move forward with this, which is certainly a good thing and improvement.

I find the API of the callback objects confusing. I think it makes sense to have concrete methods:
- set_size(func, args). Note that the length of a list/set/bytes is always fast, so I wonder where being lazy helps
- relative_update(val)
- absolute_update(val)
- normalise_path(path)
- branch(path1, path2) ; I don't think we need the function too.
rather than having to look up string values.

I assume you already have a tqdm callback in your code - can it be included? It should be possible to set the default callback to that instead of no-op. Or at least we should provide a dot-printing callback. If so, we need to try with all backends, to make sure we don't pass `callback=` to methods that don't expect it.

isidentical:
> I find the API of the callback objects confusing. I think it makes sense to have concrete methods:

Being able to delegate other objects' methods into a self-contained object is something that I find useful, especially considering the objective here (being able to support multiple plug in and out different implementations). I can observe that using strings when calling is a bit weird, and error-prone (typos) (but it is very unlikely that you will miss out this, since callbacks tend to be very interacted with the overall UI) but I'd prefer them over restricted methods in fsspec itself (since callbacks might vary depending on the filesystem). 

> set_size(func, args). Note that the length of a list/set/bytes is always fast, so I wonder where being lazy helps

For example you might want to make a `set_size` call on a cloud provider where you don't have immediate access to the object details, so you would waste an API call even if the hook is not set (the same goes to making a stat call). I'd agree on `len()` is cheap compared to others (e.g a stat call). 

> branch(path1, path2) ; I don't think we need the function too.

Would you mind expanding this further? `branch()` is very useful, and very simplifying for certain codepaths (e.g `put`/`get`) but is not a callback method, just a utility. 

> I assume you already have a tqdm callback in your code - can it be included? It should be possible to set the default callback to that instead of no-op. Or at least we should provide a dot-printing callback. If so, we need to try with all backends, to make sure we don't pass callback= to methods that don't expect it.

I don't think fsspec should display a progress bar by default. It would break a lot of assumptions about the side effects of those methods.

martindurant:
Quick answers:

> Would you mind expanding this further? branch() is very useful, and very simplifying for certain codepaths

It feels like a method, since it introspects the callbacks attributes and compares to a specific value of the callback

> I don't think fsspec should display a progress bar by default.

Sorry, that's not what I suggested; I thought we should provide the no-op callback and at least one more. The user should be able to specify which callback is the default using the config or even a function. They would now have to set the global module attribute.

isidentical:
> It feels like a method, since it introspects the callbacks attributes and compares to a specific value of the callback

But it is not a callback by itself (as well as `normalize_path`), so I feel like they should be separated from the callback object. They are simply convenience utilities for filesystem developers. 

> Sorry, that's not what I suggested; I thought we should provide the no-op callback and at least one more. The user should be able to specify which callback is the default using the config or even a function. They would now have to set the global module attribute.

Right, it is a private variable now but we might make it part of the filesystem config to make it accessible by the users (maybe not on this PR directly, since it is huge as is).

fsspec/tests/test_async.py
isidentical: It used use asyncio.gather() internally, but now it switched to asyncio.wait so I thought renaming makes sense, though can revert that

fsspec/spec.py
isidentical: Makes sense, considering we can seek on these sort of local files.

fsspec/spec.py
isidentical: Makes sense, thanks for noticing

fsspec/implementations/sftp.py
```diff
@@ -119,12 +120,14 @@ def ls(self, path, detail=False):
             paths = [stat["name"] for stat in stats]
             return sorted(paths)
 
-    def put(self, lpath, rpath):
+    def put(self, lpath, rpath, **kwargs):
         logger.debug("Put file %s into %s" % (lpath, rpath))
+        callback = as_callback(kwargs.pop("callback", None))  # noqa: F841
```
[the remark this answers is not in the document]
isidentical: I was a bit tempted to do change any signatures, but if you agree in that way I can inline all callbacks in the implementations and keep signature-change-free stuff in the specs (regular and async base file system)

fsspec/implementations/reference.py
```diff
@@ -199,9 +200,12 @@ async def _get_file(self, rpath, lpath, **kwargs):
             f.write(data)
 
     def get_file(self, rpath, lpath, **kwargs):
+        callback = as_callback(kwargs.pop("callback", None))
         data = self.cat_file(rpath, **kwargs)
```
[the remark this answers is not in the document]
isidentical: Right, though there is no guarantee any of the functions will have the true 'progress'. I didn't try to spend too much on the external implementations like reference, but can certainly try to make it improved.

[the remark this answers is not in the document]
isidentical: It seems like it is not easily possible to show true progress, since `get_file` needs to be about both reading+writing, I don't think we should show progress of cat_file. So I left it as is.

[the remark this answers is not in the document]
martindurant: I think w are safe to assume that the reading will never be faster than the writing, and usually much slower, so maybe showing the progress of cat_file is the right thing to do after all.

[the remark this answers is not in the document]
isidentical: Btw i just noticed that `ReferenceFileSystem.cat_file` doesn't show progress, since it is just `f.read()`. So I am not sure how we can achieve that here.

isidentical:
@martindurant I've tried to address all the general points, let me know if there are any more comments (or directions that you want this patch take). I'd really love to pass this before 2021.7.0.

martindurant:
I think this is ready (except for the small conflict that's come up), I'll give it a once-over tomorrow, likely will merge.

martindurant:
So, I'm going to merge this now, as it doesn't have downsides for anyone. However, I think it needs a lot of documentation and examples for usage. How do we actually get a TDQM set of progress bars from this?

I also pushed alternative code for you to consider, along the lines of my previous comments. To my (possibly old-fashioned brain), my version seems a lot easier to understand.

PR #676, opened 2021-06-18
avoid isfile calls on find as much as possible

skshetry:
In `find`, if `walk` also includes the `path` if it's a file, it should already be included in the `out`.
So, if `out` is not empty, we can safely assume that it is either a directory or a file. So, we don't need
to make an extra `isfile` call.

Though, if the path was an empty directory or if the walk does not return the file, we'll still be making an `isfile` call, which is unfortunate but better than before, where we were making an additional call for each find.

This was noticed in DVC when trying to migrate to WebdavFS, which was making remote cache querying noticeably slower when traversing through the caches.

Please let me know if I'm missing something here, or anything else that needs to be considered here. :)

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -445,7 +445,7 @@ class AbstractFileSystem(up, metaclass=_Cached):
             if withdirs:
                 files.update(dirs)
             out.update({info["name"]: info for name, info in files.items()})
-        if self.isfile(path) and path not in out:
+        if not out and self.isfile(path):
             # walk works on directories, but find should also return [path]
             # when path happens to be a file
             out[path] = {}
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -581,7 +581,7 @@ class AsyncFileSystem(AbstractFileSystem):
             if withdirs:
                 files.update(dirs)
             out.update({info["name"]: info for name, info in files.items()})
-        if (await self._isfile(path)) and path not in out:
+        if not out and (await self._isfile(path)):
             # walk works on directories, but find should also return [path]
             # when path happens to be a file
             out[path] = {}
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -232,6 +232,14 @@ def test_find_details():
         assert details[filename] == test_fs.info(filename)
 
 
+def test_find_file():
+    test_fs = DummyTestFS()
+
+    filename = "misc/foo.txt"
+    assert test_fs.find(filename) == [filename]
+    assert test_fs.find(filename, detail=True) == {filename: {}}
+
+
 def test_cache():
     fs = DummyTestFS()
     fs2 = DummyTestFS()


martindurant:
It looks good and passes tests; let me think about your argument, whether is covers all cases.

martindurant:
Looking at my *own* code, I'm not totally certain what that condition is supposed to achieve. Could you please make sure that there's a test that does `find` on a path that happens to be a file, with both `detail=True|False`?

fsspec/spec.py
```diff
@@ -445,7 +445,7 @@ def find(self, path, maxdepth=None, withdirs=False, **kwargs):
             if withdirs:
                 files.update(dirs)
             out.update({info["name"]: info for name, info in files.items()})
-        if self.isfile(path) and path not in out:
+        if not out and self.isfile(path):
```
skshetry, line 448: Let me try to explain a bit more. The `path` passed in the find could be:
1. An empty directory.
2. A directory with some files/dirs inside.
3. A file.
4.  Does not exist at all.

So, `walk` works on directories usually, but `fsspec.walk` seems to be flexible to also allow files.
https://github.com/intake/filesystem_spec/blob/0c469bfe0f4ef1e0644777791ec36e87519c6eb2/fsspec/spec.py#L406-L408

Given that, if `out` is not empty, it means either that the path was a directory or the `walk` implementation also returned info for the file itself. In both of those cases, we don't need to check for `isfile`, whereas before we were always checking for it.

If `out` is empty, it's either because the `walk` implementation only works based on directories and does not return for the files, or the directory is empty. It's unfortunate that we cannot distinguish between those two, but better than before.

skshetry: There are still a few issues that I see here with the `find` that'd be good to get rid of but could be discussed separately.
1. `find` and `walk` suppressing `FileNotFoundError`.
2. `find` not returning `info` for the file itself.

martindurant: OK, I got it; so the basic point is that we already do this check in `walk`, while having access to the full listing. Sorry for the slightly odd return API. 

For 2., there are specific problematic implementations? Perhaps we merge this and make a new issue to fix them? Note #651 , a general effort to make implementation-agnostic tests and apply them to many/all backends.

martindurant: Also would be happy to see a PR where `find` does not catch FileNotFound, and see if we depend on that behaviour anywhere.

skshetry: > there are specific problematic implementations?

I was relying on the `AbstractFileSystem.find` for the `ls` command implementation, so it is definitely an issue there.  I had to use ugly logic to fix this.

https://github.com/skshetry/webdav4/blob/3307fd568e9ad8fe59bd96894bde6f2575143c44/src/webdav4/cli.py#L579-L586

> Perhaps we merge this and make a new issue to fix them? 

>  would be happy to see a PR where find does not catch FileNotFound

If you are okay with this, let's merge this. I'll create a few more issues that I got in the way.

Especially, since `find` uses `walk`, it's `walk` where we need to fix that.

PR #678, opened 2021-06-18
LocalFileSystem.rm support for a sequence of string arguments

d-v-b:
`LocalFileSystem.rm(path)` now works if `path` is a sequence of strings. Resolves https://github.com/intake/filesystem_spec/issues/677 but not the deeper issue -- the tests in `test_mapping` are not parametrized over different filesystems, and so there is no general mechanism way to catch this kind of API mismatch.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -122,14 +122,18 @@ class LocalFileSystem(AbstractFileSystem):
         shutil.move(path1, path2)
 
     def rm(self, path, recursive=False, maxdepth=None):
-        path = self._strip_protocol(path).rstrip("/")
-        if recursive and self.isdir(path):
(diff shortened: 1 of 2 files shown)


fsspec/implementations/local.py
```diff
@@ -122,14 +122,18 @@ def mv_file(self, path1, path2, **kwargs):
         shutil.move(path1, path2)
 
     def rm(self, path, recursive=False, maxdepth=None):
```
d-v-b, line 124: `maxdepth` is not used in this function...

martindurant:
Ref: #651 about parameterised testing of many/all filesystems

martindurant:
Thanks for turning this around quickly

PR #680, opened 2021-06-21
Update entrypoint selection

jrbourbeau:
Newer versions of Python and the `importlib_metadata` backport have deprecated `get(...)` in favor of `.select(...)`. This helps avoid the following warning (xref [this CI build](https://github.com/intake/filesystem_spec/runs/2863049046))

```
 fsspec/__init__.py:49
  /home/runner/work/filesystem_spec/filesystem_spec/fsspec/__init__.py:49: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    for spec in entry_points.get("fsspec.specs", []):
```

### DIFF
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -42,11 +42,15 @@ __all__ = [
 
 if entry_points is not None:
     try:
-        entry_points = entry_points()
+        eps = entry_points()
     except TypeError:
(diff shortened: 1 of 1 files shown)


martindurant:
Bahh, this thing has changed too m many times! thanks for digging it out.

PR #682, opened 2021-06-23
Fix test_unmodel to accept ujson encoding without whitespace

mgorny:
The ujson encoding for test_unmodel data is b'"Conventions":"UGRID-0.9.0"'
(without a space after ':') that breaks the test for me.  Update
the test to make the whitespace optional.

That said, I have tried multiple ujson versions, going back to <3,
and was not able to reproduce the output with space.


martindurant:
Fascinating!

PR #683, opened 2021-06-27
Use isal if present to decompress .gz files

olivier-lacroix:
Hello,

[python-isal](https://python-isal.readthedocs.io/en/stable/index.html#) offers a faster (~25-30%) gzip compression & decompression.

This PR proposes to use python-isal if present instead of the gzip module

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -73,6 +73,11 @@ register_compression("bz2", BZ2File, "bz2")
 try:
     from isal import igzip
 
+    # igzip is meant to be used as a faster drop in replacement to gzip
+    # so its api and functions are the same as the stdlib’s module. Except
+    # where ISA-L does not support the same calls as zlib
+    # (See https://python-isal.readthedocs.io/).
+
     register_compression("gzip", igzip.IGzipFile, "gz")
 except ImportError:
     from gzip import GzipFile
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -1,6 +1,5 @@
 """Helper functions for a standard streaming compression API"""
 from bz2 import BZ2File
-from gzip import GzipFile
 from zipfile import ZipFile
 
 import fsspec.utils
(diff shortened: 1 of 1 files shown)


martindurant:
Sorry, I have not forgotten!

olivier-lacroix:
@martindurant, no worries. I opened an issue as I realised this may be better, not to rush you :-).

I have fixed the whitespace issue. other build failure seem unrelated to this PR

martindurant:
I am happy to merge this. I guess we can't test, since the target architecture on CI isn't guaranteed - but perhaps you could add a test marked with `xfail`, or at least put a comment in the code linking to `isal`'s page promising full compatibility with stdlib gzip?

Out of interest, do you now how well this compares with [`cramjam`](https://github.com/milesgranger/pyrus-cramjam)? That also gets a good fraction speed up for gzip at least in some cases (especially is the decompressed size happens to be known), but doesn't have a streaming file interface.

olivier-lacroix:
@martindurant ,

I have added a comment on isal as suggested.

There may be differences in terms of buffer allocation & such, but at first glance, isal seems faster than cramjam on a decompression task (a 700Mo CSV file). 

```
with open(filename, "rb") as f:
    file = io.BytesIO(f.read()).read()

isal.igzip.decompress(file)
# 1.93 s ± 31.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

cramjam.gzip.decompress(file)
# 3.18 s ± 12.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

gzip.decompress(file)
# 3.72 s ± 10.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

fsspec/compression.py
```diff
@@ -70,7 +69,22 @@ def unzip(infile, mode="rb", filename=None, **kwargs):
 
 register_compression("zip", unzip, "zip")
 register_compression("bz2", BZ2File, "bz2")
-register_compression("gzip", lambda f, **kwargs: GzipFile(fileobj=f, **kwargs), "gz")
+
+try:
```
martindurant, line 73: could you please mark this block as not covered, since we won't install isal on CI?

olivier-lacroix: @martindurant, what do you mean exactly? just a comment ?

martindurant: Add `  # pragma: no cover` to the end of the line with `try:` will indicate to `coverage` that we understand that this block is not used in the CI.

olivier-lacroix: Ok thanks. done!

ghuls:
@martindurant It might be worth to restrict igzip usage for decompression only as for compression the compression ratio is not so good as gzip (igzip only supports 3 compression levels). So an optin for fast, but bad compression ratio might be needed.

PR #686, opened 2021-06-28
Make remote directories before put_files

martindurant:
(async implementations, see https://github.com/dask/s3fs/issues/491 )

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -371,7 +371,7 @@ class AsyncFileSystem(AbstractFileSystem):
         lpaths = fs.expand_path(lpath, recursive=recursive)
         dirs = [l for l in lpaths if os.path.isdir(l)]
         rdirs = other_paths(dirs, rpath)
-        await asyncio.gather([self._makedirs(d, exist_ok=True) for d in rdirs])
+        await asyncio.gather(*[self._makedirs(d, exist_ok=True) for d in rdirs])
         files = sorted(set(lpaths) - set(dirs))
         rpaths = other_paths(files, rpath)
         batch_size = kwargs.pop("batch_size", self.batch_size)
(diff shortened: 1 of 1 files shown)


fsspec/asyn.py
```diff
@@ -369,12 +369,16 @@ async def _put(self, lpath, rpath, recursive=False, **kwargs):
             lpath = make_path_posix(lpath)
         fs = LocalFileSystem()
         lpaths = fs.expand_path(lpath, recursive=recursive)
-        rpaths = other_paths(lpaths, rpath)
+        dirs = [l for l in lpaths if os.path.isdir(l)]
+        rdirs = other_paths(dirs, rpath)
+        await _throttled_gather([self._makedirs(d, exist_ok=True) for d in rdirs])
```
isidentical, line 374: Since `makedirs()` doesn't do anything relevant with fds, I think it is safe to just use `asyncio.gather` for this.

martindurant: Is there any cost to always using _throttled (or whatever we finally call it) ?

isidentical: We will add callback management to it, so it is better just avoid using it if not necessary (and in this case, it isnt).

fsspec/asyn.py
```diff
@@ -369,12 +369,16 @@ async def _put(self, lpath, rpath, recursive=False, **kwargs):
             lpath = make_path_posix(lpath)
         fs = LocalFileSystem()
         lpaths = fs.expand_path(lpath, recursive=recursive)
-        rpaths = other_paths(lpaths, rpath)
+        dirs = [l for l in lpaths if os.path.isdir(l)]
+        rdirs = other_paths(dirs, rpath)
+        await asyncio.gather([self._makedirs(d, exist_ok=True) for d in rdirs])
```
isidentical, line 374:
```suggestion
        await asyncio.gather(*[self._makedirs(d, exist_ok=True) for d in rdirs])
```

martindurant: Already got it :)

isidentical:
This change broke adlfs, since AzureBlobFileSystem doesn't implement _makedirs.

PR #689, opened 2021-07-01
async: add _mkdir/_makedirs to the async methods list

### LINKED ISSUE #688: AsyncFileSystem regression: _makedirs
https://github.com/intake/filesystem_spec/pull/686 introduced a requirement over `_makedirs()` of the downstream filesystems, which is something optional according to this list;
https://github.com/intake/filesystem_spec/blob/ce0b70320d5d8704a97aafa45b17b8e60bd8ecd5/fsspec/asyn.py#L210-L230

I think there should be empty methods `_mkdir`/`_makedirs` that does nothing: 
https://github.com/intake/filesystem_spec/blob/ce0b70320d5d8704a97aafa45b17b8e60bd8ecd5/fsspec/spec.py#L264-L296

isidentical:
Resolves #688

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -226,6 +226,8 @@ async_methods = [
     "_find",
     "_du",
     "_size",
+    "_mkdir",
+    "_makedirs",
 ]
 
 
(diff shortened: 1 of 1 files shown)


martindurant:
Merging, thank.

PR #697, opened 2021-07-09
Callbacks - alternate form

martindurant:
cc @isidentical 

(needs docs, examples)


isidentical:
I've already written docs for the callbacks / other utilities so you can just copy of the docstrings from them. Regarding examples, I feel like they can be handled in a separate PR (we can probably use the snippet I shared). IMHO it is much better to get a version on the main as soon as possible and work out non-code details later since we are also going to spend some time to implement this on other filesystems and for that a released fsspec version would be great to have.

martindurant:
OK, agreed. I'll come back and fix the windows test shortly, merge and we can release.

PR #701, opened 2021-07-12
Pass around HTTP kwargs better

martindurant:
Fixes https://github.com/intake/filesystem_spec/issues/699

@rsignell-usgs : the correct kwarg you wanted is `ssl=False`. This is the aiohttp variant; for requests it would have been `verify=False`. With this PR, you can pass it either when instantiating the file system (in which case it will be in effect for all files) or in `open()` (for the one file).

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -282,7 +282,7 @@ class HTTPFileSystem(AsyncFileSystem):
         kw = self.kwargs.copy()
         kw["asynchronous"] = self.asynchronous
         kw.update(kwargs)
-        size = size or self.size(path)
+        size = size or self.info(path, **kwargs)["size"]
         session = sync(self.loop, self.set_session)
         if block_size and size:
             return HTTPFile(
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
@martindurant , when I tried to run this I got dropped into the debugger.  I hit "c" a couple of times and got the same error:

Did the right thing get merged?

martindurant:
D'oh. Pushed fix to master. Note that the kwarg should have been `ssl=False` (I don't know where `insecure` comes from!).

rsignell-usgs:
YES!!!!!   So cool Martin -- check this out!

martindurant:
One without References - yet :)

fsspec/implementations/http.py
```diff
@@ -317,16 +317,19 @@ async def _info(self, url, **kwargs):
...
-                )
+                size = await _file_size(url, size_policy=policy, session=session, **kw)
                 if size:
                     break
-            except Exception:
-                pass
+            except Exception as e:
+                import pdb
+
+                pdb.set_trace()
```
ocefpaf, line 331: Nice!

martindurant: Yeah, but it went away very quickly https://github.com/intake/filesystem_spec/commit/64144ed25c33aee4d7b5442847b289a810e5170a

ocefpaf: I'm not complaining! That is the price we pay for fast implementations :smile:

martindurant: I'm probably supposed to use pycharm or similar to do this kind of thing automatically

ocefpaf: BTW, I use pre-commits for that. I checked and the one I used only catches the new `breakpoint` syntax. I'm sure there is one for pdb too.

PR #703, opened 2021-07-12
Add `create_parents` flag and `makedirs()` method to `FTPFileSystem`

### LINKED ISSUE #702: `FTPFileSystem.mkdir()` has no support for `create_parents=True`
The interface in `spec.py` defines the signature for `mkdir` as

```python
def mkdir(self, path, create_parents=True, **kwargs):
    """
    Create directory entry at path
    For systems that don't have true directories, may create an for
    this instance only and not touch the real filesystem
    Parameters
    ----------
    path: str
        location
    create_parents: bool
        if True, this is equivalent to ``makedirs``
    kwargs:
        may be permissions, etc.
    """
    pass  # not necessary to implement, may not have directories
```

The implementation in `implementations/ftp.py` is

```python
def mkdir(self, path, **kwargs):
    self.ftp.mkd(path)
    self.invalidate_cache(self._parent(path))
```

which doesn't make use of the `create_parents` argument.

I propose adding support for this parameter by traversing up the tree and creating the directories as needed.

matteosantama:
+  `README.md` has been updated with fresh instructions for setting up a dev environment.
+ `FTPFileSystem.mkdir()` now takes a `create_parents` flag, as defined in the spec
+ `FTPFileSystem` now has a `makedirs()` method, as defined in the spec.

I didn't add any tests because I wasn't sure if there were already tests I could piggyback off of.

I added type hints to the two methods, which is inconsistent with the rest of the library (but I think helpful). 

Let me know where you want to go from here!
 
Closes #702


README.md
```diff
@@ -39,17 +39,17 @@ environments. First, install conda with tox and tox-conda in a base environment
 (eg. `conda install -c conda-forge tox tox-conda`). Calls to `tox` can then be
 used to configure a development environment and run tests.
 
-First, setup a development conda environment via `tox -e dev`. This will
-install fspec dependencies, test & dev tools, and install fsspec in develop
-mode. Then, activate the dev environment under `.tox/dev` via `conda activate .tox/dev`.
+First, setup a development conda environment via `tox -e {env}` where `env` is one of `{py36,py37,py38,py39}`.
+This will install fspec dependencies, test & dev tools, and install fsspec in develop
+mode. Then, activate the dev environment under `.tox/{env}` via `conda activate .tox/{env}`.
```
martindurant, line 44: You don't actually need to activate the env, tox can automatically run your test in the env by calling `tox` alone with no arguments. The `TOXENV` environment variable defines which of the defined environments gets used.

fsspec/implementations/ftp.py
```diff
@@ -179,11 +180,20 @@ def _rm(self, path):
...
+        parent = self._parent(path)
+        if parent != self.root_marker and not self.exists(parent) and create_parents:
+            self.mkdir(parent, create_parents=create_parents)
+
         self.ftp.mkd(path)
         self.invalidate_cache(self._parent(path))
 
+    def makedirs(self, path: str, exist_ok: bool = False) -> None:
+        if self.exists(path) and not exist_ok:
+            raise FileExistsError(f"{path} exists and {exist_ok=}")
```
martindurant, line 194: This fstring is causing a syntax error during linting

martindurant:
Thanks for producing this so quickly. Would you mind adding a simple test in test_ftp?

martindurant:
> if there were already tests I could piggyback off 

Yep, copy from one of the others in test_ftp that uses the writable fixture.

martindurant:
TODO: ever one of those calls to the ftp instance should be in a block to catch the errors and translate them to PermissionError, FileNotFoundError, etc.

matteosantama:
Nice! Thanks for wrapping this up @martindurant. I got caught up in some other things

Issue #704, opened 2021-07-14
SMB Protocol support for storage_options


mattinbits:
When using `fsspec` with the `smb` protocol, e.g. `with fsspec.open("smb://myserver/share/file.txt") as f:` it does not seem to be possible to pass storage options to the underlying `smbprotocol` library.

The implementation calls `smbclient.open_file` via `_open` here: https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/smb.py#L202

This method on `smbclient` supports options such as `share_access`: https://github.com/jborean93/smbprotocol/blob/master/smbclient/_os.py#L311

However, although `_open` in `smb.py` passes `kwargs` to `smbclient.open_file`, the `kwargs` are not propagated from the outer call to `fsspec.open` https://github.com/intake/filesystem_spec/blob/master/fsspec/core.py#L102 

Is there an established pattern for passing these implementation-specific options through to the `_open` call in the implementation? Should the implementation capture them in the constructor of `SMBFileSystem` since the `storage_options` appear to be passed when an instance of `AbstractFileSystem` is instantiated here? https://github.com/intake/filesystem_spec/blob/master/fsspec/core.py#L376

martindurant:
Yes, you are quite right: the "storage options" in `fsspec.open` refer to global parameters used in creating the file-system, and do not fo to its file-open method. We could reasonably add another optional argument to `fsspec.open` specifically for this purpose. Note that most versions of `_open()` don't take any/many options.

The trouble that will need to be faced, is that `fsspec.open` can create multiple file systems in the case of a chained URL. It can already pass arguments to the various components, but whether and where `_open` will be called isn't clear or consistent (e.g., zip operates on one open file, but simplecache uses get() to download files and never uses open at all). Maybe the extra argument in `fsspec.open` would always only go to the "outermost" `_open`.

> Should the implementation capture them in the constructor

This is the only way to do it now

mattinbits:
Thank you for the rapid response!

> This is the only way to do it now

I will have a go at a PR on this basis.

PR #706, opened 2021-07-14
Support share_access in smb implementation

mattinbits:
To solve #704

### DIFF
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -66,6 +66,7 @@ class SMBFileSystem(AbstractFileSystem):
         password=None,
         timeout=60,
         encrypt=None,
+        share_access=None,
         **kwargs,
     ):
         """
@@ -83,13 +84,22 @@ class SMBFileSystem(AbstractFileSystem):
             Port to connect with. Usually 445, sometimes 139.
         username: str or None
             Username to connect with. Required if Kerberos auth is not being used.
-        password: str of None
+        password: str or None
             User's password on the server, if using username
         timeout: int
             Connection timeout in seconds
         encrypt: bool
             Whether to force encryption or not, once this has been set to True
             the session cannot be changed back to False.
+        share_access: str or None
+            Specifies the default access applied to file open operations
+            performed with this file system object.
+            This affects whether other processes can concurrently open a handle
+            to the same file.
+              None (the default): exclusively locks the file until closed.
+              'r': Allow other handles to be opened with read access.
+              'w': Allow other handles to be opened with write access.
+              'd': Allow other handles to be opened with delete access.
         """
         super(SMBFileSystem, self).__init__(**kwargs)
         self.host = host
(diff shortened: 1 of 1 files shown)


fsspec/implementations/smb.py
```diff
@@ -83,13 +84,20 @@ def __init__(
...
+        password: str or None
             User's password on the server, if using username
         timeout: int
             Connection timeout in seconds
         encrypt: bool
             Whether to force encryption or not, once this has been set to True
             the session cannot be changed back to False.
+        share_access: str or None
+            specifies the type of access that is allowed when a handle to this 
+            file is opened by another process.
```
martindurant, line 96: The wording here is a little unclear - we are defining what the default share access will be for any open-file operations, unless otherwise specified.

fsspec/implementations/smb.py
```diff
@@ -199,7 +208,13 @@ def _open(
         if "w" in mode and autocommit is False:
             temp = _as_temp_path(self.host, path, self.temppath)
             return SMBFileOpener(wpath, temp, mode, block_size=bls, **kwargs)
-        return smbclient.open_file(wpath, mode, buffering=bls, **kwargs)
+        return smbclient.open_file(
```
martindurant, line 211: I suggest that kwargs should still allow for overriding the default value for those that use fs.open directly. One way:
```python
    share_access = kwargs.pop("share_access", self.share_access)
    return smbclient.open_file(.., share_access=share_access, **kwargs)
```
(or you can include it as an explicit kwarg 
```python
_open(.., share_access=None, **kwargs):
    share_access = share_access or self.share_access
    ...
```

fsspec/implementations/smb.py
mattinbits: Yes you're right. I've reworded it.

fsspec/implementations/smb.py
mattinbits: Right, good idea. I've opted for the first option, since this method overrides from AbstractFileSystem, I prefer not to change its signature.

mattinbits:
This is the basic test I've run locally to check that it works:

```
import fsspec
file = "smb://my.server:445/MYSHARE/smb_test/test.txt"

with fsspec.open(file, share_access="r") as f:
    with fsspec.open(file, share_access="r") as f2:
        print(f2.read())

smb_fs = fsspec.filesystem(
    "smb",
    host="my.server",
    port=445,
)

with smb_fs.open("/MYSHARE/smb_test/test.txt", share_access="r") as f:
    with smb_fs.open("/MYSHARE/smb_test/test.txt", share_access="r") as f2:
        print(f2.read())
```

fsspec/implementations/smb.py
```diff
@@ -193,16 +204,24 @@ def _open(
...
+            wpath, mode, buffering=bls, share_access=share_access, **kwargs
+        )
 
     def copy(self, path1, path2, **kwargs):
-        """ Copy within two locations in the same filesystem"""
```
mattinbits, line 205: This edit is a result of running "black" against the file, due to the linting failures shown on a previous build. I'm a bit unsure about it since this isn't a line I changed, so not sure why linting wasn't a problem already on master.

martindurant: Maybe a new version of `black`

martindurant:
Thanks @mattinbits

PR #710, opened 2021-07-15
check for bytes return type and dictifiy it

### LINKED ISSUE #707: return type of fs.cat
`fsmap.getitems` assumes that `fs.cat` returns a dict: see https://github.com/intake/filesystem_spec/blob/master/fsspec/mapping.py#L95. However, there is a branch in `fs.cat` that returns bytes, not a dict: see https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py#L728. Is this deliberate? Should `fsmap.getitems` convert the non-dict value to a dict, or should `fs.cat` always return a dict (as per its docstring)?

And maybe some type annotations would be helpful here.

(This came up while testing zarr against the latest fsspec, which causes a test failure because of this API issue)

d-v-b:
Fixes #707

`self.fs.cat` returns an instance of `bytes` if `keys2` has one element, which causes an error when `fsmap.getitems` attempts to access the `items` property of that value. This PR checks if `self.fs.cat` returned an instance of `bytes`, and if so, sticks it in a dict. This kind of problem might be an argument for type annotations, which would have flagged this issue earlier.

### DIFF
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -88,6 +88,8 @@ class FSMap(MutableMapping):
         oe = on_error if on_error == "raise" else "return"
         try:
             out = self.fs.cat(keys2, on_error=oe)
+            if isinstance(out, bytes):
+                out = {keys2[0]: out}
         except self.missing_exceptions as e:
             raise KeyError from e
         out = {


martindurant:
> might be an argument for type annotations

I am not against it in principle, but where I've been forced to use it myself (pandas, xarray), it took far more effort to twist the type system to what I need, and several errors in that process. Perhaps my coding style is inherently type-unfriendly.

martindurant:
I suppose it's worth a test? 
NB: the s3fs failure is some intermittent conda problem.

martindurant:
Can someone confirm that this fixes xarray? @keewis

keewis:
it does! I can confirm that the `xarray` test fails with `2021.07.0` and passes with this branch

martindurant:
Rerunning tests, will merge when green.

martindurant:
Thank you @d-v-b . I still don't quite understand why this became an issue now, but never mind!

jakirkham:
Would it be possible to tag a release with this change @martindurant ? 🙂 No worries if you are busy

martindurant:
Waiting on #731

PR #713, opened 2021-07-20
Support rpath list in asyn put

baxen:
Fixes intake/filesystem_spec#712

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -378,18 +378,20 @@ class AsyncFileSystem(AbstractFileSystem):
             lpath = make_path_posix(lpath)
         fs = LocalFileSystem()
         lpaths = fs.expand_path(lpath, recursive=recursive)
-        dirs = [l for l in lpaths if os.path.isdir(l)]
-        rdirs = other_paths(dirs, rpath)
+        rpaths = other_paths(lpaths, rpath)
+
+        is_dir = {l: os.path.isdir(l) for l in lpaths}
+        rdirs = [r for l, r in zip(lpaths, rpaths) if is_dir[l]]
+        file_pairs = [(l, r) for l, r in zip(lpaths, rpaths) if not is_dir[l]]
+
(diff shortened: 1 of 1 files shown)


baxen:
I couldn't really find a good place to add a test for this in fsspec, looks like this is more something that is deferred to tests in the concrete async filesystems like GCSFS? I could possibly extend https://github.com/dask/gcsfs/blob/main/gcsfs/tests/test_core.py#L460 to cover this regression

martindurant:
Is this an issue for `_get` too? HTTPFileSystem is the only implementation within this repo that's async, and it's read-only. So any test would need to be reading against that.

baxen:
Nope no issue with _get - this was coming just from the code that was handling making directories. And I think https implementation has a good test of this already in test_multi_download

fsspec/asyn.py
```diff
@@ -378,18 +378,19 @@ async def _put(
             lpath = make_path_posix(lpath)
         fs = LocalFileSystem()
         lpaths = fs.expand_path(lpath, recursive=recursive)
-        dirs = [l for l in lpaths if os.path.isdir(l)]
-        rdirs = other_paths(dirs, rpath)
+        rpaths = other_paths(lpaths, rpath)
+
+        rdirs = [r for l, r in zip(lpaths, rpaths) if os.path.isdir(l)]
+        file_pairs = [(l, r) for l, r in zip(lpaths, rpaths) if not os.path.isdir(l)]
```
martindurant, line 384: Could we possibly avoid calling `isdir` on every file twice?

martindurant (commented):
All looks good and function. Just the one question (see below).

I agree now that this will only get tested in gcsfs, but that's OK. Note that gcsfs tests run in CI here, so a test there ( dask/gcsfs#401 ) will show up here too after merging.

fsspec/asyn.py
baxen: This is now called just once per file

PR #718, opened 2021-07-30
Issue 717 fix error handling file size

### LINKED ISSUE #717: Error reading pyarrow parquet dataset with HTTPFilesystem and AWS presigned urls
When trying out [databricks delta sharing](https://databricks.com/blog/2021/05/26/introducing-delta-sharing-an-open-protocol-for-secure-data-sharing.html) I encountered a bug when reading the delta file into a pyarrow pandas dataset.

Investigation lead to discovering a problem in the determination of the file size within the [fsspec](https://pypi.org/project/fsspec/) python library.

In https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/http.py#L673-L681 there is no error handling present. 
And with the presigned url a HEAD request might return a 403 is the signing was done based on the GET method.

Working on making a proper PR. 

for reproducing the error I create a smal repo: https://github.com/krisgeus/filesystem_spec_issue_reproduced

krisgeus:
fixes #717

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -685,7 +685,7 @@ async def _file_size(url, session=None, size_policy="head", **kwargs):
                 return int(r.headers["Content-Length"])
             elif "Content-Range" in r.headers:
                 return int(r.headers["Content-Range"].split("/")[1])
-        except requests.HTTPError:
+        except aiohttp.ClientResponseError:
             logger.debug("Error retrieving file size")
             return None
         r.close()
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -573,13 +573,16 @@ class HTTPFile(AbstractBufferedFile):
         pass
 
     def __reduce__(self):
-        return reopen, (
-            self.fs,
-            self.url,
-            self.mode,
-            self.blocksize,
-            self.cache.name,
-            self.size,
+        return (
+            reopen,
+            (
+                self.fs,
+                self.url,
+                self.mode,
+                self.blocksize,
+                self.cache.name,
+                self.size,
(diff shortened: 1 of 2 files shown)


martindurant:
Thanks for the contribution and diagnosing the original problem.
Please fix the lint, and this will be good to go.

krisgeus:
@martindurant Fixed linting. Can you approve the workflow?

fsspec/implementations/http.py
```diff
@@ -671,13 +671,19 @@ async def _file_size(url, session=None, size_policy="head", **kwargs):
...
+            r.raise_for_status()
+
+            # TODO:
+            #  recognise lack of 'Accept-Ranges', or  'Accept-Ranges': 'none' (not 'bytes')
+            #  to mean streaming only, no random access => return None
+            if "Content-Length" in r.headers:
+                return int(r.headers["Content-Length"])
+            elif "Content-Range" in r.headers:
+                return int(r.headers["Content-Range"].split("/")[1])
+        except:
```
martindurant, line 684: I think flake will still complain here; indeed it would be better to know what sort of exception we expect. `aiohttp.ClientError` ?

krisgeus: Indeed. Fixed it and ran the pre-commit hook locally. Should be all good now.

martindurant: I see you've used a `requests` error - but these should come from aiohttp, no? I don't think the two are compatible. Your test does pass somehow... can you please ensure that this block actually activates during your test?

krisgeus: Fixed it. Was mistakenly thinking it was a reqeusts.session that was used.

PR #719, opened 2021-07-30
tiny bugfix to WholeFileCacheFileSystem.cat

AlpAribal:
Variable `path` defined in the for loop shadows the argument `path` of the `cat` method. Even if the original argument `path` is a list, after the for loop, the path has become a `str`. This affects the `isinstance(path, str)` check below, changing the behavior of the method.

This PR fixes this by renaming the variable used in the for loop.


martindurant:
Thanks for picking this up.I'm sure I've made this mistake plenty of times, would hope that my IDE notices.

Issue #588, opened 2021-03-29
Feature Request: Copy between filesystems


c-thiel:
It would be great if we could copy between two different remote filesystems with fsspec.

I think this can be achieved with minimal effort if we check if the `lpath` is a path in a known filesystem in `get/put_file` functions in [fsspec.spec](https://github.com/intake/filesystem_spec/blob/f0971839ae93cfeba603bd90f206e81210089d2c/fsspec/spec.py#L707). Then we could just use `lpath.fs.open()` instead of open(lpath).

martindurant:
It is a reasonable request, but not I think that easy to do. Indeed, https://github.com/intake/filesystem_spec/issues/41 discusses the idea of a "dispatch" filesystem, where you always give the full URLs where now you give fs-specific paths. The trouble is, most implementations need arguments to work. Either you would have to pass through sets of args for every user method (twice for copy, for origin and target), or have a concept of the "current" instance of each file system type. It would perhaps work better with fsspec path-like objects, where the filesystem is encoded in the object.

aside: I would call this operation "copy" rather than get/put, which to me are synonymous with download/upload.

Issue #41, opened 2019-04-24
add dispatch filesystem


martindurant:
Like `open`/`open_files`, which find the file-system of interest and use it with parameters, could implement a generic top-level file-system which finds the correct instance and use it depending on the protocol implicit in the path given. This would become the primary user-facing API.

This can also be the place to implement globs or recursive where multiple files make sense.

Should allow copy/move/(sync?) between file-systems.

d4l3k:
This would be very helpful to have. Specifically having methods for file put/get without having to create a filesystem via `fsspec.get_fs_token_paths` would be very convenient.

PR #723, opened 2021-08-13
Add top-level `copy`/`cp` function

cisaacstern:
First draft for the feature proposed by @martindurant in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/179#issuecomment-896039997. The included test passes, as do the CI tests for https://github.com/pangeo-forge/pangeo-forge-recipes/pull/179 (with fsspec installed from here).

Remaining TODO:

- [ ] Add callbacks

Martin, my understanding is that within `fsspec.cp`, this should look like:
```python
while True:
    data = a.read(a.fs.blocksize, callback=callback)
```
Except that [`AbstractBufferedFile.read`](https://github.com/intake/filesystem_spec/blob/3cdfd6cf8386857815535a32f747cd2033b30da7/fsspec/spec.py#L1463) does not implement callbacks. (And neither does and [`HTTPFile.read`](https://github.com/intake/filesystem_spec/blob/3cdfd6cf8386857815535a32f747cd2033b30da7/fsspec/implementations/http.py#L477), as one backend example.) Do I understand correctly that this is because callbacks are new and sporadically implemented, and therefore the next step is to add callbacks to `AbstractBufferedFile.read` following this example:

https://github.com/intake/filesystem_spec/blob/5c58feb616a1e39c234fe85394d990ae36737203/fsspec/spec.py#L730-L743

?

- [ ] Special-casing for FTP servers that don't allow random access reads

This is a second step after completing the callbacks implementation, but my impression is it should be something like:

```python
# fsspec.cp
while True:
    data = a.read(a.fs.blocksize, callback=callback)
    if not data:
        break
    if not isinstance(a.fs, FTPFileSystem):
        b.write(data)
```

...and above that, we'll need to make sure `b.write(data)` is included in the callback if `a.fs` is an FTPFileSystem.

Two questions about this:
1. Do we want to force writing within the callback for __*all*__ FTPFileSystems? Or just the subset that don't allow random access?
2. For those that do require writing within the callback, might a good way to implement that be to pass the callback to a new `FTPFileSystem._fetch_all` method (following the naming pattern of `HTTPFileSystem._fetch_all`), which does something like:
```python
def _fetch_all(self, rpath, callback):    
    with self.ftp.transfercmd("RETR %s" % rpath) as conn:
        data = conn.recv(self.blocksize)
        callback.relative_update(data)  # assumes that a write function is one of the callback hooks
```
?

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -468,6 +468,35 @@ def open_local(url, mode="rb", **storage_options):
     return paths
 
 
+def cp(filea, fileb: OpenFile, callback=None, **open_kwargs):
+    """Copy ``filea`` to ``fileb``.
+
+    Parameters
+    ----------
+    filea: str
+        The path to a local or remote file.
+    fileb: OpenFile
+        The target to write to. Must be in write mode.
+    callback:
+
+    open_kwargs: dict
+        Additional arguments passed to ``fsspec.open``
+        when opening ``filea``.
+    """
+    with open(filea, **open_kwargs) as a:
+        with fileb as b:
+            while True:
+                data = a.read(a.fs.blocksize)
+                if not data:
+                    break
+                b.write(data)
+
+
+def copy():
+    """Alias for `cp`"""
+    pass
+
+
 def get_compression(urlpath, compression):
     if compression == "infer":
         compression = infer_compression(urlpath)
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -10,7 +10,7 @@ except ImportError:  # python < 3.8
 from . import caching
 from ._version import get_versions
 from .callbacks import Callback
-from .core import get_fs_token_paths, open, open_files, open_local, copy, cp
+from .core import copy, cp, get_fs_token_paths, open, open_files, open_local
 from .exceptions import FSTimeoutError
 from .mapping import FSMap, get_mapper
 from .registry import (
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -10,7 +10,7 @@ except ImportError:  # python < 3.8
 from . import caching
 from ._version import get_versions
 from .callbacks import Callback
-from .core import get_fs_token_paths, open, open_files, open_local
+from .core import get_fs_token_paths, open, open_files, open_local, copy, cp
 from .exceptions import FSTimeoutError
 from .mapping import FSMap, get_mapper
 from .registry import (
@@ -37,6 +37,8 @@ __all__ = [
     "open",
     "open_files",
     "open_local",
+    "copy",
+    "cp",
     "registry",
     "caching",
     "Callback",
--- a/fsspec/tests/test_api.py
+++ b/fsspec/tests/test_api.py
@@ -187,6 +187,21 @@ def test_chained_fs():
     assert os.listdir(d2) == ["f1"]
 
 
+def test_cp():
+    d = tempfile.mkdtemp()
+    f1 = os.path.join(d, "f1")
+    f2 = os.path.join(d, "f2")
+    with open(f1, "wb") as f:
+        f.write(b"test")
+
+    target = fsspec.open(f2, mode="wb")
+    fsspec.cp(f1, target)
+    target.close()
+
+    with fsspec.open(f2) as f:
+        assert f.read() == b"test"
+
+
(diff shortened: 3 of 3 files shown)


martindurant:
cc @isidentical for questions about callbacks. As far as I understand, callbacks are not implemented for any file-like objects, only for get/put/cat/pipe and potentially bulk rm/cp.

fsspec/core.py
```diff
@@ -468,6 +468,35 @@ def open_local(url, mode="rb", **storage_options):
     return paths
 
 
+def cp(filea, fileb: OpenFile, callback=None, **open_kwargs):
```
martindurant, line 471: Note that the filesystem API `cp` has `recursive` and other parameters - it can do batch jobs. I had imagined the code here as a file-specific copy method of a filesystem which could be overidden in various specific implementations (like FTP!).

This code doesn't use the given callback.

fsspec/core.py
```diff
@@ -468,6 +468,35 @@ def open_local(url, mode="rb", **storage_options):
...
+    with open(filea, **open_kwargs) as a:
+        with fileb as b:
+            while True:
+                data = a.read(a.fs.blocksize)
+                if not data:
+                    break
+                b.write(data)
+
+
+def copy():
```
martindurant, line 495: `copy = cp`

martindurant (commented):
My intent for a top-level function would be that you can do `cp("ftp://path/*", "s3://bucket/path")`, just like with a (posix) command line. 

That means that you would need to pass multiple open_kwargs for the two filesystems - but not for methods such as `rm` that only need one path. Maybe we could solve this by encouraging people to use the configuration system to define defaults for each protocol, or have some other mechanism for registering the "current" instance(s) for such operations.

#588 #41

isidentical:
> As far as I understand, callbacks are not implemented for any file-like objects, only for get/put/cat/pipe and potentially bulk rm/cp.

No. We do not use callbacks on any file-like objects. We only implement them to raw data transfer methods (e.g `get_file` callback will return `transferred_bytes / len(data)`, unit: bytes) and bulk file transfer methods (e.g `put` callback will return `transferred_files / len(total_files)`, unit: whole files).

fsspec/core.py
brl0: Thanks for this, I think it will be great to have this useful feature!

Sorry for the nitpick, but might be nice to use `src` and `dst` as argument names similar to the python [shutil.copyfile](https://docs.python.org/3/library/shutil.html#shutil.copyfile) function, also maybe mark them as positional only as well.

martindurant: Certainly fine with src/dst. I'm less sure there is a need to make them position-only.

cisaacstern:
> just like with a (posix) command line.

Gotcha, happy to re-work the draft to align with this. Setting aside momentarily where `src_fs_kwargs`, `src_open_kwargs`, and `dst_open_kwargs` come from (perhaps they are passed directly to the top-level function, or configured elsewhere), just want to confirm I'm on the right track with the design:

```python
# fsspec/core.py

def cp(src, dst, src_fs_kwargs, src_open_kwargs, dst_open_kwargs, callback=None):
    src_fs, _ = url_to_fs(src, **src_fs_kwargs)
    open_dst = open(dst, **dst_open_kwargs)
    src_fs.cp(src, open_dst, src_open_kwargs, callback=callback)
```
```python
# fsspec/implementations/some_implementation.py

def cp_file():
    # already exists in some implementations, for intra-filesystem copying

def cp_file_x():
    # the new inter-filesystem copy function, non-recursive
    # callbacks implemented here

def cp():
    # recursive wrapper for `cp_file_x` and/or `cp_file`
    # called by fsspec/core.py

```
Does this seem to be the right general direction?

martindurant:
> Does this seem to be the right general direction?

Yes I think so, except maybe I would do `expand_paths` right in the generic `cp`, and then call the `cp_file_x` method rather than the fs-specific `cp` (to keep the latter pure). 

Have you seen #732 ?

cisaacstern:
@martindurant, should we close this? Hopefully the discussion was useful, but I assume you'll be starting fresh when you begin working on a top-level `copy`?

martindurant:
Yes, I am starting fresh (but my initial version is not that different, of course!).

Issue #727, opened 2021-08-20
glob reads irrelevant directories; can fail on permissions


drorspei:
The implementation of glob calls find on the maximal path not including magic, which reads all files in the matching subdirectories, possibly up to a maximal depth.

This is a large pessimization, both on performance and on permissions:
1. Say the glob path is `/data/year=*/month=1/*.csv`, and each "year" directory has 12 "month" subdirectories. Currently glob will read x12 directories more than it needs to, since it will reads all month subdirectories before pattern matching.
2. Say the glob path is `/users/year=*/user=drorspei/*`, where each "year" directory has many user directories, and I only have permissions to list my own user's files. Then glob can fail since it will try to list files in subdirectories which I don't have permissions for. This won't happen on all "filesystems", but on something like S3 it's common to have `List*` permissions on some prefixes and not others.

Is it possible for glob to read only the subdirectories that match the pattern?

PR #728, opened 2021-08-20
Make glob only read relevant paths

drorspei:
Hi!

This is to resolve issue #727 that I opened.

The code here is copied from CPython's pathlib glob implementation. I don't know anything about licenses, and don't know if the work needs to be attributed somehow somewhere in the code.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -212,6 +212,126 @@ async def _run_coros_in_chunks(
     return results
 
 
+def _make_selector(pattern_parts, fs):
+    pat = pattern_parts[0]
+    child_parts = pattern_parts[1:]
+    if pat == "**":
+        cls = _RecursiveWildcardSelector
+    elif "**" in pat:
+        raise ValueError("Invalid pattern: '**' can only be an entire path component")
+    elif "*" in pat or "?" in pat or "[" in pat:
+        cls = _WildcardSelector
+    else:
+        cls = _PreciseSelector
+    return cls(pat, child_parts, fs)
+
+
+class _Selector:
+    """A selector matches a specific glob pattern part against the children
+    of a given path."""
+
+    def __init__(self, child_parts, fs):
+        self.fs = fs
+        self.child_parts = child_parts
+        if child_parts:
+            self.successor = _make_selector(child_parts, fs)
+            self.dironly = True
+        else:
+            self.successor = _TerminatingSelector()
+            self.dironly = False
+
+    async def select_from(self, parent_path, kwargs):
+        """Iterate over all child paths of `parent_path` matched by this
+        selector.  This can contain parent_path itself."""
+        if parent_path != self.fs.root_marker and not await self.fs._isdir(parent_path):
+            return
+        kwargs = dict(kwargs)
+        kwargs.pop("detail", None)
+        async for path in self._select_from(parent_path, kwargs):
+            yield path
+
+
+class _TerminatingSelector:
+    async def _select_from(self, parent_path, kwargs):
+        yield parent_path
+
+
+class _PreciseSelector(_Selector):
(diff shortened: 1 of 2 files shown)


martindurant:
First running the tests to see if your method works :) 
Then we can talk about whether this is a reasonable thing to do.

drorspei:
Cool cool

I see some tests indeed fail. I didn't run the full suite for lack of docker setup on my laptop; I'll get a beefier machine and make sure these pass as well.

Note that there are inconsistencies in what is expected from glob, shown in these test failures. Specifically, my version returns exactly what is returned from ls, but it's not clear which calls are expected to return the beginning front slash and which don't.

martindurant:
I believe for HDFS, paths should always start with `"/"`. Perhaps the implementation is missing `root_marker = "/"` as a class attribute.

drorspei:
Cool. I'm actually having trouble running the hdfs tests locally, trying to resolve ^_^

Do you know if I need any extra configuration for any of these tests?

In the ftp implementation, using `ls(path, detail=True)` will return results with the root marker, and using `ls(path, detail=False)` will return without. To make it compatible, I'll update the pull request to only use `detail=True` in all calls to `ls`.

martindurant:
> To make it compatible, I'll update the pull request to only use detail=True

Better would be to fix it!

At this point, I hope that other people chime in here about the general idea of this PR (but it is Friday, so this might take a while). In the general case, skipping `ls` on directories that we know aren't required - as would be done by default find/walk - seems like a good idea. I would at this point probably put all the relevant code in a separate module and indeed include whatever license/notice requirements. Also worth mentioning, is that the current `glob` is derived from python's builtin `glob.glob` and why pathlib's `glob` is different. 

Note that some backends (s3, ...) can do `find` with a single call, instead of the potentially many calls to walk a directory, so they should *not* be changed. For them, it will almost always be much faster with `find`. It should be checked what happens is some sub-tree is permission locked (I reckon nothing, they just won't show up).

drorspei:
Huh, I'm actually doing all of this because my queries to S3 are slow, and this pull request reduces my call times by more than x10.

The implementation in s3fs is a bit complicated, it's far from a single call. First, if there's no "**" in the path, then it is in fact many calls to `walk`. Second, if there is a "**" in the path, the performance will depend on how many objects are in the directory that do not match the pattern: if a small amount, then sure, the call to `list_objects_v2` with pagination will be fast. But if there are many objects, as I have, then again it is much slower. I'm not sure if pagination is considered "a single call", but as far as the network layer goes it's definitely not.

I've changed the code to pass the (s)ftp and cache tests. I still need to configure hdfs locally somehow to test that, do you maybe have a reference for this?

drorspei:
Hej hej!

Ok! I thought a bit more, tell me what you think about the following change to the code and whether you think it will fit well with s3fs as well as other systems: find the first "\*\*" in the pattern, and use the new "selectors" for everything up to the "\*\*" - meaning only precise matches and using `fs.ls` for "\*". Then from the "\*\*" use exactly a single call to `fs.find` as it is in the current implementation of `glob`. This way patterns without "\*\*" benefit from reading only relevant directories, and those with it call `list_objects_v2` only the part of the path that needs it, so a pattern like `year=*/month=01/**/*.csv` won't get x12 objects to match over (which is similar to the data I'm dealing with daily).

I think this is definitely better than the current pull request, and for some use cases it's better than the current fsspec implementation, though not all. For example, continuing with my example above of `year=*/month=01/**/*.csv`, if there is only the single "month=01" in all the "years=*", then a single call to `find` is better than a call to `ls` and then multiple find calls.

We can also play with the design and let users have a choice.

What do you think?

drorspei:
Hej hej
Sorry, I don't have enough time to push this. Maybe some time in the future ^_^

PR #731, opened 2021-08-23
http: general enhancements

isidentical:
TODO:
- [x] add tests
- [x] add docs

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-2021.08.0
+2021.08.1
 ---------
 
 Enhancements
@@ -11,7 +11,7 @@ Enhancements
     - get_file/put_file APIs now support callbacks
     - New put_file API for transferring data to the remote server (chunked)
     - Customizable client initializers (through passing ``get_client`` argument)
-    - Support for ``checksum`` (either ``ETag`` or ``MD5-Checksum``) on ``info()`` calls
+    - Support for various checksum / fingerprint headers in ``info()`` (``ETag``, ``MD5-Checksum``, ``Digest``)
 
 2021.07.0
 ---------
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+2021.08.0
+---------
+
+Enhancements
+
+    HTTTP
+
+    - get_file/put_file APIs now support callbacks
+    - New put_file API for transferring data to the remote server (chunked)
+    - Customizable client initializers (through passing ``get_client`` argument)
+    - Support for ``checksum`` (either ``ETag`` or ``MD5-Checksum``) on ``info()`` calls
+
 2021.07.0
 ---------
 
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -361,7 +361,7 @@ class HTTPFileSystem(AsyncFileSystem):
         corresponding file will not work).
         """
         info = {}
-        failed = None
+        failed = exc = None
         session = await self.set_session()
 
         for policy in ["head", "get"]:
@@ -379,10 +379,12 @@ class HTTPFileSystem(AsyncFileSystem):
                     break
             except Exception as e:
                 failed = policy
+                exc = e
                 logger.debug((str(e)))
 
-        if failed and info.get("size") is None:
-            raise FileNotFoundError(url)
+        if failed == "get":
+            # If get failed, then raise a FileNotFoundError
+            raise FileNotFoundError(url) from exc
         else:
             return {"name": url, "size": None, **info, "type": "file"}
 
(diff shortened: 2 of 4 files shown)


martindurant:
Sorry I didn't comment yet. I can look in detail, but currently `test_entry_points_registered_on_import ` is failing in py36/37

isidentical:
Ah, didnt notice that. Seems irrelevant though

isidentical:
Oh, you already fixed, nvm.

docs/source/changelog.rst
```diff
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+2021.08.0
```
martindurant, line 4: This will be 2021.08.1, because s3fs has a release in between due to aiobotocore; I'd like the versions to match exactly.

docs/source/changelog.rst
```diff
@@ -1,6 +1,18 @@
...
+---------
+
+Enhancements
+
+    HTTTP
+
+    - get_file/put_file APIs now support callbacks
+    - New put_file API for transferring data to the remote server (chunked)
+    - Customizable client initializers (through passing ``get_client`` argument)
+    - Support for ``checksum`` (either ``ETag`` or ``MD5-Checksum``) on ``info()`` calls
```
martindurant, line 14: ETag maybe needs more explanation?

fsspec/implementations/http.py
```diff
@@ -223,18 +230,55 @@ async def _cat_file(self, url, start=None, end=None, **kwargs):
...
+    async def _get_file(
+        self, rpath, lpath, chunk_size=5 * 2 ** 20, callback=_DEFAULT_CALLBACK, **kwargs
+    ):
         kw = self.kwargs.copy()
         kw.update(kwargs)
         logger.debug(rpath)
         session = await self.set_session()
         async with session.get(rpath, **self.kwargs) as r:
+            try:
+                size = int(r.headers["content-length"])
```
martindurant, line 242: What are the chances of this being wrong due to transfer compression? (`_file_info` has `head["Accept-Encoding"] = "identity"`)

fsspec/implementations/http.py
```diff
@@ -223,18 +230,55 @@ async def _cat_file(self, url, start=None, end=None, **kwargs):
...
+                    yield chunk
+                    callback.relative_update(len(chunk))
+                    chunk = f.read(64 * 1024)
+
+        kw = self.kwargs.copy()
+        kw.update(kwargs)
+        session = await self.set_session()
+        meth = getattr(session, method)
+
+        async with meth(lpath, data=gen_chunks(), **kw) as resp:
```
martindurant, line 280: This is clever! Does it work for both POST and PUT?

fsspec/implementations/http.py
```diff
@@ -316,22 +360,33 @@ async def _info(self, url, **kwargs):
...
+                        session=session,
+                        **self.kwargs,
+                        **kwargs,
+                    )
+                )
+                if info.get("size") is not None:
                     break
             except Exception as e:
+                failed = policy
+                exc = e
```
martindurant, line 382: Future reference: any place where we capture the exception and carry it out of the block like this, we are likely making circular references via the traceback frames. This might cause memory leaks.

fsspec/implementations/http.py
```diff
@@ -666,29 +722,38 @@ async def _file_size(url, session=None, size_policy="head", **kwargs):
...
+        # TODO:
+        #  recognise lack of 'Accept-Ranges',
+        #                 or 'Accept-Ranges': 'none' (not 'bytes')
+        #  to mean streaming only, no random access => return None
+        if "Content-Length" in r.headers:
+            info["size"] = int(r.headers["Content-Length"])
+        elif "Content-Range" in r.headers:
+            info["size"] = int(r.headers["Content-Range"].split("/")[1])
+
+        checksum = r.headers.get("ETag") or r.headers.get("Content-MD5")
```
martindurant, line 745: Do you ever have both? 
Since we are now returning a set of information, maybe can consider what other likely keys are of interest (without assuming their format or otherwise processing them).

fsspec/implementations/http.py
```diff
@@ -666,29 +722,38 @@ async def _file_size(url, session=None, size_policy="head", **kwargs):
...
-            elif "Content-Range" in r.headers:
-                return int(r.headers["Content-Range"].split("/")[1])
-        except aiohttp.ClientResponseError:
-            logger.debug("Error retrieving file size")
-            return None
-        r.close()
+async def _file_size(url, session=None, *args, **kwargs):
+    if session is None:
+        session = await get_client()
+    info = _file_info(url, session=session, *args, **kwargs)
```
martindurant, line 755: `await` ?

docs/source/changelog.rst
martindurant: Do we actually check the value anywhere?

martindurant (commented):
I like what's here, but have some comments.

docs/source/changelog.rst
isidentical: > Do we actually check the value anywhere?

No, we don't. This is primarily useful for end user consuption (e.g in our use case, we use this to ensure the remote file is not changed).

fsspec/implementations/http.py
isidentical: Yes, I think so. (will add a test, if I can easily implement this)

fsspec/implementations/http.py
isidentical: > Do you ever have both?

You can technically get both of them, since these are basically headers.

> Since we are now returning a set of information, maybe can consider what other likely keys are of interest (without assuming their format or otherwise processing them).

This is what we have been using so far for DVC's identification, but can check other keys as well, sure.

fsspec/implementations/http.py
isidentical: 🤦🏻‍♀️ my bad!

fsspec/implementations/http.py
isidentical: Humm, not sure. This seemed like the easiest way to predict the general length, though there might certainly some catches about it. Do you have other suggestions?

isidentical:
Thanks for the review @martindurant, all should be addressed now!

fsspec/implementations/http.py
martindurant: If the purpose is to report on the transfer, then I suppose it's not important - but the number of bytes written to disk won't match. I don't know if aiohttp makes available the actual number of bytes read on the stream before decompression (probably no).

martindurant:
Thanks for making HTTP r/w! Maybe the File object can be made writable too (with the limitation of being able to PUT the file in one call!). We'll see if the file size for the callback ends up being a problem.

Issue #730, opened 2021-08-22
LocalFileSystem implement rm, not _rm


chrisroat:
`AbstractFileSystem` implements `rm` and `rm_file`, which both end up calling `_rm`.  `LocalFileSystem` does not implement `_rm` (but does override `rm`), and its call to `rm_file` fails.

What is the right way forward here?  I think it's an easy change to make things consistent, but a decision needs to be made.  Should `LocalFileSystem` be implementing `_rm`, or should the `AbstractFileSystem` implementing `rm_file` to call `rm` and get rid of `_rm`?

martindurant:
LocalFileSystem should probably have an `rm_file`. The reason for the difference, is that locally, you can call `shutil.rmtree`, whereas for most/all other implementations, you need to explicitly find all the files and folders to remove first, and then remove them.
Whether the method is called `_rm` or `rm_file` in the base class is not important, I see it just as an alias (like `delete`, which also exists!). The former exists for backward compatibility, and the latter fits better with the style elsewhere.

chrisroat:
Gotcha.  I will put together a PR.

You've piqued my interest about the other implementations.  For example, do you know how the performance of a recursive delete using `gcsfs` compares to `gsutil -m rm -r` (where `-r` does recursive delete and `-m` causes parallel operations)?

PR #736, opened 2021-08-25
All local implemention for rm_file

chrisroat:


chrisroat:
Closes #730

isidentical (approved):
LGTM

fsspec/implementations/local.py
```diff
@@ -128,6 +128,9 @@ def mv_file(self, path1, path2, **kwargs):
         path2 = self._strip_protocol(path2).rstrip("/")
         shutil.move(path1, path2)
 
+    def rm_file(self, path):
+        self.rm(path)
```
martindurant, line 132: This ends up being sort of circular. I know that for this implementation, the actual removal is in `rm`, but that's not the normal pattern, so should we just use `os.remove` directly here? It'll make the right kind of exception if the path is a directory.

chrisroat: This implementation does raise the IsADirectory error if the path is a directory (it falls through to the os.remove, since recursive is false).  I don't think it's truly circular -- it's just been implemented the opposite of the other classes.

But you've made me realize an actual problem -- this implementation accepts a list when it shouldn't.

I've updated to your suggestion, which is correct.  I also added a couple more checks in the test.

PR #741, opened 2021-09-02
isdir to give false when http url does not exist

martindurant:
@tdhopper I think this was the fix you were after. Here is the s3 counterpart: https://github.com/dask/s3fs/pull/522 (already released).


tdhopper:
Thank you!

PR #744, opened 2021-09-08
start

martindurant:


martindurant:
cc @rjzamora @jrbourbeau 

The idea is to use the `max_gap` arg to decide whether to merge bytes ranges for the same file:
- if None, don't do any merging
- if 0, merge overlaps and contiguous pieces
- is >0, merge pieces where the gap is smaller than this number. Maybe should consider the maximum size of the request, but I would say no.

fsspec/caching.py
```diff
@@ -422,6 +422,30 @@ def _fetch(self, start, end):
         return self.data[start:end]
 
 
+class KnownPartsOfAFile(BaseCache):
+    name = "parts"
+
+    def __init__(self, blocksize, fetcher, size, data={}, **_):
```
rjzamora, line 428: It looks like you are assuming that the user has already read the necessary byte ranges from the file and has used them to populate the values of `data` before the file is open and this cache is initialized. Am I mistaken?

To clarify, I want to make sure that we can specify only the known byte ranges (not the actual data stored at those byte ranges) when we call `fs.open`.  I would expect the necessary byte ranges to be fetched automatically from the file (using cat_ranges) when this cache is initialized.  Does that make sense, or am I misunderstanding the API?

martindurant:
@rjzamora , as agreed. We can iterate on this yet, since it's not very prominent in the API.

rjzamora:
I'll be happy to help add the necessary logic to leverage these new features in Dask once [dask#8072](https://github.com/dask/dask/pull/8072) gets in.

PR #745, opened 2021-09-09
Add DirFileSystem

### LINKED ISSUE #395: Allow setting the root of a filesystem
It would be useful to be able to specify a filesystem with an arbitrary root - e.g.
```python
fs = LocalFileSystem(root='C:/temp')
```
All operations would then be performed relative to this root - e.g. `fs.ls('/')` would then list the contents of `C:/temp`

lucmos:
The DirFileSystem is a filesystem-wrapper. It assumes every path it is dealing with
is relative to its `path`. After performing the necessary paths operations it delegates
everything to the wrapped filesystem.

Resolves #395


martindurant:
Since this is quite substantial, it will take us some time to review. Please bear with us. @isidentical , do you or any of your team have an interest here?

@lucmos , am I right in thinking that this implementation wouldn't participate in URL unpacking, i.e., there would be no point in doing something like `fsspec.open("prefix://mypath::s3://", prefix={"root": "mybucket"})`

isidentical:
CC: @efiop

fsspec/implementations/tests/test_prefix.py
```diff
@@ -0,0 +1,135 @@
...
+def test_isdir():
+    fs = PrefixFileSystem(prefix=".", filesystem=fsspec.filesystem("file"))
+    with filetexts(files, mode="b"):
+        for f in files.keys():
+            assert fs.isdir(os.path.dirname(os.path.abspath(f)))
+            assert not fs.isdir(f)
+        assert not fs.isdir("not-a-dir")
+
+
+@pytest.mark.parametrize("prefix", ["/", "/tmp"])
```
efiop, line 121: This needs to run on windows too, so we can't make assumptions about root/tmp dirs.

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,135 @@
...
+            return protocol + "://" + path if protocol is not None else path
+        elif isinstance(path, Iterable):
+            return [self._add_fs_prefix(x) for x in path]
+        assert False
+
+    def _remove_fs_prefix(self, path: Union[str, Path]) -> Union[str, Sequence[str]]:
+        if isinstance(path, (str, Path)):
+            path = stringify_path(path)
+            protocol, path = split_protocol(path)
+            path = os.path.relpath(path, start=self.prefix)
```
efiop, line 37: Might be missing something, but I don't thin we can use `os.path` here because this could be ran on windows and because `relpath` is slow (esp on windows and esp when dealing with large number of files, e.g. in `ls()`). We should be able to just do something like `path = path[len(self.prefix]:]` instead.

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,135 @@
...
+        filesystem: fsspec.AbstractFileSystem,
+        *args,
+        **storage_options,
+    ) -> None:
+        super().__init__(*args, **storage_options)
+        self.prefix = prefix
+        self.filesystem = filesystem
+
+    def _add_fs_prefix(self, path: Union[str, Path]) -> Union[str, Sequence[str]]:
+        if isinstance(path, (str, Path)):
```
efiop, line 24: Type annotations say that `path` is str in all methods like `ls/makedirs/etc`, so we need to either remove it or propagate Union to those methods too (I suppose this is only relevant for LocalFileSystem right now, right?).

efiop:
From dvc perspective, we don't have a notion of `cwd` in our internals, and choose to operate on absolute paths, which has its problems. We might consider trying this out for localfs, but it is not going to be soon 🙁 prefixfs is cool though 🙂

lucmos:
Hello, thanks for the replies and comments!

@martindurant yes, the idea is that `PrefixFileSystem` would only wrap another filesystem already instantiated, without any additional URL unpacking, and modify all paths to be relative to its own prefix. 
I think in the future it may make sense to consider a built-in wrapping, e.g. `fs = fsspec.filesystem('file', prefix='myroot')` or  `fsspec.open("s3://my-path-relative-to-mybucket", prefix="mybucket")` would internally wrap the file system instantiated with `PrefixFileSystem` , but I think it is out of scope for this MR

@efiop Thanks for the review! I will work on it as soon as possible

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,135 @@
...
+            path = stringify_path(path)
+            protocol, path = split_protocol(path)
+            path = os.path.join(self.prefix, path)
```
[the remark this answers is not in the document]
lucmos: I see, there are some special cases that we need to handle though (that `os.path.join` or `relpath` does for us), e.g.

- If `prefix==sep` we must not add the prefix, otherwise we will have a double separator
- if `prefix` is empty we must not add the sep
- if the path is already absolute, we should not modify it. i.e. `os.path.join('c/d', '/a/b')` returns `/a/b`.  ..or should we assume all paths are relative to the prefix regardless if they are absolute or not?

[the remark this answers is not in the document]
dhirschfeld: > *should we assume all paths are relative to the prefix*

That would be my expectation - i.e. if I specified `/` with a prefix of `/tmp`, I would expect that to reference `/tmp`. Similarly, I would expect `/a/b` to reference `/tmp/a/b`

fsspec/implementations/tests/test_prefix.py
lucmos: Ok I refactored this test.

In general, I would like to re-use the tests already implemented in other file systems, wrapping it with the PrefixFileSystem. Do you think it is feasible? 
Or maybe it is better to just wait for https://github.com/intake/filesystem_spec/pull/651

fsspec/implementations/prefix.py
lucmos: I changed from `relpath` to your suggestion!

fsspec/implementations/prefix.py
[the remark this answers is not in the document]
lucmos: Ok I implemented this behavior, the test `test_directories` now tests also for this

fsspec/implementations/prefix.py
lucmos: I removed the `Union` annotation!

lucmos:
I would like to increase the number of tests by reusing the tests written for other file system, and wrapping that filesystem in this `PrefixFileSystem`. 
Do you have any pointer on how to do that? 

I tried to do this with the `LocalFileSystem` but it is not straightforward to re-parametrize everything to work with both `LocalFileSystem` and `PrefixFileSystem`

fsspec/implementations/prefix.py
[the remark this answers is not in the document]
martindurant: I think prefix=empty and prefix=sep should raise an exception, or at least, a user has no right to assume they do anything useful :)

martindurant:
> reusing the tests written for other file system, and wrapping that filesystem in this PrefixFileSystem.
Do you have any pointer on how to do that?

Not really! Again, the idea in #651 might give a framework that makes such a thing easier, but it's still not super-clear.

martindurant:
(I don't think #651 is making any progress, though)

fsspec/implementations/prefix.py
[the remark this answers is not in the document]
lucmos: At the moment if `prefix=empty` it defaults to `prefix=sep`. If `prefix=sep` the PrefixFileSystem should be a passthrough that does not change any path (if everything works).

fsspec/implementations/tests/test_prefix.py
efiop: Reusing sounds great. IMPO, there is no need to test prefixfs with every single fs implementation, as that would be excessive. Since all filesystems follow the same interface and prefixfs is meant to be generic and doesn't make any fs-specific assumptions, your approach with only testing with localfs looks good to me.

fsspec/implementations/prefix.py
efiop: But we still have the check here and I suppose this is all meant to work with localfs, which accepts Path-like objects too. Looks like there type annotations  are not yet ready (https://github.com/intake/filesystem_spec/issues?q=is%3Aissue+is%3Aopen+type+annotations), so maybe not worth adding them at all for now? 🤷 CC @martindurant

efiop:
@lucmos Btw, what's the plan with `AbstractFile` (maybe also something else that I'm missing?) that has `path` "public" property, would we need to wrap that as well somehow?

lucmos:
@efiop not sure I got what you are referring to. Do you have a link to `AbstractFile`?

fsspec/implementations/tests/test_prefix.py
lucmos: Ok makes sense! 

I fear that also the tests that use the localfs are not enough though, it would be nice to reuse the entire `LocalFileSystem` test suite for this `PrefixFileSystem`, because re-implementing them means duplicating a lot of code. But probably it is not possible right now...

martindurant:
That would be fsspec.spec.AbstractBufferedFile , used by several backends which don't have their own native file-like object. The instances have an `fs` attribute holding the parent filesystem, so it makes sense if their `.path` is in the terms of that filesystem. Except that we are probably making this file-like object via the wrapped filesystem's `.open` method, which won' know that the filesystem has been wrapped.

However, where there *is* a native file of file wrapper (like LocalFileOpener),  probably that instance contains details of the path in terms of the wrapped filesystem and we can do nothing about it. If it's not in a `.path` attribute, it might not matter.

lucmos:
I see, thanks for the explanation! 

Would something like this be enough?

https://github.com/intake/filesystem_spec/blob/c7f763f3810abbb84707b602ba37435ca6afa559/fsspec/implementations/prefix.py#L158-L163

martindurant:
That would work, in the sense that you would be able to open the right file, but its `.path` attribute would be the full one, and `.fs` attribute, if present, would be the wrapped filesystem, not the prefix filesystem. Maybe that's OK (but the docstring for `open` should say this clearly).

lucmos:
I see, I think it may be OK to do this and document in the docstrings. 
What do you think?

martindurant:
> do this and document in the docstrings

That's probably the easiest thing to do, I don't have a problem with it.

lucmos:
Sorry for the delay, I updated the `open` docstring

fsspec/spec.py
```diff
@@ -944,6 +944,10 @@ def open(self, path, mode="rb", block_size=None, cache_options=None, **kwargs):
         """
         Return a file-like object from the filesystem
 
+        The file-like object returned ignores an eventual PrefixFileSystem:
```
martindurant, line 947: I wouldn't put this here, the vast majority of people calling this won't be using Prefix. I would put it into the docstring of prefix's open()

martindurant:
Please add your implementation to the API docs.

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,163 @@
...
+        prefix: str,
+        filesystem: fsspec.AbstractFileSystem,
+        *args,
+        **storage_options,
+    ) -> None:
+        super().__init__(*args, **storage_options)
+        self.prefix = stringify_path(prefix)
+
+        if not self.prefix:
+            self.prefix = self.sep
```
efiop, line 28: Do I understand correctly that `sep` is used here as a root indicator?

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.filesystem = filesystem
```
efiop, line 30: How about
```suggestion
        self.fs = fs
```
same in the args, just makes it much easier to read in the rest of the code. Also `filesystem` is often used in this codebase for `fsspec.filesystem` helper, so will make grepping easier too.

fsspec/implementations/prefix.py
efiop: @lucmos Your type annotation says `str`, but you check for `Path` too 🙁

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,163 @@
...
+            return [self._remove_fs_prefix(x) for x in path]
+        assert False
+
+    def mkdir(self, path: str, create_parents: bool = True, **kwargs) -> None:
+        path = self._add_fs_prefix(path)
+        return self.filesystem.mkdir(path=path, create_parents=create_parents, **kwargs)
+
+    def makedirs(self, path: str, exist_ok: bool = False):
+        path = self._add_fs_prefix(path)
+        return self.filesystem.mkdirs(path=path, exist_ok=exist_ok)
```
efiop, line 70:
```suggestion
        return self.filesystem.makedirs(path=path, exist_ok=exist_ok)
```
for symmetry, since we are overriding `makedirs`

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,163 @@
...
+        path = self._add_fs_prefix(path)
+        return self.filesystem.created(path)
+
+    def modified(self, path: str):
+        path = self._add_fs_prefix(path)
+        return self.filesystem.modified(path)
+
+    def sign(self, path: str, expiration=100, **kwargs):
+        path = self._add_fs_prefix(path)
+        return self.filesystem.sign(path, expiration=100, **kwargs)
```
efiop, line 141:
```suggestion
        return self.filesystem.sign(path, expiration=expiration, **kwargs)
```

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__qualname__}(prefix='{self.prefix}', filesystem={self.filesystem})"
+
+    def open(
+        self,
+        path,
+        **kwargs,
+    ):
+        return self.filesystem.open(self._add_fs_prefix(path), **kwargs)
```
efiop, line 163: Doesn't match the signature of the parent class. Probably not intended.

fsspec/spec.py
```diff
@@ -1455,7 +1459,7 @@ def _upload_chunk(self, final=False):
         # may not yet have been initialized, may need to call _initialize_upload
 
     def _initiate_upload(self):
-        """ Create remote file/upload """
```
efiop, line 1458: Not clear why some of the docstrings got re-formatted, but others didn't. Probably some pre-commit hooks your your editor? Just looks like unrelated uncomplete changes, better to avoid.

fsspec/implementations/prefix.py
lucmos: I completely overlooked `root_marker`, now I am doing:

```python
        if not self.prefix:
            self.prefix = self.root_marker
```

Does it work for you?

fsspec/implementations/prefix.py
lucmos: Ok I completely removed the `Path` and any `pathlib` import!

lucmos:
Thanks for the review! :smile:

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,196 @@
...
+        pass
+
+
+class PrefixFileSystem(AbstractFileSystem):
+    """A meta-filesystem to add a prefix and delegate to another filesystem"""
+
+    def __init__(
+        self,
+        prefix: str,
+        fs: fsspec.AbstractFileSystem,
```
efiop, line 21:
```suggestion
        fs: AbstractFileSystem,
```

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,196 @@
...
+
+        fs: fsspec.AbstractFileSystem
+            An instantiated filesystem to wrap. All operations are delegated to
+            this filesystem after appending the specified prefix
+        """
+        super().__init__(*args, **storage_options)
+        self.prefix = stringify_path(prefix)
+
+        if not self.prefix:
+            self.prefix = self.root_marker
```
efiop, line 39: Continuing from https://github.com/intake/filesystem_spec/pull/745#discussion_r722297683

Is there a use case for this? I'm not sure this works right now, and doesn't seem like there are any tests for it. So maybe easier to just error out?

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,196 @@
...
+        return path
+
+    def _add_fs_prefix(self, path: str) -> Union[str, Sequence[str]]:
+        if isinstance(path, str):
+            path = stringify_path(path)
+            protocol, path = split_protocol(path)
+
+            path = self._get_relative_path(path)
+
+            if self.prefix == self.sep:
```
efiop, line 55: I guess this is again used in `root` sense, right?

fsspec/implementations/prefix.py
lucmos: The idea was to give the possibility to easily let the `PrefixFileSystem` be a passthrough that doesn't change any path

fsspec/implementations/prefix.py
lucmos: Mmmh yes, but the distinction between `root_marker` and `sep` is a bit confusing

So I guess I should use `root_marker` also here?

https://github.com/intake/filesystem_spec/blob/2d3179b401b019f6d9cb28be018a34f6e9a7e28c/fsspec/implementations/prefix.py#L43-L46

martindurant: Following this conversation, I do tend to go back to my original argument, that a prefixfilesystem with no prefix should be an exception, and then you don't have to deal with this ambiguity.

lucmos: Ok I (think) I fixed several issues with the prefix/sep/root_marker

- If there is not prefix I raise a `ValueError`
- Any trailing `sep` on prefix is removed
- If prefix is equal to the `root_marker` add only the root marker once
- Use the `sep` and `root_marker` of the wrapped fs

martindurant:
We want paths relative to the filesystem root, and there's no particular reason we should support "." at all, since most filesystems don't have a concept of current directory.

efiop (dismissed):
Other than a few minor comments above and a failing test, looks good to me 👍 Thanks @lucmos ! 🙏

benlindsay:
Hi all, now that @martindurant pointed me to this PR, I'm super excited about it. This will be a really nice feature to add. Being able to have a filesystem limited to a bucket or subdirectory within that bucket will be great for limiting the scope of potential errors and reducing the amount of information needed to interact with a filesystem. Keep up the good work!

lucmos:
Thank you for your appreciation!

Unfortunately in this period, I don't have any bandwidth to finish this PR. If someone is willing to take over feel free to do so!

martindurant:
@efiop , do you happen to have time and motivation to finish this off?

efiop:
@martindurant Sure, will take a look during weekend. I think we even have a use case for this for the near future.

fsspec/implementations/tests/test_prefix.py
```diff
@@ -0,0 +1,238 @@
...
+        with OpenFile(fs, fn, mode="rb"):
+            pass
+
+
+def test_isfile():
+    fs = PrefixFileSystem(prefix=".", fs=fsspec.filesystem("file"))
+    with filetexts(files, mode="b"):
+        for f in files.keys():
+            assert fs.isfile(f)
+            assert fs.isfile("file://" + f)
```
efiop, line 141: Hm, I don't think PrefixFileSystem should be aware of full urls for the filesystem it is wrapping. E.g. `s3://one/two` will mean very different things depending on whether or not `one` is a bucket or if it is a dir within prefixfs wrapper for s3, so to handle that the same way we handle `file://` here, we would have to know how `s3fs` works inside. What we might want though is something name-addresable, like `fs://myfs/path`, which would understand that `myfs` is a PrefixFileSystem instance and would pass `path` to it, which would then be further resolved.

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def cp_file(self, path1, path2, **kwargs):
+        path1 = self._add_fs_prefix(path1)
+        path2 = self._add_fs_prefix(path2)
+        return self.fs.cp_file(path1, path2, **kwargs)
+
+    def get_file(self, path1, path2, callback=None, **kwargs):
+        path1 = self._add_fs_prefix(path1)
+        path2 = self._add_fs_prefix(path2)
+        return self.fs.get_file(path1, path2, callback, **kwargs)
```
efiop, line 92:
```suggestion
        return self.fs.get_file(path1, path2, callback=callback, **kwargs)
```

efiop:
@martindurant I've adjusted the implementation and some tests to make as few assumptions about the filesystem as possible. I think this could definitely be adjusted further, but we can go with this as a start, and adjust when people start using it (us included, have a use-case already, will be playing with it this week). I've added a comment to mark this implementation as experimental and incomplete, since all used methods need to be wrapped one-by-one.

martindurant:
I see "requested changes" in addition to the commits to this branch - is the request still relevant?

efiop:
@martindurant Nope, I think we are good for now and ready.

fsspec/implementations/prefix.py
```diff
@@ -0,0 +1,148 @@
...
+    def mv_file(self, path1, path2, **kwargs):
+        path1 = self._add_fs_prefix(path1)
+        path2 = self._add_fs_prefix(path2)
+        return self.fs.mv_file(path1, path2, **kwargs)
+
+    def rm_file(self, path):
+        path = self._add_fs_prefix(path)
+        return self.fs.rm_file(path)
+
+    def rm(self, path, recursive=False, maxdepth=None):
```
martindurant, line 108: I think this and methods following should work based on the superclass implementation

fsspec/implementations/tests/test_prefix.py
```diff
@@ -0,0 +1,198 @@
+from __future__ import absolute_import, division, print_function
```
martindurant, line 1: don't need this

fsspec/implementations/tests/test_prefix.py
```diff
@@ -0,0 +1,198 @@
...
+from pathlib import Path
+
+import pytest
+
+import fsspec
+from fsspec.core import OpenFile
+from fsspec.implementations.local import make_path_posix
+from fsspec.implementations.prefix import PrefixFileSystem
+
+files = {
```
martindurant, line 16: This and the contexts are copied from somewhere?

martindurant: Seem more like a fixture than a context

martindurant (commented):
I still had a few comments. 

We can still merge as-is, if that makes life easier for you, @efiop (feel free to press the button)

fsspec/implementations/tests/test_prefix.py
```diff
@@ -0,0 +1,198 @@
+from __future__ import absolute_import, division, print_function
+
```
efiop, line 2:
```suggestion
```

fsspec/implementations/tests/test_prefix.py
efiop: Yep, this was copied before from `test_local`. Didn't touch it much, as I believe we need to rework these tests and introduce common fsspec fixtures. At least it is isolated from `test_local` for now.

For prefixfs, some unit tests that would make sure that the path is properly joined and properly passed to the underlying filesystem should be enough (plus some for transforming stuff back, e.g. `ls`). So I plan on introducing those along with support for new methods in the follow-up.

fsspec/implementations/prefix.py
efiop: I agree, just wanted to avoid changing this too much right now. But might as well, alongside with `cat` too. Will do.

martindurant: Something I'm not sure about here: are we taking any care of sync/async? For example, [here](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/reference.py#L213) we have both `_cat_file` and `cat_file` for the two cases.

martindurant: And [here](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/reference.py#L246) we decide which to call. Maybe this pattern can be reused. In fact, there is a lot in common between prefix and reference.

martindurant: Sorry :|

efiop:
For the record: getting distracted by other stuff, but this is still on my radar, will get back to this tomorrow.

efiop:
prefixfs also brings up an interesting point about path normalization. Most of the implementations call `_strip_protocol` in every public method, which means that, for example, paths returned from `ls` will also be normalized, which means that we need to call `fs._strip_protocol` on the prefix as well, to be able to properly remove them. And this is fine for most filesystems, but breaks with odd filesystems like localfs with windows paths. For example:

Say `prefix = C:\my\path`, then prefixfs needs to be able to handle both `rel/path` and `rel\\path`. To do that, ideally, it needs to normalize both the prefix and the relpath with `_strip_protocol`, but relative path might not be a valid fspath for that particular filesystem, so this might be fragile. So I think that the original assumption that both `rel/path` and `rel\\path` are valid paths for prefixfs is wrong, and we should instead say that `prefixfs.sep` is `/` and thus only `rel/path` is a valid path for it. Otherwise, we will be opening a giant can of worms that will pretty much require implementing `cwd` support in every filesystem so that it could have some opinion about handling its relpaths.

I can totally see this `rel/path` vs `rel\\path` thing being a gotcha for some users of prefixfs. It seems like the name itself is a bit confusing, as what we are really implementing here is closer to `chdir` or maybe `chroot` rather than `prefix`(e.g. someone might want to use `s3://company-` as a prefix to get paths like `s3://company-project/path`). Maybe `Chdir/Chroot/DirFileSystem` would be more appropriate here? @lucmos @martindurant WDYT?

Plus (unrelated) localfs says that root_marker is `/`, but it is wrong for windows paths as `_strip_protocol` still produces paths like `C:/some/path`.

These localfs ntpath inconsistencies (unless I'm missing something) are one of the reasons why `localfs` is the last filesystem that we use that has not yet been migrated to fsspec. We will be getting to it soon though.

martindurant:
> Plus (unrelated) localfs says that root_marker is /, but it is wrong for windows paths as _strip_protocol still produces paths like C:/some/path.

The root marker is only used when the user passes empty paths, it doesn't need to be the start of other paths - so I think this is OK.

I tend to agree with the rest of what you say, but I'll think a little. I'm OK to rename this implementation to something clearer. Elsewhere ( #843 I think) there are requests for a more general "mount", which I think pyfisystems2 supports. I don't know whether that's a good idea.

dhirschfeld:
On the pathsep front, can't you just normalise all paths to use `/` which also works on Windows AFAIK

dhirschfeld:
As one of the people requesting this feature, all I'm looking for is for prefix to be a valid directory on the filesystem so I'm fine with disallowing arbitrary strings which have to be joined together to make a valid path.

lucmos:
@efiop I agree that chroot-like support is enough. Actually, this is what I had in mind all along; in the first (problematic) revisions with `Path` I was using `relative_to` to achieve this

martindurant:
OK, so lets keep this simple and push it in so people can start using it and comment back

benlindsay:
Agreed that arbitrary string prefixes are probably not necessary, just subdirs. I'd support a name change for clarity, like SubdirFileSystem or RelativePathFileSystem or ChrootFileSystem or some such.

efiop:
Thanks for the feedback. I'll get back to this this week.

efiop:
For the record: I've reworked prefixfs into dirfs with sync, async and external async support, tested with some tests with those, but decided to remove existing tests for now, as they will likely be unmaintainable longterm.

I will give dirfs a try in some more of our projects this week and will try to get to a nicer testing suit for it.

martindurant:
Thanks for working on this, @efiop

fsspec/implementations/dirfs.py
```diff
@@ -0,0 +1,311 @@
...
+        fs: AbstractFileSystem
+            An instantiated filesystem to wrap.
+        """
+        super().__init__(*args, **storage_options)
+
+        if self.asynchronous and not fs.async_impl:
+            raise ValueError("can't use asynchronous with non-async fs")
+
+        if self.asynchronous != fs.asynchronous:
+            raise ValueError("both dirfs and fs should be in the same sync/async mode")
```
efiop, line 20: Decided not to try to wrap asynchronous fs with sync like reffs does, because it makes little sense in dirfs use case, as there is just one fs that we are dealing with and user expects to use dirfs in the same mode as fs it is wrapping. Though that also means that we could just mimic `fs.asynchronous` instead of passing it explicitly, but I kept it explicit for now.

Also, it is clear that some of these attributes (e.g. asynchronous/batch_size/etc) might be expected to be mirrorred by the dirfs, but they are obviously not right now. We can add them when someone needs them.

martindurant: Agreed to keep it simple until we see this being used and issues naturally arise. Since we are not aiming for generalised "mount" here, I suspect issues will be few, and this will mainly be used locally.

efiop:
@lucmos @benlindsay Hey guys! It would be great if you could give the current implementation a shot, just to make sure it works for you as expected.

efiop:
Ok, I've added unit tests for all methods for sync/async_impl/asynchronous filesystems, to keep it lightweight and maintainable. Ideally, some func tests would be really nice too, but we don't yet have nice fixtures in fsspec (we have some in dvc, but didn't get to contributing them yet) to reuse different filesystems in a unified way (e.g. fixtures that will automatically generate temp paths in every filesystem, etc), so leaving it for later.

I'm not happy with some `*args, **kwargs` in methods, because they try to preserve positional args, but sync/async specs lack `*` arg guards in them, so we'll likely get back to that later. There are also some questionable transformations or lack thereof in results of some methods (e.g. `name` is transformed in `ls` because it is useful and is not transformed in `info`, as well as `destination`).

PR #749, opened 2021-09-10
hdfs: general enhancements

isidentical:
Individual commits describes the changes. But can provide more detail / explanation if asked.

### DIFF
--- a/fsspec/implementations/hdfs.py
+++ b/fsspec/implementations/hdfs.py
@@ -1,5 +1,8 @@
+import os
+import re
+import subprocess
 import weakref
-from functools import partialmethod
+from functools import partial
 
 from pyarrow.hdfs import HadoopFileSystem
 
@@ -8,16 +11,19 @@ from fsspec.utils import infer_storage_options
 
 
 def inherits(*methods):
-    def client_method(self, method, *args, **kwargs):
-        return getattr(self.client, method)(*args, **kwargs)
+    def client_getter(name, self):
+        return getattr(self.client, name)
 
     def wrapper(cls):
         for method in methods:
-            wrapped_method = partialmethod(client_method, method)
-            setattr(cls, method, wrapped_method)
+            wrapped_method = partial(client_getter, method)
+            setattr(cls, method, property(wrapped_method))
         return cls
 
     return wrapper
+
+
+CHECKSUM_REGEX = re.compile(r".*\t.*\t(?P<checksum>.*)")
 
 
 @inherits(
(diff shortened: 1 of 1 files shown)


fsspec/implementations/hdfs.py
```diff
@@ -95,94 +241,52 @@ def _open(
             **kwargs,
         )
 
-    def __reduce_ex__(self, protocol):
-        return PyArrowHDFS, self.pars
+    def checksum(self, path, env=None, **kwargs):
```
efiop, line 244: For fsspec's purposes hash of `info` or something like that will do just fine. Let's not introduce CLI checksum here, we will stop using it ourselves in the future as well anyway, so it doesn't make sense to adopt such a complication.

fsspec/implementations/hdfs.py
```diff
@@ -1,17 +1,93 @@
...
+    "df",
+    "chmod",
+    "chown",
+    "disk_usage",
+    "download",
+    "upload",
+    "read_parquet",
+    "rm",
+    "stat",
+    "upload",
```
efiop, line 84: Duplicate. Looks like it was here before.
```suggestion
```

fsspec/implementations/hdfs.py
```diff
@@ -1,17 +1,93 @@
...
+from functools import partial, wraps
 
 from pyarrow.hdfs import HadoopFileSystem
 
-from ..spec import AbstractFileSystem
-from ..utils import infer_storage_options
+from fsspec.spec import AbstractFileSystem
+from fsspec.utils import infer_storage_options
 
+_NOT_FOUND = os.strerror(errno.ENOENT)
```
efiop, line 16: os.strerror is fast enough to just inline this.

fsspec/implementations/hdfs.py
```diff
@@ -1,17 +1,93 @@
...
 
+
+def wrap_exceptions(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except OSError as exception:
```
efiop, line 24: actually, this might no longer be needed, https://issues.apache.org/jira/browse/ARROW-6248 was fixes in april and looks like the fix was included in the already released versions.

jorisvandenbossche:
Note that `pyarrow.hdfs.HadoopFileSystem` is deprecated (in favor of `pyarrow.fs.HadoopFileSystem`, which has a different API) and will be removed in a future pyarrow version (probably early next year). See https://github.com/intake/filesystem_spec/issues/663. 
(this of course doesn't mean that the current implemenation can't be further maintained/improved, as this PR does, to be clear! I just want to make people aware of the situation)

efiop:
@jorisvandenbossche Thanks for the heads up! We are actually already using pyarrow.fs.HadoopFileSystem in our own implementation in dvc, and will look into migrating fsspec as well, as our end goal is to migrate to fsspec's hdfs fs anyway :slightly_smiling_face:

isidentical:
Thanks @jorisvandenbossche, noticed it when writing tests and seeing the deprecation warnings pop up. I am currently working on preparing a patch (based on your branch referenced in the same issue) to support `pyarrow.fs` like interfaces (including Local and Hadoop). I would assume it is going to be a separate PR than this, since this might be still useful but I have no preferences (it is a on a separate branch anyway).

fsspec/implementations/hdfs.py
```diff
@@ -42,21 +118,91 @@ def __init__(
...
+
+    @wrap_exceptions
+    def cp_file(self, lpath, rpath, **kwargs):
+        with self.open(lpath) as lstream:
+            tmp_fname = "/".join([self._parent(rpath), f".tmp.{secrets.token_hex(16)}"])
+            try:
+                with self.open(tmp_fname, "wb") as rstream:
+                    shutil.copyfileobj(lstream, rstream)
+                self.client.mv(tmp_fname, rpath)
+            except BaseException:  # noqa
```
martindurant, line 189: This seems odd to me. I would at least document the process in the method docstring.

fsspec/implementations/hdfs.py
martindurant: agree

fsspec/implementations/hdfs.py
```diff
@@ -95,94 +241,52 @@ def _open(
...
 
-    def close(self):
-        self.pahdfs.close()
+        return match.group("checksum")
 
-    @classmethod
-    def _strip_protocol(cls, path):
-        ops = infer_storage_options(path)
-        return ops["path"]
+    def _run_command(self, cmd, env=None):
```
martindurant, line 258: If this is here only to support checksum, I recommend removing it

fsspec/implementations/hdfs.py
```diff
@@ -95,94 +241,52 @@ def _open(
...
+        out, err = p.communicate()
+
+        if p.returncode != 0:
+            raise subprocess.CalledProcessError(p.returncode, cmd, out, err)
         else:
-            # attributes of the superclass, while target is being set up
-            return super().__getattribute__(item)
+            return out.decode("utf-8")
+
+    def _as_url(self, path):
```
martindurant, line 285: I'm not sure we need this anymore, but it may be worth considering having something like this in AbstractFileSystem (c.f., [`_unstrip_protocol`](https://github.com/intake/filesystem_spec/pull/744/files#diff-3f0d077f95c8b71995c7d1c0815d95d64aaa5c516411caa8a91b6f783867d86aR452) , which doesn't include any netloc things, and doesn't know about chained FSs).

martindurant (commented):
No serious issues here, just a couple of comments.
Given the introduction of the arrow FS, do you think we should start referring to this one as "hdfs_legacy" or something (and maintain the original "hdfs" too, for now).

isidentical:
> Given the introduction of the arrow FS, do you think we should start referring to this one as "hdfs_legacy" or something (and maintain the original "hdfs" too, for now).

Let's defer this after the merge of both of my PRs.

Issue #759, opened 2021-09-17
`SFTPFileSystem.ls()` in root directory hides names


matteosantama:
Consider the following methods

```python
@staticmethod
def _decode_stat(stat, parent_path=None):
    if S_ISDIR(stat.st_mode):
        t = "directory"
    elif S_ISLNK(stat.st_mode):
        t = "link"
    else:
        t = "file"
    out = {
        "name": "",
        "size": stat.st_size,
        "type": t,
        "uid": stat.st_uid,
        "gid": stat.st_gid,
        "time": datetime.datetime.utcfromtimestamp(stat.st_atime),
        "mtime": datetime.datetime.utcfromtimestamp(stat.st_mtime),
    }
    if parent_path:
        out["name"] = "/".join([parent_path.rstrip("/"), stat.filename])
    return out

def ls(self, path, detail=False):
    logger.debug("Listing folder %s" % path)
    stats = [self._decode_stat(stat, path) for stat in self.ftp.listdir_iter(path)]
    if detail:
        return stats
    else:
        paths = [stat["name"] for stat in stats]
        return sorted(paths)
```

And now suppose we want to run `fs.ls("")`, where `""` is the root marker for `SFTPFileSystem`.  The crucial part is here, in `_decode_stat`. Note that we are calling `self._decode_stat(state, parent_path="")`

```python
out = {
    "name": "",
    "size": stat.st_size,
    "type": t,
    "uid": stat.st_uid,
    "gid": stat.st_gid,
    "time": datetime.datetime.utcfromtimestamp(stat.st_atime),
    "mtime": datetime.datetime.utcfromtimestamp(stat.st_mtime),
}
if parent_path:  # <-- This evaluates to false when parent_path=""
    out["name"] = "/".join([parent_path.rstrip("/"), stat.filename])
```

The result is that all the items in the root directory come back as `["", "", "". ...]`. The condition should likely be `if parent_path is not None:`

martindurant:
Do you think you can formulate this as a test and show that `if not None` get's the right result, but the current code does not?

On September 17, 2021 10:23:10 AM EDT, Matteo Santamaria ***@***.***> wrote:
>Consider the following methods
>
>```python
***@***.***
>def _decode_stat(stat, parent_path=None):
>    if S_ISDIR(stat.st_mode):
>        t = "directory"
>    elif S_ISLNK(stat.st_mode):
>        t = "link"
>    else:
>        t = "file"
>    out = {
>        "name": "",
[...]

PR #760, opened 2021-09-21
fix sftp ls() method from root

matteosantama:
See comments in #759

### DIFF
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -108,6 +108,8 @@ class SFTPFileSystem(AbstractFileSystem):
         }
         if parent_path:
             out["name"] = "/".join([parent_path.rstrip("/"), stat.filename])
+        else:
+            out["name"] = stat.filename
         return out
 
     def ls(self, path, detail=False):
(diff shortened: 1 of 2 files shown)


matteosantama:
Tests are failing because with this sftp server, paramiko does not let you create paths that start with a `/`. Given that `SFTPFileSystem.root_marker == ""`, this seems reasonable to me.

martindurant:
So what do you suppose we should do?

martindurant:
Yeah, sorry, still a bit confused here. The test you add to show the problem doesn't actually run, and other things are failing which don't seem to have changed.

Weren't you thinking of changing the condition mentioned in #759 to `is not None`?

matteosantama:
There are a few things to resolve. First, the test I added is failing with `fixture 'sftp_server' not found`. How can I add the `mocksftp` to the testing environment? 

Here's the problem, when I simply make it `is not None:`, I get 

```python
fs.touch("a.txt")
assert fs.ls(fs.root_marker) == ["/a.txt"]
```
to pass. However, if I instead run `fs.touch("/a.txt")`, I get a `PermissionError`. `paramiko` does not like the leading slash. 

Now to be honest, I'm not sure if this is a just quirk of my mock SFTP server. This same problem is what's breaking the old `sftp` tests. So the question is, should  SFTP paths start with a leading slash, or not?

martindurant:
> So the question is, should SFTP paths start with a leading slash, or not?

I don't think it matters too much, so long as we are consistent. I am happy to go with what paramiko does, apparently not allowing a leading "/". Of course, the paths the user can pass and the paths reported back by the implementation should match too.

PR #762, opened 2021-09-21
webhdfs: implement cp_file/rm_file

isidentical:
Implements `cp_file` (copying through the user) and `rm_file`. Also removes a logging call, since it might contain non-string arguments.


fsspec/implementations/webhdfs.py
```diff
@@ -122,7 +125,6 @@ def _call(self, op, method="get", path=None, data=None, redirect=True, **kwargs)
         args = kwargs.copy()
         args.update(self.pars)
         args["op"] = op.upper()
-        logger.debug(url, method, args)
```
martindurant, line 125: Perhaps still log the url and method?

isidentical:
I think it makes sense, creating https://github.com/intake/filesystem_spec/issues/763. Btw @martindurant if could you make a new release with all new hdfs enhancements, it would be really cool.

martindurant:
>  if could you make a new release

On my list!

PR #764, opened 2021-09-21
http: allow put_file()ing file-like objects

isidentical:
Since writing to `HTTPFile` is not supported, currently you can not upload arbitrary streams (unlike other filesystems where you could open a new file in writing mode through `fs.open(path, 'wb')` and copy the original stream through `shutil.copy_file_obj()` to the new file). Not sure if this would be worth documenting at the moment, but would be extremely useful to compensate for the missing `w` method at the moment while we are thinking about other solutions.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -255,8 +255,8 @@ class HTTPFileSystem(AsyncFileSystem):
 
     async def _put_file(
         self,
-        rpath,
         lpath,
+        rpath,
         chunk_size=5 * 2 ** 20,
         callback=_DEFAULT_CALLBACK,
         method="post",
@@ -265,11 +265,11 @@ class HTTPFileSystem(AsyncFileSystem):
         async def gen_chunks():
             # Support passing arbitrary file-like objects
             # and use them instead of streams.
-            if isinstance(rpath, io.IOBase):
-                context = nullcontext(rpath)
+            if isinstance(lpath, io.IOBase):
+                context = nullcontext(lpath)
                 use_seek = False  # might not support seeking
             else:
-                context = open(rpath, "rb")
+                context = open(lpath, "rb")
                 use_seek = True
 
             with context as f:
@@ -294,8 +294,8 @@ class HTTPFileSystem(AsyncFileSystem):
             )
 
         meth = getattr(session, method)
-        async with meth(lpath, data=gen_chunks(), **kw) as resp:
-            self._raise_not_found_for_status(resp, lpath)
+        async with meth(rpath, data=gen_chunks(), **kw) as resp:
+            self._raise_not_found_for_status(resp, rpath)
 
     async def _exists(self, path, **kwargs):
         kw = self.kwargs.copy()
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -276,6 +276,8 @@ class HTTPFileSystem(AsyncFileSystem):
                 if use_seek:
                     callback.set_size(f.seek(0, 2))
                     f.seek(0)
+                else:
+                    callback.set_size(getattr(f, "size"))
 
                 chunk = f.read(64 * 1024)
                 while chunk:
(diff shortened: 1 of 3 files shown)


isidentical:
I agree. It might require a bigger design discussion, but something I wondered when I started to the migration to the fsspec was why this was not supported natively (since that would potentially render the requirement to the callbacks redundant, since we could simply wrap `read()` or `write()` methods of the original stream with the callback updates).

fsspec/implementations/http.py
```diff
@@ -262,9 +263,19 @@ async def _put_file(
         **kwargs,
     ):
         async def gen_chunks():
-            with open(rpath, "rb") as f:
-                callback.set_size(f.seek(0, 2))
-                f.seek(0)
+            # Support passing arbitrary file-like objects
+            # and use them instead of streams.
+            if isinstance(rpath, io.IOBase):
+                context = nullcontext(rpath)
```
martindurant, line 269: Although you have it the right way around, the signature of this method is wrong: rpath should be the remote one, and lpath the local one. We could fix this now.

fsspec/implementations/http.py
```diff
@@ -262,9 +263,19 @@ async def _put_file(
...
+                use_seek = False  # might not support seeking
+            else:
+                context = open(rpath, "rb")
+                use_seek = True
```
martindurant, line 273: The file instance might happen to have an `.size` attribute - many fsspec implementations do use this.

fsspec/implementations/tests/test_http.py
```diff
@@ -465,6 +465,11 @@ def test_put_file(server, tmp_path, method, reset_files):
     fs.get_file(server + "/hey", dwl_file)
     assert dwl_file.read_bytes() == data
 
+    src_file.write_bytes(b"xxx")
+    with open(src_file, "rb") as stream:
+        fs.put_file(stream, server + "/hey_2", method=method)
```
martindurant, line 470: Should also test with a non-standard file, e.g., io.BytesIO or MemoryFile

fsspec/utils.py
```diff
@@ -466,3 +467,12 @@ def wrapper(cls):
         return cls
 
     return wrapper
+
+
+try:
+    from contextlib import nullcontext
+except ImportError:
+
+    @contextmanager
```
martindurant, line 476: It's so simple, may as well just keep our own. That way, we can make the argument required.

fsspec/implementations/http.py
```diff
@@ -262,9 +263,19 @@ async def _put_file(
         **kwargs,
     ):
         async def gen_chunks():
-            with open(rpath, "rb") as f:
-                callback.set_size(f.seek(0, 2))
-                f.seek(0)
+            # Support passing arbitrary file-like objects
+            # and use them instead of streams.
+            if isinstance(rpath, io.IOBase):
```
martindurant, line 268: I *think* all of our own file-like instances pass this, but we could consider testing if the path is a string instead.

isidentical: then we would give away PathLike objects, we can test it against `isinstance(rpath, os.PathLike)` but I am not sure if that would break compatibility in any way, so this is seemed more right. But if we decide to formalize what can be passed to `get_file`/`put_file` (adding streams as an official fsspec feature for example), then we can offer some sort of utility to either open the path, or return the stream directly and do the checks locally with against these defined standards.

martindurant: Good point

martindurant:
Should be `callback.set_size(getattr(f, "size", None))` ? Or `if hasttr(fm "size"):`

fsspec/implementations/http.py
```diff
@@ -231,21 +232,21 @@ async def _cat_file(self, url, start=None, end=None, **kwargs):
         return out
 
     async def _get_file(
```
martindurant, line 234: I thunk this one was right! 
It says, "get file from remote to local", and should be the revers of put_file. That's how we have it in spec.

isidentical: Ah, missed that! Thanks for the correction.

PR #767, opened 2021-09-23
Mount chained fs as a local directory

xyb:
* add `fsspec.fuse` command line:

    `python3 -m fsspec.fuse memory /usr/share /tmp/mem`

* add chained-URL fuse mount feature:

    python3 -m fsspec.fuse 'filecache::zip::file://data.zip' \
            / /tmp/zip
            -o 'filecache-cache_storage=/tmp/simplecache'

* fix ftp root dir issue
* fix chmod issue

### DIFF
--- a/fsspec/fuse.py
+++ b/fsspec/fuse.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import logging
 import os
 import stat
@@ -5,28 +7,37 @@ import threading
 import time
 from errno import EIO, ENOENT
 
-from fuse import FUSE, FuseOSError, Operations
+import click
+from fuse import FUSE, FuseOSError, LoggingMixIn, Operations
+
+from fsspec._version import get_versions
+from fsspec.core import url_to_fs
 
 logger = logging.getLogger("fsspec.fuse")
 
 
 class FUSEr(Operations):
-    def __init__(self, fs, path):
+    def __init__(self, fs, path, ready_file=False):
         self.fs = fs
         self.cache = {}
         self.root = path.rstrip("/") + "/"
         self.counter = 0
         logger.info("Starting FUSE at %s", path)
+        self._ready_file = ready_file
 
     def getattr(self, path, fh=None):
         logger.debug("getattr %s", path)
+        if self._ready_file and path in ["/.fuse_ready", ".fuse_ready"]:
+            return {"type": "file", "st_size": 5}
+
         path = "".join([self.root, path.lstrip("/")]).rstrip("/")
         try:
             info = self.fs.info(path)
         except FileNotFoundError:
             raise FuseOSError(ENOENT)
-        data = {"st_uid": 1000, "st_gid": 1000}
-        perm = 0o777
+
+        data = {"st_uid": info.get("uid", 1000), "st_gid": info.get("gid", 1000)}
+        perm = info.get("mode", 0o777)
 
         if info["type"] != "file":
             data["st_mode"] = stat.S_IFDIR | perm
@@ -61,6 +72,9 @@ class FUSEr(Operations):
 
     def read(self, path, size, offset, fh):
         logger.debug("read %s", (path, size, offset))
+        if self._ready_file and path in ["/.fuse_ready", ".fuse_ready"]:
+            return b"ready"
+
         f = self.cache[fh]
         f.seek(offset)
         out = f.read(size)
@@ -118,10 +132,21 @@ class FUSEr(Operations):
         return 0
 
     def chmod(self, path, mode):
+        if hasattr(self.fs, "chmod"):
+            path = "".join([self.root, path.lstrip("/")])
+            return self.fs.chmod(path, mode)
         raise NotImplementedError
 
 
-def run(fs, path, mount_point, foreground=True, threads=False):
+def run(
+    fs,
+    path,
+    mount_point,
+    foreground=True,
+    threads=False,
+    log_filename="",
+    ready_file=False,
+):
     """Mount stuff in a local directory
 
     This uses fusepy to make it appear as if a given path on an fsspec
@@ -141,7 +166,7 @@ def run(fs, path, mount_point, foreground=True, threads=False):
         character.
     mount_point: str
         An empty directory on the local file-system where the contents of
-        the remote path will appear
+        the remote path will appear.
     foreground: bool
         Whether or not calling this function will block. Operation will
         typically be more stable if True.
@@ -149,12 +174,35 @@ def run(fs, path, mount_point, foreground=True, threads=False):
         Whether or not to create threads when responding to file operations
         within the mounter directory. Operation will typically be more
         stable if False.
+    log_filename: str
+        The FUSE log file name. If not provided, the logging feature is
+        disabled.
+    ready_file: bool
+        Whether the FUSE process is ready. The `.fuse_ready` file will
+        exist in the `mount_point` directory if True. Debugging purpose.
 
     """
+    if log_filename:
+        logging.basicConfig(
+            level=logging.DEBUG,
+            filename=log_filename,
+            format="%(asctime)s %(message)s",
+        )
+
+        class LoggingFUSEr(FUSEr, LoggingMixIn):
+            pass
+
+        fuser = LoggingFUSEr
+    else:
+        fuser = FUSEr
+
     func = lambda: FUSE(
-        FUSEr(fs, path), mount_point, nothreads=not threads, foreground=foreground
+        fuser(fs, path, ready_file=ready_file),
+        mount_point,
+        nothreads=not threads,
+        foreground=foreground,
     )
-    if foreground is False:
+    if not foreground:
         th = threading.Thread(target=func)
         th.daemon = True
         th.start()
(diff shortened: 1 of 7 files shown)


isidentical:
Introducing a new dependency for a simple usage that can be replaced by standard library (`argparse` instead of `click`) might cause more problems on the downstream (direct or indirect version conflicts).

xyb:
@isidentical @martindurant good point, extra dependency removed, please take another look.

fsspec/fuse.py
```diff
@@ -141,20 +167,43 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        the remote path will appear.
     foreground: bool
         Whether or not calling this function will block. Operation will
         typically be more stable if True.
     threads: bool
         Whether or not to create threads when responding to file operations
         within the mounter directory. Operation will typically be more
         stable if False.
+    log_filename: str
+        The FUSE log file name. If not provided, the logging feature is
```
martindurant, line 179: Presumably, this function could be called from code which sets up logging in some other custom way. Possibly this should be done in the app main function rather than here.

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
             func()
         except KeyboardInterrupt:
             pass
+
+
+def mount(
+    url,
```
martindurant, line 219: Is this meant to be an internal function? If not, some docs would be a good idea.

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        else:
+            kwargs[key] = value
+
+    fs, url_path = url_to_fs(url, **kwargs)
+    logger.debug("Mounting %s to %s", url_path, mount_point)
+    run(fs, source_path, mount_point, log_filename=log_file, ready_file=ready_file)
+
+
+def main(args):
+    """Mount filesystem from chained URL to MOUNT_POINT.
```
martindurant, line 251: Include this text in the docs that get printed when no options (or "-h" option) passed?

fsspec/implementations/ftp.py
```diff
@@ -126,6 +126,8 @@ def info(self, path, **kwargs):
         try:
             out = [f for f in files if f["name"] == path][0]
         except IndexError:
+            if path == "/":
+                return {"name": "/", "size": 0, "type": "dir"}
```
martindurant, line 130: Did you find this path gets inspected?

martindurant (commented):
@isidentical , do you have time to go through this again?

fsspec/fuse.py
```diff
@@ -1,32 +1,44 @@
+from __future__ import print_function
```
isidentical, line 1: We do not support Python 2, this is redundant

fsspec/fuse.py
```diff
@@ -1,32 +1,44 @@
...
+import pathlib
 import stat
 import threading
 import time
 from errno import EIO, ENOENT
 
-from fuse import FUSE, FuseOSError, Operations
+from fuse import FUSE, FuseOSError, LoggingMixIn, Operations
+
+from fsspec._version import get_versions
```
isidentical, line 14: You should be able to replace this with `fsspec.__version__`.

fsspec/fuse.py
isidentical: Agreed.

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+    mount_point,
+    log_file,
+    option,
+    foreground=True,
+    threads=False,
+    ready_file=False,
+):
+    kwargs = {}
+    for item in option:
+        key, value = item.split("=", 1)
```
isidentical, line 230: Use `str.partition` for better error messages when the `item` doesn't include `=`.

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        if value.lower().endswith("[int]"):
+            value = int(value[: -len("[int]")])
+        elif value.lower().endswith("[bool]"):
+            value = value[: -len("[bool]")].lower() in ["1", "yes", "true"]
```
isidentical, line 234: Another alternative to this approach is just using `ast.literal_eval`. Instead of trying to manually cast types, you can get options in this form `a=1`, `a=True`, `a='xxx'` and then just call `ast.literal_eval(value)` which would just get you the actual python literal.

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        def format_help(self):
+            usage = super(RawDescriptionArgumentParser, self).format_help()
+            parts = usage.split("\n\n")
+            parts[1] = self.description.rstrip()
+            return "\n\n".join(parts)
+
+    parser = RawDescriptionArgumentParser(prog="fsspec.fuse", description=main.__doc__)
+    parser.add_argument(
+        "--version", action="version", version=get_versions()["version"]
+    )
```
isidentical, line 287: Do you use this option anywhere?

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        action="append",
+        help="Any options of protocol included in the chained URL",
+    )
+    parser.add_argument(
+        "-l", "--log-file", type=pathlib.Path, help="Enable FUSE debug logging"
+    )
+    parser.add_argument(
+        "-f",
+        "--foreground",
+        default=True,
```
isidentical, line 303: This suggests the default is `True`, but the comment below says the default is `False`. Is this a typo or the underlying logic handles it that way? (I think it makes sense to just mark this argument with `action="store_false"`)

fsspec/fuse.py
```diff
@@ -164,3 +213,122 @@ def run(fs, path, mount_point, foreground=True, threads=False):
...
+        "--ready-file",
+        action="store_true",
+        help="The `.fuse_ready` file will exist after FUSE is ready.",
+    )
+    args = parser.parse_args(args)
+    mount(
+        args.url,
+        args.source_path,
+        str(args.mount_point),
+        str(args.log_file),
```
isidentical, line 323: If you are going to cast them to `str()` without (as far as I can see) using any of the pathlib methods, would it make sense to modify the argparse code to `type=str`?

fsspec/fuse.py
xyb: Oops, I wrote this last year when python 2 was still supported. Will delete it.

fsspec/fuse.py
xyb: `fsspec.__version__` is way better, will use it.

fsspec/fuse.py
xyb: OK.

fsspec/fuse.py
xyb: I will merge it with the `main` function.

fsspec/fuse.py
xyb: Good idea!

fsspec/fuse.py
xyb: In fact, I intentionally avoided forcing Python syntax on the command line. Also, instead of use `ast.literal_eval`, which involves more code to handle exceptions, I think it would be more simpler to check plain strings. But yes, I should add more document to explain that hard coded boolean values.

fsspec/fuse.py
xyb: Options like `-h` still work. The `RawDescriptionArgumentParser` just uses the original `ArgumentParser.description` instead of formatted paragraph to avoid breaking command line examples in the docstring.

fsspec/fuse.py
xyb: No. But I think it is standard behavior on command line to provide version information.

fsspec/fuse.py
xyb: Good catch, thanks!

fsspec/implementations/ftp.py
xyb: Yes. If you run the ftp fs example in docstring with logging enabled and `ls` local mounted directory, you will find these messages in log file:
```
2021-09-26 13:00:14,778 FUSE operation getattr raised a <class 'fuse.FuseOSError'>, returning errno 2.
  File "/Users/xyb/Projects/fsspec/fsspec/implementations/ftp.py", line 127, in info
    out = [f for f in files if f["name"] == path][0]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/Users/xyb/Projects/fsspec/fsspec/fuse.py", line 34, in getattr
    info = self.fs.info(path)
  File "/Users/xyb/Projects/fsspec/fsspec/implementations/ftp.py", line 131, in info
    raise FileNotFoundError(path)
FileNotFoundError: /historic/packages/RPMS/.DS_Store

During handling of the above exception, another exception occurred:

  File "/Users/xyb/.virtualenvs/fsspec/lib/python3.8/site-packages/fuse.py", line 734, in _wrapper
    return func(*args, **kwargs) or 0
  File "/Users/xyb/.virtualenvs/fsspec/lib/python3.8/site-packages/fuse.py", line 774, in getattr
    return self.fgetattr(path, buf, None)
  File "/Users/xyb/.virtualenvs/fsspec/lib/python3.8/site-packages/fuse.py", line 1027, in fgetattr
    attrs = self.operations('getattr', self._decode_optional_path(path), fh)
  File "/Users/xyb/.virtualenvs/fsspec/lib/python3.8/site-packages/fuse.py", line 1076, in __call__
    return getattr(self, op)(*args)
  File "/Users/xyb/Projects/fsspec/fsspec/fuse.py", line 36, in getattr
    raise FuseOSError(ENOENT)
fuse.FuseOSError: [Errno 2] No such file or directory
2021-09-26 13:00:15,150 getattr /
```
unless you add the code above.

fsspec/fuse.py
xyb: Changed to `type=str`

fsspec/fuse.py
xyb: Added explanatory information to docstring.

martindurant:
@isidentical , all good with you here?

isidentical:
Yep, LGTM.

PR #769, opened 2021-09-23
Declare more extras; fix various Sphinx warnings

### LINKED ISSUE #768: What FUSE library does `filesystem_spec` use?
[The docs for this library currently state](https://filesystem-spec.readthedocs.io/en/latest/features.html#mount-anything-with-fuse):

> Any path of any file-system can be mapped to a local directory using pyfuse and `fsspec.fuse.run()`.

However, there is no library called "pyfuse" on PyPI.  There's a [pyfuse3](https://pypi.org/project/pyfuse3/), but it uses the module name "pyfuse3", whereas [`filesystem_spec` imports "fuse"](https://github.com/intake/filesystem_spec/blob/b0ce4015481cea0d68fa45db8d53fd511d7b0922/fsspec/fuse.py#L8).  The FUSE libraries that I can find that name their module "fuse" are [fuse-python](https://pypi.org/project/fuse-python/) and [fusepy](https://pypi.org/project/fusepy/).  I suspect that `filesystem_fuse` intends to use fusepy, as it is the only one of the two that provides classes named `FuseOSError` and `Operations`.  `filesystem_spec` should make it clear what fuse library it uses by updating the docs and, preferrably, by adding the library as a (possibly extra) dependency.

Side note: I see in the source for [`docs/source/api.rst`](https://github.com/intake/filesystem_spec/blob/master/docs/source/api.rst) the line `.. autofunction:: fsspec.fuse.run`, but the documentation for this function does not appear in [the Readthedocs site](https://filesystem-spec.readthedocs.io/en/latest/api.html).

jwodder:
Fixes #768 and then some.

In addition to adding fusepy as an extra dependency and correcting the mention of the dependency in the docs (as discussed in #768), this PR also adds extra dependencies for installing libarchive-c and panel, and it also fixes a number of warnings generated when building the docs with Sphinx v4.0.2.

Unfortunately, building the docs still produces a number of warnings of the form "stub file not found \*. Check your autosummary_generate setting." which I was unable to resolve (regardless of whether `autosummary_generate` was set to `True` or `False`).  A possibly related issue is that, when a class inherits one or more methods, those methods will be listed in the class's method summary, but they will not be hyperlinked, and no documentation for the inherited methods will be included in the subclass; this can be fixed with the `:inherited-members:` flag of `.. autoclass::`, but that results in the documentation for inherited methods appearing under both the parent and child classes, so I leave it up to you to decide how best to handle this.

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -143,7 +143,7 @@ is a harder problem to solve, and the implementation described here is only part
 Mount anything with FUSE
 ------------------------
 
-Any path of any file-system can be mapped to a local directory using pyfuse and
+Any path of any file-system can be mapped to a local directory using `fusepy <https://pypi.org/project/fusepy/>`_ and
 :func:`fsspec.fuse.run`. This feature is experimental, but basic file listing with
 details, and read/write should generally be available to the extent that the
 remote file-system provides enough information. Naturally, if a file-system is read-only,
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -168,8 +168,8 @@ is currently a simple dict, but could in the future be LRU, or something more co
 to fine-tune instance lifetimes.
 
 Since files can hold on to write caches and read buffers,
-the instance cache may cause excessive memory usage in some situations; but normally, files
-will get ``close``\d, and the data discarded. Only when there is also an unfinalised transaction or
+the instance cache may cause excessive memory usage in some situations; but normally, files'
+``close`` methods will be called, discarding the data. Only when there is also an unfinalised transaction or
 captured traceback might this be anticipated becoming a problem.
 
 To disable instance caching, i.e., get a fresh instance which is not in the cache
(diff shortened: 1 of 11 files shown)


docs/source/features.rst
```diff
@@ -169,7 +169,7 @@ to fine-tune instance lifetimes.
 
 Since files can hold on to write caches and read buffers,
 the instance cache may cause excessive memory usage in some situations; but normally, files
-will get ``close``d, and the data discarded. Only when there is also an unfinalised transaction or
+will get ``close``\d, and the data discarded. Only when there is also an unfinalised transaction or
```
martindurant, line 172: This construct is a little weird, feel free to reword to avoid the extra syntax

martindurant (approved):
Thank you, this looks perfect!

martindurant:
>  I leave it up to you to decide how best to handle this

If there's no obvious workaround (link, but don't duplicate), they we can leave it be.

docs/source/features.rst
jwodder: Reworded.

jwodder:
@martindurant I am not aware of such a workaround, though the next best thing would probably be to use the `:show-inheritance:` flag of `.. autoclass::` so that readers at least know where inherited methods are inherited from.

isidentical (approved):

martindurant:
Let's shelve further changes, so I can do a release today. Thanks again, @jwodder

PR #771, opened 2021-09-23
spec: put_file should use _parent()

isidentical:
`os.path.dirname()` is a OS specific handling of the parent logic, but in this case `rpath` is a remote-fs path so it should be handled using `_parent()` (which can be overriden by the downstream).

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -802,7 +802,7 @@ class AbstractFileSystem(up, metaclass=_Cached):
             callback.set_size(f1.seek(0, 2))
             f1.seek(0)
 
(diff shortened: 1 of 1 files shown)


martindurant:
Seems to fail for Path objects

martindurant:
I think this is the only one holding a new release.

isidentical:
> I think this is the only one holding a new release.

Should be done now, thanks for reminding.

martindurant:
Anything else before release?

isidentical:
> Anything else before release?

Not from my side.

PR #778, opened 2021-09-29
hdfs: ensure the protocol is stripped for each path

### LINKED ISSUE #777: FSSpec 2021.9.0 Introduces Breaking Changes to the HDFS Api
it looks like one or more PRs merged as part of the 2021.9.0 release cause breaking changes in how FSSpec is used to access HDFS.

The breaking behaviors we observe are:

- Calling `get("hdfs://CLUSTER/a/b/c.txt", "local.txt")` would previously copy the contents of `/a/b/c.txt` on CLUSTER to the local file `local.txt`. Now it tries to copy it to `local.txt/a/b/c.txt`, which does not exist, causing a `FileNotFoundError`.

- Calling `copy("hdfs://CLUSTER/a/somedir/", "somelocaldir", recursive=True)` would previously recursively copy the contents of remote path `/a/somedir/` to the local directory `somelocaldir`. Now it fails with `pyarrow.lib.ArrowInvalid`. 

Some example code snippets that reproduce this behavior (paths anonymized):

With fsspec==2021.9.0:
```python
import fsspec

my_fs = fsspec.filesystem('hdfs')
my_fs.get("hdfs://CLUSTER/a/b/c.txt", "local.txt", recursive=True)
>> FileNotFoundError: [Errno 2] No such file or directory: '/user/home/local.txt/a/b/c.txt'
```

With fsspec==2021.9.0:
```python
import pandas as pd
pd.read_parquet("hdfs://CLUSTER/dir")  # read_parquet wraps pyarrow, which wraps fsspec
>> ArrowInvalid: GetFileInfo() yielded path 'hdfs://CLUSTER/dir/dir.snappy.parquet', which is outside base dir '/dir'
```

Whereas, with fspec==2021.8.1 both examples execute correctly.

Our guess is that the offending change may be https://github.com/isidentical/filesystem_spec/blob/88bc8f1ff1737e9e83cdb7b8aabae3ed331b6bcf/fsspec/implementations/hdfs.py#L181.

This appears to affect all users of `pandas` who use `read_parquet` with partitioned datafiles, so it could be a big impact. Suggest reverting #749 .

isidentical:
There a couple minor changes, as well as the primary one which is under `_adjust_entry`. In the previous revision (08.01) we called `_strip_protocol()` for each item returned from the `ls()`. When I was refactoring this part of the codebase, I unified this logic with `info()` but then we only started to do it when the `'name'` was missing. The primary fix that should resolve #777 is that, now we do it unconditonally on the `_adjust_entry` if the given entry has any `name` field at all.

### DIFF
--- a/fsspec/implementations/tests/test_hdfs.py
+++ b/fsspec/implementations/tests/test_hdfs.py
@@ -75,7 +75,7 @@ def test_read(hdfs):
         assert f.read(100) + f.read() == data
 
 
-def test_copying(hdfs):
+def test_copy(hdfs):
     fs = fsspec.filesystem("hdfs")
 
     fs.mkdir(basedir + "/test_dir")
@@ -92,7 +92,7 @@ def test_copying(hdfs):
     ]
 
 
-def test_getting(hdfs, tmpdir):
+def test_put_get(hdfs, tmpdir):
     fs = fsspec.filesystem("hdfs")
 
     src_dir = Path(tmpdir / "source")
@@ -123,3 +123,17 @@ def test_getting(hdfs, tmpdir):
     assert {
         (dst_dir / file).read_text() for file in files if (dst_dir / file).is_file()
     } == {"file_1", "file_2", "file_3", "file_4"}
+
+
+def test_put_file_get_file(hdfs, tmpdir):
+    fs = fsspec.filesystem("hdfs")
+
+    src_file = Path(tmpdir / "src_file")
+    dst_file = Path(tmpdir / "dst_file")
+
+    src_file.write_bytes(b"heyhey")
+
+    fs.put_file(src_file, basedir + "/src_file")
+    fs.get_file(basedir + "/src_file", dst_file)
+
+    assert dst_file.read_bytes() == b"heyhey"
--- a/fsspec/implementations/tests/test_hdfs.py
+++ b/fsspec/implementations/tests/test_hdfs.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import pytest
 
 import fsspec
@@ -30,9 +32,19 @@ def hdfs(request):
 
 
 def test_ls(hdfs):
-    h = fsspec.filesystem("hdfs")
-    out = [f["name"] for f in h.ls(basedir)]
-    assert out == hdfs.ls(basedir)
+    fs = fsspec.filesystem("hdfs")
+    fs.touch(basedir + "/file_2")
+    fs.mkdir(basedir + "/dir_1")
+    fs.touch(basedir + "/dir_1/file_1")
+    fs.mkdir(basedir + "/dir_2")
+
+    out = {(f["name"], f["kind"]) for f in fs.ls(basedir)}
+    assert out == {
+        (basedir + "/file", "file"),
+        (basedir + "/file_2", "file"),
+        (basedir + "/dir_1", "directory"),
+        (basedir + "/dir_2", "directory"),
+    }
 
 
 def test_walk(hdfs):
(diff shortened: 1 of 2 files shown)


isidentical:
Also CC: @jaedoucette, @j3doucet it would be great if you could test out my branch against your use cases with a real server since I just used the local one.

martindurant:
@j3doucet , will you have the chance to try this?

fsspec/implementations/tests/test_hdfs.py
```diff
@@ -61,3 +73,53 @@ def test_read(hdfs):
...
+def test_copying(hdfs):
+    fs = fsspec.filesystem("hdfs")
+
+    fs.mkdir(basedir + "/test_dir")
+    fs.touch(basedir + "/test_dir/a")
+    fs.touch(basedir + "/test_dir/b")
+    fs.mkdir(basedir + "/test_dir/c")
+    fs.touch(basedir + "/test_dir/c/d")
+
+    fs.copy(basedir + "/test_dir", basedir + "/copy_dir", recursive=True)
```
jaedoucette, line 87: This test should cover one of the new behaviors we saw.

fsspec/implementations/tests/test_hdfs.py
```diff
@@ -61,3 +73,53 @@ def test_read(hdfs):
...
+    files = [file.relative_to(dst_dir) for file in dst_dir.glob("**/*")]
+
+    assert set(map(str, files)) == {
+        "file_1.txt",
+        "file_2.txt",
+        "dir_1/file_3.txt",
+        "dir_1/file_4.txt",
+        "dir_1",
+        "dir_2",
+    }
```
jaedoucette, line 121: I think this test should cover the other new behavior we saw.

isidentical: Which one?

jaedoucette:
@martindurant @isidentical I'll have to check with some other folks about whether I can test this code on our system, but hopefully will be able to.

jaedoucette:
> @j3doucet , will you have the chance to try this?

Yes, I will try this out now.

jaedoucette:
> > @j3doucet , will you have the chance to try this?
> 
> Yes, I will try this out now.

@martindurant @isidentical unfortunately it is much more complex than I expected to run unpublished code on the machines that I can use to access HDFS. 

It may take me a day or more to test this. If you have another setup for testing, that may be preferable.

fsspec/implementations/tests/test_hdfs.py
jaedoucette: If I'm reading the test correctly, this should cover the relative vs. absolute path behavior when copying directories, like the example shown in my most recent comment on #777 .

jaedoucette:
> > > @j3doucet , will you have the chance to try this?
> > 
> > 
> > Yes, I will try this out now.
> 
> @martindurant @isidentical unfortunately it is much more complex than I expected to run unpublished code on the machines that I can use to access HDFS.
> 
> It may take me a day or more to test this. If you have another setup for testing, that may be preferable.

I was able to get a testing setup with a remote HDFS cluster, and verified that `get` now works as hoped.

I was also able to verify that `pandas.read_parquet` works as before when called with a directory against this version.

Thanks for the quick turnaround on this. This fix appears to address all the issues we saw.

isidentical:
> Thanks for the quick turnaround on this. This fix appears to address all the issues we saw.

Cool to hear that! I'll try to take a look at what you meant with the new test case, then we can hopefully land this.

martindurant:
> what you meant with the new test case

I think it was just to confirm that both the previously-working and into-directory writes work, i.e., you have already covered the bases.

isidentical:
If you mean like put_file/get_file with a single file instead of a directory, that is also added now.

jaedoucette:
Yep, my comment was meant to confirm that all the test cases I was hoping to see were already there. This look to me like it will solve all the issues we were having.

isidentical:
Thank you again for testing it out, and verifying. Also sorry for the inconvenience the initial bug has caused.

jaedoucette:
@isidentical thanks you for improving `fsspec`, and for the quick fix! Using this package has simplified so much of the work we do across storage systems.

PR #779, opened 2021-09-29
http: only close connector if it is available

isidentical:
Stuff like `asyncio-retry` offer their own clients which proxy the original `ClientSession`, and they do not bind stuff like `_connector` directly. Since this is a private-attribute, it might be nicer if we access it optionally and only try to close if we can find it.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -123,9 +123,10 @@ class HTTPFileSystem(AsyncFileSystem):
                 return
             except (TimeoutError, FSTimeoutError):
                 pass
-        if session._connector is not None:
+        connector = getattr(session, "_connector", None)
+        if connector is not None:
             # close after loop is dead
-            session._connector._close()
+            connector._close()
 
     async def set_session(self):
         if self._session is None:
--- a/fsspec/implementations/tests/test_http.py
+++ b/fsspec/implementations/tests/test_http.py
@@ -488,13 +488,6 @@ async def get_proxy():
     return ProxyClient()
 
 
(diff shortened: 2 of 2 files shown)


martindurant:
For py 3.6, I think you need [`loop.run_until_complete`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_until_complete)

isidentical:
Ah, right. I'll just skip these under 3.6 (as we do in test_asyn, https://github.com/intake/filesystem_spec/blob/e9bd63f795d84c35898bef3223fc7e055d49c774/fsspec/tests/test_async.py#L56)

martindurant:
That's another solution :)

martindurant:
I am running the tests again, but the windows async [failure](https://github.com/intake/filesystem_spec/pull/779/checks?check_run_id=3748399312) is a little worrying, even if it was a one-off.

isidentical:
I recall seeing this error before too. Would you mind explaining what `test_async_other_thread` does?

martindurant:
> what test_async_other_thread does

The filesystem's loop is running in another explicit thread, and we manipulate the loop object directly to get it to do things (i.e., not using async/await). This is actually quite similar in effect to specifying no loop/asyncronous flags at all and calling the blocking functions.

martindurant:
Perhaps try changing
```
loop = asyncio.get_event_loop()
to
loop = asyncio.new_event_loop()
```
(don't know why this should show up just now...)

isidentical:
The only thing that I guess might affect is the order of tests, let me temporarily move the test to the end and see if it passes or not.

martindurant:
OK, well it passed - we can revisit if it happens again.

PR #780, opened 2021-09-29
Add merge_offset_ranges utility

rjzamora:
This implements a simple `merge_offset_ranges` utility to support the `max_gap` option for the `cat_ranges` method added in #744. Note that this utility also recognizes a `max_block=` argument to help preserve parallelism for large reads.  I intend to use this utility (via `cat_ranges`) to help improve remote-file-system performance in Dask.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -12,7 +12,7 @@ from glob import has_magic
 from .callbacks import _DEFAULT_CALLBACK
 from .exceptions import FSTimeoutError
 from .spec import AbstractFileSystem
-from .utils import PY36, is_exception, other_paths, merge_offset_ranges
+from .utils import PY36, is_exception, other_paths
 
 private = re.compile("_[^_]")
 
@@ -360,16 +360,15 @@ class AsyncFileSystem(AbstractFileSystem):
 
     async def _cat_ranges(self, paths, starts, ends, max_gap=None, **kwargs):
         # TODO: on_error
+        if max_gap is not None:
+            # to be implemented in utils
+            raise NotImplementedError
         if not isinstance(paths, list):
             raise TypeError
         if not isinstance(starts, list):
             starts = [starts] * len(paths)
         if not isinstance(ends, list):
             ends = [starts] * len(paths)
-        if max_gap is not None:
-            paths, starts, ends = merge_offset_ranges(
-                paths, starts, ends, max_gap=max_gap, **kwargs
-            )
         if len(starts) != len(paths) or len(ends) != len(paths):
             raise ValueError
         return await asyncio.gather(
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -12,7 +12,7 @@ from glob import has_magic
 from .callbacks import _DEFAULT_CALLBACK
 from .exceptions import FSTimeoutError
 from .spec import AbstractFileSystem
-from .utils import PY36, is_exception, other_paths
+from .utils import PY36, is_exception, other_paths, merge_offset_ranges
 
 private = re.compile("_[^_]")
 
@@ -360,15 +360,16 @@ class AsyncFileSystem(AbstractFileSystem):
 
     async def _cat_ranges(self, paths, starts, ends, max_gap=None, **kwargs):
         # TODO: on_error
-        if max_gap is not None:
-            # to be implemented in utils
-            raise NotImplementedError
         if not isinstance(paths, list):
             raise TypeError
         if not isinstance(starts, list):
             starts = [starts] * len(paths)
         if not isinstance(ends, list):
             ends = [starts] * len(paths)
+        if max_gap is not None:
+            paths, starts, ends = merge_offset_ranges(
+                paths, starts, ends, max_gap=max_gap, **kwargs
+            )
         if len(starts) != len(paths) or len(ends) != len(paths):
             raise ValueError
         return await asyncio.gather(
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -345,37 +345,6 @@ def test_mirror_from():
     mock.func_3.assert_not_called()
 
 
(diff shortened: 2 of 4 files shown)


fsspec/asyn.py
```diff
@@ -360,15 +360,16 @@ async def _cat(self, path, recursive=False, on_error="raise", **kwargs):
...
-            # to be implemented in utils
-            raise NotImplementedError
         if not isinstance(paths, list):
             raise TypeError
         if not isinstance(starts, list):
             starts = [starts] * len(paths)
         if not isinstance(ends, list):
             ends = [starts] * len(paths)
+        if max_gap is not None:
+            paths, starts, ends = merge_offset_ranges(
```
martindurant, line 370: The problem here is, that the returned list does not align with the input list; we need a way to tell the caller what ranges we are actually fetching. *Unless* we run a second pass and unbundle the bytes blocks into the original start/stop lists (this would involve memory copies, at least, and the pieces might get stuck back together again if used as the cache for a file).

martindurant (commented):
Great to see this! 

I have the one API question that I think needs to be answered. 
Note that even without solving that question, the code would already be useful, but require the called to themselves call marge_offset_ranged, and then cat_ranges with the result (i.e., not actually use the max_gap arg of the latter).

fsspec/utils.py
```diff
@@ -483,3 +483,40 @@ def wrapper(cls):
...
+
+    if paths:
+        # Loop through the coupled `paths`, `starts`, and
+        # `ends`, and merge adjacent blocks when appropriate
+        new_paths = paths[:1]
+        new_starts = starts[:1]
+        new_ends = ends[:1]
+        for i in range(1, len(paths)):
+            if (
+                paths[i] != paths[i - 1]
```
martindurant, line 506: There is an implicit requirement here that the paths be ordered and that the offsets per path also be ordered. That should at least be documented.

fsspec/asyn.py
rjzamora: Ah - Great point.  The user is asking to merge ranges, but still needs to know what those ranges are after merging occurs.  In the interest of keeping the `cat_ranges` API simple, perhaps the best solution is to remove the `max_gap=` argument, and require the user to call `merge_offset_ranges` seperately?

martindurant: That's definitely the easiest option

rjzamora: >the code would already be useful, but require the called to themselves call marge_offset_ranged, and then cat_ranges with the result (i.e., not actually use the max_gap arg of the latter).

I guess you already pointed this out :)

fsspec/utils.py
rjzamora: Good point - I added an explicit sorting step to `merge_offset_ranges`, which the user can skip with `sort=False` if they know their input is already ordered.

martindurant:
Thank you for the changes, @rjzamora !

Can I please request that you remove the copyright and licensing blocks, since this whole project is already licensed as BSD-3.

That aside, this is good to go. I was about to say you should add the to the API docs, but maybe that can wait until we see any usage aside from our own.

fsspec/tests/test_utils.py
```diff
@@ -342,3 +343,68 @@ def func_3(self):
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
```
rjzamora, line 376:
```suggestion
```

fsspec/utils.py
```diff
@@ -483,3 +483,84 @@ def wrapper(cls):
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
```
rjzamora, line 516:
```suggestion
```

rjzamora:
>Can I please request that you remove the copyright and licensing blocks, since this whole project is already licensed as BSD-3.

Sounds good. Thanks for the review @martindurant !

PR #784, opened 2021-10-04
Don't set module variables while processing entrypoints

martindurant:


efiop:
@martindurant Hi 🙂  Would it be possible to release a new version with this fix? Running into it in our packages, blocking release 🙁

martindurant:
You are thinking of 2021.10.0 ; https://pypi.org/project/fsspec/2021.10.1/ was on Friday

efiop:
Thanks, @martindurant ! 🙏

PR #796, opened 2021-10-11
[ENH] Temporary files

### LINKED ISSUE #797: [ENH] Temporary Files
Hello,

Is it possible to implement temporary file  features:

On close file, remove itself, like [Tempfile](https://docs.python.org/3/library/tempfile.html)

```python
import tempfile
fp = tempfile.TemporaryFile()

# Would be like now
fs = build_filesystem('')
fp = fs.TemporaryFile()
```

Would be comfy and nice since your module is able to extend to many fs, while only local for tempfile

Br
Nicolas

ghost:
Add Temporary Files impl

open_temporary, open_temporary_files, TemporaryOpenFile

Closes #797

### DIFF
--- a/fsspec/temporary.py
+++ b/fsspec/temporary.py
@@ -1,4 +1,4 @@
-from .core import OpenFile, OpenFiles
+from .core import OpenFile, OpenFiles, get_fs_token_paths
 
 
 __all__ = (


isidentical:
It would be great if you could provide some description about the potential new APIs that you want to propose in a separate issue, with why they are needed / what are the current solutions etc.

martindurant:
This seems to have stalled. Can someone remind me where we are up to here? @ghost , did you also consider the behaviour of filesystem transactions or opening files with `autocommit=False` ?

isidentical:
The GitHub account seems to be deleted. Closing the PR.

PR #805, opened 2021-10-17
hdfs: switch to pyarrow.fs

ArchangeGabriel:
Should fix `FutureWarning: pyarrow.hdfs.HadoopFileSystem is deprecated as of 2.0.0, please use pyarrow.fs.HadoopFileSystem instead.`.

### DIFF
--- a/fsspec/implementations/hdfs.py
+++ b/fsspec/implementations/hdfs.py
@@ -3,7 +3,7 @@ import shutil
 import weakref
 from contextlib import suppress
 
-from pyarrow.hdfs import HadoopFileSystem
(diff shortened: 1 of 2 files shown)


isidentical:
We already have a separate implementation for `pyarrow.fs` (see `fsspec.implementations.arrow`) :)

ArchangeGabriel:
OK, so I guess when `pyarrow.hdfs.HadoopFileSystem` gets removed upstream, that’s also what should happen here.

isidentical:
> that’s also what should happen here.

Probably. or we could just move it to it's own package with pinning `pyarrow < last_supported_version`.

PR #806, opened 2021-10-20
Add fsspec.parquet.open_parquet_file 

rjzamora:
This is a POC, and a proposal to add a Parquet-specific version of `AbstractFileSystem.open` within a new `fsspec.parquet` module.

**Why not require this logic to live in Dask?**: The original plan was to implement most (perhaps all) of the logic from this PR in `dask.dataframe`.  We can certainly do this.  However, these optimizations are also beneficial to users who are working with pandas and/or cudf on remote file sysems without Dask.  Therefore, the only way to avoid code duplication in Dask, cudf, NVTabular, ... is to add the optimized `open_parquet_file` directly to fsspec.

**Pandas Example**:

```python
from fsspec.core import get_fs_token_paths
from  fsspec.parquet import open_parquet_file
import pandas as pd

path = "gs://my-bucket/criteo-parquet/day_0.parquet"
columns = ["I10"]
engine = "fastparquet"

fs = get_fs_token_paths(path, storage_options=...)[0]

%%time
with fs.open(path, mode="rb") as f:
    df = pd.read_parquet(f, columns=columns, engine=engine)
```
```
CPU times: user 6.75 s, sys: 1.33 s, total: 8.07 s
Wall time: 52.4 s
```

```python
%%time
with open_parquet_file(path, fs=fs, columns=columns) as f:
    df = pd.read_parquet(f, columns=columns, engine=engine)
```
```
CPU times: user 3.9 s, sys: 510 ms, total: 4.41 s
Wall time: 5.23 s
```


martindurant:
I like these numbers:
```
Wall time: 52.4 s

Wall time: 5.23 s
```
Which engine was this (or do both do so well)?

I feel like this functionality should go into a separate module, since it's all quite separate from the core functionality, and maybe even the entry point for calling it should be a standalone function rather than a method on the main file system base class.

Do you think it's possible to open a file using a metadata object from a _metadata file, rather than requiring to download the footer?

martindurant:
> perhaps fsspec.parquet

Sounds right

> Is this what you are suggesting

Exactly. fastparquet.core.read_row_group  or ParquetFile.read_row_group_file do this.

(aside, I notice the following block in dask, fastparquet.pyL1019:
```
                dfs = []
                for row_group in row_groups:
                    dfs.append(
                        parquet_file.read_row_group_file(
                            row_group,
                            columns,
                            categories,
                            index=index,
                            **kwargs.get("read", {}),
                        )
                    )
                return concat(dfs, axis=0) if len(dfs) > 1 else dfs[0]
```
This would be better done by assigning the memory and filling it in, as ParquetFile.to_pandas currently does )

rjzamora:
@martindurant - I feel that I am getting pretty close to a reasonable "stopping point" for this particular PR.

I expect that we will want to continue evolving/improving this new `parquet` module a bit, but I think the current state achieves what we (perhaps I) "need" in dask/cudf/nvtabular.  For example, I agree that we should avoid transfering the footer bytes if we already have a metadata object handy. However, this feature will not be used for the applications I am focused on, and will likely take a fair amount of work to get "right".

rjzamora:
> OK, fair enough. I haven't had a chance to look at your more recent commits - is it fair to say that they were all changes related to previous comments?

Yes - All changes were either intended to address your code review comments, or to debug problems I ran into while experimenting in a real GCS environment.

> Do you plan on publishing an article about this progress, because I think it's a pretty big deal to show what a little python duck-typing can do!

Yes - I certainly plan to write this all up (and will be happy to include you if you are interested) :)

rjzamora:
**Correction**: The current state of this PR is sufficient for the "pyarrow" engine in Dask (and for dask_cudf/NVTabular). However, we will probably need further changes to optimize the "fastparquet" engine in Dask. There was no harm in merging this, but I will try to do a bit of digging today and tomorrow to figure out if the API needs to change at all for the fastparquet case.

martindurant:
This piece of functionality is all yours, but let me know if I can help, even with code changes in fastparquet.

PR #811, opened 2021-11-01
Faster references, git URL fix

### LINKED ISSUE #809: get_fs_token_paths doesnt use path from git:// uri
When `get_fs_token_paths` is used with `"git://<some_local_path>"`, it should return an instance of `GitFileSystem` in that local path. But `_get_kwargs_from_urls` of that class does not extract `path` option from uri, so when the [class is instantiated later inside `get_fs_token_paths`](https://github.com/intake/filesystem_spec/blob/da89f163c94d17deacbe2cecfc37e20009580573/fsspec/core.py#L640), the `options` dict is empty, and default value for `path` is used, which is `os.getcwd` (and also wrong if uri was not just `"git://"`). And if there is no git repo in cwd, an exception is raised.
One-liner to reproduce:
`mkdir aaa && cd aaa && git init && cd ../ && python -c "from fsspec import get_fs_token_paths; get_fs_token_paths('git://aaa')"`

martindurant:
Fixes #809

### DIFF
--- a/fsspec/config.py
+++ b/fsspec/config.py
@@ -63,7 +63,7 @@ def set_conf_files(cdir, conf_dict):
                 conf_dict.setdefault(key, {}).update(dict(js[key]))
 
 
-def apply_config(cls, kwargs, conf_dict=conf):
+def apply_config(cls, kwargs, conf_dict=None):
     """Supply default values for kwargs when instantiating class
 
     Augments the passed kwargs, by finding entries in the config dict
@@ -80,6 +80,8 @@ def apply_config(cls, kwargs, conf_dict=conf):
     -------
     dict : the modified set of kwargs
     """
+    if conf_dict is None:
+        conf_dict = conf
     protos = cls.protocol if isinstance(cls.protocol, (tuple, list)) else [cls.protocol]
     kw = {}
     for proto in protos:
--- a/fsspec/fuse.py
+++ b/fsspec/fuse.py
@@ -71,6 +71,7 @@ class FUSEr(Operations):
     def read(self, path, size, offset, fh):
         logger.debug("read %s", (path, size, offset))
         if self._ready_file and path in ["/.fuse_ready", ".fuse_ready"]:
+            # status indicator
             return b"ready"
 
         f = self.cache[fh]
@@ -79,7 +80,7 @@ class FUSEr(Operations):
         return out
 
     def write(self, path, data, offset, fh):
-        logger.debug("read %s", (path, offset))
+        logger.debug("write %s", (path, offset))
         f = self.cache[fh]
         f.write(data)
         return len(data)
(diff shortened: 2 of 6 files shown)


martindurant:
@mike0sv : the correct invocation is with a URL
```
get_fs_token_paths("git://myfile::file://local_git")
```
i.e., find the file `myfile` within the repo at local path `local_git`. I do not believe `pygit` can be persuaded to work with anything other than local paths, but still. Feel free to suggest alternate ways to encode the repo location and starting hash (and see the example of the github implementation).

mike0sv:
Maybe just look for `.git` dir in `path`, `os.path.dirname(path)`,  `os.path.dirname(os.path.dirname(path))` until we find it? Then it will be `path` for pygit, and the rest (path relative to git root) we can return as `paths`

mike0sv:
Also, when I first saw that fsspec supports git filesystem, I thought it supports everything from here https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols, but I guess this is another big topic

martindurant:
> Maybe just look for .git dir in path, os.path.dirname(path), os.path.dirname(os.path.dirname(path)) until we find it?

git will already look in the parents, that's not my question. I am asking for suggestions of how you would expect to partition the "location of the git repo" from "location of a file within the repo" in a single URL. I have given a form that works now, as of this URL.

mike0sv:
Yes, I propose to just find it before providing to `pygit`. For example, for URL `git://some/path/to/git_repo_root_here/some/path/inside/repo` it will yield `fs=GitFileSystem(path="some/path/to/git_repo_root_here"), token=...,paths=["some/path/inside/repo"]`. Rev could be provided maybe like `git://rev@...` or `git://rev::...`

martindurant:
@mike0sv , see the GitFileSystem docstring for my proposed compromise.

mike0sv:
I am still convinced that what I am proposing is easier and cleaner. Am I missing something here?
Also, I think in your solution it makes more sense to use `git://<rev>@<repo>::<path>`, than `git://<rev>@<path>::<repo>`

martindurant:
Docstring:

> may be of the form
> "git://[path-to-repo[:]][ref@]path/to/file" (but the actual
> file path should not contain "@" or ":").

The thing with "::" is general across fsspec and not very relevant to git, which only works with local paths.

mike0sv:
Yes, but in tests its still
```python
    fs, _, paths = fsspec.core.get_fs_token_paths(f"git://file1::file://{d}")
    assert make_path_posix(d) in make_path_posix(fs.repo.path)
    assert paths == ["file1"]
```
Or you just didn't implement this yet?
Still, you can just extract fo from regular local path like this `git://rev@local_path` -> strip protocol, split by @ -> `options['fo'] = Repository(local_path).workdir, paths = [os.path.relpath(...)]`. Without separately specifying repo path

martindurant:
> you can just extract fo from regular local path like

With which API are you expecting to use this? 

I added the newer alternate URL type to the same test.

martindurant:
I'm putting this in now, but we can still iterate on the best spelling for git URLs afterwards.

PR #813, opened 2021-11-03
Add metadata option to open_parquet_file

rjzamora:
This is a follow-up to #806 (where a new `fsspec.parquet` module was added). The most important changes in this PR are the following:

1. Adds a new `metadata` argument to `open_parquet_file`
2. Expands the `row_group` argument of `open_parquet_file` so that the user may optionally pass in a list of distinct engine-specific row-group metadata (rather than a list of row-group indices).
3. Adds a `strict` option to `KnownPartsOfAFile` (defaulting to `True`). When `strict=False`, reads that extend beyond the boundary of a known part will be zero-buffered. When `strict=True`, the missing range will be explicitly fetched from the remote source. Note that the `strict` logic only applies to reads that begin within a known part. It seems that Fastparquet will often seek to an offset within a known part, but then read slightly beyond the end of the buffer/file. I experimented with various ways to deal with this, but using `strict=False` (within `open_parquet_file`) seems to be the simplest way to avoid constant errors/warnings and unnecessary data transfer. Note that the user can always set `strict=True` for debugging purposes.

The primary purpose of this PR is to make it possible to utilize `open_parquet_file` within Dask for better remote-file perfomance in `dask.dataframe.read_parquet`.  Without these changes, only the "pyarrow" engine can benefit from the new optimization.

Note that these chnges do not (yet) add support for pyarrow-based metadata input. I will target that feature in a separate PR.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -426,7 +426,7 @@ class AllBytes(BaseCache):
 class KnownPartsOfAFile(BaseCache):
     name = "parts"
 
-    def __init__(self, blocksize, fetcher, size, data={}, strict=False, **_):
+    def __init__(self, blocksize, fetcher, size, data={}, strict=True, **_):
         super(KnownPartsOfAFile, self).__init__(blocksize, fetcher, size)
         self.strict = strict
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -442,6 +442,15 @@ class KnownPartsOfAFile(BaseCache):
                 else:
                     offsets.append((start, stop))
                     blocks.append(data.pop((start, stop)))
+
+            if self.size:
+                # If our last range goes to the end of the
+                # file, extend the range in case the read
+                # goes beyond the end of the file
+                last_key = offsets[-1]
+                if last_key[1] >= self.size:
+                    offsets[-1] = (last_key[0], max(last_key[1], self.size * 2))
+
             self.data = dict(zip(offsets, blocks))
         else:
             self.data = data
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -426,8 +426,9 @@ class AllBytes(BaseCache):
 class KnownPartsOfAFile(BaseCache):
     name = "parts"
 
-    def __init__(self, blocksize, fetcher, size, data={}, **_):
+    def __init__(self, blocksize, fetcher, size, data={}, strict=False, **_):
         super(KnownPartsOfAFile, self).__init__(blocksize, fetcher, size)
+        self.strict = strict
 
         # simple consolidation of contiguous blocks
         if data:
@@ -443,26 +444,31 @@ class KnownPartsOfAFile(BaseCache):
                     offsets.append((start, stop))
                     blocks.append(data.pop((start, stop)))
 
-            if self.size:
-                # If our last range goes to the end of the
-                # file, extend the range in case the read
-                # goes beyond the end of the file
-                last_key = offsets[-1]
-                if last_key[1] >= self.size:
-                    offsets[-1] = (last_key[0], max(last_key[1], self.size * 2))
-
             self.data = dict(zip(offsets, blocks))
         else:
             self.data = data
 
     def _fetch(self, start, stop):
         for (loc0, loc1), data in self.data.items():
-            if loc0 <= start < loc1 and loc0 <= stop <= loc1:
+            # If self.strict=False, use zero-padded data
+            # for reads beyond the end of a "known" buffer
+            out = b""
+            if loc0 <= start < loc1:
                 off = start - loc0
                 out = data[off : off + stop - start]
-                # reads beyond buffer are padded with zero
-                out += b"\x00" * (stop - start - len(out))
-                return out
+                if not self.strict or loc0 <= stop <= loc1:
+                    # The request is within a known range, or
+                    # it begins within a known range, and we
+                    # are allowed to pad reads beyond the
+                    # buffer with zero
+                    out += b"\x00" * (stop - start - len(out))
+                    return out
+                else:
+                    # The request ends outside a known range,
+                    # and we are being "strict" about reads
+                    # beyond the buffer
+                    start = loc1
+                    break
 
         # We only get here if there is a request outside the
         # known parts of the file. In an ideal world, this
@@ -476,7 +482,7 @@ class KnownPartsOfAFile(BaseCache):
             f"Read is outside the known file parts: {(start, stop)}. "
             f"IO/caching performance may be poor!"
         )
-        return super()._fetch(start, stop)
+        return out + super()._fetch(start, stop)
 
 
 caches = {
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -424,6 +424,28 @@ class AllBytes(BaseCache):
 
 
 class KnownPartsOfAFile(BaseCache):
+    """
+    Cache holding known file parts.
+
+    Parameters
+    ----------
+    blocksize: int
+        How far to read ahead in numbers of bytes
+    fetcher: func
+        Function of the form f(start, end) which gets bytes from remote as
+        specified
+    size: int
+        How big this file is
+    data: dict
+        A dictionary mapping explicit `(start, stop)` file-offset tuples
+        with known bytes.
+    strict: bool, default True
+        Whether to fetch reads that go beyond a known byte-range boundary.
+        If `False`, any read that ends outside a known part will be zero
+        padded. Note that zero padding will not be used for reads that
+        begin outside a known byte-range.
+    """
+
     name = "parts"
 
     def __init__(self, blocksize, fetcher, size, data={}, strict=True, **_):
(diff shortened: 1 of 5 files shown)


martindurant:
> Fastparquet will often seek to an offset within a known part, but then read slightly beyond the end of the buffer/file.

I suspect this must be thrift's internal buffering at work. https://github.com/dask/fastparquet/pull/633 would not behave like this, should I ever get around to merging it.

rjzamora:
To add a bit of "performance motivation" here, I just ran a fresh "fastparquet" test from GCS storage:

```python
from fsspec.core import get_fs_token_paths
from  fsspec.parquet import open_parquet_file
import fastparquet as fp
from functools import partial

path = "gs://my-bucket/criteo-parquet/day_0.parquet"
columns = ["I10"]
engine = "fastparquet"

fs = get_fs_token_paths(path, storage_options=...)[0]
pf = fp.ParquetFile(paths[0], open_with=fs.open)

%%time
df = pf.to_pandas(columns=columns)
```
```
CPU times: user 8.34 s, sys: 2.08 s, total: 10.4 s
Wall time: 45.2 s
```

```python
%%time
pf.open = partial(
    open_parquet_file,
    fs=fs,
    metadata=pf,
    columns=columns,
)
df = pf.to_pandas(columns=columns)
```
```
CPU times: user 3.39 s, sys: 431 ms, total: 3.83 s
Wall time: 3.94 s
```

rjzamora:
@martindurant - Any thoughts on how likely this is to get in before the next fsspec release? I think the PR is in pretty-good shape and is ready for final review, but I certainly have tunnel vision here (as I often do) :)

fsspec/caching.py
```diff
@@ -426,8 +426,9 @@ def _fetch(self, start, end):
 class KnownPartsOfAFile(BaseCache):
     name = "parts"
 
-    def __init__(self, blocksize, fetcher, size, data={}, **_):
+    def __init__(self, blocksize, fetcher, size, data={}, strict=True, **_):
```
martindurant, line 429: Should add minimal docs for these arguments

fsspec/parquet.py
```diff
@@ -54,10 +65,19 @@ def open_parquet_file(
...
+        engine is used).
     storage_options : dict, optional
         Used to generate an `AbstractFileSystem` object if `fs` was
         not specified.
+    strict : bool, optional
+        Whether the resulting `KnownPartsOfAFile` cache should
+        fetch reads that go beyond a known byte-range boundary.
+        If `False` (the default), any read that ends outside a
+        known part will be zero padded. Note that using
+        `strict=True` may be useful for debugging.
```
martindurant, line 80: (should have a logger in KnownPartsCache?)

fsspec/parquet.py
```diff
@@ -78,11 +98,20 @@ def open_parquet_file(
     if fs is None:
         fs = url_to_fs(path, storage_options=(storage_options or {}))[0]
 
+    # For now, `columns == []` not supported. Just use
+    # default `open` command with `path` input
+    if columns is not None and len(columns) == 0:
```
martindurant, line 103: `if not columns` ?

fsspec/parquet.py
```diff
@@ -91,20 +120,35 @@ def open_parquet_file(
         footer_sample_size=footer_sample_size,
     )
 
+    # If data is empty, just use default
+    # `open` command with `path` input
+    if len(data) != 1:
+        return fs.open(path, mode=mode)
```
martindurant, line 126: I'm not sure you want this shortcut. You still might have multiple requests sequentially if the one block is big compared to the readahead size (no opportunity to pass other arguments here)

fsspec/caching.py
martindurant: (or just refer to the parquet calling function, which I see does have the doc)

fsspec/parquet.py
```diff
@@ -119,8 +163,24 @@ def _get_parquet_byte_ranges(
...
 
-    # Set the engine
-    engine = _set_engine(engine)
+    # Pass to specialized function if metadata is defined
+    if metadata is not None:
+
+        # Use the provided parquet metadata object
+        # to avoid transfering/parsing footer metadata
+        return _get_parquet_byte_ranges_from_metadata(
+            metadata,
```
martindurant, line 172: Derive value of engine from the type of the metadata? Will get weird error on mismatch. Not urgent.

fsspec/parquet.py
```diff
@@ -220,29 +283,81 @@ def _get_parquet_byte_ranges(
...
+):
+    """Simplified version of `_get_parquet_byte_ranges` for
+    the case that an engine-specific `metadata` object is
+    provided, and the remote footer metadata does not need to
+    be transfered before calculating the required byte ranges.
+    """
+
+    # Set engine if necessary
+    if isinstance(engine, str):
+        engine = _set_engine(engine)
```
martindurant, line 313: Engine is not used in multiple places? Is it possible to use one engine for the metadata and another for the data read?

fsspec/parquet.py
```diff
@@ -220,29 +283,81 @@ def _get_parquet_byte_ranges(
...
+
+    # Add b"PAR1" to header
+    _add_header_magic(result)
+
+    return result
+
+
+def _transfer_ranges(fs, blocks, data_paths, data_starts, data_ends):
+    # Use cat_ranges to gather the data byte_ranges
+    for i, data in enumerate(fs.cat_ranges(data_paths, data_starts, data_ends)):
```
martindurant, line 344: looks like a `zip`

fsspec/parquet.py
```diff
@@ -220,29 +283,81 @@ def _get_parquet_byte_ranges(
...
+        if data_ends[i] > data_starts[i]:
```
martindurant, line 345: Is this ever not true?

fsspec/parquet.py
```diff
@@ -220,29 +283,81 @@ def _get_parquet_byte_ranges(
...
+def _add_header_magic(data):
+    # Add b"PAR1" to file headers
+    for i, path in enumerate(list(data.keys())):
+        add_magic = True
+        for k in data[path].keys():
+            if k[0] == 0 and k[1] >= 4:
+                add_magic = False
+                break
+        if add_magic:
+            data[path][(0, 4)] = b"PAR1"
```
martindurant, line 358: (fastparquet doesn't check this by default)

fsspec/parquet.py
rjzamora: Well, we don't want to do this if `columns=None`, only if `columns=[]` I guess. Perhaps we can just assume columns will be a list

martindurant: sorry, misread

fsspec/parquet.py
rjzamora: I think it may only be cudf that checks this :/

I do think we need the functionality to make this optional, but I was hoping to attack this later since the overhead is pretty negligible

fsspec/parquet.py
rjzamora: Good point - Do you prefer something like:

```python
def _transfer_ranges(fs, blocks, paths, starts, ends):
    ranges = (paths, starts, ends)
    for path, start, stop, data in zip(*ranges, fs.cat_ranges(*ranges)):
        blocks[path][(start, stop)] = data
```
or even

```python
def _transfer_ranges(fs, blocks, *ranges):
    if len(ranges) != 3:
        raise ValueError("Incorect arguments passed to _transfer_ranges")
    for path, start, stop, data in zip(*ranges, fs.cat_ranges(*ranges)):
        blocks[path][(start, stop)] = data
```
?

martindurant: The first one looks nice

fsspec/parquet.py
rjzamora: I think this was a band-aid for an old version of this code with bugs :) removed.

fsspec/parquet.py
rjzamora: I removed this particular engine check, but I agree that there is further work to be done to properly check/infer that the correct engine is used with user-specified metadata.

fsspec/parquet.py
rjzamora: Yes - We absolutely should do this, but I'd prefer to do it in a stand-alone PR if possible.

fsspec/parquet.py
rjzamora: Your are right - Removed.

fsspec/parquet.py
rjzamora: Good call - Added logger when an "unknown" range is fetched

fsspec/caching.py
rjzamora: Agree - I added a new docstring to KnownPartsOfAFile

rjzamora:
Okay - I think the more-critical review comments have been addressed. I still want to clean up the general approach to engine handling, but I'd prefer to do that work separately.

rjzamora:
Woohoo - Thanks for the prompt and valuable review here @martindurant !

PR #814, opened 2021-11-04
fix 'find' with protocol striped

JinrongGuo1:
**Problem**: This exception was found by reading parquet from HDFS dir with `dask.dataframe` and nothing returned in my case. After some digging, I think it might be raised by the function `glob` in `spec.py`. 

**Caused by**: In `spec.py`, line 537 parses all files in the given input dir, the files returned by `find` are presented with protocol included. However, the `re.match` operation in line 564, the match pattern starts with path without protocol, finally, `glob` returns None.

**Solution**: As the PR

**Note**: I've tested in my case, after modifying the function `find`, I can read my HDFS correctly. But I'm not sure by now whether this modification would affect other cases.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -446,7 +446,7 @@ class AbstractFileSystem(up, metaclass=_Cached):
         for _, dirs, files in self.walk(path, maxdepth, detail=True, **kwargs):
             if withdirs:
                 files.update(dirs)
-            out.update({info["name"]: info for name, info in files.items()})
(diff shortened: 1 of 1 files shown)


martindurant:
On the failures, the HDFS paths should probably *not* have the terminating "/" characters for directories, so the tests are wrong, not your code.

martindurant:
On the original question: would it be fair to day that the actual bug is HDFS returning names with protocol? I am loath to add protocol stripping to the output, because it might be expensive across potentially many paths.

PR #815, opened 2021-11-05
Update the GitHub url with the new namespace

isidentical:


martindurant:
You got there quickly :)

I was planning on releasing all three with updated URLs this afternoon (although github is good about redirects, so nothing will break).

martindurant:
You are finding quite a few! Also, there are some `/intake/` ones (for s3fs, gcsfs, adlfs when it too moves).

fsspec/tests/test_mapping.py
```diff
@@ -72,7 +72,7 @@ def test_pickle():
 
 
 def test_keys_view():
-    # https://github.com/intake/filesystem_spec/issues/186
+    # https://github.com/fsspec/fsspec/issues/186
```
martindurant, line 75: I guess this one is wrong

isidentical:
```suggestion
    # https://github.com/fsspec/filesystem_spec/issues/186
```

Something to consider :) (since the package is already named fsspec on pypi)

isidentical:
> Also, there are some /dask/ ones (for s3fs, gcsfs, adlfs when it too moves).

`/dask/` s3fs/gcsfs ones are fixed here (but not adlfs)

PR #817, opened 2021-11-08
Fix range-index bug in open_parquet_file

rjzamora:
Simple bug fix in `fsspec.parquet.read_parquet_file` for parquet files with range-index information in the pandas metadata. Note that the bug is only for the "pyarrow" engine.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -471,10 +471,10 @@ class KnownPartsOfAFile(BaseCache):
             self.data = data
 
     def _fetch(self, start, stop):
+        out = b""
         for (loc0, loc1), data in self.data.items():
             # If self.strict=False, use zero-padded data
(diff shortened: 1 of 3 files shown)


fsspec/caching.py
```diff
@@ -471,10 +471,10 @@ def __init__(self, blocksize, fetcher, size, data={}, strict=True, **_):
             self.data = data
 
     def _fetch(self, start, stop):
+        out = b""
```
rjzamora, line 474: Another minor bug fix for the unlikely edge case that `self.data` is empty.

rjzamora:
(cc @martindurant for viz) It looks like this fix (although small) is a blocker for dask and cudf/dask-cudf to use `fsspec.parquet`.

PR #821, opened 2021-11-11
Fix footer re-sampling bug in fsspec.parquet

rjzamora:
This PR fixes yet another small bug in the new `fsspec.parquet` module. Although the existing logic to update the footer-metadata sample is correct, it does not properly update the "start" offset for the metadata sample before adding the corresponding footer key to the `data` dictionary.

### DIFF
--- a/fsspec/parquet.py
+++ b/fsspec/parquet.py
@@ -214,14 +214,12 @@ def _get_parquet_byte_ranges(
 
         # Check our footer samples and re-sample if necessary.
         missing_footer_starts = footer_starts.copy()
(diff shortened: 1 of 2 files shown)


martindurant:
I don't exactly follow after a brief glance. Would the changed test fail without this update?

rjzamora:
Yes, exactly. We were previously collecting a “correct” footer sample after an initial undersample. However, we were failing to adjust the “key” for the footer byte range to reflect the correct offset. Therefore, if we used a silly footer-sample size, like 8, the key would incorrectly reflect that only the last 8 bytes were being cached for the footer.

PR #824, opened 2021-11-16
Allow batches in more async operations

martindurant:


martindurant:
cc @rabernat

martindurant:
The problem is that the keyword "batch_size" is not in the non-async implementations. We could include it there in all the calls, but ignore it. If users only ever touch the default value, this is not an issue.

rabernat:
I like the idea of setting this via global configuration options for fsspec. That would seem to be easier than passing config options through every function.

martindurant:
Here we enable both (but needs documentation, of course). The variables `fsspec.asyn._DEFAULT_BATCH_SIZE` and `fsspec.asyn._NOFILES_DEFAULT_BATCH_SIZE` and corresponding config keys "gather_batch_size", "nofiles_gather_batch_size" (config takes precedence, if key exists; function kwargs take ultimate precedence).

martindurant:
(thank you, CI, for spotting the subtle bug!)

PR #828, opened 2021-11-17
Generics

martindurant:
Enables `info(url)` where the URL is for any known implementation.

Open questions:
- to what extent can we recreate URLs for the complex chained case, and should we? My idea is, you should be able to feed the name back into `info` and get the same result
- should the functions belong to a full filesystem class, so that some behaviours can be inherited from AbstractFileSystem? I am thinking no. However, many of the functions will have a rather similar form, so we might need at least a decorator.
- some functions (only `cp`?) will operate across filesystems, so you would need more than one "protocol", if using that kwarg
- There are a few simple options for how to pick a filesystem for each protocol, I think they cover all reasonable bases.


fsspec/spec.py
```diff
@@ -196,6 +196,17 @@ def _strip_protocol(cls, path):
...
+    def _unstrip_protocol(self, name):
+        # could be overridden for FSs that make work with URLs
+        if isinstance(self.protocol, str):
+            if name.startswith(self.protocol):
+                return name
+            return self.protocol + "://" + name
+        else:
+            if name.startswith(tuple(self.protocol)):
+                return name
+            return self.protocol[0] + "://" + name
```
efiop, line 208: For the record: we've also implemented this before in our wrappers, e.g. https://github.com/iterative/dvc/blob/28f57933db4873b2f6fcb466e479754aa016de2f/dvc/fs/s3.py#L162

efiop:
One note about localfs on windows: right now _strip_protocol will return you path that starts with `C:`, which doesn't start with the `localfs.root_marker`(`/`), which makes it hard to create generic path handling. At the same time `C:` is clearly not a root marker, as localfs should be able to worth with both `C:` and, e.g., `D:` paths, so we probably need to introduce a convention that they get normalized to `/C:/some/path`, but the leading `/` will need to be stripped before passing over to system methods. Maybe this has been discussed somewhere before?

fsspec/spec.py
```diff
@@ -196,6 +197,17 @@ def _strip_protocol(cls, path):
         # use of root_marker to make minimum required path, e.g., "/"
         return path or cls.root_marker
 
+    def _unstrip_protocol(self, name):
```
d4l3k, line 200: Can we make this an exported method? (without _) since it's very useful in conjunction with url_to_fs

martindurant:
I will put this in now as an experimental feature with appropriate warning in the docstring, and we can iterate on pushing it forward.

PR #830, opened 2021-11-19
Remove BlockSizeError

martindurant:
So that we can still stream from the start of a file without error
(will cancel transfer once enough bytes are received). Does *not*
allow random access from not the start of the download.

Fixes https://github.com/dask/dask/issues/8402


martindurant:
If no one can think of a reason *not* to do this, I'll merge by the end of this day.

PR #831, opened 2021-11-21
FileNotFoundError [LocalFileSystem] on unix-like root / directory - closes #829

### LINKED ISSUE #829: FileNotFoundError [LocalFileSystem] on unix-like root / directory
Hello,

in case of `LocalFileSystem().ls('/')`, a `FileNotFoundError` is being raised. 

```python
FileNotFoundError: [Errno 2] No such file or directory: ''
```

This looks quite normal as path is stripped right 
https://github.com/fsspec/filesystem_spec/blob/b68be6c57f54a71c6ffc89e669c410da353ebc1f/fsspec/implementations/local.py#L54

Is this an expected behavior?

e-roux:
Hello, 
proporsal for fixing issue #829

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -52,6 +52,8 @@ class LocalFileSystem(AbstractFileSystem):
 
     def ls(self, path, detail=False, **kwargs):
         path = self._strip_protocol(path)
+        # root_marker if stripped path return an empty string
+        path = path or self.root_marker
         if detail:
             with os.scandir(path) as it:
                 return [self.info(f) for f in it]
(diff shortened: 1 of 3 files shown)


fsspec/implementations/local.py
```diff
@@ -52,6 +52,8 @@ def rmdir(self, path):
 
     def ls(self, path, detail=False, **kwargs):
         path = self._strip_protocol(path)
+        # root_marker if stripped path return an empty string
+        path = path or self.root_marker
```
martindurant, line 56: Do you think this should live in `_strip_protocol`, like 
```
        return make_path_posix(path).rstrip("/") or cls.root_marker
```
?

e-roux: Indeed, this specific issue is linked to `_strip_protocol`, not to `ls`

PR #839, opened 2021-11-26
Fix typos found by codespell

DimitriPapadopoulos:
I have left out `cachable-> cacheable` because it would modify the API.

### DIFF
--- a/fsspec/implementations/github.py
+++ b/fsspec/implementations/github.py
@@ -23,7 +23,7 @@ class GithubFileSystem(AbstractFileSystem):
       may specify sha in the extra args
     - 'github://org:repo@/precip/catalog.yml', where the org and repo are
       part of the URI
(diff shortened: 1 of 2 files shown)


martindurant:
Thank you.

> I have left out cachable-> cacheable because it would modify the API.

we can argue about what's right for that one :)

DimitriPapadopoulos:
Most sources say it's `cacheable`, for example:
* https://www.lexico.com/definition/cacheable

Also the general accepted spelling seems to be `cacheable`:
*  [1.440.000](https://www.google.com/search?q=cacheable) vs. [207.000](https://www.google.com/search?q=cachable) hits in Google
* [15M](https://github.com/search?q=cacheable&type=code) vs. [2M](https://github.com/search?q=cachable&type=code) hits in GitHub

But I know it's debatable...
https://howtospell.co.uk/spelling-words-with-ible-or-able

Issue #834, opened 2021-11-22
[Feature Request] Typehints


alanhdu:
Hi all! I was wondering whether `fsspec` would be open to adding typehints to its codebase (both for internal checking, and for packaging externally via PEP 561). This has a number of advantages (better IDE integration, might catch bugs, etc), so this is definitely something I think would improve the end-user experience.

If you're interested, I'd be happy to send along some patches setting up some of the infrastructure and starting to create some of the type annotations.

Let me know what you think!

martindurant:
I am not opposed to typehints in the repo, but I'm not likely to put in any effort myself to add them.

alanhdu:
> I am not opposed to typehints in the repo, but I'm not likely to put in any effort myself to add them.

That's ok! I can try and carve out some time at work to try and contribute some patches for this then.

ianthomas23:
There are a number of issues about adding type annotations. I am closing this as a duplicate of #625 as that has the most relevant discussion.

PR #842, opened 2021-11-30
Add mypy to the test suite

alanhdu:
This is the first step towards https://github.com/fsspec/filesystem_spec/issues/834, where we add `mypy` to the tox config. For now, we do the minimal thing and only add some typehints to a couple global variables. I intend to send some follow-up PRs gradually adding typehints (and hopefully setting `check_untyped_def=True` at some point).


martindurant:
Importing typings annoys flake

PR #851, opened 2021-12-13
BF: do not hardcode /tmp -- use tempfile.gettempdir()

yarikoptic:
In this version it is incomplete since other locations needing change exist, e.g.:

	fsspec/implementations/ftp.py:        tempdir="/tmp",
	fsspec/implementations/sftp.py:        self.temppath = ssh_kwargs.pop("temppath", "/tmp")

so it would be nice to centralize.  Where should such "constant" be defined?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,5 @@ ENV/
 
 # docker artifacts
 .docker
+
+build/
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -1,4 +1,3 @@
-import os
 import uuid
 from ftplib import FTP, Error, error_perm
 from typing import Any
@@ -266,7 +265,7 @@ class FTPFile(AbstractBufferedFile):
         )
         if not autocommit:
             self.target = self.path
-            self.path = os.path.join(kwargs["tempdir"], str(uuid.uuid4()))
+            self.path = "/".join([kwargs["tempdir"], str(uuid.uuid4())])
 
     def commit(self):
         self.fs.mv(self.path, self.target)
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -1,3 +1,4 @@
+import tempfile
 import uuid
 from ftplib import FTP, Error, error_perm
 from typing import Any
@@ -21,7 +22,7 @@ class FTPFileSystem(AbstractFileSystem):
         password=None,
         acct=None,
         block_size=None,
-        tempdir="/tmp",
+        tempdir=tempfile.gettempdir(),
         timeout=30,
         **kwargs,
     ):
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -1,5 +1,4 @@
 import os
-import tempfile
 import uuid
 from ftplib import FTP, Error, error_perm
 from typing import Any
@@ -23,7 +22,7 @@ class FTPFileSystem(AbstractFileSystem):
         password=None,
         acct=None,
         block_size=None,
-        tempdir=str(tempfile.gettempdir()),
+        tempdir=None,
         timeout=30,
         **kwargs,
     ):
@@ -56,7 +55,7 @@ class FTPFileSystem(AbstractFileSystem):
         super(FTPFileSystem, self).__init__(**kwargs)
         self.host = host
         self.port = port
-        self.tempdir = tempdir
+        self.tempdir = tempdir or "/tmp"
         self.cred = username, password, acct
         self.timeout = timeout
         if block_size is not None:
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -1,3 +1,4 @@
+import os
 import tempfile
 import uuid
 from ftplib import FTP, Error, error_perm
@@ -266,7 +267,7 @@ class FTPFile(AbstractBufferedFile):
         )
         if not autocommit:
             self.target = self.path
-            self.path = "/".join([kwargs["tempdir"], str(uuid.uuid4())])
+            self.path = os.path.join(kwargs["tempdir"], str(uuid.uuid4()))
 
     def commit(self):
         self.fs.mv(self.path, self.target)
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -23,7 +23,7 @@ class FTPFileSystem(AbstractFileSystem):
         password=None,
         acct=None,
         block_size=None,
-        tempdir=tempfile.gettempdir(),
+        tempdir=str(tempfile.gettempdir()),
         timeout=30,
         **kwargs,
     ):
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -43,7 +43,7 @@ class SFTPFileSystem(AbstractFileSystem):
         if self._cached:
             return
         super(SFTPFileSystem, self).__init__(**ssh_kwargs)
-        self.temppath = ssh_kwargs.pop("temppath", tempfile.gettempdir())
+        self.temppath = ssh_kwargs.pop("temppath", str(tempfile.gettempdir()))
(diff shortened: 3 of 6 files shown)


martindurant:
Sounds like a config option. We currently have implementation-specific config, but not general. That should be fixed.

You will see that the three instances of this are doing the same thing (writing the file to local until it is committed), so there is probably some code deduplication that could happen.

Simply using tempfile in each place would be fine with me for now.

yarikoptic:
ok, rewrote my commit, and adjusted commit message to include

    This is just a temporary fix, since
    
    - location of such directory should be centralized better (e.g. via config)
    
    - how it could be specified should be unified (now WebHDFS has is just as a
      class level attribute, while FTPFileSystem as an option tempdir in signature,
      and SFTPFileSystem completely hides it from signature but pops temppath kwarg
      and binds as .temppath)

martindurant:
Thank you, and I agree with those statements.

Note that something went wrong with one of the windows paths in testing

yarikoptic:
yeah, most likely that "real" tmpdir on windows is not pointing to `/tmp` but somewhere else (may be on another drive), thus leading to
```
>           raise error_perm(resp)
E           ftplib.error_perm: 550 "/" points to a path which is outside the user's root directory.
``` 
... actually stderr might be giving a possible reason/answer:
```
2021-12-13T16:43:07.9818449Z ---------------------------- Captured stderr call -----------------------------
2021-12-13T16:43:07.9819064Z [I 2021-12-13 16:40:39] 127.0.0.1:50399-[] FTP session opened (connect)
2021-12-13T16:43:07.9819404Z 
2021-12-13T16:43:07.9819875Z [I 2021-12-13 16:40:39] 127.0.0.1:50399-[user] USER 'user' logged in.
2021-12-13T16:43:07.9820206Z 
2021-12-13T16:43:07.9820929Z [I 2021-12-13 16:40:39] 127.0.0.1:50399-[user] MKD C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\test_transaction0\tmp 257
2021-12-13T16:43:07.9821574Z 
2021-12-13T16:43:07.9822325Z [I 2021-12-13 16:40:39] 127.0.0.1:50399-[user] RMD C:\Users\RUNNER~1\AppData\Local\Temp\36185765-3152-4d9e-883f-a3569179bfb9 550 '"/" points to a path which is outside the user\'s root directory'
```
so you can spot difference in full vs abbreviated paths: `C:\Users\runneradmin` vs `C:\Users\RUNNER~1`.  I remember we had something similar in DataLad. May be @adswa could shine the light?

FTR test was `test_transaction` . I might try to reproduce locally in VM later on.

martindurant:
> "/" points to a path

actually suggests to me that either the temporary directory (which should certainly be writable) is not making it to the local path, or that mixing "/" and "\" is doing something weird.

adswa:
Hey, I lack context, and I'm also not an expert on Windows. I just spent a few fun hours in resolving a big pile of problems with temporary directories on Windows. I can only dump a quick collection of thoughts - please ignore them if they're not applicable or don't make sense. Looking at https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir I can't think of a reason why this function would return ``/``. Its docs say 

> Python searches a standard list of directories to find one which the calling user can create files in. The list is:
> * The directory named by the TMPDIR environment variable.
> * The directory named by the TEMP environment variable.
> * The directory named by the TMP environment variable.
> * A platform-specific location:
>    * On Windows, the directories C:\TEMP, C:\TMP, \TEMP, and \TMP, in that order.
>    * On all other platforms, the directories /tmp, /var/tmp, and /usr/tmp, in that order.
>
> As a last resort, the current working directory.

Those env variables with highest precedence are generally defined as far as I know, and should point to something like ``C:\<user>\AppData\Local\Temp``. I'm not aware that any other of those choices could become ``/``. On a real Windows box, it looks like this:
```
Python 3.9.9 (tags/v3.9.9:ccb0e6a, Nov 15 2021, 18:08:50) [MSC v.1929 64 bit (AMD64)] on win32
>>> import tempfile
>>> tempfile.gettempdir()
'C:\\Users\\adina\\AppData\\Local\\Temp
```
 
 I took a look into the [source code of gettempdir]( 
 https://github.com/python/cpython/blob/a62be77266b1beadd42d4952186332bc0847b7d6/Lib/tempfile.py#L156-L171), and here is what they do

 ```
 def _candidate_tempdir_list():
    """Generate a list of candidate temporary directories which
    _get_default_tempdir will try."""
    dirlist = []

    # First, try the environment.
    for envname in 'TMPDIR', 'TEMP', 'TMP':
        dirname = _os.getenv(envname)
        if dirname: dirlist.append(dirname)

    # Failing that, try OS-specific locations.
    if _os.name == 'nt':
        dirlist.extend([ _os.path.expanduser(r'~\AppData\Local\Temp'),
                         _os.path.expandvars(r'%SYSTEMROOT%\Temp'),
                         r'c:\temp', r'c:\tmp', r'\temp', r'\tmp' ])
 ```
 
The one potential problematic thing I am aware of is that ``expanduser`` since Python 3.8 relies on the ``$USERPROFILE`` variable, and if this variable is altered in the test setup somehow things could become confusing. Given that knowledge, I also noticed that there are two different users (``RUNNER~1`` and ``runneradmin``) but that may be because I lack all context here and two users is how it should be. But in case those users look funny, maybe there's something involving the env vars ``USERPROFILE``, ``HOMEDRIVE`` or ``HOMEPATH`` variables - from expandusers [docs](https://docs.python.org/3/library/os.path.html#os.path.expanduser):
> On Windows, USERPROFILE will be used if set, otherwise a combination of HOMEPATH and HOMEDRIVE will be used. 

My guess is in the fact that I spotted a mix of slashes in the logs ([here](https://github.com/fsspec/filesystem_spec/runs/4509089874?check_suite_focus=true#step:4:987))

```
 cmd = 'STOR C:\\Users\\RUNNER~1\\AppData\\Local\\Temp/36185765-3152-4d9e-883f-a3569179bfb9'
 ```
which looks a bit like something patched this sha (?) at the end of the tempdir, maybe an ``os.path.join``, e.g. https://stackoverflow.com/questions/16333569/mixed-slashes-with-os-path-join-on-windows? 
I just tried and such a mixed-slash path resolves when I give them to a ``CMD`` on Windows 10, but maybe some other tool gets caught off guard by it?

Sorry for the loose collection of thoughts, maybe there was something helpful in it.

**Comment by martindurant** [comment]
> I just tried and such a mixed-slash path resolves when I give them to a CMD on Windows 10

If I remember right, python itself detects whether a path is "posix-like" on windows, and converts as needed. That might be where this breaks. For a local path, we should be using os.path.join, not "/".join, unless we have first explicitly converted any possibly-windows path to be posixy.

**Comment by yarikoptic** [comment]
Thank you @adswa -- you helped me to see obvious (with a mix of `\` and `/`).  Fixed some spots for that - let's see if it would be of any help.

**Comment by yarikoptic** [comment]
FTR: unfortunately f7b6f3fe607c454f600d321250fb3d78140352f9 provided no description, so I was confused on "why?" since I was certain we should be receiving a `str`:
```
$> python-all -c 'import tempfile; print(type(tempfile.gettempdir()))'
Testing using /usr/bin/python2.7
<type 'str'>
Testing using /usr/bin/python3.6
<class 'str'>
Testing using /usr/bin/python3.8
<class 'str'>
Testing using /usr/bin/python3.9
<class 'str'>
Testing using /usr/bin/python2.7-dbg
<type 'str'>
Testing using /usr/bin/python3.9-dbg
<class 'str'>
```
Apparently , according to docs https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir  

    Changed in version 3.10: Always returns a str. Previously it would return any tempdir value
    regardless of type so long as it was not None.

so indeed could have been not a `str` if `tempdir` was misassigned some other data type.  So, I guess `str()`ing might indeed be useful.  But I wonder if you encountered non-str value in wild life @martindurant ?

PR #853, opened 2021-12-13
Update smb.py

rolveb:
Default port to smbclient is 445 not None. Example will now work without explicitly specifying port as int.

### DIFF
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -119,7 +119,7 @@ class SMBFileSystem(AbstractFileSystem):
             self.host,
             username=self.username,
             password=self.password,
-            port=self.port,
+            port=445 if self.port is None else self.port,
             encrypt=self.encrypt,
             connection_timeout=self.timeout,
         )


fsspec/implementations/smb.py
```diff
@@ -119,7 +119,7 @@ def _connect(self):
             self.host,
             username=self.username,
             password=self.password,
-            port=self.port,
+            port=445 if self.port is None else self.port,
```
martindurant, line 122: Perhaps 
```
self.port = 445 if port is None else port
```
is clearer, or set a class attribute `port = 445`?

rolveb: Yes both works.  

Also default username and password should be retrived from fsspec.config.conf if set using configuration system:
```bash
export FSSPEC_smb_username = 'someuser'
export FSSPEC_smb_password = 'somepassword'
```
If set they are available as:
```python
fsspec.config.conf['smb']['username']
fsspec.config.conf['smb']['password']
```
so to retrieve something like this:

```python
username = self.username
if username is None and ('smb' in fsspec.config.conf and 'username' in fsspec.config.conf['smb']): username = fsspec.config.conf['smb']['username']
```
and similar for password

A workaround is however to set defaults before initializing filesystem
```python
smbclient.ClientConfig(username=username, password=password) 
```

Not sure what is the best logic here.

Best,

martindurant: You don't need to explicitly look in the conf, the kwargs are set at instantiation automatically, if the user doesn't provide overrides.

Setting the smb config is a perfectly fine too, could be mentioned in the docstring.

PR #855, opened 2021-12-14
Fix a bug in `CachingFileSystem.pop_from_cache()`

jwodder:
`pop_from_cache()` would crash if asked to pop a path that wasn't in the cache.  This PR fixes that.


martindurant:
OK, yes this makes sense. Do you think `pop` should accept an extra argument for the default return value when the key doesn't exist?

jwodder:
@martindurant By `pop`, do you mean `_check_file()`?  Either way, I don't see the need.

martindurant:
Thank you

PR #859, opened 2021-12-16
asyn: ls: match signature

efiop:
To match AbstractFileSystem's ls.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -560,7 +560,7 @@ class AsyncFileSystem(AbstractFileSystem):
     async def _info(self, path, **kwargs):
         raise NotImplementedError
 
-    async def _ls(self, path, **kwargs):
+    async def _ls(self, path, detail=True, **kwargs):
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


martindurant:
Well, the existing signature was, technically, compatible :)

+1

fsspec/asyn.py
```diff
@@ -560,7 +560,7 @@ async def _exists(self, path):
     async def _info(self, path, **kwargs):
         raise NotImplementedError
 
-    async def _ls(self, path, **kwargs):
+    async def _ls(self, path, detail=True, **kwargs):
```
efiop, line 563: I also noticed that even though spec says `detail=True`, implementations are all over the place with using either `True` or `False`(this seems to be more common). @martindurant Is there some kind of intent here or is this something that we should ideally fix in spec or in implementations?

efiop: Context: i'm working on prefixfs and so going through the specs and noticing such small inconsistencies. The one mentioned above is causing default behavior of `ls()` to be implementation-specific depending on `detail` being False/True, which is a little inconvenient 🙂

PR #861, opened 2021-12-18
implementations: local: info for broken symlink

pared:
In the case of local filesystem, `info` called on broken symlink will result in FileNotFounError. If `info` is to imitate `ls` behavior, it should not fail in such use case.

Reproduction script:
```
#!/bin/bash

set -ex

rm -rf test
mkdir test 
pushd test

echo -e "
fs = LocalFileSystem()

print(fs.info('link'))
" >> print_info.py

echo data >> source
ln -s source link

python print_info.py

rm source

python print_info.py
```

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -63,16 +63,17 @@ class LocalFileSystem(AbstractFileSystem):
         return super().glob(path, **kwargs)
 
     def info(self, path, **kwargs):
+        t = None
         if isinstance(path, os.DirEntry):
             # scandir DirEntry
             out = path.stat(follow_symlinks=False)
             link = path.is_symlink()
+            if link:
+                t = "link"
             if path.is_dir(follow_symlinks=False):
                 t = "directory"
             elif path.is_file(follow_symlinks=False):
                 t = "file"
-            else:
-                t = "other"
             path = self._strip_protocol(path.path)
         else:
             # str or path-like
@@ -80,17 +81,16 @@ class LocalFileSystem(AbstractFileSystem):
             out = os.stat(path, follow_symlinks=False)
             link = stat.S_ISLNK(out.st_mode)
             if link:
-                out = os.stat(path, follow_symlinks=True)
+                out = os.lstat(path)
+                t = "link"
             if stat.S_ISDIR(out.st_mode):
                 t = "directory"
             elif stat.S_ISREG(out.st_mode):
                 t = "file"
-            else:
-                t = "other"
         result = {
             "name": path,
             "size": out.st_size,
-            "type": t,
+            "type": t or "other",
             "created": out.st_ctime,
             "islink": link,
         }
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -508,8 +508,8 @@ def test_linked_files(tmpdir):
 
     fs = LocalFileSystem()
     assert fs.info(fn0)["type"] == "file"
-    assert fs.info(fn1)["type"] == "file"
-    assert fs.info(fn2)["type"] == "file"
+    assert fs.info(fn1)["type"] == "link"
+    assert fs.info(fn2)["type"] == "link"
 
     assert not fs.info(fn0)["islink"]
     assert fs.info(fn1)["islink"]
@@ -526,6 +526,11 @@ def test_linked_files(tmpdir):
     of = fsspec.open(fn2, "rb")
     with of as f:
         assert f.read() == data
+
+    os.unlink(fn0)
+
+    assert fs.info(fn1)["size"] == 0
+    assert fs.info(fn2)["size"] == 0
 
 
 def test_linked_files_exists(tmpdir):
@@ -548,7 +553,7 @@ def test_linked_files_exists(tmpdir):
 
     os.unlink(origin)
 
-    assert not fs.exists(copy_file)
+    assert fs.exists(copy_file)
     assert fs.lexists(copy_file)
 
     os.unlink(copy_file)
@@ -577,8 +582,8 @@ def test_linked_directories(tmpdir):
 
     fs = LocalFileSystem()
     assert fs.info(subdir0)["type"] == "directory"
-    assert fs.info(subdir1)["type"] == "directory"
-    assert fs.info(subdir2)["type"] == "directory"
+    assert fs.info(subdir1)["type"] == "link"
+    assert fs.info(subdir2)["type"] == "link"
 
     assert not fs.info(subdir0)["islink"]
     assert fs.info(subdir1)["islink"]


fsspec/implementations/local.py
```diff
@@ -63,51 +63,49 @@ def glob(self, path, **kwargs):
         return super().glob(path, **kwargs)
 
     def info(self, path, **kwargs):
+        t = None
         if isinstance(path, os.DirEntry):
             # scandir DirEntry
             out = path.stat(follow_symlinks=False)
             link = path.is_symlink()
+            if link:
+                t = "link"
```
pared, line 72: Looking at `smb` and `sftp` I assumed `link` type is desired. This kind of makes `islink` field obsolete, but I guess removing it would result in breaking the api.

martindurant:
Unfortunately, this seems to have broken something on windows in `test_linked_files` - the windows "link" claims to have no size.

I am uncertain of replacing `exists` with the os.path version, which will return `False` for a broken symlink. It could be argued that the path of the ilnk does exists, since we are able to get information for it after all.

pared:
> Unfortunately, this seems to have broken something on windows in test_linked_files - the windows "link" claims to have no size.

Researching that, regretfully I have a problem finding any official docs saying how big symlink should be on windows.

> I am uncertain of replacing exists with the os.path version, which will return False for a broken symlink. It could be argued that the path of the ilnk does exists, since we are able to get information for it after all.

I agree, I was not sure what should `exists` return, but looking at `lexists` I assumed it should behave as `os.path.exists/lexists` - failing `fsspec/implementations/tests/test_local.py::test_linked_files_exists` led me to that conclusion. However I would agree that link, even if broken, does exist. I will change that.

martindurant:
> I have a problem finding any official docs saying how big symlink should be on windows.

I believe windows does not have posix-style symlinks, but something else, so I am not surprised that this fails. In any case, the size of the symlink itself is not that interesting, so you could apply this part of the test to posix-only, or just assert that the value is different than the size of the file pointed at.

fsspec/implementations/local.py
```diff
@@ -63,46 +63,41 @@ def glob(self, path, **kwargs):
...
         }
         for field in ["mode", "uid", "gid", "mtime"]:
             result[field] = getattr(out, "st_" + field)
         if result["islink"]:
             result["destination"] = os.readlink(path)
-            try:
-                out2 = os.stat(path, follow_symlinks=True)
-                result["size"] = out2.st_size
-            except IOError:
-                result["size"] = 0
```
efiop, line 105: Hm, could you elaborate on why we are removing this size? The size of the link itself (and not the file it points to) seems useless to me, unless I'm missing something.

fsspec/implementations/tests/test_local.py
```diff
@@ -508,16 +508,19 @@ def test_linked_files(tmpdir):
...
     assert fs.info(fn2)["islink"]
 
     assert fs.info(fn0)["size"] == len(data)
-    assert fs.info(fn1)["size"] == len(data)
-    assert fs.info(fn2)["size"] == len(data)
+
+    if not sys.platform == "win32":
+        # Posix symbolic link size is the lenght of the path it contains
+        assert fs.info(fn1)["size"] == len(fn0)
+        assert fs.info(fn2)["size"] == len(fn0)
```
efiop, line 523: Previously we would get the size of the file it points to, which is what I would expect. And now we are getting the size of the link itself, which is platform-specific and I can't even think of a scenario where anyone would care about it.

fsspec/implementations/tests/test_local.py
```diff
@@ -527,6 +530,11 @@ def test_linked_files(tmpdir):
     with of as f:
         assert f.read() == data
 
+    # After unlinking source we should still get the info about link
+    os.unlink(fn0)
+    assert fs.info(fn1)
+    assert fs.info(fn2)
```
efiop, line 536: This is probably where you wanted to check that the size of the broken link, that should clearly be not equal to the size of data.

fsspec/implementations/local.py
pared: https://filesystem-spec.readthedocs.io/en/latest/api.html?highlight=info#fsspec.archive.AbstractArchiveFileSystem.info
Accordign to docs, `info` is supposed to return ls-like information - which returns link size. 
But I agree, that this is probably not that useful. I will roll it back

efiop: > Returns a single dictionary, with exactly the same information as ls would with detail=True.

> The default implementation should calls ls and could be overridden by a shortcut. kwargs are passed on to `ls().

@pared There might be a misunderstanding here. The docs talk about `fs.ls`, not CLI ls. Unless I'm missing something.

fsspec/implementations/local.py
efiop: @pared But how can I now tell that this link is pointing to a file or a directory?

fsspec/implementations/tests/test_local.py
```diff
@@ -577,8 +582,8 @@ def test_linked_directories(tmpdir):
 
     fs = LocalFileSystem()
     assert fs.info(subdir0)["type"] == "directory"
-    assert fs.info(subdir1)["type"] == "directory"
-    assert fs.info(subdir2)["type"] == "directory"
```
efiop, line 581: This is also a desired information though. Link is an fs implementation detail, so users should be able to tell that this is a directory first and also that this is a link. So it looks like it should stay as `type==directory` here and so `islink` is not actually obsoleted.

fsspec/implementations/local.py
```diff
@@ -63,34 +63,34 @@ def glob(self, path, **kwargs):
...
             if link:
-                out = os.stat(path, follow_symlinks=True)
+                out = os.lstat(path)
+                t = "link"
             if stat.S_ISDIR(out.st_mode):
                 t = "directory"
             elif stat.S_ISREG(out.st_mode):
                 t = "file"
-            else:
-                t = "other"
```
efiop, line 89: The path might be a special dev file (e.g. see S_ISBLK/S_ISFIFO/etc) still, so we shouldn't remove this.

efiop (changes requested):
Taking a closer look at this, I think `info` behaves correctly here (you've made a good point about smb and sftp, it looks like they shouldn't use `type=link`). The confusion is coming from https://github.com/iterative/dvc/issues/7120 where we do the listing ourselves and it is us who should simply handle broken symlink case gracefully (we do that in a few other places already).

fsspec/implementations/local.py
pared: It would be handled in return statement

pared:
So I have been thinking about our discussion and here are my thoughts:

- If we are returning `size` of a target in case of `link` then calling `lstat` does not make sense, because as a result, when calling `info(symlink)` we will get a mix of link and file properties (eg modification times from link and size from target file)

- If we agree that `info` should return target properties, then we should keep previous behaviour (calling `stat` with `follow_symlink=True`) and properly handle potential error that might occur in case of dangling symlink.

- We should get rid of `link` type and just provide `islink` field (create issue for `smb` and `sftp`)

So the question is:

1. Do we agree that info should provide target properties? 

Looking at how it was working before, it seems to me that yes and I should just handle potential error in here:
https://github.com/fsspec/filesystem_spec/blob/bd932c30f6998532002885aa0b98a1ed6f18bd14/fsspec/implementations/local.py#L83

efiop:
@pared I agree with you, but let's narrow it down a bit by going from a particular problem we are trying to solve here: we get the link in `fs.ls()`(or walk that is using it) but following `fs.info()` on it errors out with `FileNotFoundError` because the link is broken. This corresponds to `os.scandir` + `os.stat`(following symlinks), and so considering that there is no `linfo` in spec, this should be handled by us (users) and localfs is not at fault here at all and behaves correctly enough for what we use it for so far. So we just need to catch this exception ourselves in user code.

Now if we go back to improvements, I agree with you that `info()` should return stuff about the link target, but it also means that the spec needs to formalize `linfo`(or `follow_symlink` flag for `info`) for when you want to get info about the link itself. This will also allow sftp/smb/etc to switch to this and for all of those (including local) to avoid doing redundant stat calls and stuff like `readlink` which is also expensive. This will be important for us in the future to switch to fsspec's localfs from our own implementation. Unless I'm missing some original intent here. CC @martindurant , would appreciate your thoughts on this.

efiop:
Ok, closing this for now, since this approach doesn't help us and potential improvements are a separate subject.

PR #863, opened 2021-12-18
Add mimetypes

### LINKED ISSUE #862: MimeTypes
Hello,

Is it possible to synchronize with [mimetypes](https://docs.python.org/3/library/mimetypes.html) package ?

For compressions with encodings, extensions

Infer compression using this package

Br
Nicolas

Platob:
add mimetypes extensions

```python
import mimetypes
mimetypes.guess_type("test.csv.gz") # ('application/vnd.ms-excel', 'gzip')
```

Fixes #862

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -1,6 +1,7 @@
 """Helper functions for a standard streaming compression API"""
 from bz2 import BZ2File
 from zipfile import ZipFile
+from mimetypes import encodings_map
 
 import fsspec.utils
 from fsspec.spec import AbstractBufferedFile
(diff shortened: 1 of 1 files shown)


martindurant:
I am generally against merging this for the following reasons:
- it introduces the `mimetypes` dependency, whereas fsspec is very careful to be as dependency-free as possible
- the user could very easily do this when they need the conversion
- I find that guessing from file names is actually pretty error-prone (a CSV file is *not* associated with excel!).

Platob:
Ok i understand thank you

PR #864, opened 2021-12-19
async: add set_session

efiop:
This is used in multiple implementations (s3fs, http, etc) that allow
_external_ async use and is documented in fsspec docs (but the example
is http-specific), so let's add it to the spec, so that we could officially
wrap it in #745

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -307,6 +307,14 @@ class AsyncFileSystem(AbstractFileSystem):
             raise RuntimeError("This class is not fork-safe")
         return self._loop
 
+    async def set_session(self, *args, **kwargs):
+        """Establish a connection.
+        Returns
+        -------
+        Session to be closed later with await .close()
+        """
+        raise NotImplementedError
+
     async def _rm_file(self, path, **kwargs):
         raise NotImplementedError
 


efiop:
@martindurant I suppose you guys were hesitant to formally include it into the spec because not all async filesystems support async use. But looks like this is de-facto the standard convention among those that do. Happy to learn about any nuances I could've missed here 🙂

martindurant:
I generally agree with the sentiment, and we have indeed made a few of these; even though it turns out is can be possible to work around it.

I don't think this should be NotImplemented, as it's conceivable that some implementation doesn't need it, even if it's async (yes, I know they can override it, but this is our place to show intent). I with it had been called `_init_async`, or somesuch, since there may be future backends that don't have a "session".

I suppose if we do this we also need to have `async def close()`, since it's explicitly mentioned ?

efiop:
>  even though it turns out is can be possible to work around it.

@martindurant Do you mean some dynamic session creation when an async method is being accessed and then caching it (e.g. fs._session) until it explicitly `fs.close()`ed? Just making sure I follow.

> I don't think this should be NotImplemented, as it's conceivable that some implementation doesn't need it, even if it's async (yes, I know they can override it, but this is our place to show intent). I wish it had been called _init_async, or somesuch, since there may be future backends that don't have a "session".

Oh, I didn't notice this before, makes sense now. I guess nothing stops us from introducing `_init_async` and calling `set_session` obsolete.

> I suppose if we do this we also need to have async def close(), since it's explicitly mentioned ?

You mean if we go with automatic `fs._session` creation, right? Because the docs only have `session.close()`, not `fs.close()`, if I'm not missing anything.

martindurant:
> @martindurant Do you mean some dynamic session creation when an async method is being accessed and then caching it (e.g. fs._session)

Yes, you can do the initial async setup on whenever the first async call is made, and so remove the explicit need to call `await fs.set_session()`. I am told explicit is better than implicit, though... 

>  I guess nothing stops us from introducing _init_async and calling set_session obsolete.

Not sure which is better. If all of them do the same thing of creating a session, then maybe it's fine, for now. We can always consolidate later.

> You mean if we go with automatic fs._session creation, right? Because the docs only have session.close(), not fs.close(), if I'm not missing anything.

Sorry, you are right, I misread. It might be useful to have an async dealloc thing too, but if we don't use one yet, then no hurry. We do care about what happens at garbage collection time, trying to clean up any async client object. This is starting to make the whole filesystem like an async context - which could be an optional improvement, but I don't think we should force the user that way.

efiop:
> This is starting to make the whole filesystem like an async context - which could be an optional improvement, but I don't think we should force the user that way.

I also was craving for a context manager for this :)

> Yes, you can do the initial async setup on whenever the first async call is made, and so remove the explicit need to call await fs.set_session(). I am told explicit is better than implicit, though...

But sync spec also doesn't really define connect/disconnect methods, so looks like implicit session creation (with some weakref.finalize magic, I suppose?) would be more in line with the current fsspec behavior. Unless I'm missing something.

Regarding this PR, if we can agree on the `set_session` alternative - I would close this PR and just propagate new approach through existing implementations. We can keep `set_session` wrapper in `dirfs` for now.

martindurant:
> looks like implicit session creation (with some weakref.finalize magic, I suppose?) would be more in line with the current fsspec behavior

More inline, but not necessarily right! Occasionally it's good to refactor and improve these things. Which is why its great to have you looking at the code and making such improvements!

efiop:
@martindurant Ah, I see, thanks for clarifying! Created #865 for further discussion.

Closing this PR, will keep `set_session` in `dirfs` for now.

Issue #773, opened 2021-09-24
[RFC] persistent filesystem id


efiop:
[fsspec currently uses](https://github.com/intake/filesystem_spec/blob/0714e52c1ce1f500f9bd88320b4e22da2689e45a/fsspec/spec.py#L67) `fs._fs_token` that is based on (among a few other things) thread id and args/kwargs passed to the filesystem constructor. This token is used to compare filesystems for caching.

In dvc, we have persistent storage for metadata associated with a particular file on a particular filesystem (e.g. we are saving that `path/to/file` has an `md5` that is `123456` and use `fs.checksum()` for verification when loading next time).  Currently we only store this for local files, but we need to support that for remote files as well, so in our experimentation, we've introduced `fs_id`, which is a hash of a config dictionary that was used for creating a specific fs instance, which is very similar to `_fs_token`, except that it doesn't include `tid`, so it is persistent across sessions.

In the simplest case (and probably by default), `fs_id`(open for discussion on the naming) could just default to the hash of the whole config, but it is clear that many filesystems could do better than that. For example:

* for local, it might be something like a hash of `b"local"`. For other scenarios where more granularity is needed, this might be based on /etc/machine-id or alternative to be able to distinguish separate machines. Also, fsspec itself doesn't distinguish real filesystems mounted to the root fs in the system, so that might also be a useful one to identify (e.g. maybe based on real fsid from fsstab), here PrefixFileSystem might come in handy too.
* for s3, `fs_id` should really depend on things like endpoint/region/bucket/etc, and not on user creds.
* for ssh, `fs_id` should depend on base url, but not on user creds.

`fs_id` for a particular filesystem could change its algorithm in the future (e.g. if developers forgot to include some important parameter), so users should be ready for this and have some mechanism to recover from this. E.g. for dvc it will simply mean that old cached metadata will be effectively dropped, which is just a temporary inconvenience to recompute md5s for requested files.

I think this functionality could be handy for fsspec users in scenarios where they want to be able to identify filesystems across sessions and maybe have persistent storage of metadata. We've used to use hash of `url` (e.g. s3://bucket/path) in some cases, but url is not able to account for all important parameters, so a standalone `fs_id` would be much better. If there is interest in adopting it, we'll be happy to contribute PRs for it. Would really appreciate any comments/suggestions.

martindurant:
[...]

PR #867, opened 2021-12-21
Allow ReferenceFS to refer to multiple FSs

martindurant:


martindurant:
cf https://github.com/fsspec/kerchunk/issues/106

fsspec/implementations/reference.py
```diff
@@ -172,11 +199,14 @@ def __init__(
         if remote_protocol is None:
             remote_protocol = target_protocol
 
-        self.fs = filesystem(remote_protocol, loop=loop, **(remote_options or {}))
+        fs = fs or filesystem(remote_protocol, loop=loop, **(remote_options or {}))
+        self.fss[remote_protocol] = fs
```
efiop, line 203: What if there are multiple filesystems with the same protocol but different locations? E.g. s3 with different endpoints, ssh with different hosts,etc ?

Seems like ideally something like https://github.com/fsspec/filesystem_spec/issues/773 should be used here instead.

martindurant: You are right, this is a problem, and a way of referring to unique combinations of arguments is very useful. But even if we can specify these filesystems, how would the individual references point to them? They have URLs like "s3://bucket/.." .

martindurant: I am thinking perhaps we could move to basing the FS choice on a *prefix* rather than protocol, so that "s3://bucket1/..." might be a different endpoint to "s3://bucket2/..." . This seems achievable, but also complex for a user to understand and wield. Perhaps that doesn't matter is the details of the options for each filesystem is not really something the user gets to see.

martindurant:
@manzt , I was thinking to use this idea to prototype a workflow in which we made a multiscale pyramid of some big dataset, putting the downsampled data in a different storage backend. With this PR, it would be possible to build a combined virtual FS with everything where it needs to be and pass it to napari or whatever. The original data doesn't need to be copied, or zarr.

manzt:
Ah interesting, so a user can specify a file system per protocol used in the references? I guess this would also allow for delineation of multiple FSs using the same protocols. E.g. `"a://"` and `"b://"` could map to two different `LocalFileSystem` with different roots?

martindurant:
I was more simply thinking of a situation where the original data is on azure or s3, say, but we only wanted to store our downsampled data on gcs; the references themselves could be elsewhere again.

> I guess this would also allow for delineation of multiple FSs using the same protocols. E.g. "a://" and "b://" could map to two different LocalFileSystem with different roots?

Interesting idea, I hadn't quite thought of that. Yes, pretty sure you can pass your dict of filesystem objects with whatever keys you like and they map to the protocols. It would be tricky to represent this declaratively!
The thing might fail if the backend needs the actual protocol (http/s). For local this might not be that interesting, since you can refer to absolute paths anyway, and there are no parameters when making the instance.

(LocalFileSystem doesn't work from a root, but DirFileSystem in https://github.com/fsspec/filesystem_spec/pull/745 would, if it was merged)

fsspec/implementations/reference.py
martindurant: I am settling for protocols only for now, but the door is open to have FSs per prefix, if we can avoid expensive string matching (.startswith) at access time.

martindurant:
Marked as "ready", but the doc should be updated.

PR #869, opened 2021-12-22
Make put_file continue on zero-read

### LINKED ISSUE #868: fs.put_file flakily uploading 0 bytes
We're running into a flaky issue where using `fsspec` to upload data to WebHDFS is uploading 0 bytes. Printing the contents of the file before and after `fs.put_file()` confirms that the file on disk is non-empty. I have also found that this line is returning an empty string on the first (and only) iteration of the loop:

https://github.com/fsspec/filesystem_spec/blob/b45ae3eddc0e5aa032b25356197c491e0ca645f0/fsspec/spec.py#L813

This is happening in CI (Circle CI), so resource/disk constraints is not out of the question, but it's very odd that I can read the file before and after. Let me know what other information would be helpful to debug and I can try to get those to you.

martindurant:
Fixes #868

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -803,13 +803,13 @@ class AbstractFileSystem(up, metaclass=_Cached):
             return None
 
         with open(lpath, "rb") as f1:
-            callback.set_size(f1.seek(0, 2))
+            size = f1.seek(0, 2)
+            callback.set_size(size)
             f1.seek(0)
 
             self.mkdirs(self._parent(os.fspath(rpath)), exist_ok=True)
             with self.open(rpath, "wb", **kwargs) as f2:
-                data = True
-                while data:
+                while f2.tell() < size:
                     data = f1.read(self.blocksize)
                     segment_len = f2.write(data)
                     callback.relative_update(segment_len)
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -803,13 +803,13 @@ def put_file(self, lpath, rpath, callback=_DEFAULT_CALLBACK, **kwargs):
...
-            callback.set_size(f1.seek(0, 2))
+            size = f1.seek(0, 2)
+            callback.set_size(size)
             f1.seek(0)
 
             self.mkdirs(self._parent(os.fspath(rpath)), exist_ok=True)
             with self.open(rpath, "wb", **kwargs) as f2:
-                data = True
-                while data:
+                while f2.tell() < size:
```
brandon-leapyear, line 812: why not `f1.tell() < size`? If `f2.write()` silently failed to write a chunk, then this would go into an infinite loop where `f1.read()` thinks it's done but `f2` never gets the whole file

martindurant: Hm, that really shouldn't happen, but I'd be OK with your suggestion. Do you want to make that follow-on PR?

Issue #380, opened 2020-08-21
md5 incompatible with FIPS enabled OS


klucar:
For [FIPS ](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security_guide/sect-security_guide-federal_standards_and_regulations-federal_information_processing_standard) enabled operating systems (think AWS GovCloud), md5 isn't allowed because FIPS deems it insecure. (even though it's just being used as a local cache lookup key) 

The below line of code fails on fips enabled systems because Python's md5 uses the system openSSL library, and FIPS blocks md5.

https://github.com/intake/filesystem_spec/blob/f29c59c2c9a8b37cf8a9ed5e3956e4a7708e235c/fsspec/utils.py#L278

FIPS will allow sha256, so that is a potential solution, or some other cache index mechanism can be developed.

cpbrophy:
63a668fb4a60d41efc66c5f6445b6a500d5e4396 appears to undo this fix.

martindurant:
Is there a middle ground, something that is fast for hashing but not falling foul of security?

cpbrophy:
Something like [this approach](https://code.djangoproject.com/ticket/28401) that django came up with might be a middle ground?

martindurant:
A little local testing actually suggests blake2b with digest_size=8 is faster than both sha and hd5, and is considered secure. Will that do?

martindurant:
I assume the `zlib` (numerical) hashes are also forbidden?

cpbrophy:
Well annoyingly it turns out that even though blake2 is considered to have a ["large security margin"](https://nvlpubs.nist.gov/nistpubs/ir/2012/NIST.IR.7896.pdf) it [still isn't FIPS approved](https://research.kudelskisecurity.com/2017/03/06/why-replace-sha-1-with-blake2/), so trying that out didn't work. For whatever reason though, there's no error using `zlib.crc32` so go figure.

cpbrophy:
Just my opinion, but this seems like a less silly workaround than going the zlib route:

```python
def tokenize(*args, **kwargs):
    """Deterministic token

    (modified from dask.base)

    >>> tokenize([1, 2, '3'])
    '9d71491b50023b06fc76928e6eddb952'

    >>> tokenize('Hello') == tokenize('Hello')
    True
    """
    if kwargs:
        args += (kwargs,)
    try:
        return blake2b(str(args).encode()).hexdigest()
    except ValueError:
[...]

PR #872, opened 2021-12-28
MD5 for FIPS

martindurant:


martindurant:
Suggested in https://github.com/fsspec/filesystem_spec/issues/380#issuecomment-982927757

isidentical (approved):

PR #889, opened 2022-01-15
drop python 3.6

efiop:
3.6 has reached the EOL.

fsspec (along with pydata) dropped support for it early 2021, but we brought
it back for dvc in #575.


martindurant:
Would you say that there's still some ongoing cost to supporting py36?

martindurant:
OK, sounds like good justification to me. Let's keep this up for a day for any dessent.

PR #891, opened 2022-01-20
Seek by offset before write for fuse

u8sand:
Writes after seeks on the fuse mounted filesystem fail, this fixes it and things like `pd.HDFStore` operability on these mounted filesystems.


isidentical:
If you can, it would be really nice if we can add a test for that so future regressions won't happen though can see that this is a clear bug.

u8sand:
@isidentical I wrote the test but discovered another bug.. the read isn't working on the memory provider (added a commented out assertion) which is odd since BytesIO seems to do seeking just fine.

isidentical (approved):
LGTM

PR #892, opened 2022-01-21
Boost references speed 

martindurant:
Fixes #118


martindurant:
Thanks for the concrete test case, @dmedv , I think things should be much faster now, both when opening the mapper and when accessing the data.

PR #893, opened 2022-01-24
Dictionary changed size during iteration

fbriol:
In multithreaded mode "self.store" can be modified.


martindurant:
Thank you @fbriol

PR #895, opened 2022-01-25
Add optional file-based listings caching

gutzbenj:
Dear @martindurant ,

this PR adds an additional file-based listings cache with the help of the beautiful diskcache[1]. This is relevant as the listings cache docs mention that the creation of certain large listings can be really expensive. 

Following changes are done:
- rename memdircache.py module to dircache.py to represent general diraching (listings caching) and add argument to choose from both (MemDirCache is set to default)
- rename DirCache class to MemDirCache
- introduce new FileDirCache that implements all abstractmethods using given methods/attributes of diskcache.Cache class[2] and places the cache at a given location (if wanted)
- parametrize listings cache tests to test both MemDirCache and FileDirCache
- adjust relevant documentation

[1] http://www.grantjenks.com/docs/diskcache/
[2] http://www.grantjenks.com/docs/diskcache/api.html#diskcache.Cache


amotl:
Dear Benjamin,

I want to thank you very much for tackling this improvement to fsspec. In order to shed more light on the background of this, I would like to reference https://github.com/earthobservations/wetterdienst/issues/243, where we are coming from.

This improvement solves a longstanding issue for us, as the dbmfile-based backend of [dogpile.cache](https://dogpilecache.sqlalchemy.org/) proved to be too brittle when operating the Wetterdienst library in concurrent workload situations. I used to work with both [Beaker] and [dogpile.cache] in the past and they really shine in environments where you can easily afford to run a proper database-based cache backend, like Memcached, Redis, or even MongoDB.

However, this is not the situation with Wetterdienst: It is mostly library- and commandline-driven, so we didn't want to impose the overhead of running a caching database on its users.

As discussed at https://github.com/earthobservations/wetterdienst/issues/243#issuecomment-840493802, I was very happy to learn about [DiskCache] by @grantjenks from you, as its [features](http://www.grantjenks.com/docs/diskcache/#features) promised it to be the absolute right thing to use in a library context.

On my ramblings
> It might even be suitable to accompany the fsspec-based caching implementation [fsspec/caching.py].

@martindurant answered:
> Certainly, fsspec would be happy to see alternative, more fully-featured caching schemes. Integration should generally be pretty simple.

In this manner, thank you again for making a start with this patch. I hope the integration will go along well.

With kind regards,
Andreas.

[Beaker]: https://beaker.readthedocs.io/
[DiskCache]: http://www.grantjenks.com/docs/diskcache/
[dogpile.cache]: https://dogpilecache.sqlalchemy.org/
[fsspec/caching.py]: https://github.com/fsspec/filesystem_spec/blob/2022.01.0/fsspec/caching.py

nbren12:
Should/is this feature disabled by default? I couldn't tell from the code what the default behavior is. I could see it being pretty confusing if not. E.g. an application that monitors for changes in a bucket using `fs.ls`. Users are always free to wrap operations in their own caching layer no? e.g.
```
from toolz import memoize
# in memory
ls = memoize(fsspec.ls)
# on disk
from joblib import Memory
cache = Memory(".cache")
ls = cache.cache(ls)
```

martindurant:
The dircache is used for more operations than simply caching the output of ls, so I don't think the simple approach above does everything you want.

nbren12:
I'm mainly just cautioning against caching by default...is it a default?

martindurant:
Yes, we cache file listings for remote filesystems, where getting such listings can be pretty expensive.

nbren12:
So this is the expected default behavior:
```
>>> fs.ls("gs://some-bucket")
[]
>>> fs.put("local_file", "gs://some-bucket")
>>> fs.ls("gs://some-bucket")
[]
```
I could imagine losing hours debugging something like that.

martindurant:
`put` is supposed to explicitly invalidate the parts of the cache it touches, so this is a bug to be fixed.

nbren12:
I'm not sure it is a bug. I didn't run the code above...I just meant it as an illustrative example. Suppose instead of `fsspec.put` someone used another tool or that they want to monitor changes made by other machines. The cache can easily become out of sync with reality.

martindurant:
Agreed, the cache can get out of sync when processes are happening elsewhere, so we have the ability to force refresh, set an expiry time or turn the listings cache off. However, the default is to cache for the duration of the python session, which seems (to me) to align with the most common use cases.

ci/environment-win.yml
```diff
@@ -20,3 +20,5 @@ dependencies:
   - python-libarchive-c
   - numpy
   - nomkl
+  - diskcache
+  - appdirs
```
amotl, line 24: I recently discovered that the successor library to `appdirs` is [`platformdirs`](https://pypi.org/project/platformdirs/).

gutzbenj:
Dear all,

finally updated the PR. Could you please give me a review?

amotl:
Hi Benjamin,

there is another discussion over at https://github.com/fsspec/filesystem_spec/pull/1127, which I wanted to make you aware about. It may happen that this patch will get rejected, or not merged in its current form.

If @martindurant agrees with my evaluation that it is indeed unlikely that it will get merged, I am thinking about shipping it on behalf of a separate utility package, in order to overcome the roadblock in making the recipe reusable beyond the [Wetterdienst](https://github.com/earthobservations/wetterdienst) package, where it is currently used.

What do you think about this proposal?

With kind regards,
Andreas.

docs/source/features.rst
```diff
@@ -181,15 +181,18 @@ Listings Caching
...
+:class:`fsspec.dircache.FileDirCache` to manage the listings.
+
+The type of cache that is used, can be controlled via the keyword ``listings_cache_type``
+that has to be one of `memdircache` or `filedircache`. The cache allows for time-based expiry
+of entries with the ``listings_expiry_time`` parameter, or LRU expiry with the ``max_paths``
+parameter. These can be set on any implementation instance that uses listings caching; or to
+skip the caching altogether, use ``use_listings_cache=False``. That would be appropriate
+when the target location is known to be volatile because it is being written to from other
+sources. If you want to use the file-based caching, you can also provide the argument
+``listings_cache_location`` to determine where the cache file is stored.
```
martindurant, line 195: Do you think there's anything that can be done at this stage to simplify all these parameters? Should we have a caching_options dict, or config options (or both) ?

fsspec/dircache.py
```diff
@@ -1,9 +1,13 @@
+import logging
 import time
 from collections.abc import MutableMapping
 from functools import lru_cache
+from pathlib import Path
 
+logger = logging.getLogger("fsspec")
```
martindurant, line 7: perhaps fsspec.dircache?

fsspec/dircache.py
```diff
@@ -93,6 +97,95 @@ def __iter__(self):
...
+        listings_expiry_time: int or float (optional)
+            Time in seconds that a listing is considered valid. If None,
+            listings do not expire.
+        listings_cache_location: str (optional)
+            Directory path at which the listings cache file is stored. If None,
+            an autogenerated path at the user folder is created.
+
+        """
+        import platformdirs
+        from diskcache import Cache
```
martindurant, line 129: Informative error on what to install if this import fails

fsspec/dircache.py
```diff
@@ -93,6 +97,95 @@ def __iter__(self):
...
+
+        listings_expiry_time = listings_expiry_time and float(listings_expiry_time)
+
+        if listings_cache_location:
+            listings_cache_location = Path(listings_cache_location) / str(
```
martindurant, line 134: Why would we use Path within a filesystem library :)

martindurant: Can diskcache handle fsspec paths? E.g., memory://cache would be a very similar solution to MemDirCache. Such an option might even be useful for semi-speedy remotes like smb:

fsspec/dircache.py
```diff
@@ -93,6 +97,95 @@ def __iter__(self):
...
+            logger.error(
+                f"folder for dircache could not be created at {listings_cache_location}"
+            )
+
+        self.cache_location = listings_cache_location
+
+        logger.info(f"Dircache located at {listings_cache_location}")
+
+        self._cache = Cache(directory=listings_cache_location)
+        self.use_listings_cache = use_listings_cache
```
martindurant, line 155: if use_listings_cache is False, should we end up here at all? Perhaps we need a NoopDirCache

fsspec/implementations/http.py
```diff
@@ -103,6 +103,8 @@ def __init__(
         request_options.pop("listings_expiry_time", None)
         request_options.pop("max_paths", None)
         request_options.pop("skip_instance_cache", None)
+        request_options.pop("listings_cache_type", None)
```
martindurant, line 106: This is problematic - where else might these kwargs pollute calls? Do we need to change the signatures of all backends?

fsspec/implementations/tests/test_http.py
```diff
@@ -80,9 +106,20 @@ def test_list_cache_with_expiry_time_purged(server):
     cached_items = h.dircache.get(server + "/index/")
     assert len(cached_items) == 1
 
+    h.dircache.clear()  # clean up
```
martindurant, line 109: Should be fixture, to make sure we clean up even on failure?

fsspec/spec.py
```diff
@@ -144,7 +144,16 @@ def __init__(self, *args, **storage_options):
         self._intrans = False
         self._transaction = None
         self._invalidated_caches_in_transaction = []
-        self.dircache = DirCache(**storage_options)
+
+        listings_cache_type = storage_options.get("listings_cache_type", "memdircache")
+        if listings_cache_type not in ("memdircache", "filedircache"):
```
martindurant, line 149: I would imagine a dircaches dict in which these are registered rather than hardcoded here. External packages might bring their own.

tox.ini
martindurant, line 1: This file is here by accident?

martindurant:
Given progress here, I think we can fix this up, and see whether it garners any interest.

docs/source/features.rst
gutzbenj: Dear @martindurant ,

for now I named it `listings_cache_options` to make clear what it refers to.

Probably options for [data] caching and listings caching should not be put together to keep it simple and because keywords e.g. "expiriy_time" are similarly named.

fsspec/asyn.py
```diff
@@ -297,15 +297,15 @@ class AsyncFileSystem(AbstractFileSystem):
     mirror_sync_methods = True
     disable_throttling = False
 
-    def __init__(self, *args, asynchronous=False, loop=None, batch_size=None, **kwargs):
+    def __init__(self, *args, asynchronous=False, loop=None, batch_size=None, listings_cache_options=None, **kwargs):
```
gutzbenj, line 300: Should we name it `dir_cache_options` instead to adhere with `MemoryDirCache/FileDirCache`?

fsspec/dircache.py
```diff
@@ -26,19 +32,15 @@ class DirCache(MutableMapping):
 
     def __init__(
         self,
-        use_listings_cache=True,
```
gutzbenj, line 29: A cache instance should probably not contain functionality that makes it not work.

fsspec/implementations/http.py
```diff
@@ -201,11 +201,13 @@ async def _ls_real(self, url, detail=True, **kwargs):
...
     async def _ls(self, url, detail=True, **kwargs):
-        if self.use_listings_cache and url in self.dircache:
+        listings_cache_disabled = self.dircache is None
+        if not listings_cache_disabled and url in self.dircache:
             out = self.dircache[url]
         else:
             out = await self._ls_real(url, detail=detail, **kwargs)
-            self.dircache[url] = out
+            if not listings_cache_disabled:
+                self.dircache[url] = out
```
gutzbenj, line 210: This would be the alternative to `use_listings_cache`. May not be as nice, but it seems more accurate.

fsspec/dircache.py
gutzbenj: The name is now taken from the file: `logging.getLogger(__name__)`

fsspec/implementations/tests/test_http.py
gutzbenj: After the latest update it now uses `tmp_path` which automatically cleans up after runtime.

fsspec/spec.py
gutzbenj: Changed this although right now the mapping happens in the `create_dircache` function. Also I created an enum for convenience but this could not be extend by users. What do you think?

fsspec/implementations/http.py
```diff
@@ -104,10 +108,6 @@ def __init__(
         # TODO: Maybe rename `self.kwargs` to `self.request_options` to make
         #       it clearer.
         request_options = copy(storage_options)
-        self.use_listings_cache = request_options.pop("use_listings_cache", False)
-        request_options.pop("listings_expiry_time", None)
-        request_options.pop("max_paths", None)
-        request_options.pop("skip_instance_cache", None)
```
gutzbenj, line 110: What about those args? I think they are not needed/used anymore.

fsspec/spec.py
```diff
@@ -146,7 +147,23 @@ def __init__(self, *args, **storage_options):
...
+            listings_cache_options = {"expiry_time": None} | listings_cache_options
+            cache_type = listings_cache_options.get("cache_type")
+            if cache_type:
+                if isinstance(cache_type, str):
+                    try:
+                        cache_type = CacheType[cache_type.upper()]
+                    except KeyError as e:
+                        raise ValueError(
+                            f"Cache type must be one of {', '.join(ct.name.lower() for ct in CacheType)}") from e
+                listings_cache_options["cache_type"] = cache_type
```
gutzbenj, line 164: I think there's too much going on here!

martindurant: I think it's fine! It could be a `norm_cache_options` util functions to pull it out of here.

fsspec/implementations/http.py
martindurant: We need to make sure not to leak any kwargs to downstream implementations, since some of them take any kwargs and assume they are to be propagated to, for example, the http session constructor. This could in principle be handled in `_Cached.__call__`, which happens before any implementation-specific `__init__`.

martindurant:
```
>           self.dircache[path.rstrip("/")] = files
E           TypeError: 'NoneType' object does not support item assignment
```
It seems like the "no cache" case should still have a no-op dict to cope with this kind of thing. So, there would be a third cache type, one that corresponds to the `use_listing_cache = False` from before.

martindurant:
@amotl @gutzbenj : this PR seems to have stalled. I know there's quite a thread above, but what do you think we can do from here to get this useful functionality in?

gutzbenj:
Dear @martindurant ,

thanks for the reminder. I think I need just a bit of time. I will work on this in about 2 weeks again and push it actively.

fsspec/implementations/http.py
gutzbenj: If we use the listings_cache_options specifically, they are only passed to the _DirCache in `AbstractFileSystem` class but do not go further down the hierarchy.

gutzbenj:
Dear @martindurant ,

still on it... 

Currently some tests are failing which I believe I'm not responsible for. Maybe some reallife backend has changed or so.

martindurant:
This seems like it might be real:
```
FAILED fsspec/implementations/tests/test_http.py::test_list_cache_memory[None] - assert (0.002689361572265625 / 0.0018138885498046875) > 1.5
```
Could it be that the file cache is causing a significant slow-down? Perhaps we just need to alter the test a bit to allow for a smaller speedup factor like 1.1 .

PR #901, opened 2022-02-10
remove py36 env from tox.ini

ap--:
The minimum supported python version of fsspec is now 3.7.
This PR removes the py36 env from tox.ini


efiop (approved):
Ah, missed it in https://github.com/fsspec/filesystem_spec/pull/889

Thank you! 🙏

PR #902, opened 2022-02-10
callbacks: use existing methods instead of .call/.lazy_call

ap--:
After implementing tests for callbacks in s3fs I noticed that there are some inconsistencies in the way callbacks are used throughout fsspec. It required overriding `.call` and `.lazy_call` to allow to pass through callbacks. See here: https://github.com/fsspec/s3fs/blob/6f844d4dfb70ad7f2397220663b907398659ccd6/s3fs/tests/test_s3fs.py#L914-L925

I believe `.lazy_call` might be a left over of a previous callback implementation and should be dropped, since it is not present in the `fsspec.callbacks.Callback` base class.

I also believe that `.call` should be reserved for calling user provided hooks, and not used for operations that have explicit methods in the `Callback` base class.

#### Changes

This commit uses `.set_size`, `.relative_update` and `.absolute_update` instead of `.call` and `.lazy_call` throughout the fsspec codebase.

After this commit `.lazy_call` is not occurring anywhere in the codebase
anymore.

Let me know what you think,
Cheers,
Andreas 😃

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -234,9 +234,7 @@ async def _run_coros_in_chunks(
         ]
         if callback is not _DEFAULT_CALLBACK:
             [
-                t.add_done_callback(
-                    lambda *_, **__: callback.call("relative_update", 1)
-                )
+                t.add_done_callback(lambda *_, **__: callback.relative_update(1))
                 for t in chunk
             ]
         results.extend(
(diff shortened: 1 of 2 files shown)


martindurant:
Yes, you are quite right that these are leftovers.

martindurant:
Unfortunately fails linting by `black`. Note that there was a recent new version, which might be causing this.

isidentical:
It might need a rebase against the current `main` branch.

ap--:
Let me know if you want me to rebase. That should fix the s3fs tests.
The hanging py37, py38, and py39 tests seem to be a GH Actions problem though.

martindurant:
please ping when you have merged in the CI fix

ap--:
@martindurant this is ready to be merged.

martindurant:
Got it, thank you.

PR #903, opened 2022-02-10
tox: s3fs: bump moto version

efiop:
s3fs requires moto >=2


martindurant:
This is all good, thanks.
I wonder if anyone knows the reason that some CI runs are so amazingly slow?

efiop:
@martindurant Also wondering about that. Waiting for logs.

PR #905, opened 2022-02-11
tests: fix hanging fuse tests

ap--:
This fixes the hanging fuse tests for me locally.

Cheers,
Andreas 😃


ap--:
This should probably require some investigation why the fuse process doesn't stop on `SIGTERM` anymore.

efiop:
@ap-- Great catch! Thank you so much for investigating! 🙏 

Not 100% sure, but it looks like a new ubuntu-latest image got released on Feb 7th https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md and might've affected this somehow 🤔

martindurant:
Great work, thanks.

ap--:
Thanks 😊 
Happy to contribute back. `fsspec` has been incredibly useful.

PR #906, opened 2022-02-11
#904 do not double (de)compress on pyarrow Filesystem >=4

### LINKED ISSUE #904: ArrowFSWrapper automagically decompresses files
From here:
https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/arrow.py#L140
fsspec call https://arrow.apache.org/docs/python/generated/pyarrow.fs.FileSystem.html#pyarrow.fs.FileSystem.open_input_stream
without haading over the compression kwarg
pyarrow defaults to 'detect' while fsspec defaults to None - aka do no decompress

this leads to.. strange situations
e.g.
`with fss.open('compressed.gz', compression='infer')`
will try to decmpress an already decompressed stream

pyarrow would have a open_input_file(self, path) which might be better here?

here is an example script to reproduce
```
import os
import gzip
import tempfile
from pyarrow.fs import LocalFileSystem
from fsspec.implementations.arrow import ArrowFSWrapper


# create dummy input test file
gzip_file = os.path.join(tempfile.gettempdir(), 'input.file.gz')
with gzip.open(gzip_file, mode='wb') as g:
    g.write(b'AWESOME TEXT')

# check that we got a gzipped file
with open(gzip_file, mode='rb') as f:
    print(next(f))


local = LocalFileSystem()
fs = ArrowFSWrapper(local)

with fs.open(gzip_file, 'rb') as f:
    print(next(f))
    # This should also be gzipped
    # while the actual output is
    # b'AWESOME TEXT'

with fs.open(gzip_file, 'rb', compression='infer') as f:
    print(next(f))
    # This should also be decompressed
    # while the actual output is:
    """
    File "/home/jesusch/git/backend/test.py", line 31, in <module>
        print(next(f))
    File "/usr/lib/python3.9/gzip.py", line 398, in readline
        return self._buffer.readline(size)
    File "/usr/lib/python3.9/_compression.py", line 68, in readinto
        data = self.read(len(byte_view))
    File "/usr/lib/python3.9/gzip.py", line 487, in read
        if not self._read_gzip_header():
    File "/usr/lib/python3.9/gzip.py", line 435, in _read_gzip_header
        raise BadGzipFile('Not a gzipped file (%r)' % magic)
    gzip.BadGzipFile: Not a gzipped file (b'AW')
    """
    # this suggest me that fsspec tries to decompress the file twice
```

jesusch:
Closes #904

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -5,7 +5,6 @@ import secrets
 import shutil
 from contextlib import suppress
 from functools import wraps
-from setuptools._vendor.packaging import version
 
 from fsspec.spec import AbstractFileSystem
 from fsspec.utils import infer_storage_options, mirror_from, stringify_path
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -153,10 +153,10 @@ class ArrowFSWrapper(AbstractFileSystem):
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
         # Check if pyarrow detects compression
-        if self.pyarrow_filesystem_compression_detect():
-            stream = method(path, compression=None)
-        else:
-            stream = method(path)
+        _kwargs = {}
+        if self._pyarrow_version >= version.parse("4.0"):
+            _kwargs["compression"] = None
+        stream = method(path, **_kwargs)
 
         return ArrowFile(self, stream, path, mode, block_size, **kwargs)
 
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -4,7 +4,7 @@ import os
 import secrets
 import shutil
 from contextlib import suppress
-from functools import wraps, lru_cache
+from functools import lru_cache, wraps
 
 from fsspec.spec import AbstractFileSystem
 from fsspec.utils import infer_storage_options, mirror_from, stringify_path
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -154,7 +154,7 @@ class ArrowFSWrapper(AbstractFileSystem):
 
         # Check if pyarrow detects compression
         _kwargs = {}
-        if self._pyarrow_version >= version.parse('4.0'):
+        if self.pyarrow_filesystem_compression_detect:
             _kwargs["compression"] = None
         stream = method(path, **_kwargs)
 
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -52,12 +52,6 @@ class ArrowFSWrapper(AbstractFileSystem):
 
         return path
 
-    def pyarrow_filesystem_compression_detect(self) -> bool:
-        from pyarrow import __version__ as pyarrow_version
-        pyarrow_default_compression_started = version.parse("4.0")
-        pyarrow_version = version.parse(pyarrow_version)
-        return pyarrow_version >= pyarrow_default_compression_started
-
     def ls(self, path, detail=False, **kwargs):
         from pyarrow.fs import FileSelector
 
@@ -145,6 +139,10 @@ class ArrowFSWrapper(AbstractFileSystem):
 
     @wrap_exceptions
     def _open(self, path, mode="rb", block_size=None, **kwargs):
+        import pyarrow
+        from packaging import version
+
+
         if mode == "rb":
             method = self.fs.open_input_stream
         elif mode == "wb":
@@ -152,9 +150,9 @@ class ArrowFSWrapper(AbstractFileSystem):
         else:
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
-        # Check if pyarrow detects compression
         _kwargs = {}
-        if self.pyarrow_filesystem_compression_detect:
+        if version.parse(pyarrow.__version__) >= version.parse("4.0"):
+            # disable compression auto-detection
             _kwargs["compression"] = None
         stream = method(path, **_kwargs)
 
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -5,6 +5,7 @@ import secrets
 import shutil
 from contextlib import suppress
 from functools import wraps
+from setuptools._vendor.packaging import version
 
 from fsspec.spec import AbstractFileSystem
 from fsspec.utils import infer_storage_options, mirror_from, stringify_path
@@ -50,6 +51,12 @@ class ArrowFSWrapper(AbstractFileSystem):
             _, _, path = path.partition("://")
 
         return path
+
+    def pyarrow_filesystem_compression_detect(self) -> bool:
+        from pyarrow import __version__ as pyarrow_version
+        pyarrow_default_compression_started = version.parse('4.0')
+        pyarrow_version = version.parse(pyarrow_version)
+        return pyarrow_version >= pyarrow_default_compression_started
 
     def ls(self, path, detail=False, **kwargs):
         from pyarrow.fs import FileSelector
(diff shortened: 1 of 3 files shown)


martindurant:
Please merge from main following #905 , else the tests will hang.

jesusch:
@martindurant I've pull from AP's branch

fsspec/implementations/arrow.py
```diff
@@ -5,6 +5,7 @@
 import shutil
 from contextlib import suppress
 from functools import wraps
+from setuptools._vendor.packaging import version
```
efiop, line 8: We probably shouldn't use private `_vendor` here. Looks like `from packaging import version` is the current correct way?

fsspec/implementations/arrow.py
```diff
@@ -139,12 +146,18 @@ def rm(self, path, recursive=False, maxdepth=None):
...
+            method = self.fs.open_output_stream
         else:
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
+        # Check if pyarrow detects compression
+        if self.pyarrow_filesystem_compression_detect():
+            stream = method(path, compression=None)
+        else:
+            stream = method(path)
+
```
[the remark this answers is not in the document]
jesusch: makes sense
I've pushed the suggested change

fsspec/implementations/arrow.py
jesusch: Basically I would agree. But this would create a new dependency, as `packaging` is not available in python-base
While setuptools is and includes this module.
Yet - Open for discussion

jesusch:
> Now only failing on `isort` :)

@martindurant fixed

fsspec/implementations/arrow.py
```diff
@@ -139,12 +146,18 @@ def rm(self, path, recursive=False, maxdepth=None):
...
         elif mode == "wb":
-            stream = self.fs.open_output_stream(path)
+            method = self.fs.open_output_stream
         else:
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
+        # Check if pyarrow detects compression
+        _kwargs = {}
+        if self._pyarrow_version >= version.parse('4.0'):
+            _kwargs["compression"] = None
```
efiop, line 158: So 4.0+ will disable compression, but older ones will `detect` it, right? That seems like an odd behavioural change for fs to have.

Also, you mention `with fss.open('compressed.gz', compression='infer')` in the original issue, which raises the question of who should receive that arg. The ArrowFile(based on io.IOBase) doesn't seem to care for it, right? Should we catch this argument in fs.open then and not pass it through?

jesusch: It's the opposite way - starting at v4.0 pyarrow has a `compression='detect'` arg - which defaults to decompress.
if `fss.open('some.gz', compression='infer')` is used, it receives a stream which is already decompressed
if `fss.open('some.gz', compression=None)` is used, it receives a decompressed stream which is unintentional

I've added a code example in the ticket

efiop: by `fss.open` you mean open in fsspec fs, right? If so, that `compression=` arg is not being passed to the method at all, hence why I'm wondering about the desired behaviour here and wether or not `fs.open` here should catch it.

efiop: Ah, ok, there is misunderstanding here on my side. I see that it is being used down the line.

efiop:
@jesusch Looks like maintainer edits are not allowed in your fork 🙁 I find long and convoluted names rather confusing in the current approach, wdyt about something like this:

```
diff --git a/fsspec/implementations/arrow.py b/fsspec/implementations/arrow.py
index 0a785bd..42fea16 100644
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -138,13 +138,23 @@ class ArrowFSWrapper(AbstractFileSystem):
 
     @wrap_exceptions
     def _open(self, path, mode="rb", block_size=None, **kwargs):
+        import pyarrow
+        from packaging import version
+
         if mode == "rb":
-            stream = self.fs.open_input_stream(path)
+            method = self.fs.open_input_stream
         elif mode == "wb":
-            stream = self.fs.open_output_stream(path)
+            method = self.fs.open_output_stream
         else:
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
+        _kwargs = {}
+        if version.parse(pyarrow.__version__) >= version.parse("4.0"):
+            # disable compression auto-detection
+            _kwargs["compression"] = None
+
+        stream = method(path, **_kwargs)
+
         return ArrowFile(self, stream, path, mode, block_size, **kwargs)
 
     @wrap_exceptions
```

jesusch:
> @jesusch Looks like maintainer edits are not allowed in your fork slightly_frowning_face
> 
> I still find long and convoluted names rather confusing in the current approach, wdyt about something like this:
> 
> ```
> diff --git a/fsspec/implementations/arrow.py b/fsspec/implementations/arrow.py
> index 0a785bd..42fea16 100644
> --- a/fsspec/implementations/arrow.py
> +++ b/fsspec/implementations/arrow.py
> @@ -138,13 +138,23 @@ class ArrowFSWrapper(AbstractFileSystem):
>  
>      @wrap_exceptions
>      def _open(self, path, mode="rb", block_size=None, **kwargs):
> +        import pyarrow
> +        from packaging import version
> +
>          if mode == "rb":
> -            stream = self.fs.open_input_stream(path)
> +            method = self.fs.open_input_stream
>          elif mode == "wb":
> -            stream = self.fs.open_output_stream(path)
> +            method = self.fs.open_output_stream
>          else:
>              raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
>  
> +        _kwargs = {}
> +        if version.parse(pyarrow.__version__) >= version.parse("4.0"):
> +            # disable compression auto-detection
> +            _kwargs["compression"] = None
> +
> +        stream = method(path, **_kwargs)
> +
>          return ArrowFile(self, stream, path, mode, block_size, **kwargs)
>  
>      @wrap_exceptions
> ```

updated the fork - dunno how/where github has the permissions for fork-origin maintainers :D

jesusch:
only thing I'm not sure / uncertain if the version check shouldd be cached (e.g. cached_property) for performance reasons while opening many files?

martindurant:
On the issue of whether or not to depend on `packaging`, is there any reason to not go with a far simpler condition like
```
if pyarrow.__version__.split(".") > ["4"]:
```
?

jesusch:
> if pyarrow.__version__.split(".") > ["4"]:

then we're again at the "how to name that property" problem :dancers:

martindurant:
PYARROW_GT4 ?

ap--:
Hi everyone, 

```
>>> ["10"] >= ["4"]
False
```

maybe `int(pyarrow.__version__.split(".")[0]) >= 4` ?

(and i think you switched from `>=` to `>` since the initial version)

martindurant:
`[int(p) for p in pyarrow.__version__.split(".")]` is what I *meant*...

fsspec/implementations/arrow.py
```diff
@@ -43,6 +43,17 @@ def __init__(self, fs, **kwargs):
         self.fs = fs
         super().__init__(**kwargs)
 
+
+    @property
+    def _pyarrow_version_tuple(self) -> tuple:
+        from pyarrow._generated_version import version_tuple
```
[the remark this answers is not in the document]
martindurant: top of file:
```python
PYARROW_VERSION = None
```

where pyarrow is imported
```python
global PYARROW_VERSION
if PYARROW_VERSION is None:
    PYARROW_VERSION = tuple(int(_) for _ in pyarrow.__version__.split("."))
```

martindurant:
Hi, @jesusch  - sorry for the stylistic back and forth here, and that the lint CI is probably running with a more recent version of black than you. Do you think you can pull it together to get this finished off?

jesusch:
sorry for the delay - hope it's fine now?

PR #907, opened 2022-02-13
Add petl package to docs section who uses

juarezr:
Petl is a general-purpose Python package for extracting, transforming, and loading tables of data.
It uses `fsspec` as optional integration for reading and writing data from remote filesystems.
Coupling both packages enable the creation of pipelines for transferring and transforming data 
between pairs of remote servers, cloud filesystems, databases, compressed files, Webservices, 
and many diverse file formats like spreadsheets, avro, jons, XML, csv and others.


martindurant:
You are welcome here!
Also, it seems like petl and [Intake](https://github.com/intake/intake/) have a lot to discuss.

PR #913, opened 2022-02-26
Add methods to retrieve available protocols and compression

ghiggi:
This is just a little PR proposal adding methods to easily retrieve the available protocols and compressions. 

Usage:
```
import fsspec
fsspec.available_protocol()
# ['file', 'memory', 'dropbox', 'http', 'https', 'zip', 'tar', 'gcs', 'gs', 'gdrive', ...]

fsspec.available_compression()
# [None, 'zip', 'bz2', 'gzip', 'lzma', 'xz']
```

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -10,8 +10,8 @@ User Functions
    fsspec.open_files
    fsspec.open
    fsspec.open_local
-   fsspec.available_compression
-   fsspec.available_protocol
+   fsspec.available_compressions
+   fsspec.available_protocols
    fsspec.filesystem
    fsspec.get_filesystem_class
    fsspec.get_mapper
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -10,6 +10,8 @@ User Functions
    fsspec.open_files
    fsspec.open
    fsspec.open_local
+   fsspec.available_compression
+   fsspec.available_protocol
    fsspec.filesystem
    fsspec.get_filesystem_class
    fsspec.get_mapper
@@ -19,6 +21,8 @@ User Functions
 .. autofunction:: fsspec.open_files
 .. autofunction:: fsspec.open
 .. autofunction:: fsspec.open_local
+.. autofunction:: fsspec.available_compression
+.. autofunction:: fsspec.available_protocol
 .. autofunction:: fsspec.filesystem
 .. autofunction:: fsspec.get_filesystem_class
 .. autofunction:: fsspec.get_mapper
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -21,8 +21,8 @@ User Functions
 .. autofunction:: fsspec.open_files
 .. autofunction:: fsspec.open
 .. autofunction:: fsspec.open_local
-.. autofunction:: fsspec.available_compression
-.. autofunction:: fsspec.available_protocol
+.. autofunction:: fsspec.available_compressions
+.. autofunction:: fsspec.available_protocols
 .. autofunction:: fsspec.filesystem
 .. autofunction:: fsspec.get_filesystem_class
 .. autofunction:: fsspec.get_mapper
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -28,7 +28,7 @@ Look-up via registry:
 
     import fsspec
 
-    fs = fsspec.filesystem('local')
+    fs = fsspec.filesystem('file')
 
 Many filesystems also take extra parameters, some of which may be options - see :doc:`api`, or use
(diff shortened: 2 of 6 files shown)


fsspec/registry.py
```diff
@@ -251,3 +251,8 @@ def filesystem(protocol, **storage_options):
     """
     cls = get_filesystem_class(protocol)
     return cls(**storage_options)
+
+
+def available_protocol():
+    """Return a list of the available protocols."""
```
martindurant, line 257: Can you please make this more specific: these are protocols that fsspec knows how to deal with, but they still might not be "available" if other required dependencies are missing.

martindurant (commented):
Thanks, I think this is useful. 
I have once comment on your docstring. Do you think these functions should appear in the API docs?

fsspec/registry.py
martindurant: > Do you mean to correct the docstring

Yes, exactly. Doing a load of imports is probably a bad idea.

> I think would be useful to add it to the API docs. Should I do a PR also for that?

You can include here - see  docs/source/api.rst .

ghiggi: Hey @martindurant. Doc updated. 
Note that I also added to the registry an additional protocol "local" as a proxy for "LocalSystem" (in addition to the existing "file").

martindurant: > I also added to the registry an additional protocol "local"

That's fine

docs/source/usage.rst
```diff
@@ -28,7 +28,7 @@ Look-up via registry:
 
     import fsspec
 
-    fs = fsspec.filesystem('file')
+    fs = fsspec.filesystem('local')
```
martindurant, line 31: I am OK to allow the "local" alias, but I don't want to change the default recommendation here

fsspec/compression.py
```diff
@@ -166,3 +166,8 @@ def zstandard_file(infile, mode="rb"):
     register_compression("zstd", zstandard_file, "zst")
 except ImportError:
     pass
+
+
+def available_compression():
+    """Return a list of the implemented compressions."""
+    return list(compr.keys())
```
martindurant, line 173: `keys` not required here

fsspec/registry.py
```diff
@@ -86,6 +86,7 @@ def register_implementation(name, cls, clobber=True, errtxt=None):
 # updated with register_implementation
 known_implementations = {
     "file": {"class": "fsspec.implementations.local.LocalFileSystem"},
+    "local": {"class": "fsspec.implementations.local.LocalFileSystem"},
```
martindurant, line 89: If you add this, you also need to add "local" to the FSs `protocols` attribute, otherwise passing "local://..." URLs will fail with some operations. Also, this is not now tested anywhere.

fsspec/registry.py
```diff
@@ -251,3 +252,8 @@ def filesystem(protocol, **storage_options):
     """
     cls = get_filesystem_class(protocol)
     return cls(**storage_options)
+
+
+def available_protocol():
+    """Return a list of the implemented protocols."""
+    return list(known_implementations.keys())
```
martindurant, line 259: `keys` not required here

martindurant (commented):
I have given a more thorough review here.
I think semantically, the functions should be `available_protocols` and `available_compressions` (plural)

On second thoughts, using "local" may be more of a change than you had realised, so maybe consider reverting this.

docs/source/api.rst
```diff
@@ -10,6 +10,8 @@ User Functions
    fsspec.open_files
    fsspec.open
    fsspec.open_local
+   fsspec.available_compression
+   fsspec.available_protocol
```
martindurant, line 14:
```suggestion
   fsspec.available_compressions
   fsspec.available_protocols
```

docs/source/api.rst
```diff
@@ -19,6 +21,8 @@ User Functions
 .. autofunction:: fsspec.open_files
 .. autofunction:: fsspec.open
 .. autofunction:: fsspec.open_local
+.. autofunction:: fsspec.available_compression
+.. autofunction:: fsspec.available_protocol
```
martindurant, line 25:
```suggestion
.. autofunction:: fsspec.available_compressions
.. autofunction:: fsspec.available_protocols
```

fsspec/registry.py
```diff
@@ -251,3 +251,8 @@ def filesystem(protocol, **storage_options):
     """
     cls = get_filesystem_class(protocol)
     return cls(**storage_options)
+
+
+def available_protocols():
+    """Return a list of the implemented protocols."""
```
martindurant, line 257:
```suggestion
    """Return a list of the implemented protocols.
    
    Note that any given protocol may require extra packages to be importable.
    """
```

PR #914, opened 2022-03-04
Add more consistent ls for TarFileSystem

aramesh117:
Add consistent ls that lists subdirectories
of directories and add tests

### DIFF
--- a/fsspec/implementations/tar.py
+++ b/fsspec/implementations/tar.py
@@ -88,7 +88,7 @@ class TarFileSystem(AbstractArchiveFileSystem):
         out = {}
         for ti in self.tar:
             info = ti.get_info()
-            info["type"] = typemap[info["type"]]
+            info["type"] = typemap.get(info["type"], "file")
             name = ti.get_info()["name"].rstrip("/")
             out[name] = (info, ti.offset_data)
 
@@ -96,14 +96,17 @@ class TarFileSystem(AbstractArchiveFileSystem):
         # TODO: save index to self.index_store here, if set
 
     def _get_dirs(self):
-
         if self.dir_cache is not None:
             return
 
-        self.dir_cache = {}
+        # This enables ls to get directories as children as well as files
+        self.dir_cache = {
+            dirname + "/": {"name": dirname + "/", "size": 0, "type": "directory"}
+            for dirname in self._all_dirnames(self.tar.getnames())
+        }
         for member in self.tar.getmembers():
             info = member.get_info()
-            info["type"] = typemap[info["type"]]
+            info["type"] = typemap.get(info["type"], "file")
             self.dir_cache[info["name"]] = info
 
     def _open(self, path, mode="rb", **kwargs):
--- a/fsspec/implementations/tests/test_tar.py
+++ b/fsspec/implementations/tests/test_tar.py
@@ -1,12 +1,17 @@
 import os
 import shutil
+import tarfile
 import tempfile
+from io import BytesIO
+from pathlib import Path
+from typing import Dict
 
 import pytest
 
 import fsspec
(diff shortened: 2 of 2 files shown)


fsspec/implementations/tar.py
```diff
@@ -88,22 +88,25 @@ def _index(self):
...
 
     def _get_dirs(self):
-
         if self.dir_cache is not None:
             return
 
-        self.dir_cache = {}
+        # This enables ls to get directories as children as well as files
+        self.dir_cache = {
+            dirname + "/": {"name": dirname + "/", "size": 0, "type": "directory"}
```
martindurant, line 104: I would discourage the use of the "/" suffix, since other implementations don't do this. We already say it's a directory, so no need to indicate it again.

fsspec/implementations/tar.py
```diff
@@ -88,22 +88,25 @@ def _index(self):
...
+            for dirname in self._all_dirnames(self.tar.getnames())
+        }
         for member in self.tar.getmembers():
             info = member.get_info()
-            info["type"] = typemap[info["type"]]
+            info["type"] = typemap.get(info["type"], "file")
```
martindurant, line 109: Are there other "type"s that we should be inferring?

fsspec/implementations/tests/test_tar.py
```diff
@@ -189,8 +193,48 @@ def test_url_to_fs_direct(recipe, tmpdir):
...
+            for tar_file_path, data in tar_data.items():
+                content = data
+                info = tarfile.TarInfo(name=tar_file_path)
+                info.size = len(content)
+                tf.addfile(info, BytesIO(content))
+    with open(temp_archive_file, "rb") as fd:
+        fs = TarFileSystem(fd)
+        assert fs.find("/", withdirs=True) == [
+            "a.pdf",
+            "b/",
```
martindurant, line 234: Fixing my comment on suffixes above would change this comparison.

martindurant (commented):
Thanks for the PR, this is useful and I only have a couple of comments.

fsspec/implementations/tar.py
aramesh117: @martindurant interesting, I actually cloned this from the ZipFileSystem's implementation: https://github.com/aramesh117/filesystem_spec/blob/f7237f8c3472db9a795213e080d709da994b4c57/fsspec/implementations/zip.py#L69. I can certainly remove it, but just wondering why the difference in behavior.

fsspec/implementations/tar.py
martindurant: There are a *few* inconsistencies between the various implementations, depending on when they were made. We still lack a comprehensive test system that guarantees that the API of each acts the same across all the implementations.

martindurant: 👍 , got it. Odd!

fsspec/implementations/tar.py
aramesh117: Done.

aramesh117: Seems like removing the '/' suffix breaks the info test. The expected value includes a / suffix. What would you recommend?

martindurant: OK, well I suppose put it back, and we can consider a separate PR where we bring TAR and other archive formats into line.

aramesh117:
@martindurant awesome, thank you for the comprehensive review.. Anything else I can do to get this merged?

martindurant:
Thanks and sorry for the back-and-forth here

aramesh117:
Not at all, thanks a lot for merging! We have a patch locally for this, but we would love to get rid of it in the next release of fsspec.

PR #917, opened 2022-03-08
Fix on_error in reference's cat

### LINKED ISSUE #915: reference: `KeyError` is propagated through zarr
Zarr says that chunks which don't exist should be read as `fill_value`. This doesn't work when using the `reference` filesystem:

```python
import json
import zarr

zarray = {
    "chunks": [1],
    "compressor": None,
    "dtype": "<f8",
    "fill_value": "NaN",
    "filters": [],
    "order": "C",
    "shape": [10],
    "zarr_format": 2,
}
refs = {".zarray": json.dumps(zarray)}

with open("refs.json", "w") as fp:
    json.dump(refs, fp)

a = zarr.open_array("reference::refs.json")
print(a[0])
```

raises:
```
  File ".../fsspec/implementations/reference.py", line 192, in _cat_common
    part = self.references[path]
KeyError: '0'
```
I'd expect the code to print the `fill_value`.

`fsspec.__version__`: `'2022.02.0'`

Thanks @lkluft for discovering this.

martindurant:
Fixes #915

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -279,25 +279,44 @@ class ReferenceFileSystem(AsyncFileSystem):
                     self, paths, lpath, recursive=recursive, **kwargs
                 )
 
-    def cat(self, path, recursive=False, **kwargs):
+    def cat(self, path, recursive=False, on_error="raise", **kwargs):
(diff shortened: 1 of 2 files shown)


d70-t:
*wohoo* 🚀 that was fast!

PR #929, opened 2022-03-25
Specify the git branch name for tests

gadomski:
On some systems, the default branch name might not be `master` (e.g. `main` is becoming more common,
https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/).

### DIFF
--- a/fsspec/implementations/tests/test_git.py
+++ b/fsspec/implementations/tests/test_git.py
@@ -17,8 +17,8 @@ def repo():
     d = tempfile.mkdtemp()
     try:
         os.chdir(d)
-        subprocess.call("git init", shell=True, cwd=d)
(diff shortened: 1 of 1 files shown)


martindurant:
Does the implicit default branch name make a difference anywhere in the tests?

gadomski:
It caused test failures on my local system, where `main` is the default, because `master` is hardcoded: https://github.com/fsspec/filesystem_spec/blob/8f62e479ddd45285b3e7126e9313151f1bebd3b9/fsspec/implementations/tests/test_git.py#L27

martindurant:
Got it. I thought it wouldn't be mentioned in the tests, but this of course is part of the setup fixture. There's probably a good pure git way to do this without reading a file, but your solution is good too.

PR #931, opened 2022-03-25
callbacks: add a tqdm callback

### LINKED ISSUE #923: Docs could use an example for callbacks.
Hi !

Thank you for fsspec ! :heart: 🙏 

I think the docs could use an example for callbacks. I've come up with this little hook with `tqdm`, not sure if it's the right way to do it:
```python

import tqdm
import fsspec


class FsspecTqdmCallback(tqdm.tqdm):
    """Provides `update_to(n)` which uses `tqdm.update(delta_n)`."""

    def update_to(self, total_size, value, **kwargs):
        """
        total_size  : int
            Total size (in tqdm units)
        value  : int, optional
            Number of blocks transferred so far.
        """
        self.total = total_size
        return self.update(value - self.n)


storage_opts = {
    "key": KEY,
    "secret": SECRET_KEY,
}

fs = fsspec.filesystem("s3", **storage_opts)
path2distant_data = "s3://your-path"
with FsspecTqdmCallback(
    desc="test",
) as pbar:
    fs.upload(
        "local_path/",
        path2distant_data,
        recursive=True,
        callback=fsspec.callbacks.Callback(hooks={"tqdm": pbar.update_to}),
    )

```

Timost:
This PR adds a TqdmCallback class to display a progress bar using tqdm during fsspec operations.

This is probably wrong in many ways, please let me know what needs to be changed.

closes #923

### DIFF
--- a/fsspec/callbacks.py
+++ b/fsspec/callbacks.py
@@ -1,3 +1,6 @@
+import importlib
+
+
 class Callback:
     """
     Base class and interface for callback mechanism
@@ -175,6 +178,44 @@ class DotPrinterCallback(Callback):
     def call(self, **kwargs):
         """Just outputs a character"""
         print(self.chr, end="")
+
+
+class TqdmCallback(Callback):
+    """
+    A callback to display a progress bar using tqdm
+
+    Examples
+    --------
+    >>> import fsspec
+    >>> from fsspec.callbacks import TqdmCallback
+    >>> fs = fsspec.filesystem("memory")
+    >>> path2distant_data = "/your-path"
+    >>> fs.upload(
+            ".",
+            path2distant_data,
+            recursive=True,
+            callback=TqdmCallback(),
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        try:
+            self._tqdm = importlib.import_module("tqdm")
+        except ImportError as exce:
+            raise ImportError(
+                "Using TqdmCallback requires tqdm to be installed"
+            ) from exce
+        super().__init__(*args, **kwargs)
+
+    def set_size(self, size):
+        self.tqdm = self._tqdm.tqdm(desc="test", total=size)
+
+    def relative_update(self, inc=1):
+        self.tqdm.update(inc)
+
+    def __del__(self):
+        self.tqdm.close()
+        self.tqdm = None
 
 
 _DEFAULT_CALLBACK = NoOpCallback()
--- a/fsspec/callbacks.py
+++ b/fsspec/callbacks.py
@@ -1,6 +1,3 @@
-import importlib
-
-
 class Callback:
     """
     Base class and interface for callback mechanism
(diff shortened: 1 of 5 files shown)


fsspec/callbacks.py
```diff
@@ -177,4 +180,42 @@ def call(self, **kwargs):
...
+            ".",
+            path2distant_data,
+            recursive=True,
+            callback=TqdmCallback(),
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        try:
+            self._tqdm = importlib.import_module("tqdm")
```
martindurant, line 203:
```suggestion
            import tqdm
            self.tqdm = tqdm
```

martindurant:
@efiop @isidentical you probably hav some typical callbacks to compare with

dtrifiro:
@martindurant:

https://github.com/iterative/dvc/blob/5ff1d768261c759bbde01d402ada12c79758e7ce/dvc/fs/_callback.py#L6-L32

fsspec/callbacks.py
```diff
@@ -177,4 +180,43 @@ def call(self, **kwargs):
...
+            path2distant_data,
+            recursive=True,
+            callback=TqdmCallback(),
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        try:
+            import tqdm
+            self.tqdm = tqdm
```
martindurant, line 201:
```suggestion
            self._tqdm = tqdm
```

martindurant:
I think it's nice to have, actually - code that lives in the code, and is immediately useful to anyone calling it.

fsspec/callbacks.py
```diff
@@ -177,4 +177,43 @@ def call(self, **kwargs):
...
+            path2distant_data,
+            recursive=True,
+            callback=TqdmCallback(),
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        try:
+            import tqdm
+            self._tqdm = tqdm
```
martindurant, line 201:
```suggestion
            import tqdm

            self._tqdm = tqdm
```

fsspec/callbacks.py
```diff
@@ -177,4 +177,44 @@ def call(self, **kwargs):
...
+
+            self._tqdm = tqdm
+        except ImportError as exce:
+            raise ImportError(
+                "Using TqdmCallback requires tqdm to be installed"
+            ) from exce
+        super().__init__(*args, **kwargs)
+
+    def set_size(self, size):
+        self.tqdm = self._tqdm.tqdm(desc="test", total=size)
```
Timost, line 210: Maybe we could the tqdm `desc` parameter somewhere ?

Timost:
Hi @martindurant thank you for doing the clean up !
🙏 I was offline the last few days.

I have a small question regarding the `desc=` parameter of tqdm.

fsspec/callbacks.py
martindurant: Oh, yes please do add this

Timost: 👍Here it is: https://github.com/fsspec/filesystem_spec/pull/932

PR #932, opened 2022-03-30
callbacks: forward tqdm args

Timost:
follow-up to #931

### DIFF
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -20,3 +20,4 @@ dependencies:
   - python-libarchive-c
   - numpy
   - nomkl
+  - tqdm
--- a/fsspec/callbacks.py
+++ b/fsspec/callbacks.py
@@ -54,7 +54,7 @@ class Callback:
 
     def relative_update(self, inc=1):
         """
-        Delta increment the internal cuonter
+        Delta increment the internal counter
 
         Triggers ``call()``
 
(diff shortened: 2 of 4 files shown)


martindurant:
I wonder if there's a decent way to test this with pytest's capture? Probably we could test instead by mocking tqdm completely (i.e., nothing is actually printed).

Timost:
Yeah both mocking and using [capfd](https://pytest.org/en/latest/reference/reference.html#std-fixture-capfd) would work I guess.
I can give it a try.

martindurant:
(please ping me when you've had a go, or perhaps if you've realised you won't have time to work on the mock)

Timost:
Here is a version using `capsys`. It's probably really brittle (wouldn't be surprised if the ci failed).
Using mock would probably be safer, what do you think ?

martindurant:
It does seem like, as written, the exact output might depend on the assumed console width and character encoding, so I would expect this to fail on some systems. Also, I was wondering where the iterations between the first and last end up. I'm not certain whether this can be hardened, so if mocking is easy, that will do.

Timost:
I think tqdm buffers the output based on the rate of processing, so in this case it only outputs the first and last "frames" of the progress bar.
I'll try an alternative based on mocks.

Timost:
Here you go ! I should have gone with mocks directly '^^.

martindurant:
Perfect

PR #933, opened 2022-03-31
Add reference from CachingFileSystem to MMapCache

djhoese:
I was reading the docs and wanted to click on MMapCache but it wasn't linked. Made this edit from the GitHub editor so not sure if the docs are rendered correctly yet.

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -25,8 +25,8 @@ class CachingFileSystem(AbstractFileSystem):
     access after the initial download. The files are stored in a given
     directory with random hashes for the filenames. If no directory is given,
     a temporary one is used, which should be cleaned up by the OS after the
-    process ends. The files themselves as sparse (as implemented in
-    MMapCache), so only the data which is accessed takes up space.
+    process ends. The files themselves are sparse (as implemented in
+    :class:`~fsspec.core.MMapCache`), so only the data which is accessed takes up space.
 
     Restrictions:
 


fsspec/implementations/cached.py
```diff
@@ -25,8 +25,8 @@ class CachingFileSystem(AbstractFileSystem):
     access after the initial download. The files are stored in a given
     directory with random hashes for the filenames. If no directory is given,
     a temporary one is used, which should be cleaned up by the OS after the
-    process ends. The files themselves as sparse (as implemented in
-    MMapCache), so only the data which is accessed takes up space.
+    process ends. The files themselves are sparse (as implemented in
+    :class:`~fsspec.core.MMapCache`), so only the data which is accessed takes up space.
```
martindurant, line 29: What does the "~" do here?

djhoese: It makes it so the actual text is `MMapCache` (the actual object/element name) instead of the full module path `fsspec.core.MMapCache`. This should result in the final readable text being unchanged except that `MMapCache` will now be clickable.

djhoese: From https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax:

> If you prefix the content with ~, the link text will only be the last component of the target. For example, :py:meth:`~Queue.Queue.get` will refer to Queue.Queue.get but only display get as the link text. This does not work with all cross-reference roles, but is domain specific.

martindurant:
Triggered manually - please remind me to look into the hook if I don't otherwise do it.

martindurant:
OK, I think I did it

djhoese:
Looks like the docs are updated, but it looks like my link is invalid. I knew I should have rendered this locally. One sec...

PR #939, opened 2022-04-04
Add all FSMap constructor arguments as class attributes

rabernat:
There are currently two constructor attributes for `FSMap` - `create` and `check` - that are not accessible as instance attributes. This makes it hard to copy an FSMap. Having these accessible would be useful for e.g. https://github.com/zarr-developers/zarr-python/pull/911; it would allow us to completely rebuild an FSMap object from another.

### DIFF
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -44,6 +44,8 @@ class FSMap(MutableMapping):
                 NotADirectoryError,
             )
         self.missing_exceptions = missing_exceptions
+        self.check = check
+        self.create = create
         if create:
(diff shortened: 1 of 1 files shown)


martindurant:
Good idea, a long unprobable string might be good. Do we really need `check`, I wonder.

rabernat:
I have used check in the past to verify that the store is actually writeable. I believe that wrapping the touch in a `try` / `except` block and providing a very verbose error message would be even more useful.

martindurant:
The rest we can leave for a follow-up

Issue #792, opened 2021-10-08
Implementation mkdir not accepting AbstractFileSystem parameters


headlessButSmart:
Some implementations (such as SFTPFileSystem) do not accept create_parents parameter, which is part of the AbstractFileSystem specification for mkdir. While this is OK for typical use cases, it breaks the abstraction logic for users of fsspec. 

An example issue appears when these implementation classes are used together with pyarrow's PyFileSystem. FSSpecHandler translates its create_dir(self, path, recursive) calls into mkdir(path, create_parents=recursive) following AbstractFileSystem standards and it fails due to missing parameter in implementation.

PR #944, opened 2022-04-14
Implementation mkdir not accepting AbstractFileSystem parameters

uunal:
Related issue is discussed at -->[( https://github.com/fsspec/filesystem_spec/issues/792)]( https://github.com/fsspec/filesystem_spec/issues/792)


fsspec/implementations/sftp.py
```diff
@@ -64,9 +64,21 @@ def _get_kwargs_from_urls(urlpath):
...
-    def mkdir(self, path, mode=511):
-        logger.debug("Creating folder %s" % path)
-        self.ftp.mkdir(path, mode)
+    def mkdir(self, path, mode=511, **kwargs):                
+        logger.info("Creating folder %s" % path)           
+        if self.exists(path):            
+            raise FileExistsError("File exists: {}".format(path))                
+        
+        # kwargs is a fix for create_parent          
+        if kwargs.get("create_parents"):
```
martindurant, line 73: Why not add `create_parents=False` to the method signature?

martindurant (commented):
Thank you for the contribution. Can we please have a test showing that you can create parents, hit the exception if the file exists, or get a different exception when making a directory without create_parents and where the parent does not exist?

martindurant:
Do you think you might have time to add tests?

martindurant:
Fails on `black` formatting

uunal:
> Do you think you might have time to add tests?

I've added a test file.  I've checked the conditions as you requested.
Is this the proper way to do it? 
Thanks

fsspec/implementations/sftp.py
```diff
@@ -64,9 +64,20 @@ def _get_kwargs_from_urls(urlpath):
         out.pop("protocol", None)
         return out
 
-    def mkdir(self, path, mode=511):
-        logger.debug("Creating folder %s" % path)
-        self.ftp.mkdir(path, mode)
+    def mkdir(self, path, create_parents=False, mode=511):
+        logger.info("Creating folder %s" % path)
+        if self.exists(path):
+            raise FileExistsError("File exists: {}".format(path))
```
martindurant, line 70: It seems unnecessary to make this extra check, we would expect the remote to give a reasonable error when you try to create the dir.

fsspec/implementations/tests/test_sftp_mkdir.py
```diff
@@ -0,0 +1,67 @@
...
+    for cmd in cmds:
+        subprocess.call(["docker", "exec", cid] + shlex.split(cmd))
+    try:
+        time.sleep(1)
+        yield dict(host="localhost", port=9200, username="root", password="pass")
+    finally:
+        stop_docker(name)
+
+
+def test_mkdir_create_parent(ssh):
```
martindurant, line 57: Could we add this function to the existing test_sftp file, rather than duplicating the fixtures here?

uunal: I moved it back into existing test file.

fsspec/implementations/sftp.py
uunal: I removed this extra check, then test are failed due to `FileExistsError`. That's why I reverted it, now all tests are successful on my end.

martindurant: Can we catch the paramiko exception and turn it into FileExists without first checking?
```
try:
...
    self.ftp.mkdir(path, mode)
except paramiko-exists-error as e:
    raise FileExistsError from e
except paramiko-path-doesn't-exist-error as e:
    raise FileNotFound from e
```

uunal: In [paramiko>sftp_client.py:891](https://github.com/paramiko/paramiko/blob/29c4d3210b745e7d56c328ffebc2513ebfb482fb/paramiko/sftp_client.py#L891), it already raises `FileNotFoundError` and `Failure` for `FileExistsError` (other exceptions included too) . If we create this type of control we double the `FileNotFoundError` and miss the exact `FileExistsError` since no identifier presented for it in the `paramiko`.
I think it's better to check FileExists in there and let `pyarrow` control the exception. Check [pyarrow>fs.py:339](https://github.com/apache/arrow/blob/306d0b4144527fb988da3b64ec74b53536d974f0/python/pyarrow/fs.py#L339)

fsspec/implementations/sftp.py
```diff
@@ -64,9 +64,20 @@ def _get_kwargs_from_urls(urlpath):
...
+
+        if create_parents:
+            parts = path.split("/")
+            path = ""
+            for part in parts:
+                path += "/" + part
+                if not self.exists(path):
+                    self.ftp.mkdir(path, mode)
```
martindurant, line 78: If we find a path that exists, we should stop the loop and not keep checking the parents (which must therefore also exist)

fsspec/implementations/sftp.py
martindurant: `pyarrow` ? You have confused me :|

uunal: Sorry:) I'm considering this scenario, that's why I guess:

> An example issue appears when these implementation classes are used together with pyarrow's PyFileSystem. FSSpecHandler translates its create_dir(self, path, recursive) calls into mkdir(path, create_parents=recursive) following AbstractFileSystem standards and it fails due to missing parameter in implementation.

fsspec/implementations/sftp.py
uunal: We will not know how parents will a directory has, that's why it checks and creates according to a parent does not exist. 

In a condition for mkdir('/a/b/c') usage, such that `create_parents=True`  and '/a/b/' exists, it will loop extra as you pointed out. But we still have no information where to split (meaning which part exists).

martindurant: If '/a/b/c' exists, then so must '/a/b', we do not need to check

martindurant: Yes, that would be better

martindurant:
```
fsspec/implementations/tests/test_sftp.py::test_mkdir_create_parent Aborted (core dumped)
```

uunal:
Hi,  should I keep merge request or close it? Let me know if it is suitable for the repo or not.
Thanks

martindurant:
Sorry this slipped through the cracks. Merging now.

PR #955, opened 2022-04-28
 Fix #954: don't crash when reading empty file from libarchive 

### LINKED ISSUE #954: libarchive filesysystem breaks on empty files
When given a tar file with an empty file inside, libarchive explodes if you try to open the empty file:

```python
import os
import fsspec
import tarfile
import tempfile
from io import BytesIO
td = tempfile.TemporaryDirectory()
LibArchive = fsspec.get_filesystem_class("libarchive")

# tar file created this way is fine
with tarfile.TarFile.open(f"{td.name}/empty.tar.gz", mode="w:gz") as tf:
    # create an empty file and add it to the archive
    io = BytesIO()
    io.write(b"")
    tf.addfile(tarfile.TarInfo("empty.txt"), io)

# extract tarball using TarFile to show it's fine
with tarfile.TarFile.open(f"{td.name}/empty.tar.gz", mode="r:gz") as tf:
    os.chdir(td.name)
    tf.extractall(td.name)
    assert "empty.txt" in os.listdir(td.name)

# now try to read the file with fsspec/libarchive:
f = LibArchive(f"{td.name}/empty.tar.gz").open("empty.txt")
```

fails with:

```
  File "/Users/dlwh/src/sprucfluo/examples/stream_lambada.py", line 45, in <module>
    f = LibArchive(f"{td.name}/empty.tar.gz").open("empty.txt")
  File "/opt/homebrew/Caskroom/miniforge/base/envs/sprucfluo/lib/python3.10/site-packages/fsspec/spec.py", line 1009, in open
    f = self._open(
  File "/opt/homebrew/Caskroom/miniforge/base/envs/sprucfluo/lib/python3.10/site-packages/fsspec/implementations/libarchive.py", line 209, in _open
    raise ValueError
ValueError
```

The relevant code is here:

```
    def _open(
        self,
        path,
        mode="rb",
        block_size=None,
        autocommit=True,
        cache_options=None,
        **kwargs,
    ):
        if mode != "rb":
            raise NotImplementedError

        data = bytes()
        with self._open_archive() as arc:
            for entry in arc:
                if entry.pathname != path:
                    continue
                for block in entry.get_blocks(entry.size):
                    data = block
                    break
                else:
                    raise ValueError
        return MemoryFile(fs=self, path=path, data=data)
```

I think it's as simple as checking if `entry.size == 0` and being happy to return an empty data.

dlwh:
As above, added a simple unit test and verified it crashes without the patch

### DIFF
--- a/fsspec/implementations/libarchive.py
+++ b/fsspec/implementations/libarchive.py
@@ -202,6 +202,11 @@ class LibArchiveFileSystem(AbstractArchiveFileSystem):
             for entry in arc:
                 if entry.pathname != path:
                     continue
+                    
+                if entry.size == 0:
+                    # empty file, so there are no blocks
+                    break
+
                 for block in entry.get_blocks(entry.size):
                     data = block
                     break
--- a/fsspec/implementations/libarchive.py
+++ b/fsspec/implementations/libarchive.py
@@ -202,7 +202,7 @@ class LibArchiveFileSystem(AbstractArchiveFileSystem):
             for entry in arc:
                 if entry.pathname != path:
                     continue
-                    
+
                 if entry.size == 0:
                     # empty file, so there are no blocks
                     break
--- a/fsspec/implementations/tests/test_archive.py
+++ b/fsspec/implementations/tests/test_archive.py
@@ -364,7 +364,7 @@ class TestAnyArchive:
 
             assert lhs_dirs == {e for e in entries if fs.isdir(e)}
             assert lhs_files == {e for e in entries if fs.isfile(e)}
-            
+
     def test_read_empty_file(self, scenario: ArchiveTestScenario):
         with scenario.provider(archive_data) as archive:
             fs = fsspec.filesystem(scenario.protocol, fo=archive)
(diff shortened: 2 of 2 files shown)


martindurant:
Please run pre commit hooks to appease the linter - otherwise looks good.

dlwh:
I'm a bit confused... I don't see a pre-commit hook? Happy to lint of course.

```
~/src/filesystem_spec (master)$ ls .git/hooks
applypatch-msg.sample		post-update.sample		pre-merge-commit.sample		pre-receive.sample		update.sample
commit-msg.sample		pre-applypatch.sample		pre-push.sample			prepare-commit-msg.sample
fsmonitor-watchman.sample	pre-commit.sample		pre-rebase.sample		push-to-checkout.sample
~/src/filesystem_spec (master)$
```

martindurant:
You would run
```
pre-commit install
```
but just fixing the whitespace as shown in https://github.com/fsspec/filesystem_spec/runs/6229522463?check_suite_focus=true#step:4:73 would do fine.

dlwh:
ah i see, it's a special package, not a git precommit hook.

martindurant:
The effect of running that is to install pre commit hooks locally :)

dlwh:
ah, ok, thanks! sorry i'm still pretty noob-ish in the python ecosystem.

PR #956, opened 2022-04-30
fixes #725: open with ZstdDecompressionReader in binary mode

### LINKED ISSUE #725: OpenFile.open() fails if the underlying file-like object has read-only attributes
Hi ! First, thanks for this amazing lib :)

Here is my issue:

The zstd decompressor has read only attributes, so I'm having this error in fsspec==2021.7.0 and zstandard==0.15.2:

```python
import fsspec

url = "https://the-eye.eu/public/AI/pile/test.jsonl.zst"
f = fsspec.open(url, "rb", compression="zstd").open()
```
raises
```python
File "/Users/user/.virtualenvs/fsspec-dev/lib/python3.7/site-packages/fsspec/core.py", line 145, in open
    out.close = close
AttributeError: 'zstd.ZstdDecompressionReader' object attribute 'close' is read-only
```
This error appears only if we use `.open()`, not if we use a `with` context.

Note that even if this error is patched, there is another error when reading a zstd file in "rb" mode that raises `UnsupportedOperation` but this is unrelated to `fsspec`. See https://github.com/indygreg/python-zstandard/issues/136

dlwh:
Implemented more or less along the lines that @martindurant suggested in the issue. I didn't try to keep pickleability, as the ZstdDecompressionReader isn't pickleable anyway, so it didn't seem like a major thing to go for.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -18,6 +18,7 @@ from .caching import (  # noqa: F401
 from .compression import compr
 from .registry import filesystem, get_filesystem_class
 from .utils import (
+    IOWrapper,
     _unstrip_protocol,
     build_name_function,
     infer_compression,
@@ -147,7 +148,11 @@ class OpenFile(object):
             closer()  # original close bound method of the final file-like
             _close(fobjects, mode)  # call close on other dependent file-likes
 
-        out.close = close
+        try:
+            out.close = close
+        except AttributeError:
+            out = IOWrapper(out, lambda: _close(fobjects, mode))
+
         return out
 
     def close(self):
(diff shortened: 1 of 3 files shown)


martindurant:
Awkward, but works! Thank you :)

dlwh:
yeah i don't love it either... The only thing that seemed cleaner was to always wrap, but it seemed more disruptive.

martindurant:
Yep. We'll keep an eye out on whether there's any downside to this, perhaps we'll have to revert and disable zstd.
On the other hand, we could be using cramjam for all decompression, and then we can do what we like!

Issue #957, opened 2022-05-02
not all globstrings are supported in open_files


ghislainp:
open_files is insensitive to some globstring:

fspec.open_files("filename*.txt") --> it's working fine.
fspec.open_files("filename?.txt") --> it's not working.
fspec.open_files("filename[0-9].txt") --> it's not working.

I'd expect all these globstrings to be expanded, as documented in the glob method.

The reason seems to be in get_fs_token_paths:

    if isinstance(paths, (list, tuple, set)):
        paths = expand_paths_if_needed(paths, mode, num, fs, name_function)
    else:
        if "w" in mode and expand:
            paths = _expand_paths(paths, name_function, num)
        elif "*" in paths:                                                   <------- HERE. Are more tests needed for "?" and "["  ?
            paths = [f for f in sorted(fs.glob(paths)) if not fs.isdir(f)]
        else:
            paths = [paths]

martindurant:
You are right, that line could probably use a `glob.has_magic` function call checking for all the characters that imply glob expansion should happen.

PR #961, opened 2022-05-08
improved handling of glob when reading files

ghislainp:
Solve #957

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -533,7 +533,7 @@ def strip_protocol(urlpath):
 
 
 def expand_paths_if_needed(paths, mode, num, fs, name_function):
-    """Expand paths if they have a ``*`` in them.
+    """Expand paths if they have a ``*`` in them (write mode) or any of ``*?[]`` in them (read mode).
 
     :param paths: list of paths
     mode: str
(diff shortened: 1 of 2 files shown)


martindurant:
One flake error:

fsspec/core.py:536:89: E501 line too long (97 > 88 characters)

ghislainp:
I don't understand the complexity of this code in get_fs_token_paths  (core.py):

```python
    if isinstance(paths, (list, tuple, set)):
        paths = expand_paths_if_needed(paths, mode, num, fs, name_function)
    else:  #->>>>
        if "w" in mode and expand:
        elif "*" in paths:
            paths = [f for f in sorted(fs.glob(paths)) if not fs.isdir(f)]
        else:
            paths = [paths]
```

It seems that expand_paths_if_needed  duplicates more or less the part of the code after #->>>> for the case of sequences. Spreading the same code in two functions seems error-prone in general, and indeed here, we have different behaviors depending on whether we have a list or a string (I'd expect the same behavior):

For instance:

```python
get_fs_token_paths(["dir/*.txt"], mode="w", expand=False)   #-> glob expansion is done, because expand_paths_if_needed don't have 'expand' arg.

get_fs_token_paths("dir/*.txt", mode="w", expand=False)   "-> glob expansion is done

```

Another unexpected difference:

```python
get_fs_token_paths(["dir/*.path"], expand=False)   #-> the directories found with glob("dir/*.path") are not excluded because expand_paths_if_needed does not exclude dirs.

get_fs_token_paths("dir/*.path", expand=False)   #->  the directories are excluded
```



I'd recommend to refactorize the code to get the same behavior for a list of a string:

```python
    if not isinstance(paths, (list, tuple, set)):   # or maybe better:  if isinstance(paths, str):

        paths = [paths]

    if expand:

        paths = expand_paths_if_needed(paths, mode, num, fs, name_function)

    paths = sorted([f for f in paths if not fs.isdir(f)])
```


This code does honor "expand" for both read and write. It also remove the directories for all modes.

PR #962, opened 2022-05-12
fix typo in asyn.py

adriangb:


martindurant:
Merging this, thanks.

Seems like a good time for us to throw out the older HDFS stuff out now.

Issue #835, opened 2021-11-22
[Feature Request] Fork Safety for async filesystems


alanhdu:
Hi there! Right now, async filesystems are explicitly incompatible with `fork`, presumably because `fsspec` tries to make sure that all the asyncio operations actually happen on a separate thread. See https://github.com/fsspec/filesystem_spec/blob/9ea19ba47dcee5c6a0c2435c9a776e8b67d7f7ef/fsspec/asyn.py#L305-L307

I was wondering whether there was some way that we could add limited support for forked processes. I understand that fork + threading gets into some tricky technical weeds, but supporting some level of fork would make for some integration with other projects easier.

In particular, I am trying to use an S3-backed zarr array with PyTorch (in order to train neural networks on some very large arrays) -- `torch.utils.data.DataLoader` does `os.fork` by default on Linux (see https://pytorch.org/docs/stable/data.html#single-and-multi-process-data-loading), and the s3fs is an async file system, making this an incompatible combination.

I have been able to hack around this with the following code:

```python
fs = ...   # S3 file system, but could be an other async filesystem
try:
    do_something(fs)
except RuntimeError as err:
    if "fork-safe" not in str(err):
        raise
    import asyncio
    from fsspec import asyn

    if fs._loop is not asyn.loop[0]:
        raise
    # fs._loop is the default loop which is running on a thread on the parent process
    # So let's spin up a new thread for this process.
    new_loop = None
    if len(asyn.loop) == 1:
        new_loop = asyncio.new_event_loop()
        asyn.loop.append(new_loop)  # Appends are thread-safe!
    if asyn.loop[1] is new_loop:   # We inserted first
        th = threading.Thread(target=new_loop.run_forever, name="fsspecIO")
        th.daemon = True
        th.start()
        asyn.iothread.append(th)

    fs._loop = asyn.loop[1]
    fs._pid = os.getpid()
```

but obviously this is reaching fairly deep into the internals of `fsspec`, which makes me uncomfortable, and we could add a more "official" way of doing this.

alanhdu:
[...]

Issue #709, opened 2021-07-15
Async http fs not working


dbernaciak:
With the latest changes the asynchronous http file system is throwing Loop not running when own loop provided. 

This is due to call to loop being set to None which is then caught in line 42 of asyn.py (2021.7.0)

'''python
async def run(loop):
   fs.fsspec.filesystem("http", asynchronous=True, loop=loop)
   fs.cat(["https://httpbin.org/ip"])

loop=...
loop.run_until_complete(run(loop))
loop.close()
'''

martindurant:
If you are explicitly running asynchronously within a coroutine, you do not need to specify the loop, and (it seems) probably should not. Your function would be
```python
async def run(loop):
   fs.fsspec.filesystem("http", asynchronous=True, loop=loop)
   await fs._cat(["https://httpbin.org/ip"])
```

dbernaciak:
Thanks. Even without loop=loop but with explicit asynchronous=True the error is still there. Seems like the events looped is None by default and then is not fetched from the coroutine.

martindurant:
Both of the following work for me with 2021.07.0. Perhaps you can be more specific about what you have done and the exception you see?
```python
import fsspec
import asyncio

loop = asyncio.get_event_loop()
async def run():
   fs = fsspec.filesystem("http", asynchronous=True)
   await fs._cat(["https://httpbin.org/ip"])

loop.run_until_complete(run()) # OK

async def run(loop):
   fs = fsspec.filesystem("http", asynchronous=True, loop=loop)
   await fs._cat(["https://httpbin.org/ip"])

loop.run_until_complete(run(loop)) # OK
```

dbernaciak:
You're right, typo on my side. This fixes the cat case. 

In my real use case I'm trying to put the http filesystem into pyarrow.ParquetDataset to fetch a file via http get request.

martindurant:
I'm pretty sure pyarrow will not support an asynchronous file system

dbernaciak:
I got it working in fsspec 0.8.7. It was performing very well. Now it seems that their fs wrapper is calling abstract asynchronous filesystem on fsspec side without setting the loop which is causing the Loop not running error I mentioned in the first message.

Note that even in the examples you got running you need to call the private method if you call public method .cat (as pyarrow does) the error appears.

martindurant:
It is not a private method, it is the asynchronous coroutine https://filesystem-spec.readthedocs.io/en/latest/async.html#using-from-async
[...]

PR #963, opened 2022-05-12
make the lock a singleton instead of allocating on import

mark-saad:


mark-saad:
This potentially fixes these issues:
https://github.com/fsspec/filesystem_spec/issues/835
https://github.com/fsspec/filesystem_spec/issues/709

martindurant:
Would you mind elaborating what difference this makes?

mark-saad:
The existing behavior allocates the threading lock whenever someone imports the asyn module [see](https://github.com/fsspec/filesystem_spec/blob/a0cd6aa8a782068e0819d95c891439246305873b/fsspec/asyn.py#L73)

This behavior is not fork friendly. When you fork a process you get the parent process's entire virtual memory, including loaded modules. If we want to be able to use this project with fork (which is the default on unix), we need to make it more friendly to fork. One way to do this is to fork early on, before initializing process variables (like the threading lock in this case).

When you run a snippet like this against the current code you will get the same lock object id
`
from multiprocessing import Pool
import time

from fsspec import asyn

def foo(x):
    time.sleep(3)
    return f"process {x} got lock {asyn.lock}"

with Pool(3) as p:
    print(p.map(foo, [1, 2, 3]))
`

If you run it against the new code you will get different object ids
`
from fsspec.asyn import get_lock

def foos(x):
    time.sleep(3)
    return f"process {x} got lock {get_lock()}"

with Pool(3) as p:
    print(p.map(foos, [1, 2, 3]))
`

martindurant:
I don't think fsspec is *ever* fork friendly. The only case (which your example just showed), is for when fsspec is imported, but not used. Is that the case you are concerned with?

mark-saad:
We actually don't import fsspec directly, we use s3fs and we traced the issue we're seeing to this line in fsspec. It used to be fork friendly, at least for our use case. Our use case is simply that we use a multiprocessing pool to get a bunch of s3 files all at once.
Generally speaking, it's better to avoid executing code on import.

martindurant:
>  we use a multiprocessing pool to get a bunch of s3 files all at once.

Why now simply use `S3FileSystem.cat(list_of_files)` (or `get` is writing to local files)? That works asynchronously internally, which is exactly what you want for an IO-bound operation like this.

> Generally speaking, it's better to avoid executing code on import.

I agree with this in general, although a lock is super cheap to make.

martindurant:
Am I right in thinking that *if* `get_lock` has been called in the parent process, then you will still fail here?

mark-saad:
That's a good point. To remedy that, we can expose a `reset_lock` function. This function can be called during the initialization of processes (either if the user is subclassing or using pool) to get a new lock for each new process.
If this makes sense to you, I can add this function fairly quickly.

martindurant:
If you explicitly have to forget the old lock, then you don't really gain anything. Perhaps get_lock should also store the PID, and take appropriate action if it doesn't match. On the other hand, you could just do what I tell most people: don't use fork! :)

mark-saad:
You're not "forgetting the old lock", you would be allocating a new lock with the creation of the new process. The old lock for the main process would still be there, unless you explicitly reset it.

I think the other alternative you're suggesting actually works better for this use case. I went ahead and implemented that. I'm essentially storing a dict `pid_to_lock` and checking if the current pid has already allocated a lock and stored it there. If not, one is created and stored. I'm not using a manager because it would make the whole thing more heavyweight. The locks are not shared between processes and we don't need to do that as far as I can tell.

swt2c:
> * if any async instance has been created, the newly forked processes must:
>   * discard references to locks, threads and event loops and make new ones
>   * not use any async fsspec instances from the parent process
>   * clear all class instance caches

Would it be reasonable to add something to the fsspec API that does all of the above (e.g. a `clear()` or `close()` that clears/closes all non-fork-safe resources)?

martindurant:
> Would it be reasonable to add something to the fsspec API

It would be reasonable to try. However, there is no way to remove instances that might be referenced from a variable in user code, or a general way to clear all classes' instance caches, unless we also had (say) a weak-dict of all classes that have ever been used, populated in `_Cached.__call__`.

mark-saad:
I think your point about this being racy is valid, but it's fixable in a number of ways. But before I get into how we can fix this, I think we need to agree on the need to have it fixed.

> the simplest solution is not to ever use fork
is not really viable for us, and for many other use cases as the other issues suggest. Fork is the default behavior on unix systems, so avoiding it altogether doesn't seem to be the right answer.

> using fork is safe, if fsspec has not been imported
This is generally not an accepted behavior in python. Code shouldn't execute on import. At least the original `get_lock` function avoided that, so if you're comfortable with that change I can revert back to it.

> if imported but not otherwise used, then the original version (before this PR) is safe
I'm not sure what you mean by that. The original version of this PR avoided the allocation on import, so it was safe to import but only use it within the multiprocessing pool. Before the PR it wasn't process-safe, and this is why we have the issue and others have it too.

Now to resolve the race condition, we can do one of these alternatives:
1. We can have that piece of getting the lock from the dict execute in a critical section.
2. Instead of having the map from process to lock, we can use the original get_lock function and expose a reset_lock that can be called on process init on systems that use fork. This would leave it up to the user to call this function if they need to.

martindurant:
Yes, I think 2) is the only reasonable answer: we provide a function that tries to do all the right things, and document that it should be called from forked child processed.

fsspec/asyn.py
```diff
@@ -16,7 +16,30 @@
...
+
 
+def reset_lock():
+    """Reset the global lock.
+
+    This should be called only on the init of a forked process to reset the lock to
+    None, enabling the new forked process to get a new lock.
+    """
+    global _lock
+    _lock = None
```
martindurant, line 42: Also
```
iothread[0] = None
loop[0] = None
```
?

mark-saad: sorry, my bad

fsspec/asyn.py
```diff
@@ -16,7 +16,35 @@
...
+    return _lock
+
 
+def reset_lock():
+    """Reset the global lock.
+
+    This should be called only on the init of a forked process to reset the lock to
+    None, enabling the new forked process to get a new lock.
+    """
+    global iothread
```
martindurant, line 41: nitpick: the lists do not need to be `global`, since we are mutating them rather than assigning

mark-saad: https://docs.python.org/3/reference/simple_stmts.html#the-global-statement
> It would be impossible to assign to a global variable without `global`

I've always understood this as "if you don't have the global statement, the change to the variable won't stick beyond the current function"

mark-saad: https://stackoverflow.com/questions/7060711/accessing-module-level-variables-from-within-a-function-in-the-module

martindurant:
```
In [1]: mylist = [None]

In [2]: def set_val(x):
   ...:     mylist[0] = x

In [3]: def get_val():
   ...:     return mylist[0]

In [4]: set_val(1)

In [5]: get_val()
Out[5]: 1
```

fsspec/asyn.py
```diff
@@ -16,6 +16,35 @@
...
+
+def reset_lock():
+    """Reset the global lock.
+
+    This should be called only on the init of a forked process to reset the lock to
+    None, enabling the new forked process to get a new lock.
+    """
+    global iothread
+    global loop
+    global _lock
```
martindurant, line 43:
```suggestion
```

fsspec/asyn.py
```diff
@@ -16,6 +16,35 @@
...
+
+    iothread = [None]
+    loop = [None]
```
martindurant, line 46:
```suggestion
    iothread[0] = None
    loop[0] = None
```

martindurant:
OK, will merge this now. For the typical non-fork case, it makes essentially no difference (the lock is created later), and anyone else who asks about fork, we can point to the new function and this discussion.

PR #964, opened 2022-05-15
uniform handling of path argument gievn as a list of strings or a single string in read mode

ghislainp:
Issue #957 was solved by PR #961 when path is a list of string but not when path is a string.
This PR solves this problem.

However, it also changes other behaviors. For instance, paths were sorted and filtered with not fs.isdir(f) only in read mode and only if a string was given with a wildcard. i.e. list of paths in read mode was not filtered+sorted. Write mode was not filtered+sorted. A single path without wildcard was not filtered+sorted.

This PR is an improvement imho with a more uniform handling of all these cases. dirs are filtered out in all the case, and sorting is also applied in all the cases. However, I hope this does not break expected behaviors that I'm ignoring.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -664,15 +664,11 @@ def get_fs_token_paths(
             update_storage_options(options, storage_options)
             fs = cls(**options)
 
-    if isinstance(paths, (list, tuple, set)):
+    if not isinstance(paths, (list, tuple, set)):
+        paths = [paths]
+    if expand:
         paths = expand_paths_if_needed(paths, mode, num, fs, name_function)
-    else:
-        if "w" in mode and expand:
-            paths = _expand_paths(paths, name_function, num)
-        elif "*" in paths:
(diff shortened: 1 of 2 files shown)


ghislainp:
In fact, it is not convenient to always filter out the directory. I'm now using get_fs_token_paths in a case where the target is a (zarr) directory. I think adding a boolean argument 'filter' like 'expand' is the most versatile option.

martindurant:
Circling back here. I think it would make a lot of sense to write out the cases you allude to as tests, so that we can confirm that the behaviour we see is what we want. 
So we can have a single path or a list of paths as input, and any of these may contain wildcard(s); these may be flagged for expanding, and we may want different behaviours on write (when the paths are expected not to yet exist of be freely over-writable) versus read.

I'm not sure we can add any additional flags to this function, or at least we can't changed the default behaviour, since many other things depend on it in this codebase and elsewhere.

PR #972, opened 2022-05-24
allow passing of atime, ctime and mtime to fuse

PatrikHlobil:
In order to support more of the Fuse compatibility, this MR allows the optional passing of the atime, ctime and mtime from the FileSystem to the Fuse protocol.


martindurant:
This is fine, but you will find that the various implementations don't agree on what to call these keys, so I expect you won't have correct information in most cases. You might be interested to look at the implementations important to you and have them add these keys to the info dict.

Issue #579, opened 2021-03-23
Inconsistent behavior with fsspec.open with mode='r' vs. mode='rb'


rabernat:
It's unclear to me how to use `fsspec.open` as a contextmanager. The behavior of this function seems inconsistent. The inconsistency is somehow related to the `mode` argument. I will try to illustrate this with an example.

### Create a test file

Bypass fsspec completely
```python
fname = 'test.txt'
with open(fname, mode='w') as f:
    f.write('hello')
```

### Open the file with a FileSystem instance

When I open the file via an instantiated FileSystem instance, everything works as I expect

```python
fs = fsspec.implementations.local.LocalFileSystem()
with fs.open(fname, mode='r') as fp:
    print(type(fp))
# -> <class '_io.TextIOWrapper'>
with fs.open(fname, mode='rb') as fp:
    print(type(fp))
# -> <class '_io.BufferedReader'>
```

The objects yielded to the context manager look like standard open file objects and can be used as such throughout python.

### Open the file via `fsspec.open`

```python
with fsspec.open(fname, mode='r') as fp:
    print(type(fp))
# -> <class '_io.TextIOWrapper'>
with fsspec.open(fname, mode='rb') as fp:
    print(type(fp))
# -> <class 'fsspec.implementations.local.LocalFileOpener'>
```

With `mode='r'`, the `fsspec.open` yields object is the same type of object as `fs.open`. **But with `mode='rb'`, we get a `LocalFileOpener`**. ⚠️  this is the key problem. In order to get a `BufferedReader`, we need an additional context manager!

```python
with fsspec.open(fname) as fp:
    with fp as fp2:
        print(type(fp2))
# -> <class '_io.BufferedReader'>
```

I can't figure out what this `LocalFileOpener` object is. It's not documented in the [API docs](https://filesystem-spec.readthedocs.io/en/latest/api.html). Most importantly for my purposes, xarray can't figure out what to do with it if I pass it to `open_dataset`. In contrast, it handles an `_io.BufferedReader` object fine.

### Proposed resolution

I would propose to remove the inconsistency and have `with fsspec.open(fname, mode='rb')` yield an `_io.BufferedReader` object. However, I recognize this could be a breaking change for some applications that rely on the `LocalFileOpener`.

martindurant:
[...]

PR #973, opened 2022-05-28
fix httpfile serialization bug

rabernat:
This implements the suggestion in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/370#issuecomment-1140028052, which fixes a bug in how `HTTPFile` is serialized.

It's a bit worrying to me that serialization of `HTTPFile` is not covered by tests sufficiently to have already surfaced this bug. Therefore, I would like to use this PR to contribute some improvements to testing as well.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -350,7 +350,7 @@ class LocalFileOpener(io.IOBase):
 
     def __enter__(self):
         self._incontext = True
-        return self.f.__enter__()
+        return self
 
     def __exit__(self, exc_type, exc_value, traceback):
         self._incontext = False
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -634,7 +634,7 @@ class HTTPFile(AbstractBufferedFile):
                 self.url,
                 self.mode,
                 self.blocksize,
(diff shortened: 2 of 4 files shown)


martindurant:
I think this PR is fine, and solves one arm of he possibilities in https://github.com/fsspec/filesystem_spec/issues/579 .

Another possible thing to consider, is making OpenFile file-like, calling into `.open()` when required. Seems like a rather large change to the core API.

fsspec/implementations/local.py
```diff
@@ -350,7 +350,7 @@ def __getattr__(self, item):
 
     def __enter__(self):
         self._incontext = True
-        return self.f.__enter__()
+        return self
```
rabernat, line 353: Am I correct in assuming that this will solve 

> `open_file = fsspec.open(local_path) -> <OpenFile '/local/path.nc'>`
> ❌  doesn't work with context manager (TypeError: cannot pickle '_io.BufferedReader' object)

from https://github.com/fsspec/filesystem_spec/issues/579#issuecomment-1140377495

rabernat: Update: just re-ran my test and the answer is YES! ✅

PR #975, opened 2022-06-06
Add truncate arg into LocalFileSystem touch()

KoyamaSohei:
We try to implement a fsspec-backend and are digging into the `AbstractFileSystem` spec.

I was reading the implementation of `LocalFileSystem` for reference, but there were some differences from the specifications.

If the implementation is as [specified](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.touch), the following test should pass, but it didn't.

This PR fix the implementation so that it does not ignore the truncate argument.

```
def test_local_touch_truncate():
    f = fsspec.filesystem("file")
    f.touch('/tmp/a', truncate=True)

    f.pipe('/tmp/a', b'a')
    f.touch('/tmp/a', truncate=True)
    assert f.cat('/tmp/a') == b''

    f.pipe('/tmp/a', b'a')
    f.touch('/tmp/a', truncate=False)
    assert f.cat('/tmp/a') == b'a'
```

```
_____________________________________________________________________ test_local_touch_truncate _____________________________________________________________________

    def test_local_touch_truncate():
        f = fsspec.filesystem("file")
        f.touch('/tmp/a', truncate=True)
    
        f.pipe('/tmp/a', b'a')
        f.touch('/tmp/a', truncate=True)
>       assert f.cat('/tmp/a') == b''
E       AssertionError: assert b'a' == b''
E         Full diff:
E         - b''
E         + b'a'
E         ?  
```

Regards,

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -158,7 +158,7 @@ class LocalFileSystem(AbstractFileSystem):
             self.makedirs(self._parent(path), exist_ok=True)
         return LocalFileOpener(path, mode, fs=self, **kwargs)
 
-    def touch(self, path, **kwargs):
+    def touch(self, path, truncate=True, **kwargs):
         path = self._strip_protocol(path)
         if self.auto_mkdir:
             self.makedirs(self._parent(path), exist_ok=True)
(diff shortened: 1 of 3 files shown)


martindurant:
Please create a test that shows this flag in action, like you have already shown in your comment. Otherwise good to go.

KoyamaSohei:
Thanks @martindurant  for the review :)
I just added the test.

martindurant:
Confirmed that the gcsfs failure is a new problem with gcsfs ( https://github.com/fsspec/gcsfs/pull/477 ) so no need for it to hold up this PR.

PR #977, opened 2022-06-09
Update setup.py

tsv19su254052:
corrected indents and hyphens


tsv19su254052:
Yes

martindurant:
For me, `black` says that the existing setup.py format is fine.

```
> black setup.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

tsv19su254052:
OK

PR #978, opened 2022-06-09
Update versioneer.py

tsv19su254052:
corrected indents and hyphens


martindurant:
Similarly, `black` (v22.3.0) is happy with the formatting in versioneer.py .

PR #980, opened 2022-06-10
spec: fix unstrip_protocol

### LINKED ISSUE #979: bug: unstrip_protocol incorrectly returns original path
In version 2022.5.0, `AbstractFileSystem.unstrip_protocol` fails to preprend `protocol://` to the path if the path happens to start with the characters of one of the class's protocols. For example:
```python
from fsspec.spec import AbstractFileSystem

AbstractFileSystem().unstrip_protocol('abstract-file')
# 'abstract-file'

from s3fs import S3FileSystem

S3FileSystem().unstrip_protocol('s3-file')
# 's3-file'

S3FileSystem().unstrip_protocol('s3a-file')
# 's3a-file'
```

It works correctly any time the beginning of the path does not match the protocol:
```python
AbstractFileSystem().unstrip_protocol('aabstract-file')
# 'abstract://aabstract-file'

S3FileSystem().unstrip_protocol('ss3-file')
# 's3://ss3-file'
```

Instead of returning the original path any time it starts with one of the protocols, the function should instead check for `protocol + "://"`.

jonburdo:
Fixes #979

Now the function only returns the original path if it starts with `f"{protocol}://"` instead of just checking for `"protocol"`.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -187,14 +187,11 @@ class AbstractFileSystem(metaclass=_Cached):
 
     def unstrip_protocol(self, name):
         """Format FS-specific path to generic, including protocol"""
-        if isinstance(self.protocol, str):
-            if name.startswith(self.protocol):
+        protos = (self.protocol,) if isinstance(self.protocol, str) else self.protocol
+        for protocol in protos:
+            if name.startswith(f"{protocol}://"):
                 return name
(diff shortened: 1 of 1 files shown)


efiop (approved):

martindurant:
Thank you

PR #983, opened 2022-06-16
memfs: add global_store flag

efiop:
By default, all instances of `MemoryFileSystem` use the same global store,
which has a few implications for users. For example:

* "memory leaks" if forgetting to delete files after use
* degrading performance (e.g. `info()` that looks through all entries in
  `store` and  `pseudo_dirs`) if storing 1K+ files (our use-case)

This PR allows one to choose whether to use a global store or not, allowing
for easy cleanup and more managable performance implications.

Having thousands of files stored in memfs would still result in a significant
performance hit, especially with complex tree structure. For cases like that,
we might want to create a trie-based memfs, that would be slightly slower to
build, but should be much faster and easier to traverse and read. We don't
have a need for this just yet, but we've ran into this problem in scmrepo's
gitfs, where we are using pygtrie ([1]).

[1] https://github.com/iterative/scmrepo/blob/main/scmrepo/fs.py

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -13,14 +13,25 @@ logger = logging.Logger("fsspec.memoryfs")
 class MemoryFileSystem(AbstractFileSystem):
     """A filesystem based on a dict of BytesIO objects
 
-    This is a global filesystem so instances of this class all point to the same
-    in memory filesystem.
+    Parameters
+    ----------
+    global_store : bool (True)
+        Whether to use global store, so instances of this class all point to
+        the same in memory filesystem. If False, every instance will have its
+        own store.
     """
 
-    store = {}  # global, do not overwrite!
-    pseudo_dirs = [""]  # global, do not overwrite!
+    store = {}  # global
+    pseudo_dirs = [""]  # global
     protocol = "memory"
     root_marker = "/"
+
+    def __init__(self, *args, **storage_options):
+        super().__init__(*args, **storage_options)
+        if not storage_options.get("global_store", True):
+            # detach from global
+            self.store = {}
+            self.pseudo_dirs = [""]
 
     @classmethod
     def _strip_protocol(cls, path):
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -2,6 +2,8 @@ import os
 
 import pytest
 
+from fsspec.implementations.memory import MemoryFileSystem
+
 
 def test_1(m):
     m.touch("/somefile")  # NB: is found with or without initial /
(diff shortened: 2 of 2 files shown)


fsspec/implementations/memory.py
```diff
@@ -9,19 +9,33 @@
 
 logger = logging.Logger("fsspec.memoryfs")
 
+_global_store = {}
+_global_pseudo_dirs = [""]
+
```
efiop, line 14: If backward compatibility wouldn't be a concern, I would rather just make `MemoryFileSystem` not use global store at all and would create a global memfs instance that people could use. But oh well 🤷

fsspec/tests/test_mapping.py
```diff
@@ -44,7 +44,7 @@ def test_getitems_errors(tmpdir):
 
 
 def test_ops():
-    MemoryFileSystem.store.clear()
```
efiop, line 47: Hm, I wonder if anyone is using this outside of tests 🤔 Worried about breaking backward compatibility for someone. An alternative might be to create `memfs2` or `LocalMemoryFileSystem` or something like that instead of altering `MemoryFileSystem` with this change. Could also keep it as it was before and overwrite `self.store` in `__init__`  if `global_store=False`

fsspec/implementations/memory.py
```diff
@@ -13,15 +13,26 @@
...
+    pseudo_dirs = [""]  # global
     protocol = "memory"
     root_marker = "/"
 
+    def __init__(self, *args, **storage_options):
+        super().__init__(*args, **storage_options)
+        if not storage_options.get("global_store", True):
+            # detach from global
+            self.store = {}
+            self.pseudo_dirs = [""]
```
efiop, line 34: It is interesting that this new fs "creation" is pretty different from what we do usually with filesystems. Reference fs seems similar, because you pass it the layout that it should be created from. Here we could also do smth like `MemoryFileSystem(store={})`, but that seems to be out of scope.

martindurant:
I don't immediately have time to look at this, but I have a couple of thoughts:
- since this is mostly used in testing, we can be certain to be able to make backward compatible changes if we choose
- I would start with the new option off by default and deprecate slowly over a couple of releases
- if `fsspec.filesystem("memory", global_store=False)` is called twice, you will get the same instance and same store anyway, due to caching of instances. You might view this as a benefit or flaw.
- `store={}` (or `None` would be better) is an interesting option, but do we also need a pseudodirs in this case? To be investigated what happens when calling  `fsspec.filesystem("memory", store=x)` twice, where `x` is the same dict, but mutated from one call to the next. Should you get the same instance, or two instances that happen to share the same store?

efiop:
Great points! Ok, so if this is mostly a debugging filesystem, I think I'll close this PR for now and we'll play a bit with Trie-based fs in https://github.com/iterative/morefs until deciding to maybe upstream it or contribute it instead of existing memfs.

PR #984, opened 2022-06-16
tox: gcsfs: fix build

efiop:
Makes so that the tests at least run. gcsfs fix is https://github.com/fsspec/gcsfs/pull/479


efiop:
@martindurant Looks like it, yes.

I'll merge this one for now so that the tests at least run.

PR #988, opened 2022-07-07
Add adl, abfs and abfss to infer_storage_options

SajidAlamQB:
Parsing filepaths in `infer_storage_options` with azure protocols, "adl", "abfs", "abfss" doesn't return the expected path. For example, the filepath, `adl://bucket/file.txt`, returns `{"path": "/file.txt"}` while the filepath, `s3://bucket/file.txt`,  returns `{"path": "bucket/file.txt"}`. 

To fix this we should add the extra protocols to the `if` statement that checks for cloud protocols.

### DIFF
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -204,7 +204,7 @@ def test_infer_options():
     # include it in the path. Test that:
     # - Parsing doesn't lowercase the bucket
     # - The bucket is included in path
-    for protocol in ["s3", "s3a", "gcs", "gs"]:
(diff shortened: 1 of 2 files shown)


martindurant:
@hayesgb @TomAugspurger , what is the expected behaviour from your point of view?

TomAugspurger:
Matching the behavior of s3 / gcs seems reasonable.

Perhaps as a followup, though, `fsspec` core shouldn't have to know anything about a particular implementation. Ideally it would have a registry or method to determine which path to take in this if / else block.

TomAugspurger:
I'll leave that to Martin. That said, it would be good to have a test here, to avoid future regressions.

martindurant:
@efiop , this is the kind of thing you were talking about :)

PR #990, opened 2022-07-12
Use cramjam for compression

### LINKED ISSUE #943: compressors to cramjam
Just as fastparquet found, having all the compressors in the fast and compact cramjam library really makes things simple. I am suggesting that, if cramjam is installed, we should get all de/compressors from there, rather than import all of the optional zstandard, snappy, lz4, ...

milesgranger:
Will close #943

This adds [cramjam](https://github.com/milesgranger/pyrus-cramjam) as a dependency, of which I'm the maintainer of.

Keeping `isal.igzip` as it's very performant on supported systems.


martindurant:
@fsspec/maintainers , what do we feel about adding a requirement? This one has no further requirements, and would guarantee (better) compression across a range of algorithms without needing the individual optional extra packages. We could leave it an optional dep, but then would anyone know to include it?

By the way, this touches on whether we should have a fsspec-complete metapackage that pulls in the most common backends and would be aplce to include cramjam is not as a direct dep.

martindurant:
@milesgranger , that's a thumbs up ^ 

Are you willing to finish this PR off? I think you'll need to add cramjam to the deps in the tox setup. Also, you should run pre-commit to appease the linters.

milesgranger:
Happy to finish it, although I'm (mostly) on vacation so any responses or updates will come sporadically. Also, not offended if you wanted to push updates. 👍

milesgranger:
Ready for a proper review, tests pass (on my machine... 😜)

martindurant:
We have had a discussion and decided that it would be better to have cramjam as optional for the following reasons:
- fsspec is special in not having any dependencies
- cramjam is binary, so although it is pretty simple to build and exists on all major platforms, there will still be a lag for new versions of python or missing architectures, whereas the pure-python fsspec would already be otherwise installable
- compression is not *that* commonly used

So, I think it's fine for cramjam to replace the current use of snappy, zstd, lzma..., but fsspec should import fine without it, and give a decent message when attempting to decompress without cramjam available. It would be nice if gzip and bzip2 still worked in these cases, using builtins.

milesgranger:
Can appreciate the reasoning. 👍 

I'm maybe not satisfied with the implementation of the conditional imports/error raising... will take another look at it later; feel free to make any suggestions if you'd like. Also, the last one failing on windows seems weird, did I miss a tox config somewhere for windows?

PR #993, opened 2022-07-19
Pass kwargs through pipe_file()

### LINKED ISSUE #992: Pass **kwargs into pipe_file()
Is there a reason why `pipe_file()` silently drops **kwargs here? https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L679

could this be changed to
```
def pipe_file(self, path, value, **kwargs):
    """Set the bytes of given file"""
    with self.open(path, "wb", **kwargs) as f:
        f.write(value)
```
Happy to open a pr for this!
I'm trying to pass in `content_type="image/svg+xml"` here and it's being silently ignored.

https://github.com/fsspec/adlfs/issues/294

hungcs:
closes #992


martindurant:
Thanks

PR #995, opened 2022-07-22
memory.py: Avoid bytes copies where possible

eode:
## Changes and reasoning
There are three cases (one of them used in a few locations) that cause a `BytesIO` (and therefore a `MemoryFile`) to become 'dirty', in the sense that CPython thinks a write has occurred, and therefore makes a copy of the underlying data when one is not necessary.

* Checking file length
  * `memoryfile.getbuffer().nbytes` creates a writable buffer
    * at this point Python assumes that a write has occurred, which forces a copy of the data when the data is read
  * Change: Created a function `_flen()` that uses `f.seek()` and `f.tell()` to check the length of the data
* Creation from another `BytesIO` or `MemoryFile` object
  * `newfile = MemoryFile(mem_fs, path, oldfile.getbuffer())` creates a writable buffer
    * this buffer is immediately copied from to create the contents of the new `BytesIO` or `MemoryFile`
    * this also forces `oldfile` to make a copy if `oldfile` is read again.
  * Change: Use `newfile = MemoryFile(mem_fs, path, oldfile.getvalue()`
    * if `oldfile` was previously modified, one copy of the data is made
    * if `oldfile` was not modified, both files use a reference to the same `bytes` object, and no copies are made
* Creation of a `MemoryFile`
  * `MemoryFile.__init__()` creates a buffer and then copies the new data into it, rather than calling `super()`
    * This causes a copy where none is needed when `data` is a `bytes` object
  * Change: Use `super().__init__(data)`
    * Whenever `data` is a `bytes` object, it will be used without a copy

## Notes
* `_flen()` could arguably be better placed as a classmethod, but I suppose that depends on project / style preferences -- happy to change it
* `_flen()` could instead be a magic method `MemoryFile.__len__()`, but this seemed to blur string and filelike object APIs too much, and would probably be considered too big of a change for a PR of this limited scope.

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -154,9 +154,7 @@ class MemoryFileSystem(AbstractFileSystem):
             filelike = self.store[path]
             return {
                 "name": path,
-                "size": filelike.size
-                if hasattr(filelike, "size")
-                else _flen(filelike),
+                "size": filelike.size if hasattr(filelike, "size") else _flen(filelike),
                 "type": "file",
                 "created": getattr(filelike, "created", None),
             }
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -38,7 +38,7 @@ class MemoryFileSystem(AbstractFileSystem):
             return [
                 {
                     "name": path,
-                    "size": self.store[path].getbuffer().nbytes,
+                    "size": _flen(self.store[path]),
                     "type": "file",
                     "created": self.store[path].created,
                 }
@@ -53,7 +53,7 @@ class MemoryFileSystem(AbstractFileSystem):
                     out.append(
                         {
                             "name": p2,
-                            "size": self.store[p2].getbuffer().nbytes,
+                            "size": _flen(self.store[p2]),
                             "type": "file",
                             "created": self.store[p2].created,
                         }
@@ -147,7 +147,7 @@ class MemoryFileSystem(AbstractFileSystem):
                 "name": path,
                 "size": filelike.size
                 if hasattr(filelike, "size")
-                else filelike.getbuffer().nbytes,
+                else _flen(filelike),
                 "type": "file",
                 "created": getattr(filelike, "created", None),
             }
@@ -193,7 +193,7 @@ class MemoryFileSystem(AbstractFileSystem):
         path1 = self._strip_protocol(path1)
         path2 = self._strip_protocol(path2)
         if self.isfile(path1):
-            self.store[path2] = MemoryFile(self, path2, self.store[path1].getbuffer())
+            self.store[path2] = MemoryFile(self, path2, self.store[path1].getvalue())
         elif self.isdir(path1):
             if path2 not in self.pseudo_dirs:
                 self.pseudo_dirs.append(path2)
@@ -246,7 +246,7 @@ class MemoryFile(BytesIO):
         self.path = path
         self.created = datetime.utcnow().timestamp()
         if data:
-            self.write(data)
+            super().__init__(data)
             self.size = len(data)
             self.seek(0)
 
@@ -263,3 +263,11 @@ class MemoryFile(BytesIO):
 
     def commit(self):
         self.fs.store[self.path] = self
+
+
+def _flen(filelike):
+    """Find length of file without causing a write condition"""
+    pos = filelike.tell()
+    size = filelike.seek(0, 2)
+    filelike.seek(pos)
+    return size
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -184,7 +184,7 @@ class MemoryFileSystem(AbstractFileSystem):
             else:
                 raise FileNotFoundError(path)
         if mode == "wb":
-            m = MemoryFile(self, path)
+            m = MemoryFile(self, path, kwargs.get('data'))
             if not self._intrans:
                 m.commit()
             return m
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -119,9 +119,7 @@ class MemoryFileSystem(AbstractFileSystem):
 
         Avoids copies of the data if possible
         """
-        with self.open(path, "wb", data=value, **kwargs) as f:
-            # value passed into initializer to avoid copy
-            assert f.getvalue() is value
+        self.open(path, "wb", data=value)
 
     def rmdir(self, path):
         path = self._strip_protocol(path)
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -114,6 +114,15 @@ class MemoryFileSystem(AbstractFileSystem):
             if not exist_ok:
                 raise
 
+    def pipe_file(self, path, value, **kwargs):
+        """Set the bytes of given file
+
+        Avoids copies of the data if possible
+        """
+        with self.open(path, "wb", data=value, **kwargs) as f:
+            # value passed into initializer to avoid copy
+            assert f.getvalue() is value
+
     def rmdir(self, path):
         path = self._strip_protocol(path)
         if path == "":


eode:
As a use-case:

We have an AWS Lambda function that handles images using `aicsimageio`.   This lambda has the bytes available to it, but not a file.  We are upgrading `aicsimageio`, and the new version uses `fsspec`, which is a great move forward.  However, their readers by and large no longer accept `bytes` or filelike objects, but expect `fsspec` paths.

`fsspec`'s `memory` implementation to the rescue -- but data handling time suffers, and each time a copy is made, there must be double the available memory for that to occur, without getting hacky.  Thus, this PR.


## Examples
```python
In [1]: from fsspec.implementations.memory import MemoryFile

In [2]: from io import BytesIO

In [3]: data = b'x' * 3_000_000   # 3 MB

In [4]: %timeit BytesIO(data)    # note -- nanoseconds
89.6 ns ± 0.774 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %timeit MemoryFile(path='x', data=data)    # note -- microseconds
427 µs ± 9.81 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

<snip -- import MemoryFIle from this PR>

In [12]: %timeit MemoryFile(path='x', data=data)   # note -- microseconds, but ~2 orders of magnitude faster
4.93 µs ± 36.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Note that the above example is with 3mb of data -- the larger the relevant files, the more of an issue this is.
```python

In [3]: data = b'x' * 30_000_000  # 30 MB

In [4]: %timeit MemoryFile(path='x', data=data)  # note -- milliseconds (!)
2.73 ms ± 41.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

<snip -- import MemoryFIle from this PR>

In [9]: %timeit MemoryFile(path='x', data=data)  # note -- microseconds
4.97 µs ± 21.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

Note with this PR, the above section's `In [9]` and the prior section's `In [12]` occur in substantially the same time, because it is simply a reference pass, and occurs in `O(1)` time, whereas the original code differs significantly between 3 MB and 30MB.

fsspec/implementations/memory.py
```diff
@@ -53,7 +53,7 @@ def ls(self, path, detail=False, **kwargs):
                     out.append(
                         {
                             "name": p2,
-                            "size": self.store[p2].getbuffer().nbytes,
```
martindurant, line 56: Since `.getvalue()` returns a view, how does `len(self.store[p2].getvalue())` compare? It's certainly simpler looking.

martindurant:
Thanks, and your improvements are definitely appreciated! This shows that memory:// has almost exclusively been used for ad-hoc and testing purposes, rather than as a full backend; but now we can make it better!

fsspec/implementations/memory.py
eode: Ok, so here were my considerations there:

I also like the look of that better, *and* it's even more performant if no buffer is extant.  However, if the user has a buffer somewhere, and the file's large, it can be sluggish.  

No writing has occurred and no buffer extant:
  * `_flen`: 228 ns
  * `getbuffer`: 169 ns
  * `getvalue`: 82.9 ns

Writing has occurred, or there is a buffer extant:
  * `_flen1: 228 ns
  * `_getbuffer`: 168 ns
  * `_getvalue`: 2.86 **ms**  < -- Ouch
  
  So the conclusion is:
  * `len(f.getbuffer())` is consistently a little on the faster side
  * `len(f.getvalue())` is faster when no writes have occurred, but terrible if dealing with a large file that has been written to or has a buffer extant
  * `_flen(f)` is consistently a little slower, but is consistent, and still decently fast.
  
That's why I went with `_flen` -- but yeah, it's not as appealing stylistically.  That's also why I considered making a magic `MemoryFile.__len__` so we could do `len(f)` (or more specifically, `len(self.store[p2])` -- but I didn't want to push for that since filelike objects don't typically support `len()`, and I thought that might be considered too weird of an API change.

eode:
I added an implementation for `pipe_file` to avoid copies when using `memfs.pipe()` or `memfs.pipe_file()`, and that may still need review.

fsspec/implementations/memory.py
```diff
@@ -114,6 +114,15 @@ def makedirs(self, path, exist_ok=False):
...
                 raise
 
+    def pipe_file(self, path, value, **kwargs):
+        """Set the bytes of given file
+
+        Avoids copies of the data if possible
+        """
+        with self.open(path, "wb", data=value, **kwargs) as f:
+            # value passed into initializer to avoid copy
+            assert f.getvalue() is value
```
martindurant, line 124: An `assert` in non-test code seems wrong to me - we know this should always work. Since this is within our own control, should we not just `self.store[path] = MemoryFile(data)` or `self.open(path, "wb", data=value, **kwargs)` (no context)? I don't actually think there are any valid kwargs here.

eode: Agreed.  ..and while I don't think there are an valid kwargs here, should we leave the `**kwargs` in place, on the potential of future changes?

eode: Followed recommendations here in this conversation, and both changes have been made.

fsspec/implementations/memory.py
eode: @martindurant 

That's on a 30 MB file, btw.

While I'm perfectly happy using `getvalue()` here, I suspect that people who have been doing writes will be disappointed with the cost of getting length (2860000ns vs 230ns is a big leap, even if occasional), plus there's the memory cost of a copy, albeit a temporary one.  I think the ~230ns range is pretty much fine, as it allows ~6k size checks per ms, and avoids copy.

Let me know your thoughts.  I really think the seek/tell method is best, but I'm happy to frame it differently.

A simple thing that would fit in with what's already there would be to implement `MemoryFile.size` as a property instead of as a variable.  This has the added benefit of providing the right size if someone has acquired the file and modified it, and then does a `memfs.ls('/containing_dir/', detail=True)`.

Let me know what you think..

martindurant: Let's replace both other uses of `_flen`, in that case. I might still add that function to fsspec.utils at another time, though, since it's an obvious thing one might want to do in some situations.

eode: With a `MemoryFile.size` property that does the same thing, yes?

martindurant: Happy to have a property, yes, but let's do it without `seek`

eode:
Not all issues from this had been addressed - specifically, the shift to a size property which was requested.

eode:
@martindurant So, it looks like this PR is rejected?

martindurant:
No not at all, I merely finished off your work in #999 and merged that instead, to get it in for the release. The mention of regression in #1004 turned out not to be due to these changes at all. Like [here](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L257) we initialise MemoryFile from data, this is your code.

martindurant:
Do you find your timings for operations on memoryfile in 2022.7.1 are back to what they were in 2022.5.0?

eode:
Sorry, that original bit was written on my phone, and I didn't have a good view of the code.  But my concerns are these:

* [line 213](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L213), which now converts the underlying bytes to a buffer and then back to bytes.  The original is `getvalue()[start:end]` and would likely be faster both with that specific call, and in that it doesn't cause a write state
* and [line 261](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L261), which causes a write state, where `len(self.getvalue()` would generally be faster

martindurant:
I am prepared to be proven wrong, but:

> [line 213](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L213), which now converts the underlying bytes to a buffer and then back to bytes. The original is getvalue()[start:end] and would likely be faster

This is not my understanding. The underlying storage of a BytesIO is *not* bytes. What we do now is get a view on that buffer, cut the part we want, and then copy that part. One possibly improvement would be to remove `[start:end]` in the case that both are `None`, but the previous version had getvalue, which made a bytes copy of the whole buffer, and then sliced it for a new copy

> and [line 261](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L261), which causes a write state, where len(self.getvalue()) would generally be faster and also avoids creating a write state

The whole point on this PR was to avoid getvalue, which makes a bytes (copy) whereas getbuffer gets a no-copy memoryview. What is a "write state"?

```
In [1]: from io import BytesIO

In [2]: data = b'x' * 30_000_000

In [3]: from fsspec.implementations.memory import MemoryFile

In [4]: %timeit MemoryFile(path='x', data=data)
10 µs ± 86.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: m = MemoryFile(path='x', data=data)

In [6]: %timeit m.size
215 ns ± 1.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

eode:
> >    [line 213](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L213), which now converts the underlying bytes to a buffer and then back to bytes. The original is getvalue()[start:end] and would likely be faster

> This is not my understanding. The underlying storage of a BytesIO is not bytes. What we do now is get a view on that buffer, cut the part we want, and then copy that part. One possibly improvement would be to remove [start:end] in the case that both are None, but the previous version had getvalue, which made a bytes copy of the whole buffer, and then sliced it for a new copy

The underlying storage of BytesIO is a buffer that uses the address of the `bytes` object given.  If a memoryview is given, a copy is made instead into a new buffer, and that new buffer is used.  That is, if you pass in a `bytes` object at address `0x12345`, then it uses `0x12345` as the underlying buffer.  If the buffer is to be written to, but is referenced elsewhere, it is detached, and a copy is made, and the copy becomes the new underlying buffer.  This is done using two different kinds of refcounts.  The states of the `BytesIO` object can be:

* No normal references.  All operations are fast.
* Some normal references.  All ***read*** operations are fast.  Writes and `memoryview` creation cause copies.
* Some export references (like `memoryview` objects) exist.  All large non-`memoryview` read operations are copies, and no non-`memoryview` writes may occur.

So to be clear: if only reading is done and no `memoryviews` are created, `BytesIO` makes no copies.

I think that the case where a `memoryview` is laying around is the responsibility of the `memoryview` creator, and is also comparatively uncommon.  Thus, I think that case can be excluded from consideration of general-use optimization.

I also think the case where *no* external references to the buffer exist can be excluded from consideration of optimization, since it is always fast, but can't be ensured to be the active state of the `BytesIO` object.  What we *can* do is ensure that we don't leave references to the data in `fsspec` code (we don't).

That leaves the case where the user has normal references to the buffer, likely as `bytes` objects. I think this is a fairly common case, but that's just an off-the-cuff guess.  In any case, it's an area we can affect pretty consistently.  The way we make that fast is by not doing write activities such as calling `x.write()` or creating a writable buffer as in `x.getbuffer()`.

On this specific issue, what that means is that when `x.getvalue()` is called, this action is not a write and has no side effects unless the user has a `memoryview` extant.  But `x.getbuffer()` has the side effect of making a copy if there are any normal references to the buffer.

However, it should be noted that when a `memoryview` ***is*** present externally, then using another `memoryview` to extract the bytes *is* faster, at least once you get past files of 8kb or so.  This, of course, becomes more dramatic as filesize increases.  However, again, my argument for this case is that the circumstance we should optimize for isn't the one where a `memoryview` is left open, but rather the one where it isn't.  Someone reading from an object that has an open `memoryview` must be careful and know what they're doing, anyways, and already has direct access to the buffer.

Example of the underlying buffer being used from the `bytes` object for a `BytesIO` object (and back):
```python
In [2]: data = b'abc' * 30                       
In [3]: data_addr = hex(id(data))                # memory address of bytes
In [4]: b = BytesIO(data)                        
In [5]: del data                                 # Delete reference to the bytes - now only the BytesIO has it
In [6]: b.seek(0)                                
Out[6]: 0                                        
In [7]: b.write(b'X')                            # Since no external reference exists, no copy is made on write.
Out[7]: 1                                        
In [8]: buffer_addr = hex(id(b.getvalue()))      # `getvalue()` returns `bytes` object of current buffer here
In [9]: buffer_addr == data_addr                 
Out[9]: True
In [10]: data = b.getvalue()
In [11]: buffer_addr == data_addr == hex(id(data))
Out[11]: True
```

>> and [line 261](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L261), which causes a write state, where len(self.getvalue()) would generally be faster and also avoids creating a write state

> The whole point on this PR was to avoid getvalue, which makes a bytes (copy) whereas getbuffer gets a no-copy memoryview. What is a "write state"?

Well, the point was more to avoid copies of memory.   The underlying situation is described above in this reply, but I was calling the following a 'write state':

* a circumstance where CPython must write to the buffer
* a circumstance where CPython must assume that the buffer has been modified
* a circumstance where CPython must assume that the buffer could be modified at any moment

Avoiding the above three circumstances is ideal, and means only references to the underlying buffer are used.

see [this BytesIO code comment](https://github.com/python/cpython/blob/main/Modules/_io/bytesio.c#L27) and related code re: exports in `_io_BytesIO_getvalue_impl` amongst others.

martindurant:
Thank you for the detailed explanation, I will take some time to digest that. 

I'll just note that `memoryview(bytes_object)` does *not* copy the original, but creates a read-only view. This is outside of the implementation of BytesIO, of course.

The main use cases of MemoryFile I foresee:
- open for write mode with an initially empty buffer (append is far less likely)
- open for random-access read mode (original file will no longer be written to)
- put/pipe to copy whole files or occasionally read file ranges.

I would expect the file's size only to be relevant during bullet 2, and when listing file details via the filesystem API (ls/info).

martindurant:
OK, I suppose I am convinced, and we should assume `bytes` for the buffer and use the get_len function for the size, except when creating the file via data=, in which case we already know the length. That should mean that put/pipe -> get/cat should be fast and also zero copy.

Issue #613, opened 2021-04-14
LocalFileSystem incompatibility / corruption when used to read FITS files


joseph-long:
When opening the same local file via `open()` and via `fsspec.open()` I get garbled floating point values from the latter when the library loads it. Confusingly, the ascii header of the file seems to come through fine, so it's at least *partially* loading the file... (I have implemented an [iRODS interface through fsspec](https://github.com/xwcl/irods_fsspec), and don't see the issue when opening files from iRODS, so as far as I know this is specific to local paths with fsspec.)

When read through fsspec instead of regular `open()`, the `astropy.io.fits.open` function is able to parse the header (a sort of ascii preamble) but the associated data are garbled. I'm trying to discourage `astropy.io.fits` from trying to memory map or seek or anything (for compatibility with other fsspec backends), hence the `mode='readonly', memmap=False` in the `fits.open` call.

I'm totally stumped, but I can reproduce with a short script, included below. It's possible this is an issue that Astropy needs to fix, but I can't figure out what would conceivably be different about the file proxy object that would let it read the header and not the data...

**To reproduce:**

1. Save this script as `repro.py`

```python
import fsspec
import irods_fsspec
irods_fsspec.register()
from astropy.io import fits

def main():
    from astropy.io import fits
    import numpy as np
    data = np.arange(10, dtype=float)
    fits.PrimaryHDU(data).writeto('ex.fits', overwrite=True)
    print('\nUsing builtin open()')
    with open('ex.fits', 'rb') as fh:
        hdul = fits.open(fh, mode='readonly', memmap=False)
        hdul.info()
        print(hdul[0].data)

    print('\nUsing fsspec.open() with local path')
    with fsspec.open('ex.fits', 'rb') as fh:
        hdul = fits.open(fh, mode='readonly', memmap=False)
        hdul.info()
        print(hdul[0].data)

if __name__ == "__main__":
    main()
```

2. Create a virtual env
```
python -m venv env
source env/bin/activate
pip install astropy fsspec
```

3. Compare output from two approaches to open `ex.fits`, e.g.

```
$ python repro.py

Using builtin open()
Filename: ex.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU       5   (10,)   float64
[0. 1. 2. 3. 4. 5. 6. 7. 8. 9.]

Using fsspec.open() with local path
Filename: /Users/josephlong/devel/repro/ex.fits
[...]

PR #1005, opened 2022-07-28
remove __fspath__ from LocalFileOpener

### LINKED ISSUE #1004: Regression in 2022.7.0 for local file handling with no bytes copy
Hey!

Seems like 2022.7.0 creates three errors in our test suite: https://github.com/AllenCellModeling/aicsimageio/runs/7566883918?check_suite_focus=true

Checked the version diffs from the yesterday's Check Upstreams and the only five package version changes were:
* `attrs-21.4.0` -> `attrs-22.1.0`
* `fsspec-2022.5.0` -> `fsspec-2022.7.0`
* `jsonschema-4.7.2` -> `jsonschema-4.8.0`
* `s3fs-2022.5.0` -> `s3fs-2022.7.0`
* `virtualenv-20.16.1` -> `virtualenv-20.16.2`

Since it's a file reading issue, the only ones that make sense to look at are fsspec or `s3fs`, from the test suite it seems like we were running local read tests and not remote read tests which to me says `fsspec` change likely culprit.

Looking through the changelog I think the likely culprit is #995 / #999 -- it also makes sense because I think @eode was working on some file reading with AICSImageIO stuff too??

Not entirely sure how to test because the error we receive on the other end is a `ValueError: seek of closed file` maybe @eode has ideas??

Much like our current "Test Upstreams" GH Action I am going to add fsspec and s3fs to the libraries tested to catch these before release.

martindurant:
Fixes #1004

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -1,5 +1,6 @@
 import datetime
 import io
+import logging
 import os
 import os.path as osp
 import posixpath
@@ -12,6 +13,8 @@ from fsspec import AbstractFileSystem
 from fsspec.compression import compr
 from fsspec.core import get_compression
 from fsspec.utils import isfilelike, stringify_path
+
+logger = logging.getLogger("fsspec.local")
 
 
 class LocalFileSystem(AbstractFileSystem):
@@ -246,6 +249,7 @@ class LocalFileOpener(io.IOBase):
     def __init__(
         self, path, mode, autocommit=True, fs=None, compression=None, **kwargs
     ):
+        logger.debug("open file: %s", path)
         self.path = path
         self.mode = mode
         self.fs = fs
@@ -343,10 +347,6 @@ class LocalFileOpener(io.IOBase):
     @property
     def closed(self):
         return self.f.closed
-
-    def __fspath__(self):
-        # uniquely among fsspec implementations, this is a real, local path
-        return self.path
 
     def __iter__(self):
         return self.f.__iter__()
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -249,6 +249,7 @@ class MemoryFile(BytesIO):
     """
 
     def __init__(self, fs=None, path=None, data=None):
+        logger.debug("open file %s", path)
         self.fs = fs
         self.path = path
         self.created = datetime.utcnow().timestamp()


martindurant:
This was causing the LocalFileOpener instances to evaluate as "pathlike", which is sometimes OK, but often means you are not operating on the same file instance you think you are.

cgohlke:
Is it expected that `insinstance(<class 'fsspec.core.OpenFile'>, os.PathLike)` still returns True?

```
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)] on win32
>>> import fsspec, os
>>> fsspec.__version__
'2022.7.1'
>>> f = fsspec.open('test.bin', 'rb')
>>> isinstance(f, os.PathLike)
True
>>> os.fspath(f)
  File "X:\Python310\lib\site-packages\fsspec\core.py", line 99, in __fspath__
    return self.open().__fspath__()
  File "X:\Python310\lib\site-packages\fsspec\implementations\local.py", line 355, in __getattr__
    return getattr(self.f, item)
AttributeError: '_io.BufferedReader' object has no attribute '__fspath__'
```

martindurant:
No, `__fspath__` was remove from LocalFileOpener. It had always been anomalous (no other file-like has it), and making `PathLike` pass the instance check was unintended - the object has no other behaviours that make it like a path, except that it does indeed point to a path string.

It looks like the method should have been removed from OpenFile too, which is causin the confusion (os.PathLike only checks for the method's existence, not that it runs).

```diff
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -94,10 +94,6 @@ class OpenFile(object):
     def __repr__(self):
         return "<OpenFile '{}'>".format(self.path)

-    def __fspath__(self):
-        # may raise if cannot be resolved to local file
-        return self.open().__fspath__()
-
     def __enter__(self):
         mode = self.mode.replace("t", "").replace("b", "") + "b"
```

barentsen:
Note: I believe a consequence of this PR is that local files opened using fsspec can no longer be passed to `numpy.fromfile`, because `fromfile` appears to rely on fspath/fileno somewhere in the C layer?!

As discussed in https://github.com/fsspec/filesystem_spec/issues/613, one workaround is to use `numpy.frombuffer(f.read())` instead of `fromfile`, so this may not be an issue that needs to be resolved, but I thought I'd document it here regardless.  See the examples below.
 
#### Example 1: `np.fromfile` worked before this PR (fsspec v2022.7.0)
```python
Python 3.9.7 (default, Sep  7 2021, 13:48:18)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> import fsspec
>>> fsspec.__version__
'2022.7.0'
>>> import numpy as np
>>> np.__version__
'1.23.1'
>>> path = "/tmp/tmp.npy"
>>> np.zeros((3, 3)).tofile(path)
>>> fh = fsspec.open(path).open()
>>> np.fromfile(fh)
array([0., 0., 0., 0., 0., 0., 0., 0., 0.])
```

#### Example 2: `np.fromfile` raises a `UnsupportedOperation` after this PR (fsspec v2022.7.1)

```python
Python 3.9.7 (default, Sep  7 2021, 13:48:18)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> import fsspec
>>> fsspec.__version__
'2022.7.1'
>>> import numpy as np
>>> np.__version__
'1.23.1'
>>> path = "/tmp/tmp.npy"
>>> np.zeros((3, 3)).tofile(path)
>>> fh = fsspec.open(path).open()
>>> np.fromfile(fh)
io.UnsupportedOperation: fileno
```

martindurant:
It is reasonable that a library that would take an fsspec file-like, but only use it to get the local filepath and then reopen the same file for its internal use, should no longer be doing that. I don't think it's what a user would expect, and they would be better off not using fsspec, or at least only use fsspec to generate the list of paths.

The specific case of numpy appears to be doing something different, though: it's trying to get a filenumber (descriptor)? This would be `fh.raw.fileno()` in this case, but I can't tell what numpy is actually calling. `f.fileno = f.raw.fileno` seems to make it work, so I propose that LocalFileOpener gets a `fileno` method.

barentsen:
The Numpy C layer is rather opaque to me, but it looks like Numpy's [array_fromfile](https://github.com/numpy/numpy/blob/45bc13e6d922690eea43b9d807d476e0f243f836/numpy/core/src/multiarray/multiarraymodule.c#L2265) calls [NpyPath_PathlikeToFspath](https://github.com/numpy/numpy/blob/3de252be1215c0f9bc0a2f5c3aebdd7ffc86e410/numpy/core/src/multiarray/methods.h#L14), which behaves differently now that `LocalFileOpener` no longer features the `__fspath__`  method.  This then causes `fileno` to be required somewhere down the chain instead.

I confirm adding the following to `LocalFileOpener` fixes the issue for me:

```python
    def fileno(self):
        return self.raw.fileno()
```

Would you like a PR for this, or would it be necessary to understand Numpy's behavior in more detail?

(Context: I came across this issue because `astropy.io.fits` is an avid user of `numpy.fromfile`.)

martindurant:
I think that should be a PR, yes, as it's a reasonable thing regardless of what numpy does with it. 

> because astropy.io.fits is an avid user of numpy.fromfile

fits does work with non-local file-like objects too, anything with seek/read. By providing the file descriptor, I suppose we are allowing memory-map access.

PR #1011, opened 2022-08-04
xrootd to root typo

ScottDemarest:
mis typed protocol name for fsspec implementation


martindurant:
Is this how xrootd connections are usually written? I have a slight worry that "root filesystem" might get misunderstood.

ScottDemarest:
i agree it's a confusing protocol name, but that is what it is. 
for example: root://eospublic.cern.ch/...

ScottDemarest:
Perhaps I should add some more detail to the error message to clear that up?

martindurant:
Yes, that seems like a good idea. More information is always good.

PR #1014, opened 2022-08-06
reversed path in _rm

ScottDemarest:
reversed path to allow proper recursion through directories like in AbstractFileSystem


nsmith-:
I think the issue here is that rm on a dir (for filesystems that have directories) requires it to be empty, which is enforced for the sync AbstractFileSystem by reversing the listing: https://github.com/fsspec/filesystem_spec/blob/e4352495bad852fff64949c183cf37757f62c232/fsspec/spec.py#L946-L948
Now indeed for async the calls are made concurrently so there may be a race later that would cause problems, so this might not be sufficient

martindurant:
OK, well I see no downside; so if it makes no difference for key-value stores but matters for root, happy to include this change.

PR #1016, opened 2022-08-08
Make sure read file closed in cached

### LINKED ISSUE #1015: SimpleCacheFileSystem: ResourceWarning: unclosed file
The snippet below raises a `ResourceWarning` (not per default - only when enabling the warning):

```python
import warnings
warnings.simplefilter("always")

import fsspec
import fsspec.implementations.cached

fs = fsspec.implementations.cached.SimpleCacheFileSystem(target_protocol="memory")

with fs.open("file1", "w") as f:
    f.write("a")
with fs.open("file2", "w") as f:
    f.write("a")

fs.cat("file*")
```

```python-traceback
.../fsspec/implementations/cached.py:591: ResourceWarning: unclosed file <_io.BufferedReader name='/tmp/...'>
  out[p] = open(fn, "rb").read()
ResourceWarning: Enable tracemalloc to get the object allocation traceback

```

I think the culprit is here

https://github.com/fsspec/filesystem_spec/blob/5d6c87e562396f352645df7cd555a2b4b6eec84a/fsspec/implementations/cached.py#L595

and could be fixed as:

```diff
-            out[p] = open(fn, "rb").read()
+            with open(fn, "rb") as f:
+                out[p] = f.read()
```

martindurant:
Fixes #1015

(there are probably more instances of this in the codebase)

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -592,7 +592,8 @@ class WholeFileCacheFileSystem(CachingFileSystem):
         out = {}
         callback.set_size(len(paths))
         for p, fn in zip(paths, fns):
-            out[p] = open(fn, "rb").read()
+            with open(fn, "rb") as f:
+                out[p] = f.read()
             callback.relative_update(1)
         if isinstance(path, str) and len(paths) == 1 and recursive is False:
             out = out[paths[0]]


mathause:
Thanks!

PR #1017, opened 2022-08-08
Allow writable ZipFileSystem

martindurant:


martindurant:
It appears to be necessary for the calling code to call `fs.zip.close()` to get the output to finalize, I am not sure why.

joshmoore:
That certainly matches the special close handling that was previously needed in Zarr: https://github.com/zarr-developers/zarr-python/blob/994f2449b84be544c9dfac3e23a15be3f5478b71/zarr/storage.py#L1225

TomasPuverle:
Hi @martindurant,

I just saw this PR in response to the discussion on one of the threads!  
This will probably make my life a lot easier.  Many thanks, I'll give this a try as soon as I can!

Tom

PR #1023, opened 2022-08-12
fix docstring typo

vmgustavo:
Just a simple docstring typo fix


martindurant:
This is actually correct. It might not be strict English, but the meaning of "falsy" is something that evaluates to False, which would include 0.

vmgustavo:
But it is counter intuitive to have it explain in the docstring it must be set to falsy. It makes more sense to indicate exactly what is the variable value that it needs to have to control the behavior.

martindurant:
Perhaps "falsy (False or 0)" ?

PR #1024, opened 2022-08-16
Make OpenFile filelike

martindurant:
@rabernat : a pickleable OpenFile in-a-context looks something like this, see what you think. Now entering `with` returns the OpenFile itself, which is now really file-like. It exists so that a chain of file-likes (compression, text...) can be closed in sequence at the right time. None of this affects what you get when you do a `fs.open()`, which objects are often pickleable (http, s3, local...), but this is not guaranteed.

Note that it breaks a couple of previous pickle tests, I will need to see whether the assumptions there are valid or not. Particularly, what should it mean to unpickle a "w"-mode file - naively, a new file gets opened with mode "w", invalidating the previous copy and truncating the file.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -18,7 +18,6 @@ from .caching import (  # noqa: F401
 from .compression import compr
 from .registry import filesystem, get_filesystem_class
 from .utils import (
-    IOWrapper,
     _unstrip_protocol,
     build_name_function,
     infer_compression,
@@ -29,7 +28,7 @@ from .utils import (
 logger = logging.getLogger("fsspec")
 
 
-class OpenFile(object):
+class OpenFile(io.IOBase):
     """
     File-like object to be used in a context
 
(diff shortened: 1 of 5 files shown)


rabernat:
Thanks a lot for working on this Martin!

I'm not sure that the `with` context manager behavior is relevant for Pangeo Forge and https://github.com/pangeo-forge/pangeo-forge-recipes/pull/391#issuecomment-1210615601. The conclusion we reached in https://github.com/fsspec/filesystem_spec/issues/579#issuecomment-1150081341 (see also https://github.com/google/xarray-beam/issues/49#issuecomment-1146693746) was that, in order to achieve the desired serializability, we should NOT use a context manager. Consequently, you will find no context managers in the Pangeo Forge [file opener code](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/beam-refactor/pangeo_forge_recipes/openers.py), but rather just `fs.open()` style.

martindurant:
This change removes `OpenFile.open.<locals>.close`, which was used for pickling, and is a consequence of calling OpenFile.open (as opposed to filesystem.open), and showed up in the error message, so probably it is fixed here.

PR #1040, opened 2022-09-07
Fix ZipFileSystem.__del__ for unsuccessful __init__

### LINKED ISSUE #1039: AttributeError: 'ZipFileSystem' object has no attribute 'zip'
When creating a `ZipFileSystem` object it is possible that `__init__` fails to create the `ZipFile` for example because of no permission to read. When python garbage collects this object there is an `AttributeError: 'ZipFileSystem' object has no attribute 'zip'` error because it tries to access a `zip` attribute that does not exist. This can be reproduced with the following:

```python
import fsspec
import os
import tempfile
import zipfile
from pathlib import Path


with tempfile.TemporaryDirectory() as tmpdir:
    os.chdir(tmpdir)

    txt_path = Path('data.txt')
    txt_path.write_text('some text\n')
    zip_path = Path('data.zip')
    zip_handle = zipfile.ZipFile(zip_path, 'w', zipfile.ZIP_DEFLATED)
    zip_handle.write(txt_path)
    zip_handle.close()

    os.chmod(zip_path, 0)

    try:
        handle = fsspec.open(f'zip://{txt_path}::file://{zip_path}', 'r')
    except Exception:
        pass
```

Running this script should not give any output, but it gives:

```shell
$ python3 fsspec_zip_issue.py 
Exception ignored in: <function ZipFileSystem.__del__ at 0x7f1a8e322040>
  File "/.../lib/python3.8/site-packages/fsspec/implementations/zip.py", line 71, in __del__
    self.zip.close()
AttributeError: 'ZipFileSystem' object has no attribute 'zip'
```

mauvilsa:
Fixes #1039.

### DIFF
--- a/fsspec/implementations/zip.py
+++ b/fsspec/implementations/zip.py
@@ -67,7 +67,8 @@ class ZipFileSystem(AbstractArchiveFileSystem):
         return super()._strip_protocol(path).lstrip("/")
 
     def __del__(self):
-        self.zip.close()
+        if hasattr(self, "zip"):
+            self.zip.close()
 
     def _get_dirs(self):
         if self.dir_cache is None:


martindurant:
I am happy to include your change and will merge when green.

I think the cause was not quite what you suggest, because apparently `__init__` for the ZipFileSystem instance did complete successfully. during interpreter shutdown, the `zip` attribute might be cleaned up before the instance itself is; so shutdown is the only time the exception might be seen.

PR #1047, opened 2022-09-16
AbstractFileSystem: add read_text/write_text api

### LINKED ISSUE #1045: cat_file (or equivalent) that returns the contents in text mode
Currently, `cat_file` only returns the contents of the file as bytes. The decoded content as a string would be really useful.

I see two ways to implement this:
1) extend `cat_file()` with `mode`/`encoding`/`errors` kwargs to pass it down to `open`.
2) introduce [`read_text()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text) or similar API.

Happy to go with either option and create a PR.

skshetry:
Closes #1045.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -649,6 +649,50 @@ class AbstractFileSystem(metaclass=_Cached):
         except:  # noqa: E722
             return False
 
+    def read_text(self, path, encoding=None, errors=None, newline=None, **kwargs):
+        """Get the contents of the file as a string.
+
+        Parameters
+        ----------
+        path: str
+            URL of file on this filesystems
+        encoding, errors, newline: same as `open`.
+        """
+        with self.open(
+            path,
+            mode="r",
+            encoding=encoding,
+            errors=errors,
+            newline=newline,
+            **kwargs,
+        ) as f:
+            return f.read()
+
+    def write_text(
+        self, path, value, encoding=None, errors=None, newline=None, **kwargs
+    ):
+        """Write the text to the given file.
+
+        An existing file will be overwritten.
+
+        Parameters
+        ----------
+        path: str
+            URL of file on this filesystems
+        value: str
+            Text to write.
+        encoding, errors, newline: same as `open`.
+        """
+        with self.open(
+            path,
+            mode="w",
+            encoding=encoding,
+            errors=errors,
+            newline=newline,
+            **kwargs,
+        ) as f:
+            return f.write(value)
+
     def cat_file(self, path, start=None, end=None, **kwargs):
         """Get the content of a file
 
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1264,6 +1264,14 @@ class AbstractFileSystem(metaclass=_Cached):
     # ------------------------------------------------------------------------
     # Aliases
 
+    def read_bytes(self, path, start=None, end=None, **kwargs):
+        """Alias of `AbstractFileSystem.cat_file`."""
+        return self.cat_file(path, start=start, end=end, **kwargs)
+
+    def write_bytes(self, path, value, **kwargs):
+        """Alias of `AbstractFileSystem.pipe_file`."""
+        self.pipe_file(path, value, **kwargs)
+
     def makedir(self, path, create_parents=True, **kwargs):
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -649,6 +649,50 @@ def isfile(self, path):
         except:  # noqa: E722
             return False
 
+    def read_text(self, path, encoding=None, errors=None, newline=None, **kwargs):
```
skshetry, line 652: A `read_text`/`write_text` in 2022 should probably just default to `utf-8`. In Python 3.10, `Path.read_text()` raises `EncodingWarning` if `-X warn_default_encoding` is set. Also, Python recommends to set encoding to "utf-8" by default for new APIs:

> please consider using UTF-8 by default (i.e. encoding="utf-8") for new APIs.

See https://docs.python.org/3/library/io.html#text-encoding.

martindurant: I didn't know that. Is this also true for builtins.open or TextIOWrapper? Certainly, utf8 is the only encoding I've come across, but people on windows still seem to have latin1 and others.

skshetry: > Is this also true for builtins.open or TextIOWrapper?

No, there's a new `io.text_encoding()` that has to be opt-in by the users of `open()` and `TextIOWrapper()` for this. That's what `Path().read_text()` does. See https://docs.python.org/3/library/io.html#opt-in-encodingwarning.

> If you are providing an API that uses [open()](https://docs.python.org/3/library/functions.html#open) or [TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) and passes encoding=None as a parameter, you can use [text_encoding()](https://docs.python.org/3/library/io.html#io.text_encoding) so that callers of the API will emit an [EncodingWarning](https://docs.python.org/3/library/exceptions.html#EncodingWarning) if they don’t pass an encoding. However, please consider using UTF-8 by default (i.e. encoding="utf-8") for new APIs.

skshetry: Reading the PEP again, yes it also applies to `builtins.open()` and `io.TextIOWrapper`. 🤦🏼 

> When the flag is set, io.TextIOWrapper(), open() and other modules using them will emit EncodingWarning when the encoding argument is omitted.

fsspec/spec.py
```diff
@@ -649,6 +649,50 @@ def isfile(self, path):
...
+        """Get the contents of the file as a string.
+
+        Parameters
+        ----------
+        path: str
+            URL of file on this filesystems
+        encoding, errors, newline: same as `open`.
+        """
+        with self.open(
```
martindurant, line 661: Same as `self.cat(path, **kwargs).decode(encoding=encoding, errors=errors)`? Does the value of newlines matter?

skshetry: To my mind, `open()` is much cleaner, and since it just works with `mode="r"`, I did not want to bother with handling that decoding _manually_. But no strong opinion here, API matters more than the implementation to me.
Are there any potential optimizations that `fs.cat()` offers?

> Does the value of newlines matter?

I haven't used `newline` that much even in `open()`, except when loading/writing `csv`s.  I see that `Path.write_text()` has `newline=""` support from 3.10 onwards, but not `read_text()`. Again, no strong opinion here, we could leave that out for now.

efiop:
Feels like if we decide to bother with `read_text/write_text` we might add `read_bytes/write_bytes` as well for symmetry.

Also this is interesting as it provides (arguably) a more pythonic interface as opposed to unix-like cat/pipe that we already have. I'm personally not a fan of creating too many different ways of doing the same thing and I'm not a fan of pathlib, but one could indeed argue that `read/write_bytes/text` should be more natural to the users. Overall I don't have very strong opinion here.

martindurant:
Thank you @efiop . Hoping for more opinions too.

martindurant:
read_bytes/write_bytes would be simple aliases of cat/pipe? I'll allow it :)

skshetry:
>  read_bytes/write_bytes would be simple aliases of cat/pipe? I'll allow it :)

Thanks, I have added those aliases. I am not sure what you mean by simple aliases, simpler form than cat/pipe or just alias with same function signatures? Since we are calling them aliases, I went with latter (and I assumed that's what you meant). 🙂

PR #1049, opened 2022-09-20
add DVCFileSystem to the registry

skshetry:
We have published [`DVCFileSystem`](https://github.com/iterative/dvc/blob/d2a31eb6603ad65188cb6451d8a88d1fc8c67b01/dvc/fs/dvc.py#L68) in dvc since 2.27.0.

Docs: https://dvc.org/doc/api-reference/dvcfilesystem
Impl: https://github.com/iterative/dvc/blob/d2a31eb6603ad65188cb6451d8a88d1fc8c67b01/dvc/fs/dvc.py#L68


fsspec/registry.py
```diff
@@ -202,6 +202,10 @@ def register_implementation(name, cls, clobber=True, errtxt=None):
         "class": "webdav4.fsspec.WebdavFileSystem",
         "err": "Install webdav4 to access WebDAV",
     },
+    "dvc": {
+        "class": "dvc.api.DVCFileSystem",
+        "err": "Install dvc to access DVCFileSystem",
+    },
```
efiop, line 208: Let's also add to https://github.com/fsspec/filesystem_spec/blob/master/docs/source/api.rst#other-known-implementations

skshetry: Should we add a link to the docs or a repository?

martindurant: No, the name of the package should be enough, people can find it from the pypi page.

martindurant:
I am happy with this change, I trust the people here that the information is correct.

fsspec/registry.py
efiop: Looks like this comment was resolved by mistake?

martindurant: Sorry, I mean: please don't change this text, it's good as it is. Indeed adding to the docs is a good idea, though.

efiop: @skshetry Btw, is this going to be operational in open/read calls? Just trying to remember if the urls will get handled correctly, since dvcfs wasn't present in the registry before.

efiop: For the record: @skshetry explained that it works with additional kwargs passed through like this:
```
with open("dvc://model.pkl", url="https://github.com/iterative/example-get-started.git", rev="main") as f:
   d = pickle.load(f)
```

efiop (approved):

PR #1051, opened 2022-09-20
Replace current hdfs with arrow_hdfs

### LINKED ISSUE #1050: The arrow_hdfs protocol violates rfc3986 in urllib
According to [rfc3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.1), schema names consist of a sequence of characters starting with a letter and followed by any combination of letters, digits, plus ("+"), period ("."), or hyphen ("-") .

The new HDFS implementation violates this rule as the protocol was defined as `arrow_hdfs` in registry.py: 

https://github.com/fsspec/filesystem_spec/blob/742a103849adf2d0c9f3ad3fe08010d8de688268/fsspec/registry.py#L132

Thus, it is difficult to validate that a url is being provided correctly. For example:

```python
import urllib.parse

urllib.parse.urlsplit("arrow_hdfs://server/path")
SplitResult(scheme='', netloc='', path='arrow_hdfs://server/path', query='', fragment='')
```

This problem can be worked around by changing the protocol name from `arrow_hdfs` to `arrow-hdfs`:

```python
import urllib.parse
urllib.parse.urlsplit("arrow-hdfs://server/path")
SplitResult(scheme='arrow-hdfs', netloc='server', path='/path', query='', fragment='')
```

hallenmaia:
This PR replaces the `hdfs` protocol with `arrow_hdfs`.

The `arrow_hdfs` protocol has been kept but will be removed in the future.

Close #1050


hallenmaia:
@martindurant I think it's ready!

martindurant:
Looks very reasonable to me. I'll leave it here a couple of days for anyone else interested.

martindurant:
Thank you

PR #1053, opened 2022-09-26
local: info: provide st_ino and st_nlink

efiop:


fsspec/implementations/local.py
```diff
@@ -97,7 +97,7 @@ def info(self, path, **kwargs):
             "created": out.st_ctime,
             "islink": link,
         }
-        for field in ["mode", "uid", "gid", "mtime"]:
+        for field in ["mode", "uid", "gid", "mtime", "ino", "nlink"]:
```
efiop, line 100: These could be provided by sshfs as well and maybe some other implementations.

I was thinking about maybe creating `stat` and `lstat` fields in info instead, but that seemed excessive for now (it will probably make sense to provide both of those in the future since we already have them, but there is a question of whether providing raw `stat_result` structure in `stat` and `lstat` info fields is a good practice or if we should add an info-like field `link_info` instead). I saw that `uid` and `gid` were provided here already, which are quite specific, so felt like inode and nlink was alright to provide too. Obviously open to suggestions. 🙂

efiop: Just to be clear: the reason why I'm proposing adding these two is that we use them in dvc and right now have to do additional stat calls.

efiop: Makes sense. Then I guess we'll end up adding `_stat_to_into()` helper and including `link_info` into the current `info()` result sometime in the future.

PR #1054, opened 2022-09-27
asyncfs: allow disabling mirroring sync methods

skshetry:
In [`AsyncLocalFileSystem`](https://github.com/iterative/morefs/blob/fe9e3ffece423a3fa527b118162cde5a601cd12c/src/morefs/asyn_local.py#L31), I want to fallback to fsspec's `LocalFileSystem` for sync methods, but the current architecture don't allow that.

I am proposing `mirror_sync_methods = True | False` to disable this mirroring and fallback to `AbstractFileSystem`'s behaviour. This would allow me to fallback to `LocalFileSystem` for sync methods instead of a generated ones.
It does feel a bit hacky, so suggestions are welcome. :)

Currently, what I am doing is something like this:
```python
class AsyncLocalFileSystem(AsyncFileSystem, LocalFileSystem):
    async_impl = False
```

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -307,6 +307,7 @@ class AsyncFileSystem(AbstractFileSystem):
     # for _* methods and inferred for overridden methods.
 
     async_impl = True
+    mirror_sync_methods = True
     disable_throttling = False
 
     def __init__(self, *args, asynchronous=False, loop=None, batch_size=None, **kwargs):
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -78,7 +78,7 @@ class _Cached(type):
             obj._fs_token_ = token
             obj.storage_args = args
             obj.storage_options = kwargs
-            if obj.async_impl:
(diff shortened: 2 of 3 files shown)


martindurant:
I don't think this is necessary - mirror_sync_methods() should only create methods where the implementation does not provide its own. The specific check is 
```
            unsync = getattr(getattr(obj, smethod, False), "__func__", None)
            is_default = unsync is getattr(AbstractFileSystem, smethod, "")
```
i.e., if the method with that name is identical to the one in AbstractFileSystem, then overwrite it; but if the subclass has its own, leave it alone.

skshetry:
yes, but in my case, most of the wrapper methods like `find()` are implemented in `AbstractFileSystem` which I don't want to be replaced by `AsyncFileSystem`.

martindurant:
Hm, OK, you have a point. 

Any objections from other watchers?

martindurant:
@efiop , is that a +1?

efiop (approved):

martindurant:
Thanks, all

PR #1058, opened 2022-10-03
localfs: fix support for pathlib/os.PathLike objects in rm

skshetry:
This fixes support for `pathlib.Path` objects in `rm`.
```python
from fsspec import filesystem
from pathlib import Path

fs = filesystem("file")
fs.rm(Path("file"))
```

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -142,7 +142,7 @@ class LocalFileSystem(AbstractFileSystem):
         os.remove(path)
 
     def rm(self, path, recursive=False, maxdepth=None):
-        if isinstance(path, str):
+        if not isinstance(path, list):
             path = [path]
 
         for p in path:


efiop:
@skshetry We don't use pathobjects though.

efiop (approved):
This implementation is more correct overall, even not thinking about pathlib. 👍

efiop:
Oops, I think I confused this with dvc-objects.fs in my comments 😅

skshetry:
`LocalFileSystem` does support `os.PathLike` objects, via `LocalFileSystem._strip_protocol` -> `stringify_paths`. This has nothing to do with DVC, just fixing support.

martindurant:
OK, this makes sense. I bet there are other instances of this in the codebase.

skshetry:
They are in `AbstractFileSystem`, especially in `put`/`expand_path`/`pipe`/`cat`/`get`. Will have to look into them separately.

PR #1060, opened 2022-10-03
registry: add asynclocalfs to the registry

### LINKED ISSUE #1052: Async LocalFileSystem
It would be nice to have an async version of LocalFileSystem for compatibility with async remote file systems and reading/writing large files locally without blocking.

Are there any plans to support async for LocalFileSystem or is there some other implementation of this I'm not aware of?

skshetry:
We need a new fsspec release to take advantage of disabled-mirroring for faster sync methods in [`AsyncLocalFileSystem`](https://github.com/iterative/morefs/blob/619809fc7d275215416c804704b17723c03ad808/src/morefs/asyn_local.py), but I think we can add it to the registry even without it (I also need to make a release there). :)

Closes #1052.

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -145,6 +145,10 @@ known_implementations = {
         "class": "ocifs.OCIFileSystem",
         "err": "Install ocifs to access OCI Object Storage",
     },
+    "asynclocalfs": {
+        "class": "morefs.asyn_local.AsyncLocalFileSystem",
+        "err": "Install 'morefs[asynclocalfs]' to use AsyncLocalFileSystem",
+    },
     "adl": {
         "class": "adlfs.AzureDatalakeFileSystem",
         "err": "Install adlfs to access Azure Datalake Gen1",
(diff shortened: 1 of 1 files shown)


martindurant:
Do you already refer to it as "asynclocalfs" elsewhere? The other protocols do not normally include "fs" in their names, even though all are filesystems.

skshetry:
I do use it on entry points, but I can change it to `asynclocal`. It did not sound good to me before, so I added `fs` to it. :)

fsspec/registry.py
```diff
@@ -145,6 +145,10 @@ def register_implementation(name, cls, clobber=True, errtxt=None):
         "class": "ocifs.OCIFileSystem",
         "err": "Install ocifs to access OCI Object Storage",
     },
+    "asynclocalfs": {
```
skshetry, line 148:
```suggestion
    "asynclocal": {
```

martindurant:
+1

skshetry:
Hey @martindurant, would you be able to tag a new release, please? 🙏🏼

martindurant:
Keep pinging me until it happens :). Now targeting Friday.

PR #1061, opened 2022-10-04
cache: added clear_expired_cache (#1055)

pl-marasco:
As discussed in #1055 time to time there is the need to keep data locally and remove them after a defined time. 
.clear_expired_cache() add the possibility to remove only the cached data that overpass the defined time.

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -159,7 +159,7 @@ class CachingFileSystem(AbstractFileSystem):
         if os.path.exists(fn):
             with open(fn, "rb") as f:
                 cached_files = pickle.load(f)
-            for k, c in cached_files.items():
+            for k, c in cached_files.copy().items():
                 if k in cache:
                     if c["blocks"] is True or cache[k]["blocks"] is True:
                         c["blocks"] = True
@@ -173,7 +173,8 @@ class CachingFileSystem(AbstractFileSystem):
                         c["blocks"] = blocks
                     c["time"] = max(c["time"], cache[k]["time"])
                     c["uid"] = cache[k]["uid"]
-
+                else:
+                    cached_files.pop(k)
             # Files can be added to cache after it was written once
             for k, c in cache.items():
                 if k not in cached_files:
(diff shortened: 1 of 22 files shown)


fsspec/implementations/cached.py
```diff
@@ -232,6 +233,41 @@ def clear_cache(self):
...
+                if self.same_names:
+                    basename = os.path.basename(detail["original"])
+                    fn = os.path.join(self.storage[-1], basename)
+                else:
+                    fn = os.path.join(self.storage[-1], detail["fn"])
+                if os.path.exists(fn):
+                    os.remove(fn)
+                    self.cached_files[-1].pop(path)
+
+        if self.cached_files[-1]:
```
martindurant, line 264: What is this bit doing?

martindurant (commented):
Thanks for this and the nice comprehensive test.

I have just a couple of small clarification questions.

fsspec/implementations/cached.py
```diff
@@ -173,7 +173,8 @@ def save_cache(self):
                         c["blocks"] = blocks
                     c["time"] = max(c["time"], cache[k]["time"])
                     c["uid"] = cache[k]["uid"]
-
+                else:
+                    cached_files.pop(k)
```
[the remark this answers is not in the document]
pl-marasco: Right, 
for my understanding k isn't in the cache but is still on the cached_file. In other words, save_cache() seems to make a comparison between the self.cache and the cache coming from the cache file.
The only option I had to not create a new (cache) file writer was to adapt what was in house.

fsspec/implementations/cached.py
pl-marasco: Not really sure what you are referring to as multiple parts are included but I'll try to explain both. 

lines 261/262 are the core component that removes the local file and the corresponding entity in the cache list.

line  264 checks if, for any reason, there are no left files and triggers a complete cleaning.

fsspec/implementations/cached.py
```diff
@@ -232,6 +233,41 @@ def clear_cache(self):
         rmtree(self.storage[-1])
         self.load_cache()
 
+
+    def clear_expired_cache(self, expiry_time=None):
+        """Remove all expired files and metadata from the cache
+        Parameters
```
martindurant, line 239:
```suggestion

        Parameters
```

martindurant:
Sorry, some blank lines around in there - please run `black` on here. I could apparently commit my suggested docstring format change, but I can't push to your master branch :)

pl-marasco:
Don't even say sorry as is my fault. I run `black` on the test component and I forgot to do the same on the cached.py

martindurant:
Would you mind investigating [the failure](https://github.com/fsspec/filesystem_spec/actions/runs/3202942410/jobs/5286076300#step:4:1049)? I am not convinced that the test is correct, and your popping from the cache might actually be the right thing to do. I am uncertain.

martindurant:
ping

pl-marasco:
I just back from COVID, I'll restart working on it today. 
I had time to though a lot about the approach and  I'm considering taking a completely different approach. I most probably will not use the 'save' method and use threads to manage the delete (if u agree on this). From my perspective threads are needed in case of processing lots of files or big files.

martindurant:
> just back from COVID

you and me both :)

I am not in general favour of using threads here, because fsspec is commonly used from within multithreaded settings anyway. It's OK if the process takes a while, since the user has explicitly asked for it. Furthermore, os operations like rm cannot actually operate in parallel anyway.

martindurant:
Thanks for sticking with it, @pl-marasco

Issue #1064, opened 2022-10-05
Multiple `open()` fails with `compression`


oakaigh:
`cifar-10-python.tar`:
```python
tfs = fsspec.filesystem('tar', fo='cifar-10-python.tar')

for p in tfs.glob('*/data_batch_*'): 
    batch = tfs.open(p, 'rb')
    batch.close()
```

`cifar-10-python.tar.gz`:
```python
tfs = fsspec.filesystem('tar', fo='cifar-10-python.tar.gz', compression='gzip')

for p in tfs.glob('*/data_batch_*'): 
    batch = tfs.open(p, 'rb')
    batch.close()
```

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [20], in <cell line: 19>()
     19 for p in tfs.glob('*/data_batch_*'): 
---> 20     batch = tfs.open(p, 'rb')
     21     batch.close()

File ~/.local/lib/python3.9/site-packages/fsspec/spec.py:1034, in AbstractFileSystem.open(self, path, mode, block_size, cache_options, compression, **kwargs)
   1032 else:
   1033     ac = kwargs.pop("autocommit", not self._intrans)
-> 1034     f = self._open(
   1035         path,
   1036         mode=mode,
   1037         block_size=block_size,
   1038         autocommit=ac,
   1039         cache_options=cache_options,
   1040         **kwargs,
   1041     )
   1042     if compression is not None:
   1043         from fsspec.compression import compr

File ~/.local/lib/python3.9/site-packages/fsspec/implementations/tar.py:132, in TarFileSystem._open(self, path, mode, **kwargs)
    130 else:
    131     newfo = copy.copy(self.fo)
--> 132 newfo.seek(offset)
    134 return TarContainedFile(newfo, self.info(path))

File /opt/conda/lib/python3.9/gzip.py:392, in GzipFile.seek(self, offset, whence)
    390 elif self.mode == READ:
    391     self._check_not_closed()
--> 392     return self._buffer.seek(offset, whence)
    394 return self.offset

ValueError: seek of closed file
```

oakaigh:
Okay did some digging. It turns out the copying of the `gzip.GzipFile` object (`self.fo` in this case) was the caveat: 
https://github.com/fsspec/filesystem_spec/blob/3969aaf7378198b92df8dee99cd464d272a0d0f4/fsspec/implementations/tar.py#L126-L131

[...]

PR #1065, opened 2022-10-05
Do not create copy of `self.fo` before seeking and passing it to `TarContainedFile`

oakaigh:
For details see https://github.com/fsspec/filesystem_spec/issues/1064.

### DIFF
--- a/fsspec/implementations/tar.py
+++ b/fsspec/implementations/tar.py
@@ -123,15 +123,9 @@ class TarFileSystem(AbstractArchiveFileSystem):
         if details["type"] != "file":
             raise ValueError("Can only handle regular files")
 
-        # `LocalFileSystem` offers its resources as `io.BufferedReader`
-        # objects, those can't be copied.
-        if isinstance(self.fo, BufferedReader):
-            newfo = self.fo
-        else:
-            newfo = copy.copy(self.fo)
-        newfo.seek(offset)
+        self.fo.seek(offset)
 
-        return TarContainedFile(newfo, self.info(path))
+        return TarContainedFile(self.fo, self.info(path))
 
 
 class TarContainedFile(object):


martindurant:
So the real culprit was `self.of.close()`, but what happens without the copy if more than one file in the archive is opened at the same time?

oakaigh:
@martindurant `self.of.close()` is not relevant to the current issue but it's never supposed to be there in the first place. The file object referencing an opened tar file is from outside the class representing a unit inside that opened file and should not be modified unless by the user or its creator. 

As for copying, I haven't had luck with it on a `gzip.GzipFile` instance hence the removal. The code below produces the same error as mentioned in the issue because it copies a `gzip.GzipFile` instance before `seek`-ing it. `read`-ing gives the same type of error:
```python
import gzip

with gzip.open('/tmp/file.txt.gz', 'wb') as f:
    f.write(b"Lots of content here")

f = gzip.GzipFile('/tmp/file.txt.gz')

import copy
f = copy.copy(f)

f.seek(0)
```

As for the case where multiple files are "opened at the same time," I don't think it's possible unless some kind of multithreading is involved, in which case it would be a totally different topic beyond the scope of this discussion. And as far as I know, `io.BufferedReader` does not support multithreading either.

martindurant:
I didn't mean quite *so* simultaneously :)
Perhaps interleaved is the right word.

```
f1 = fs.open(...)
f2 = fs.open(...)
f1.read(5)
f2.read(5)
```

oakaigh:
@martindurant If that’s the case some kind of rewinding mechanism would be befitting and it should be done in `TarContainedFile` not `TarFileSystem`. Alternatively you can create a wrapper class for the file objects to standardize this behavior and make the whole ordeal transparent (instead of type-checking if they can be copied which has the potential to introduce unpredicted behavior).

martindurant:
Does going the other way and always copying not solve anything? LocalFileSystem no longer has the restriction stated in the comment, and that branch should not get called.

martindurant:
Note that in your snippet, the original file was only closed because it lost its reference. The following works fine 
```
import gzip

with gzip.open('/tmp/file.txt.gz', 'wb') as f:
    f.write(b"Lots of content here")

f = gzip.GzipFile('/tmp/file.txt.gz')

import copy
f2 = copy.copy(f)

f2.seek(0)
```
but the two do seem to share some underlying buffer:
```
f2.read(10)
b'Lots of co'

f.read(10)
b'ntent here'
```

oakaigh:
@martindurant [`tarfile`](https://docs.python.org/3/library/tarfile.html)’s [`extractfile`](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractfile) method seems promising. You might want to look into it to get some insight on how to go about the issue you mentioned earlier - or to determine whether or not to replace the current implementation with `tarfile`'s altogether.

martindurant:
You are right, the file-in-file that `.extract()` deals with seems ideal.

Issue #1057, opened 2022-10-02
Failed: [undefined]OSError: Could not find a backend to open <fsspec.implementations.local.LocalFileOpener object at * > with iomode 'r' since version 2022.8.0


barmettl:
## Description
Starting with version 2022.8.0 the 'file' protocol does not interoperate with imageio.v3. Reading a file from `fs.open(filename,mode='rb')` leads to the exeption

```
Failed: [undefined]OSError: Could not find a backend to open <fsspec.implementations.local.LocalFileOpener object at * > with iomode 'r'
```
Despite the fact that it occurs in the imageio read call, I guess this is an fsspec issue since the issue does not occur on fsspec==2022.7.1. Also, other fsspec protocols, such as abfs work just fine.

Releated issue may be #579 .

## Minimal reproducible sample

```python
import fsspec
import numpy as np
import uuid
import tempfile
import imageio.v3 as iio
import pathlib
with tempfile.TemporaryDirectory() as dir:
    filename = f"{dir}/{str(uuid.uuid4())}.bmp"
    fs = fsspec.filesystem('file',auto_mkdir=True)
    img =(np.random.rand(100,100,3) * 255).astype('uint8')
    with fs.open(filename, mode='wb') as file:
        iio.imwrite(file, img, extension=pathlib.Path(filename).suffix)
    with fs.open(filename, mode='rb') as file:
        img_ = iio.imread(file, extension=pathlib.Path(filename).suffix)
        np.testing.assert_array_equal(img,img_)
```

martindurant:
I haven't yet found the problem root, but it is certainly on the write call not the read - the BMP extension is not making it to pillow, and so whatever is default is getting written, so then BMP fails to read it. A file does get created, but nothing can read it as a BMP. Yet, you can pass the fsspec file-like to pillow for writing or reading just fine. Perhaps we can ping someone at imagio to ask why the extension information isn't reaching the required place?

barmettl:
@martindurant: I agree that the problem is in the write and not in the read. I nailed down the problem to #1010. Due to the new method fileno, imageio will directly use the integer file descriptor rather than the file object to write to. This happens in https://github.com/python-pillow/Pillow/blob/main/src/PIL/ImageFile.py:520

```
try:
        fh = fp.fileno()
        fp.flush()
        _encode_tile(im, fp, tile, bufsize, fh)
    except (AttributeError, io.UnsupportedOperation) as exc:
[...]

PR #1070, opened 2022-10-10
explicitly wrap flush  LocalFileOpener

barmettl:
Solving #1057.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -367,10 +367,6 @@ class LocalFileOpener(io.IOBase):
     def flush(self) -> None:
         self.f.flush()
 
-    @property
-    def name(self) -> str:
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks for the fix. I suppose getattr is actually never called when the method is available (but not implemented) in the superclasses, you are right. You are welcome to add more specific known methods of a file here, I don't mind - or we can merge as is, since it solves your problem.

Thanks for the digging, I like the weird effect the bug has, of swapping the header and data portions of the output.

barmettl:
I suggest to merge just this fix. Removing getattr without replacement breaks a couple of tests. Replacing it by explicit overrides will most likely resolve it. But I won't do that right now.

martindurant:
OK, merging when green.

Issue #1071, opened 2022-10-11
ssh filesystem and put_files are incompatible


ghislainp:
Using a ssh filesystem  and put_files gives an error

```python
import fsspec
import sshfs
fs = fsspec.filesystem("ssh", host='xx.xx.xx.xx', username='eouser') 
fs.put_file(lpath="t", rpath="/home/me/t")
```
I receive an exception:
```
File "/home/me/miniconda3/envs/olci/lib/python3.9/site-packages/fsspec/callbacks.py", line 66, in relative_update
    self.value += inc
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```
The reason is because the 'write' method in paramiko does not return the length of written data.
https://github.com/jbouse-debian/paramiko/blob/master/paramiko/file.py#L385

Unless paramiko changes their function, the method 'put_file' in spec.py should test segment_len to be None and skip the update... or use len(data) as a default as follows:

```python
                    segment_len = f2.write(data)
                    if segment_len is None:
                          segment_len = len(data)
                    callback.relative_update(segment_len)
```

I'm not sure if this is a good solution...

martindurant:
@efiop , time to ditch paramiko-based ssh?

@ghislainp , would you like to try with https://github.com/fsspec/sshfs ?

ghislainp:
I'm using prefect which is using fsspec in a generic way...

martindurant:
I suppose you could try temporarily with 
```
import fsspec
import sshfs
fsspec.register_implementation("ssh", sshfs.SSHFileSystem)
```
@efiop , @pmrowla : I note tha sshfs is not noarch (surprised me) and has not been released for osx-arm64.

efiop:
@martindurant Hm, that is indeed surprising. We only have `any` wheel in https://pypi.org/project/sshfs/#files , which I thought meant noarch (same with underlying asyncssh). Are you sure it is not some other package that is getting compiled?  I'm also on m1, but don't see anything being compiled when I install sshfs.

efiop:
@martindurant Ah, right 😅 I remember we tried `noarch` before, but maybe it wasn't with sshfs conda package specifically and so we copied it over from another place with no second thoughts. We'll double-check.

efiop:
@martindurant Yeah, looks like we just copypasted it back then without trying `noarch` first. Fixed in https://github.com/conda-forge/sshfs-feedstock/pull/20 . Thanks for the heads up! 🙏

[...]

PR #1072, opened 2022-10-12
solve a bug when the 'write' method in an implementation does not ret…

ghislainp:
…urn the actual length of written data (the case of SFTP with paramiko)


martindurant:
I suppose it would be appropriate to add a test to test_ssh ? I'm happy to merge this on your sayso, if it fixes #1071 .

martindurant:
Perfect, thank you.

PR #1079, opened 2022-10-21
avoid ESA snappy

edrodrigues-ag:
This patch avoids an error if one has the European Space Agency (ESA) snappy ( https://senbox.atlassian.net/wiki/spaces/SNAP/pages/19300362/How+to+use+the+SNAP+API+from+Python ) instead of the snappy compression library.

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -140,7 +140,7 @@ try:
     # standard implementation.
     register_compression("snappy", SnappyFile, [])
 
-except (ImportError, NameError):
+except (ImportError, NameError, AttributeError):
     pass
 
 try:


martindurant:
ESA snappy is completely unrelated to compression?

edrodrigues-ag:
ESA snappy is not related with compression (it is used to process Sentinel satellite images), but unfortunately it has the same name (snappy). So, if one has it installed instead of snappy for compression, this piece of code generates an `AttributeError` because ESA snappy does not have an attribute `compress`.

martindurant:
Makes sense, thanks.

PR #1082, opened 2022-10-25
Implement recursive parameter in the get method of SFTP 

maximfridental:
Current implementation does not support downloading directories recursively in the get method. Unfortunately, this is what prefect.io is doing so that prefect agent is crashing when using SFTP storage block. 

I've implemented a simple recursive downloading as well as a test for this case.

Also, prefect is passing a URL to this method, not the remote FTP path, so that I've added URL parsing to this method too. I am not sure if this URL parsing should be added to all other methods of the class too, so I've tried to keep changes to a bare minimum required for Prefect to work.

### DIFF
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -1,9 +1,8 @@
 import datetime
 import logging
+import os
 import types
 import uuid
-import os
-from pathlib import Path
 from stat import S_ISDIR, S_ISLNK
 
 import paramiko
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -2,6 +2,8 @@ import datetime
 import logging
 import types
 import uuid
+import os
+from pathlib import Path
 from stat import S_ISDIR, S_ISLNK
 
 import paramiko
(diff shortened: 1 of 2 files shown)


fsspec/implementations/tests/test_sftp.py
```diff
@@ -93,6 +93,21 @@ def test_with_url(protocol, ssh):
         assert f.read() == b"hello"
 
 
+@pytest.mark.parametrize("protocol", ["sftp", "ssh"])
+def test_get_dir(protocol, ssh, root_path):
+    f = fsspec.get_filesystem_class(protocol)(**ssh)
```
martindurant, line 98: This is the same as
```
f = fsspec.filesystem(protocol, **ssh)
```

fsspec/implementations/tests/test_sftp.py
```diff
@@ -93,6 +93,21 @@ def test_with_url(protocol, ssh):
...
+    f.mkdirs(root_path + "deeper", exist_ok=True)
+    f.touch(root_path + "deeper/afile")
+    f.get(root_path + "deeper", ".", recursive=True)
+
+    f.get(
+        protocol + "://{username}:{password}@{host}:{port}"
+        "{root_path}deeper".format(root_path=root_path, **ssh),
+        ".",
+        recursive=True,
+    )
```
martindurant, line 108: Shouldn't we test that the transfer did what was expected, as opposed to simply proceeding without errors?

fsspec/implementations/sftp.py
```diff
@@ -130,8 +132,29 @@ def put(self, lpath, rpath, callback=None, **kwargs):
         self.ftp.put(lpath, rpath)
 
     def get(self, rpath, lpath, callback=None, **kwargs):
-        logger.debug("Get file %s into %s" % (rpath, lpath))
-        self.ftp.get(rpath, lpath)
+        # parse rpath if needed
```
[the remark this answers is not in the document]
maximfridental: Oh, wow, thanks, this works too!

fsspec/implementations/tests/test_sftp.py
maximfridental: Fixed

fsspec/implementations/tests/test_sftp.py
maximfridental: Yes, of course. I've added a couple of asserts.

fsspec/implementations/sftp.py
```diff
@@ -2,6 +2,8 @@
 import logging
 import types
 import uuid
+import os
+from pathlib import Path
 from stat import S_ISDIR, S_ISLNK
```
martindurant, line 7:
```suggestion
import logging
import os
import types
import uuid
from stat import S_ISDIR, S_ISLNK
```

fsspec/implementations/tests/test_sftp.py
```diff
@@ -2,6 +2,7 @@
 import subprocess
 import time
 from tarfile import TarFile
+import os
```
martindurant, line 6:
```suggestion
import os
import subprocess
import time
from tarfile import TarFile

```

martindurant: Actually, should be one line higher above shlex, but the UI here won't let me do that!

PR #1084, opened 2022-10-25
Pin docutils so that sphinx renders lists correctly

### LINKED ISSUE #1083: Lists in project docs rendered incorrectly
The project changelog at https://filesystem-spec.readthedocs.io/en/latest/changelog.html isn't rendered correctly as the unordered lists do not have bullet points and are spaced too far apart. Screenshot:

Screenshot 2022-10-25 at 14 32 00

This is apparently a well-known problem related to `sphinx-rtd-theme` and `docutils`: https://stackoverflow.com/questions/67542699/readthedocs-sphinx-not-rendering-bullet-list-from-rst-file

If I pin `docutils<0.17` and generate the docs locally then the problem is fixed. I think this is probably the best temporary solution until up-to-date `sphinx-rtd-theme` conda packages are available.

The question is where to put the pin? I think it should go in `docs/environment.yml` but it is not clear to me if the docs get built in CI and automatically uploaded or if it is a more manual process.

ianthomas23:
Fixes #1083. Pins `docutils` so that sphinx correctly renders lists.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -3,4 +3,5 @@ channels:
   - defaults
 dependencies:
   - python=3.9
+  - docutils<0.17
   - numpydoc


martindurant:
Please merge from latest to get tests to pass - due to an upstream release in pytest.

PR #1087, opened 2022-10-27
Sort when merging offsets for referenceFS

martindurant:
Fixes https://github.com/fsspec/kerchunk/issues/246


arongergely:
awesome

PR #1089, opened 2022-10-28
Support append mode for ArrowFSWrapper

leoleoasd:
This PR adds support append mode for ArrowFSWrapper.

https://github.com/apache/arrow/blob/apache-arrow-1.0.0/python/pyarrow/_fs.pyx#L673

It seems that from v1.0.0, pyarrow supports `open_append_stream`.

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -149,6 +149,8 @@ class ArrowFSWrapper(AbstractFileSystem):
             method = self.fs.open_input_stream
         elif mode == "wb":
             method = self.fs.open_output_stream
+        elif mode == "ab":
+            method = self.fs.open_append_stream
         else:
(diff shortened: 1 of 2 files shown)


martindurant:
Can we please add a test for this in test_arrow.py?

leoleoasd:
Hi, I've added tests. Works fine in my repo (https://github.com/leoleoasd/filesystem_spec/actions/runs/3346115981), but I need approval for running tests in PR.
image

leoleoasd:
Hi, I would like to ask when will we release this into PyPI, is there an ETA? 
Since our cluster only supports installing packages from PyPI, I need to apply a patch into `site-packages` every time I submit a new job before the release.

martindurant:
We have a few things coming in now, so maybe within the next week.

PR #1091, opened 2022-11-02
Add an optional 'seekable' argument for `ArrowFileSystem` to open as file, rather than stream.

sighingnow:
Actually HDFS does support `seek` operation and it is useful for partial reading / chunk reading.

This pull request add a `seekable` argument to `_open()` in `ArrowFSWrapper` (default to `False` to keep a compatible behaviour) and expose the `.size()` method on pyarrow's `NativeFile` as well (see https://arrow.apache.org/docs/python/generated/pyarrow.NativeFile.html#pyarrow.NativeFile.size)

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -144,9 +144,12 @@ class ArrowFSWrapper(AbstractFileSystem):
             self.fs.delete_file(path)
 
     @wrap_exceptions
-    def _open(self, path, mode="rb", block_size=None, **kwargs):
+    def _open(self, path, mode="rb", block_size=None, seekable=False, **kwargs):
         if mode == "rb":
-            method = self.fs.open_input_stream
+            if seekable:
+                method = self.fs.open_input_file
+            else:
+                method = self.fs.open_input_stream
         elif mode == "wb":
             method = self.fs.open_output_stream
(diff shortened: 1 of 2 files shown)


fsspec/implementations/arrow.py
```diff
@@ -155,9 +158,10 @@ def _open(self, path, mode="rb", block_size=None, **kwargs):
             raise ValueError(f"unsupported mode for Arrow filesystem: {mode!r}")
 
         _kwargs = {}
-        if PYARROW_VERSION[0] >= 4:
-            # disable compression auto-detection
-            _kwargs["compression"] = None
+        if mode != "rb" or not seekable:
+            if PYARROW_VERSION[0] >= 4:
+                # disable compression auto-detection
+                _kwargs["compression"] = None
```
martindurant, line 164: I guess we don't test this, since only one version of pyarrow is installed? I don't see what the practical consequences will be.

sighingnow: It was introduced in #911 to fix issue #904.

As we state `"pyarrow >= 1"` in setup.py, if we remove the check here would be a backwards compatibility issue, I guess.

martindurant:
Merging this - thank you!
We may well wish to make this the default in the future.

PR #1094, opened 2022-11-03
Add missing await statement to GenericFileSystem._cp_file

### LINKED ISSUE #1093: Generic Filesystem does not await call to async function
**Describe the problem**
In generic.py on line 173, the function _cp_file of GenericFileSystem makes the following call:
`callback.set_size(maybe_await(f1.size))`
maybe_await is a wrapper that decides whether or not it's parameter is a Coroutine, and therefore if the parameter should be called with await or not. The maybe_await function itself is an asynchronous function, and thus the call to it on line 173 should be awaited, which it is currently not.

**Propose a solution**
Simply add the await statement in front of the call to maybe_await:
`callback.set_size(await maybe_await(f1.size))`

Metamess:
### Reference Issues/PRs
Closes #1093 

### What did you change/fix?
The function _cp_file of GenericFileSystem was missing an await statement for an asynchronous function. This PR adds the missing await statement.

### DIFF
--- a/fsspec/generic.py
+++ b/fsspec/generic.py
@@ -170,7 +170,7 @@ class GenericFileSystem(AsyncFileSystem):
                 if hasattr(fs, "open_async")
                 else fs.open(url, "rb", **kw)
             )
-            callback.set_size(maybe_await(f1.size))
+            callback.set_size(await maybe_await(f1.size))
             f2 = (
                 await fs2.open_async(url2, "wb")
                 if hasattr(fs2, "open_async")


martindurant:
+1
Clearly this line was not covered by tests. Do you think you con contrive a test which does touch it?

Metamess:
From what I can trace, the mirror_sync_methods function in asyn.py should replace the cp_file method of a GenericFileSystem with the _cp_file method it implements. This means that test_cp_async_to_sync, when calling cp on the GenericFileSystem should end up calling the _cp_file defined in generic.py. 

The only way this code is not covered by the test would be if the two filesystems involved are the same, in which case the `if fs is fs2` on line 160 of generic.py would cause the execution of _cp_file to branch away from it. But as far as I can tell, this is not the case, so I don't know why the test code would not touch the line fixed in this PR.

I apologize for lacking the time/skill to be of more use right now. If the lack of an extra test is considered blocking for this PR, perhaps someone else in the community reading this could have a look at this?

martindurant:
I guess I'll merge this now and leave making a more thorough test suits for this module for future work.
I believe that this would have been triggered by:
- cp with two different filesystems
- the source filesystem is async
- a non-default callback is used

PR #1095, opened 2022-11-04
Add yarl to readthedocs dependencies

### LINKED ISSUE #1090: HTTPFileSystem missing from docs
https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations

ianthomas23:
Fixes #1090.

I've opted for the simplest change which is to explicitly add `yarl` to the list of dependencies for the readthedocs build.

There is another possibility which is to use a more complicated readthedocs build that follows the usual package build process so that the package and docs dependencies cannot get out of sync, but this would be a much bigger change that should probably be part of a longer-term strategy. Such an approach might also want to include building the docs explicitly as part of CI so that docs failures are reported directly as CI failures.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,3 +5,4 @@ dependencies:
   - python=3.9
   - docutils<0.17
   - numpydoc
+  - yarl


ianthomas23:
I have tested this locally following the process outlined in issue #1090 to confirm it solves the problem.

martindurant:
+1, merging as is.

I totally agree that it would be good to have a docsbuild in CI. Also it would be good to ditch tox, which is a pre GHA holdover. If you are looking for a task...

ianthomas23:
> I totally agree that it would be good to have a docsbuild in CI. Also it would be good to ditch tox, which is a pre GHA holdover. If you are looking for a task...

I've put both items on my todo list.

PR #1097, opened 2022-11-04
Route LocalFileSystem.put to LocalFileSystem.cp

### LINKED ISSUE #967: Unable to put a file to a directory in LocalFileSystem
I tried to copy a file to another directory with the put function of LocalFileSystem, the code is 
```

fs = LocalFileSystem()
fs.mkdir('/tmp/test_dir0')
fs.mkdir('/tmp/test_dir1')
with open('/tmp/test_dir0/a', 'w') as fp:
    fp.write('test')
fs.put('/tmp/test_dir0/a', '/tmp/test_dir1/')
```
but I get an Error traceback like this:
```Traceback (most recent call last):
  File "/home/user/anaconda3/envs/py38/lib/python3.8/site-packages/fsspec/spec.py", line 823, in put
    self.put_file(lpath, rpath, **kwargs)
  File "/home/user/anaconda3/envs/py38/lib/python3.8/site-packages/fsspec/implementations/local.py", line 127, in put_file
    return self.cp_file(path1, path2, **kwargs)
  File "/home/user/anaconda3/envs/py38/lib/python3.8/site-packages/fsspec/implementations/local.py", line 117, in cp_file
    shutil.copyfile(path1, path2)
  File "/home/user/anaconda3/envs/py38/lib/python3.8/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
IsADirectoryError: [Errno 21] Is a directory: '/tmp/test_dir1'
```

However the doc of put tell me that 
> If rpath ends with a "/", it will be assumed to be a directory, and target files will go within.

I guess directory is supported by design, so is it a bug?

ianthomas23:
This is a work in progress to fix #967.

`LocalFileSystem.put` calls `LocalFileSystem.cp` as they are essentially the same thing. I am checking for a `callback` kwarg as `put` accepts it but `cp` doesn't, then removing it from the kwargs and warning the user. Alternatively it might be acceptable to pass it through anyway, or indeed raise an exception.

Strictly speaking the signature of `put` in the abstract base class allows for `callback` to be the third positional arg, whereas I am assuming it is always a kwarg here. This may or may not be a problem?

There is a question of where to the tests for this.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -8,6 +8,7 @@ import re
 import shutil
 import stat
 import tempfile
+import warnings
 
 from fsspec import AbstractFileSystem
 from fsspec.compression import compr
@@ -132,6 +133,13 @@ class LocalFileSystem(AbstractFileSystem):
 
     def put_file(self, path1, path2, callback=None, **kwargs):
         return self.cp_file(path1, path2, **kwargs)
+
+    def put(self, lpath, rpath, **kwargs):
+        if kwargs.pop("callback", None):
+            warnings.warn(
+                f"'callback' kwarg is ignored by {self.__class__.__name__}.put()"
+            )
+        self.cp(lpath, rpath, **kwargs)
 
     def mv_file(self, path1, path2, **kwargs):
         path1 = self._strip_protocol(path1).rstrip("/")


martindurant:
I have a feeling that `put()` already works, including the passing of callbacks, except the noted directory "copy" problem. Another way to fix this could be

```
    def put_file(self, path1, path2, callback=None, **kwargs):
        if self.isdir(path1):
            self.mkdir(path2)
        else:
            return self.cp_file(path1, path2, **kwargs)
```

ianthomas23:
Yes, you are correct. I will close this PR then as it is unnecessary, and look at your recommended solution instead.

PR #1099, opened 2022-11-06
Strip protocol in arrowfs.ls and add tests for calling arrowfs with protocol

leoleoasd:
Fixed stripping protocol in `ArrowFsWrapper.ls`.
Also, all tests for arrowfs will run twice now, one with protocol and one without.


martindurant:
Lovely, thank you.

PR #1101, opened 2022-11-07
speedup FSMap._key_to_str

ap--:
Hello everyone,

I had a look at the FSMap code and noticed that `_key_to_str` is calling `fs._strip_protocol` on every call. Also there's code that casts tuples to strings, and it looks wrong (and maybe obsolete) to me. Is there a scenario where FSMap keys will include protocols or where FSMap keys will be tuples?
It seems that simplifying `_key_to_str` still passes the tests.

If there's a use case where keys are prefixed with protocols or tuples I will change this PR and work add a test to the test suite.

Cheers,
Andreas :smiley:

#### Updated TODOs:

- [x] should non-str keys be allowed? **No**
      _"This was probably a mistake to allow and should be removed. I don't believe it is used, but we should have a long depr cycle around it if we decide to remove, to give the few a chance to object."_   

- [x] should protocols be allowed in keys? **No**

- [x] should `m['/abc']` and `m['abc']` return different things?
      _"Only if they evaluate to the same path in the target FS, we shouldn't put extra logic around ensuring it."_

- [x] do `abc/` and `abc` potentially return different things using the FSMap interface?
      _"Since keys refer to files not directories, the former will not be allowed by most FS implementations."_

### DIFF
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -134,6 +134,8 @@ class FSMap(MutableMapping):
             if isinstance(key, list):
                 key = tuple(key)
             key = str(key)
+        if ":" in key and "://" in key:
+            raise ValueError(f"keys must not contain protocols, got: {key!r}")
         return f"{self._root_prefix}{key}"
 
     def _str_to_key(self, s):
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -1,5 +1,6 @@
 import array
 import os.path
+import warnings
 from collections.abc import MutableMapping
 
 from .core import url_to_fs
@@ -122,6 +123,15 @@ class FSMap(MutableMapping):
 
     def _key_to_str(self, key):
         """Generate full path for the key"""
+        if not isinstance(key, str):
+            # raise TypeError("key must be of type `str`, got `{type(key).__name__}`"
+            warnings.warn(
+                "from fsspec 2023.5 onward FSMap non-str keys will raise TypeError,"
+                f" got: {key!r} of type `{type(key).__name__}`",
+                FutureWarning,
+                stacklevel=2,
+            )
+            key = str(key)
         return f"{self._root_prefix}{key}"
 
     def _str_to_key(self, s):
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -126,7 +126,7 @@ class FSMap(MutableMapping):
             warnings.warn(
                 "from fsspec 2023.5 onward FSMap non-str keys will raise TypeError,"
                 f" got: {key!r} of type `{type(key).__name__}`",
-                FutureWarning,
+                DeprecationWarning,
                 stacklevel=2,
             )
             if isinstance(key, list):
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -1,4 +1,5 @@
 import array
+import os.path
 from collections.abc import MutableMapping
 
 from .core import url_to_fs
@@ -37,6 +38,7 @@ class FSMap(MutableMapping):
         self.root = fs._strip_protocol(root).rstrip(
             "/"
         )  # we join on '/' in _key_to_str
+        self._root_prefix = fs._strip_protocol(os.path.join(root, "x"))[:-1]
         if missing_exceptions is None:
             missing_exceptions = (
                 FileNotFoundError,
@@ -120,11 +122,7 @@ class FSMap(MutableMapping):
 
     def _key_to_str(self, key):
         """Generate full path for the key"""
-        if isinstance(key, (tuple, list)):
-            key = str(tuple(key))
-        else:
-            key = str(key)
-        return self.fs._strip_protocol("/".join([self.root, key]) if self.root else key)
+        return f"{self._root_prefix}{key}"
 
     def _str_to_key(self, s):
         """Strip path of to leave key name"""
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -131,6 +131,8 @@ class FSMap(MutableMapping):
                 FutureWarning,
                 stacklevel=2,
             )
+            if isinstance(key, list):
+                key = tuple(key)
             key = str(key)
         return f"{self._root_prefix}{key}"
 
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -1,5 +1,5 @@
 import array
-import os.path
+import posixpath
 import warnings
 from collections.abc import MutableMapping
 
@@ -37,7 +37,7 @@ class FSMap(MutableMapping):
     def __init__(self, root, fs, check=False, create=False, missing_exceptions=None):
         self.fs = fs
         self.root = fs._strip_protocol(root).rstrip("/")
-        self._root_key_to_str = fs._strip_protocol(os.path.join(root, "x"))[:-1]
+        self._root_key_to_str = fs._strip_protocol(posixpath.join(root, "x"))[:-1]
         if missing_exceptions is None:
             missing_exceptions = (
                 FileNotFoundError,
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -124,10 +124,8 @@ class FSMap(MutableMapping):
         if not isinstance(key, str):
             # raise TypeError("key must be of type `str`, got `{type(key).__name__}`"
             warnings.warn(
-                "from fsspec 2023.5 onward FSMap non-str keys will raise TypeError,"
-                f" got: {key!r} of type `{type(key).__name__}`",
+                "from fsspec 2023.5 onward FSMap non-str keys will raise TypeError",
                 DeprecationWarning,
-                stacklevel=2,
             )
             if isinstance(key, list):
                 key = tuple(key)
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -1,6 +1,5 @@
 import array
 import os.path
-import platform
 import warnings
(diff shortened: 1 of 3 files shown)


martindurant:
It appears you are quite right, this must be left over from something that no longer exists.

ap--:
Seems I was wrong regarding the passing tests. I'll investigate what I was missing locally and update the PR.

ap--:
Okay, so this is more of a rabbit hole than I thought initially 😅 

I will spend some time working on more tests for FSMap and see if there's an easy way to avoid calling `_strip_protocol` for every `__getitem__` call.

The table below shows the `_strip_protocol` behavior for strings `protocol://`, `''`, `'a'`, and `protocol:///`, `'/'`, `'/a'`,  and `'a/'`.

| class                    | `'xxx://'`   | `''`     | `'a'`      | `'xxx:///'`   | `'/'`   | `'/a'`   | `'a/'`     |
|--------------------------|--------------|----------|------------|---------------|---------|----------|------------|
| HTTPFileSystem           | `'http://'`  | `''`     | `'a'`      | `'http:///'`  | `'/'`   | `'/a'`   | `'a/'`     |
| LocalFileSystem          | `'/cwd'`     | `'/cwd'` | `'/cwd/a'` | `'/'`         | `'/'`   | `'/a'`   | `'/cwd/a'` |
| FTPFileSystem            | `'/'`        | `'/'`    | `'/a'`     | `'/'`         | `'/'`   | `'/a'`   | `'/a'`     |
| WebHDFS                  | `''`         | `''`     | `'a'`      | `'/'`         | `'/'`   | `'/a'`   | `'a/'`     |
| SFTPFileSystem           | `''`         | `''`     | `'a'`      | `'/'`         | `'/'`   | `'/a'`   | `'a/'`     |
| HadoopFileSystem         | `''`         | `''`     | `'a'`      | `'/'`         | `'/'`   | `'/a'`   | `'a/'`     |
| GCSFileSystem            | `''`         | `''`     | `'a'`      | `'/'`         | `'/'`   | `'/a'`   | `'a/'`     |
| AzureDatalakeFileSystem  | `''`         | `''`     | `'a'`      | `'/'`         | `'/'`   | `'/a'`   | `'a/'`     |
| AzureBlobFileSystem      | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'a'`    | `'a/'`     |
| ZipFileSystem            | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'a'`    | `'a'`      |
| GitFileSystem            | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'a'`    | `'a'`      |
| WholeFileCacheFileSystem | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| TarFileSystem            | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| SimpleCacheFileSystem    | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| S3FileSystem             | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| ReferenceFileSystem      | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| JupyterFileSystem        | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| GithubFileSystem         | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| GenericFileSystem        | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| DropboxDriveFileSystem   | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| DatabricksFileSystem     | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| DaskWorkerFileSystem     | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| CachingFileSystem        | `''`         | `''`     | `'a'`      | `''`          | `''`    | `'/a'`   | `'a'`      |
| MemoryFileSystem         | `''`         | `''`     | `'/a'`     | `''`          | `''`    | `'/a'`   | `'/a'`     |

The current `__getitem__` implementation is equivalent to:

```python
def _key_to_str(fs, root, key):
    # happens in __init__
    root = fs._strip_protocol(root)
    root = root.rstrip("/")

    # happens in every __getitem__ call
    if root:
        key = "/".join([root, key])
    return fs._strip_protocol(key)

```

martindurant:
- should non-str keys be allowed? (the gcsfs tests currently expect int keys to work [see here](https://github.com/fsspec/gcsfs/blob/67670f18a6773221a6fb36713541f446f7f27be6/gcsfs/tests/test_mapping.py#L60-L63))
- This test also used extended slicing syntax with integers, which explains the tuple check in the current _key_to_str implementation

This was probably a mistake to allow and should be removed. I don't believe it is used, but we should have a long depr cycle around it if we decide to remove, to give the few a chance to object. Libraries like zarr that use this API generate a full list of (str) keys to fetch rather than passing slices.

- should protocols be allowed in keys?

No

- should m['/abc'] and m['abc'] return different things?

Only if they evaluate to the same path in the target FS, we shouldn't put extra logic around ensuring it.

- do abc/ and abc potentially return different things using the FSMap interface?

Since keys refer to files not directories, the former will not be allowed by most FS implementations.

ap--:
This PR is ready for review.

- speeds up `_key_to_str` by removing the overhead of the `_strip_protocol` call.
- throws a FutureWarning if the provided key is not a string. And announces the switch to raising a TypeError starting with version 2023.5 (6+ months from now)
- raises a ValueError if the key contains a protocol
- adds PermissionError to FSMap.missing_exceptions on windows so that KeyError is raised on directory access (consistent with other OSes)

fsspec/mapping.py
```diff
@@ -34,16 +37,18 @@ class FSMap(MutableMapping):
...
+            missing_exceptions = [
                 FileNotFoundError,
                 IsADirectoryError,
                 NotADirectoryError,
-            )
-        self.missing_exceptions = missing_exceptions
+            ]
+            if platform.system() == "Windows":
+                # see: https://bugs.python.org/issue43095
+                missing_exceptions.append(PermissionError)
```
martindurant, line 50: Actually, this is problematic. The simple problem is, that windows users will of course use this API with FSs other than local. Secondly, there may be real permission errors - should they really appear missing?

fsspec/mapping.py
```diff
@@ -120,11 +125,20 @@ def delitems(self, keys):
...
+                "from fsspec 2023.5 onward FSMap non-str keys will raise TypeError,"
+                f" got: {key!r} of type `{type(key).__name__}`",
+                FutureWarning,
+                stacklevel=2,
+            )
+            if isinstance(key, list):
+                key = tuple(key)
             key = str(key)
-        return self.fs._strip_protocol("/".join([self.root, key]) if self.root else key)
+        if ":" in key and "://" in key:
```
martindurant, line 139: Honestly, I wouldn't bother checking, and let the backend FS error. Since the Map has a root, this protocol string would appear midway through the final URL. 

(I would note that the first condition is redundant, covered by the second)

fsspec/mapping.py
ap--: That makes sense. I'll revert the commit and skip the relevant test on windows.

fsspec/mapping.py
ap--: I'll revert that commit too 👍

ap--:
The s3fs-pytest failure seems to be unrelated. Can you rerun the workflow?

martindurant:
@efiop , the test failure seems to be specific to the new versioned ls. Do you have time to look?

@ap-- : it would be nice to stop all the warnings due to non-str keys showing up in the testing log.

ap--:
s3fs tests seem to require moto>=4 now. I updated it in tox.ini

martindurant:
Perfect, thank you.

fsspec/mapping.py
```diff
@@ -34,9 +36,8 @@ class FSMap(MutableMapping):
 
     def __init__(self, root, fs, check=False, create=False, missing_exceptions=None):
         self.fs = fs
-        self.root = fs._strip_protocol(root).rstrip(
-            "/"
-        )  # we join on '/' in _key_to_str
+        self.root = fs._strip_protocol(root).rstrip("/")
+        self._root_key_to_str = fs._strip_protocol(os.path.join(root, "x"))[:-1]
```
martindurant, line 40: Will `os.path` not do the wrong thing on windows? How about directly using `posixpath.join`?
(sorry for all the comments on this change!)

ap--: Good question. I made the change now.
tbh, i'm not sure though what happens with r"C:\abc\" as root or other windows style paths.
I'll try to spawn a windows VM and play around with it a bit.

martindurant:
We do run tests on windows, so maybe it doesn't matter when the root path isn't empty (which would be an illegal path on Windows anyway).

ap--:
In my benchmark it reduced _key_to_str total runtime by a factor of ~7: 2 seconds to ~0.3 seconds.

I'm moving down the call stack now, trying to reduce the overhead when accessing a zarr array, via reference fs to get as close as possible to the time it takes to do the IO.

The biggest overhead is currently due to `ReferenceFileSystem.cat` (next on the list 😃 )

Issue #1100, opened 2022-11-07
Some implementations don't have same function signatures as base class


leoleoasd:
Hi, I noticed some implementations change the default value of some parameters, for example: `AbstractFilesystem` have `detail=True` in `ls`, however `LocalFileSystem` have `detail=False`. (https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L301, https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/local.py#L56)
This is inconsistent, for example, people may expect all filesystems to behave the same in `ls` operations. Also, as `listdir` just calls `ls`, if implementations changed the signature of `ls`, calling it with `listdir` will use the default value specified in `AbstractFileSystem` since it's defined there and implementations don't rewrite that. This behavior inconsistency already caused issues (for example, Lightning-AI/lightning#3805).

So I wrote a test to find out if there is any more inconsistency, and I found a lot. (https://github.com/leoleoasd/filesystem_spec/blob/master/fsspec/implementations/tests/test_common.py#L65, https://github.com/leoleoasd/filesystem_spec/actions/runs/3407906000/jobs/5667983849)

Is this a bug, should we fix that? This fix is definitely a breaking change and may cause some code to stop working. Maybe we should, first, find what implementation has different signatures, and warn users that use its default value about the change?

martindurant:
I agree that the implementations should have the same defaults as each other, and that this default should be defined in the base class. There _might_ be occasional cases where different defaults are appropriate for some given backend, but that should be rare and need justification.

I expect much of our test suite is explicit about the optional arguments like `detail=`, and this is why we don't see any problem. Likewise, users that access multiple backends are probably explicit. 

So yes, we should fix it, but, as you say, this is a tricky thing to do without breaking downstream code. Listing the known instances in the description of this issue would be a great start.

leoleoasd:
I'll make a list tomorrow(it's 12 pm in China), but if I remember it correctly, `ls` in multiple implementation have this problem. (Maybe local filesystem changed the default value first and other implementations started with copying code from local fs?)

leoleoasd:
Methods that have a different default value:
```
test_signature[file-get_file] - AssertionError: (self, rpath, lpath, callback=<fsspec.callbacks.NoOpCallback object at 0x7fd0a8033eb0>, outfile=None, **kwargs) != (self, path1, path2, callback=None, **kwargs)
[...]

PR #1104, opened 2022-11-08
Add tests for implementation signatures

leoleoasd:
see #1100

### DIFF
--- a/fsspec/implementations/tests/conftest.py
+++ b/fsspec/implementations/tests/conftest.py
@@ -33,9 +33,11 @@ def fs(request):
     cls = FILESYSTEMS[request.param]
     return cls()
 
+
 @pytest.fixture(scope="function")
 def fscls(request):
     return fsspec.get_filesystem_class(request.param)
+
 
 @pytest.fixture(scope="function")
 def temp_file():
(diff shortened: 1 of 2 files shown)


martindurant:
That's a lot of tests :)
Fortunately, *most* of them pass.

fsspec/implementations/tests/conftest.py
```diff
@@ -24,15 +25,23 @@ class MultiProtocolFileSystem(LocalFileSystem):
...
     cls = FILESYSTEMS[request.param]
     return cls()
 
 
+@pytest.fixture(scope="function")
+def fscls(request):
+    try:
+        return fsspec.get_filesystem_class(request.param)
+    except ImportError:
+        pytest.skip(f"filesystem {request.param} not installed")
```
yarikoptic, line 42: I would have done 
```suggestion
    except ImportError as exc:
        pytest.skip(f"{request.param} not installed? Got {exc}")
```
since it is an assumption that `{request.param}` is the one causing ImportError but could be e.g. some module that one uses/relies upon

martindurant: It is only a skip message, so I'm not worried. Most accurate would be "{} could not be imported".

PR #1117, opened 2022-11-16
Fix LocalFileSystem put of file to directory

ianthomas23:
This is a candidate fix for issue #967, which is a `LocalFileSystem.put()` of a file to a directory.

There are a number of places that a fix could be applied. I have traced the problem to `AbstractFileSystem.put()` which calls `other_paths()` on the target path. This wasn't passing an `is_dir` kwarg which I think is the fundamental bug behind the OP's issue.

The fix is to pass the `is_dir` kwarg but bearing in mind the `other_paths` docstring which states
```
is_dir: bool (optional)
    For the special case where the input in one element, whether to regard the value
    as the target path, or as a directory to put a file path within. If None, a
    directory is inferred if the path ends in '/'
```
so `is_dir` is only passed as True if the length of the source paths is 1.

I've added a specific test for the OP's issue, which fails before this PR and passes with it. 

This definitely needs to be checked by someone with more `fsspec` experience than I have to confirm the fix doesn't cause problems elsewhere.

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -809,3 +809,17 @@ def test_symlink(tmpdir):
 
     fs.symlink(target, link)
     assert fs.islink(link)
+
+
+# https://github.com/fsspec/filesystem_spec/issues/967
+def test_put_file_to_dir(tmpdir):
+    src_file = os.path.join(str(tmpdir), "src")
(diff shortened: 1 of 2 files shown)


martindurant:
I think this is right.

Also, for further reference, similar issues in gcsfs and s3fs probably come down to the early stripping of "/" before hading to `other_paths`, and so might be easily fixed. https://github.com/fsspec/s3fs/issues/659

PR #1122, opened 2022-11-23
spec: introduce fsid

### LINKED ISSUE #773: [RFC] persistent filesystem id

efiop:
Adding fsid to the spec and implementing it for a few filesystems that we actually will use.

Fixes #773

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -4,10 +4,10 @@ import os
 import secrets
 import shutil
 from contextlib import suppress
-from functools import wraps
+from functools import cached_property, wraps
 
 from fsspec.spec import AbstractFileSystem
-from fsspec.utils import infer_storage_options, mirror_from
+from fsspec.utils import infer_storage_options, mirror_from, tokenize
 
 
 def wrap_exceptions(func):
@@ -49,6 +49,10 @@ class ArrowFSWrapper(AbstractFileSystem):
         PYARROW_VERSION = tuple(map(int, __version__.split(".")))
         self.fs = fs
         super().__init__(**kwargs)
+
+    @cached_property
+    def fsid(self):
+        return "hdfs_" + tokenize(self.fs.host, self.fs.port)
 
     @classmethod
     def _strip_protocol(cls, path):
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -107,6 +107,10 @@ class HTTPFileSystem(AsyncFileSystem):
         request_options.pop("skip_instance_cache", None)
         self.kwargs = request_options
 
+    @property
+    def fsid(self):
+        return "http"
+
     def encode_url(self, url):
         return yarl.URL(url, encoded=self.encoded)
 
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -36,6 +36,10 @@ class LocalFileSystem(AbstractFileSystem):
         super().__init__(**kwargs)
         self.auto_mkdir = auto_mkdir
 
+    @property
+    def fsid(self):
+        return "local"
+
     def mkdir(self, path, create_parents=True, **kwargs):
         path = self._strip_protocol(path)
         if self.exists(path):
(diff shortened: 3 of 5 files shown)


efiop:
Please excuse the delay, got distracted.

> So this is different from the fs._fs_token (or hash()) in that is should be human readable? Those tokens are more specific in that they include all the kwargs.

Human readable to some extent and persistent across sessions (fs_token depends on tid, so it is not), yes. This is meant to identify the storage/filesystem across multiple runs, no matter what credentials you are using. So for example, fsid will be the same if Alice and Bob both use, say, a shared server or s3 endpoint, each with their own credentials. There's been a bit more discussion in #773, but the way I think about fsid is like a portable fsid from your fstab.

> Is there an argument that we should try to reproduce a URL which, passed to url_to_fs reproduces the instance? Should there be a search function that, given a fsid searches all fsspec instances to find one that matches?

Totally, and `fsid` could be based on that `url` (e.g. as a hash of it). But at the same time it is trickier to build, as condensing multiple config options to make that url unique is kinda involved and results in complex urls. For example, for s3 and similar with default endpoint `url` might be something like `s3://`, but if you use a custom endpoint then, I guess, smth like `s3+https://myendpoint.com:8080`?

I can think of a few ways to identify filesystems across sessions:

1) a short fsid, which is easy to use for caching and kinda easy to implement

2) more complex but descriptive fs `url`, which can only be used for caching (as there are no creds/user info/etc), but is significantly harder to implement and will likely require non-trivial escaping to be able to use it as a part of a local path (since I assume people would want to use it as a path to cache something across sessions). Or you could just hash it and use that, but that is us going back to 1)

3) protocol and fs kwargs to be able to fully re-create the filesystem using the registry. This is more general, and easier to implement that 1 and 2, but usually contains too much info (e.g. creds/user/etc) that makes it not usable as is for caching purposes.

martindurant:
OK, I've had a chance to think about it, and I'm OK with including this. At some point it may be nice to include some text in the prose docs about how we hash, cache and compare filesystem instances, and where this attribute fits in.

Issue #1126, opened 2022-11-29
 Cached filesystem not concurrency-safe


yarikoptic:
This is #1107 reincarnated -- please see it for more details: #1111 provided remedy to only one aspect.

PR #1127, opened 2022-11-29
Wrap cache-file-management operations for CachingFileSystem in thread locks

jwodder:
Tries to address #1126.

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -7,6 +7,7 @@ import pickle
 import tempfile
 import time
 from shutil import rmtree
+from threading import RLock
 
 from fsspec import AbstractFileSystem, filesystem
 from fsspec.callbacks import _DEFAULT_CALLBACK
@@ -120,6 +121,7 @@ class CachingFileSystem(AbstractFileSystem):
             if isinstance(target_protocol, str)
             else (fs.protocol if isinstance(fs.protocol, str) else fs.protocol[0])
         )
+        self.lock = RLock()
         self.load_cache()
         self.fs = fs if fs is not None else filesystem(target_protocol, **self.kwargs)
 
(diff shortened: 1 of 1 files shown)


fsspec/implementations/cached.py
```diff
@@ -134,73 +136,76 @@ def _mkcache(self):
...
         """Reload caches if time elapsed or any disappeared"""
-        self._mkcache()
-        if not self.cache_check:
-            # explicitly told not to bother checking
-            return
-        timecond = time.time() - self.last_cache > self.cache_check
-        existcond = all(os.path.exists(storage) for storage in self.storage)
-        if timecond or not existcond:
-            self.load_cache()
+        with self.lock:
```
yarikoptic, line 200: `self.load_cache` has a lock and I think all the other points are either safe for parallel execution (nothing binding to self, and file presence check is ok to not be locked), so why to lock here?

Overall -- please analyze the code and see if may be we could avoid as much locking, and may be even avoid using reentering lock (RLock) by clearly identifying "problematic" code blocks

jwodder: @yarikoptic I've removed this lock.  I don't think any more can be removed, and `RLock` is needed due to things like `clear_cache()`.

yarikoptic:
is this PR enough to revert https://github.com/datalad/datalad-fuse/pull/83 and get dandi healthcheck work without errorring out?  if so -- do we see any more sensible performance from healthcheck?

martindurant:
The continuous writing of the cache file is designed for the situations that you might have simultaneous access to the cache from multiple processes. If you only ever have the one one FUSE process using the cache, I don't suppose there's any good reason to write the cache until program end (if you can guarantee not to crash).

yarikoptic:
Thanks @martindurant , so it feels like ideally it could be an option like `cache_dump : ["always", "exit", "periodic"]` where `"always"` is current implementation, `"exit"` where it would dump it only `atexit` (/ deconstructor call?), and "periodic" - have some thread which would look after the cache and if modified since last save -- save it (but then would need some extra setting for how often :-/ ).  May be we could start with getting `"always", "exit"` implemented @jwodder ?

martindurant:
Exactly. Again, it would be nice to decompose the caching logic into 
- what is stored
- where it is stored
- how/often it is updated
- how target files map to local files

yarikoptic:
@martindurant that sounds like a good amount of work to overhaul entire caching layer. Would you consider PR which would first just augment life cycle of the cache with "always" and "exit" strategies to when it is dumped?

yarikoptic:
@martindurant before any other major RF -- what do you think about this particular PR? IMHO it is "the best we could do" ATM to ensure thread-safety, seems to work for us in that at least we do not crash (right @jwodder ?) and should not really have noticeable penalty on users (run time performance of which would be bottlenecked primarily by all those cache read/writes anyways)

martindurant:
I think it's OK... but still share the worry that the locking is too aggressive. Are the list operations really problematic, or just the file ops (which is what the original issue was about)?

amotl:
Hi. I did not look too deep into the details of the struggles here, but I wanted to take the chance to humbly remind you about GH-895 by @gutzbenj. The patch probably needs a refresh, but the core idea was to start using [DiskCache](https://grantjenks.com/docs/diskcache/), as it promises to be a rock-solid (thread-safe and process-safe) solution to disk-based caching, without needing to run any kind of cache- or database-server. On this matter, we think the package is one of its own kind.

martindurant:
You may be right, but DiskCache should only be optional, as fsspec is very careful not to require any dependencies.

Also, you should be aware that we wish to be able to cache files to *remote* stores too! It may seem counter intuitive, but some remotes are faster than others, and egress fees often mean that copying into near-remote once is worthwhile if there are to be many reads.

amotl:
@martindurant: Thank you for sharing those important details we have not been aware of. Happy new year everyone!

martindurant:
So where shall we go from here? A full rewrite of the cache classes seems to beyond anyone's capacity right now :(

yarikoptic:
note that diskcache might be unfit for cases where location is on NFS. https://github.com/grantjenks/python-diskcache/issues/198 was "resolved" but not clear to me if there is NFS-safe way to use diskcache.

I hope that @jwodder could have a look and may be see how to refactor/structure the interface to encapsulate caching so alternative implementations could be provided.

Issue #1043, opened 2022-09-08
Improve the behavior of `du(...)` and fix comment under `du(...)` 


hozblok:
1. https://github.com/fsspec/filesystem_spec/blob/2633445fc54797c79a7ac96d213bd24dfbdfcdc2/fsspec/spec.py#L465 Passed to `find` not `ls`
1. https://github.com/fsspec/filesystem_spec/blob/2633445fc54797c79a7ac96d213bd24dfbdfcdc2/fsspec/spec.py#L455 No information about `withdirs=True|False` - I suggest adding it to the comment.
1. If we use `du(directory_path, withdirs=True)` - root directory `directory_path` will not contribute to the result and will not be taken into account. Only all nested entries will be taken into account. Looks like a bug for the disk usage.
1. If we use `du(directory_path, withdirs=True, total=False)` - sub-directories do not contain the contribution to the size of nested files. Is it ok? For example, a [du command in Linux](https://man7.org/linux/man-pages/man1/du.1.html) works by collecting sizes from directories cumulatively - I think the last is expected behaviour.

martindurant:
Thanks for the notes. Were you planning to implement your suggestion?

martindurant:
ping @hozblok

hozblok:
Hi @martindurant 
I have a lack of resources and time now. If someone is willing to do this - I will be very happy.

martindurant:
Thanks for letting us know. Hopefully someone comes to knock off at least the easier bullets! I would get to it eventually...

ianthomas23:
Most of this has been implemented in #1128 with the exception of directories including the size of their nested constituents. As explained in that PR I think this is undesirable as it would require potentially large recursion, so I am inclined to close this issue unless anyone has an objection.

martindurant:
If anyone has further improvements they desire, I suggest opening a new issue.

PR #1128, opened 2022-11-30
Improve du

ianthomas23:
Fixes most of #1043.

Changes:

1. Calling `fs.du(path, withdirs=True)` with `path` a directory will include the directory in the output as the first line (i.e. in alphabetical order).
2. Added `withdirs` as an explicit `kwarg` to function to make it easier for users to understand.
3. Docstring improvements.
4. Extended `test_api.py::test_du` and added `test_local.py::test_du`.

There was the question of whether the reported size of a directory should include the size of its contained files and directories. Linux command `du` does include the size of its contents, e.g. using something like `du -Ah -B1 some_directory`. However, I am not sure it is a good idea. It is fine if `maxdepth=None` as the directory walk will fully recurse and can add up directory totals as it walks, but what should happen if `maxdepth=1` for example? Command-line `du` gives the fully-recursed directory size, so if we are aiming for the same functionality here then to get that information we would have to fully recurse, even though the user has specified a `maxdepth`. That seems an undesirable thing to do on a potentially large filesystem. So I have  left it as it was and made it clearer in the docstring that directory size does not include its contents.

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -823,3 +823,36 @@ def test_put_file_to_dir(tmpdir):
     fs.put(src_file, target_dir)
 
     assert fs.isfile(target_file)
+
+
+def test_du(tmpdir):
+    file = os.path.join(tmpdir, "file")
+    subdir = os.path.join(tmpdir, "subdir")
+    subfile = os.path.join(subdir, "subfile")
+
+    fs = LocalFileSystem()
+    with open(file, "wb") as f:
+        f.write(b"4444")
+    fs.mkdir(subdir)
+    with open(subfile, "wb") as f:
+        f.write(b"7777777")
+
+    assert fs.du(tmpdir) == 11
+    assert fs.du(tmpdir, total=False) == {file: 4, subfile: 7}
+    # Note directory size is OS-specific, but must be positive
+    assert fs.du(tmpdir, withdirs=True) > 11
+
+    d = fs.du(tmpdir, total=False, withdirs=True)
+    assert len(d) == 4
+    assert d[file] == 4
+    assert d[subfile] == 7
+    assert d[tmpdir] > 0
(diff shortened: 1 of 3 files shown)


fsspec/implementations/tests/test_local.py
```diff
@@ -823,3 +823,36 @@ def test_put_file_to_dir(tmpdir):
...
+    d = fs.du(tmpdir, total=False, withdirs=True)
+    assert len(d) == 4
+    assert d[file] == 4
+    assert d[subfile] == 7
+    assert d[tmpdir] > 0
+    assert d[subdir] > 0
+
+    assert fs.du(tmpdir, maxdepth=2) == 11
+    assert fs.du(tmpdir, maxdepth=1) == 4
+    assert fs.du(tmpdir, maxdepth=0) == 4
```
ianthomas23, line 854: It surprised me initially that `maxdepth=0` gives the same output as `maxdepth=1`. Maybe it does makes sense though, `maxdepth` effectively has a minimum value of 1.

martindurant: Is maxdeph <=0 a ValueError then?

ianthomas23: No, currently `maxdepth <= 1` proceeds as though `maxdepth=1`.

Linux command-line `du` is maybe relevant here. A `maxdepth` less than `0` gives an error, but `maxdepth=0` when specifying a directory path gives you just the directory back without any recursion. So the equivalent behaviour in `fsspec` would be for line 854 above
```python
assert fs.du(tmpdir, maxdepth=0) == 0
```

We could classify this as a bug which should raise an error on `maxdepth<0` and handle `maxdepth=0` differently.

martindurant: OK, not going to worry about the edge case here. Anyone trying this is just being silly :)

PR #1129, opened 2022-12-01
Use builtin types.MappingProxyType rather than custom ReadOnlyRegistry

jamesmyatt:
Seems better to use the built-in immutable mapping proxy type rather than making a custom one. Was added in Python 3.3. https://docs.python.org/3/library/types.html#types.MappingProxyType

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -1,6 +1,6 @@
 import importlib
-import warnings
 import types
+import warnings
 
 __all__ = ["registry", "get_filesystem_class", "default"]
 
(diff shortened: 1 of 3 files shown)


fsspec/registry.py
```diff
@@ -80,9 +51,6 @@ def register_implementation(name, cls, clobber=True, errtxt=None):
         _registry[name] = cls
 
 
-registry = ReadOnlyRegistry(_registry)
-default = "file"
```
martindurant, line 84: This was not used anywhere?

jamesmyatt: I moved it to the top where it was more obvious. Is that OK?

martindurant: Oh yes, simply didn't notice where it had gone.

PR #1133, opened 2022-12-09
up coverage on referenceFS

martindurant:


martindurant:
Strips out some lingering dataframe related code, which will be redone shortly.

PR #1138, opened 2022-12-16
references with dataframes

martindurant:
cc @d70-t 

This implements, very experimentally, a parquet/dataframe based referenceFS, inspired by and hopefully compatible with preffs. Hoping to get the best of both! `fsspec.implementations.reference.DFReferenceFileSystem` (is not yet in the fsspec registry)

I have a companion function for kerchunk to convert JSONs to parquet. Still need to decide what to do about templates.

The plan here is to include the following features:
- multiple target filesystems
- keys with multiple targets if allow_multi=True. These can be a combination of raw data and references on multiple targets. The bits are combined with multi_func kwarg
  - a combination of whole file and file ranges
- optional chunk_func applied to the final bytes buffer (solves the different-scale-per-input problem for kerchunk)
- lazy loading of parquet subsets of references when lazy=True keyed by path prefix by default (not yet implemented)
- merging of ranges that are close enough together in target files to reduce the number of calls
- extracting out of the arrays from the parquet dataframe for speed of indexing (actually, fastparquet or pyarrow can load directly to arrays and bypass slow pandas)

This all requires a lot of testing! I am hoping to get preffs users involved now at this very early stage to make sure everything comes out right

Still to implement
- get to files
- any writing functionality and saving the new reference set; could we have a reference set built on a "manifest" where the parquet files are not in the same directory?
- probably other stuff

Optional stretch goal
- zarr does not allow for variable-length chunk sizes, but we could have columns for chunk sizes, and solve it in the storage layer
- a lot of optimisations, since there's loads of loop/comprehensions; maybe they don't matter since we may have tons of references but don't expect to be fetching too many of them in a single call.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Setup conda
         uses: mamba-org/provision-with-micromamba@main
@@ -45,6 +47,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Setup conda
         uses: mamba-org/provision-with-micromamba@main
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -71,6 +71,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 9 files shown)


martindurant:
Conversation at https://github.com/fsspec/kerchunk/pull/272 shows that this work is needed sooner rather than later!

rsignell-usgs:
This dataframe/parquet approach worked great for me with this dataset:
https://nbviewer.org/gist/rsignell-usgs/77609e86825c9c6163af15f90590b440

PR #1139, opened 2022-12-19
Update `pyarrow` version parsing

jrbourbeau:
When using `ArrowFSWrapper` with the nightly release of `pyarrow`, I got the failing error with the current version parsing logic since, for the nightly release of `pyarrow`, `pyarrow=11.0.0.dev294`:

```python
self = <fsspec.implementations.arrow.ArrowFSWrapper object at 0x13245f2e0>, fs = <pyarrow._fs.LocalFileSystem object at 0x131030270>, kwargs = {}
__version__ = '11.0.0.dev294'

    def __init__(self, fs, **kwargs):
        from pyarrow import __version__

        global PYARROW_VERSION
>       PYARROW_VERSION = tuple(map(int, __version__.split(".")))
E       ValueError: invalid literal for int() with base 10: 'dev294'

../../../mambaforge/envs/dask/lib/python3.10/site-packages/fsspec/implementations/arrow.py:49: ValueError
```

This PR updates the version parsing logic to handle this case (in addition to the usual `X.Y.Z` version format)

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -7,7 +7,12 @@ from contextlib import suppress
 from functools import cached_property, wraps
 
 from fsspec.spec import AbstractFileSystem
-from fsspec.utils import infer_storage_options, mirror_from, tokenize
+from fsspec.utils import (
+    get_package_version_without_import,
+    infer_storage_options,
+    mirror_from,
+    tokenize,
+)
 
 
 def wrap_exceptions(func):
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Initially I went to use `packaging` for version parsing, but saw it wasn't a dependency for `fsspec`. Happy to update this PR to use `packaging` if folks would like that (it'd be more robust and follow Python's recommendation for version parsing) but the current set of changes seemed more conservative.

martindurant:
We really enjoy not having any dependencies at all for this package.

jrbourbeau:
Sure, thanks @martindurant

PR #1141, opened 2022-12-21
Error if HTTP range not supported and not reading from start

### LINKED ISSUE #1140: HTTP files read the wrong data when range requests aren't supported
The description on #830 says:

> Does *not* allow random access from not the start of the download.

However, nothing appears to check for this, so it can return the wrong data. For example:

```python
import fsspec
url = 'http://piweb.ooirsn.uw.edu/das/data/Optasense/SouthCable/TransmitFiber/South-C1-LR-95km-P1kHz-GL50m-SP2m-FS200Hz_2021-11-01T16_09_15-0700/South-C1-LR-95km-P1kHz-GL50m-SP2m-FS200Hz_2021-11-01T231514Z.h5'

with fsspec.open(url) as f:
  f.seek(100)
  b = f.read(50)

print(b)
```

The data starts with `b'\x89HDF'`, the magic number at the start of an HDF5 file, instead of starting 100 bytes in. If range requests aren't supported, I'd expect `.read()` to error if you're not starting from 0.

This is somewhat complicated by the read-ahead caching - currently making multiple small reads (`[f.read(100), f.read(100)]`) does the right thing so long as the data is in the cache, but it will go wrong when it needs to fetch new data.

Ideally I guess it should use one of the streaming (non-seekable) file classes in this case, but I don't know if the `Accept-Ranges` header is used consistently enough that you can assume its absence means range requests won't work.

takluyver:
If you're happy with the basic shape of this, I can work on the details - e.g. is a generic ValueError OK, or would you prefer to derive a specific exception, and do we want to let the user override the heuristic?

Closes #1140.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -633,12 +633,29 @@ class HTTPFile(AbstractBufferedFile):
                 # range request outside file
                 return b""
             r.raise_for_status()
-            if r.status == 206:
+
+            # If the server has handled the range request, it should reply
+            # with status 206 (partial content). But we'll guess that any
+            # Content-Range header or a Content-Length no more than the
+            # requested range also mean we have got the desired range.
+            response_is_range = (
+                r.status == 206
+                or "Content-Range" in r.headers
+                or int(r.headers.get("Content-Length", end + 1)) <= end - start
+            )
+
+            if response_is_range:
                 # partial content, as expected
                 out = await r.read()
-            elif int(r.headers.get("Content-Length", 0)) <= end - start:
-                out = await r.read()
+            elif start > 0:
+                raise ValueError(
+                    "The HTTP server doesn't appear to support range requests. "
+                    "Only reading this file from the beginning is supported. "
+                    "Open with block_size=0 for a streaming file interface."
+                )
             else:
+                # Response is not a range, but we want the start of the file,
+                # so we can read the required amount anyway.
                 cl = 0
                 out = []
                 while True:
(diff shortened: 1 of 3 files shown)


martindurant:
How do you fancy playing with the test HTTP server to catch each of these possibilities?

Does anyone else have any comment on server behaviour regarding Range requests?

martindurant:
[This](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/tests/conftest.py#L30) is where you would define the various server behaviours so that you can test them.

takluyver:
Yup, I'll get on that. :slightly_smiling_face:

takluyver:
I've added some tests, and they're passing locally.

martindurant:
>  is a generic ValueError OK

Yes, since it's the message that the user really cares about. There isn't an obvious workaround that calling code should do (block_size=0??). Maybe we'll get a request later.

> do we want to let the user override the heuristic?

I can't think how they would do that. You mean a "I trust my server to give the right bytes even when the headers are wrong" flag? I don't think we should bother.

takluyver:
> There isn't an obvious workaround that calling code should do (block_size=0??)

Great, I've already put the `block_size=0` option in the error message.

> You mean a "I trust my server to give the right bytes even when the headers are wrong" flag? I don't think we should bother.

Yup, that's what I was thinking. :+1: to not bothering. The specs are clear that ignoring Range entirely is valid behaviour, so hopefully it's clear to anyone implementing a server that the response needs an indication that it's the requested range.

martindurant:
Open to any further comments for one day.

PR #1145, opened 2023-01-03
Fix type returned by calling .ls on memory file

FHTMitchell:
`MemoryFileSystem.ls()` would always return a list with a single detailed dict in it, even when `detail=False` (the default). It should return a list with a single str as per the docs.


martindurant:
Thanks for this. Can we also change the default to `detail=True`, which is what AbstractFileSystem and the others have?

FHTMitchell:
OK done, I suppose it's ok for me to fix this test?

martindurant:
Yes please

Issue #1062, opened 2022-10-05
fs.put() behaves inconsistently depending on whether the rpath already exists


tgaddair:
When calling `fs.put()` to upload a local directory to a remote filesystem, if the remote directory does not exist then only the local directory contents will be uploaded. However, if the remote directory does exist already, then the entire directory will be uploaded to the rpath. This makes certain sync patterns (like periodically calling `put()` to sync local files to the remote) difficult to implement.

Repro:

```
import os
import tempfile
from pathlib import Path
import fsspec

fs = fsspec.filesystem("file")

# at the beginning, /tmp/test does not exist
rpath = "/tmp/test"
with tempfile.TemporaryDirectory() as tmpdir:
    Path(os.path.join(tmpdir, "foo.txt")).touch()

   # in this call the file is uploaded as /tmp/test/foo.txt
    fs.put(tmpdir, rpath, recursive=True)

    # in this call the file is uploaded as /tmp/test/tmp045wchhz/foo.txt
    fs.put(tmpdir, rpath, recursive=True)
```

I've verified this behavior also happens when using s3fs. I assume it is true for all filesystem backends.

The expected behavior is that the file in the above should be written as `/tmp/test/foo.txt` for both calls. This is, for example, what happens when we wrap the fsspec filesystem with pyarrow:

```
pyarrow.fs.copy_files(
    tmpdir, rpath, destination_filesystem=pyarrow.fs.PyFileSystem(pyarrow.fs.FSSpecHandler(fs))
)
```

Using `fsspec==2022.8.2`, but also happened in `2022.7.1`.

martindurant:
The behaviour you describe as suggested is not that provided by typical CLI shells, like posix `cp -r`: if a directory exists, the whole source tree is put inside it, but if the target directory doesn't exist, the root directory is copied directly to the target path. This **may** depend on whether the terminating "/" is entered as part of the target path.

martindurant:
I guess it's a separate question as to what should be done when requesting a recursive copy on a single file...

tgaddair:
I did try with the trailing `/`, but that didn't change the result.

I think even if your goal is be consistent with `cp`, there should probably be a way to achieve a conventional "sync" operation with fsspec. Even with `cp`, there are ways to make the behavior consistent, for example:

```
mkdir test1
touch test1/foo.txt
cp -r test1/ test2/
# test2/
#.   foo.txt

touch test1/bar.txt
cp -r test1/ test2/
# test2/
[...]

PR #1148, opened 2023-01-05
Consistent recursive cp, get and put into non-existent target directory

ianthomas23:
This is a WIP for #1062. Running the test script from that issue gives the following output:
```
==> cp
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/source', '/tmp/target/source/a']
==> cp, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> rsync
  loop 0: ['/tmp/target/source', '/tmp/target/source/a']
  loop 1: ['/tmp/target/source', '/tmp/target/source/a']
==> rsync, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> scp
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/source', '/tmp/target/source/a']
==> scp, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> fs.cp
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/source', '/tmp/target/source/a']
==> fs.cp, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> fs.get
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/source', '/tmp/target/source/a']
==> fs.get, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> fs.put
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/source', '/tmp/target/source/a']
==> fs.put, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> mem.cp
  loop 0: ['/remote_target/a']
  loop 1: ['/remote_target/a', '/remote_target/remote_source', '/remote_target/remote_source/a']
==> mem.cp, slash
  loop 0: ['/remote_target/a']
  loop 1: ['/remote_target/a']
==> mem.get
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a', '/tmp/target/remote_source', '/tmp/target/remote_source/a']
==> mem.get, slash
  loop 0: ['/tmp/target/a']
  loop 1: ['/tmp/target/a']
==> mem.put
  loop 0: ['/remote_target/a']
  loop 1: ['/remote_target/a', '/remote_target/source', '/remote_target/source/a']
==> mem.put, slash
  loop 0: ['/remote_target/a']
  loop 1: ['/remote_target/a']
```
showing that the `fsspec` functions `cp`, `get` and `put` are consistent with command-line `cp` and `scp`.

The solution is mostly about passing the correct arguments to `utils.other_paths()` to obtain the required behaviour.

It is only a WIP for two reasons:

1. I have broken the use case of `fs.copy(['/tmp/source/a', '/tmp/source/b'], '/tmp/target')` and don't yet have a fix for it.
2. I have only fixed `other_paths()` calls in `spec.py` not in `asyn.py`, which presumably needs similar logic.

### DIFF
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -214,3 +214,23 @@ def test_cp_get_put_directory_recursive(m, tmpdir, funcname):
         func(src + "/", target, recursive=True)
         assert target_fs.isdir(target)
         assert target_fs.find(target) == [make_path_posix(os.path.join(target, "file"))]
+
+
+def test_cp_two_files(m):
+    src = "/src"
+    file0 = os.path.join(src, "file0")
+    file1 = os.path.join(src, "file1")
+    m.mkdir(src)
+    m.touch(file0)
+    m.touch(file1)
+
+    target = "/target"
+    assert not m.exists(target)
+
+    m.cp([file0, file1], target)
+
+    assert m.isdir(target)
+    assert sorted(m.find(target)) == [
+        "/target/file0",
+        "/target/file1",
+    ]
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -2,6 +2,8 @@ import os
 
 import pytest
 
+from fsspec.implementations.local import LocalFileSystem, make_path_posix
+
 
 def test_1(m):
     m.touch("/somefile")  # NB: is found with or without initial /
@@ -26,9 +28,11 @@ def test_put_single(m, tmpdir):
     os.mkdir(fn)
     open(os.path.join(fn, "abc"), "w").write("text")
     m.put(fn, "/test")  # no-op, no files
+    assert m.isdir("/test")
     assert not m.exists("/test/abc")
     assert not m.exists("/test/dir")
-    m.put(fn + "/", "/test", recursive=True)
+    m.put(fn, "/test", recursive=True)
+    assert m.isdir("/test/dir")
     assert m.cat("/test/dir/abc") == b"text"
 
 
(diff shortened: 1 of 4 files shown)


ianthomas23:
This now works for the synchronous case, including for copying 2+ named files. I still need to update and test the async side of things.

fsspec/implementations/tests/test_memory.py
```diff
@@ -159,3 +163,86 @@ def test_remove_all(m):
     m.touch("afile")
     m.rm("/", recursive=True)
     assert not m.ls("/")
+
+
+@pytest.mark.parametrize("funcname", ["cp", "get", "put"])
+def test_cp_get_put_directory_recursive(m, tmpdir, funcname):
```
ianthomas23, line 169: I aimed to keep this test the same as the equivalent one in `test_local.py` but it has become rather complex to get it to work on Windows. I could always split it into 3 constituent tests (`cp`, `get` and `put`).

martindurant: I like splitting it

fsspec/spec.py
```diff
@@ -878,12 +878,18 @@ def get(self, rpath, lpath, recursive=False, callback=_DEFAULT_CALLBACK, **kwarg
...
 
         if isinstance(lpath, str):
             lpath = make_path_posix(lpath)
         rpaths = self.expand_path(rpath, recursive=recursive)
-        lpaths = other_paths(rpaths, lpath)
+        isdir = isinstance(lpath, str) and LocalFileSystem().isdir(lpath)
+        lpaths = other_paths(
+            rpaths,
+            lpath,
+            exists=isdir and isinstance(rpath, str) and not rpath.endswith("/"),
```
martindurant, line 890: Is the path ever not a string here, do we test this?

martindurant (commented):
All looks good! 
One question.

fsspec/spec.py
ianthomas23: Yes and yes. The path can be a sequence of strings, e.g. `get(["file1", "file2"], "target_dir")`, and the multiple paths case is tested in the new `test_cp_two_files` functions in both `test_local.py` and `test_memory.py`.

martindurant: Got it, thanks. Good to be reminded of my design choices :)

PR #1149, opened 2023-01-10
Fix read the docs build process

ianthomas23:
Read the docs builds are failing as the version 2 config file wasn't correct, and the function `fsspec_loop` has been removed.

PR #1150, opened 2023-01-10
Workaround for RTD not showing colons for function args

### LINKED ISSUE #790: API docs on RTD meld parameters and types without separator (such as `: `)
see e.g. https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.reference.ReferenceFileSystem

image

ianthomas23:
Fixes #790 which is that Read the Docs does not display colons between argument names and types.

There is a discussion about this on the `sphinx_rtd_theme` github https://github.com/readthedocs/sphinx_rtd_theme/issues/766. The workaround here is the one recommended in that issue.

This PR is made on top of #1149 so that RTD will successfully build the docs so the workaround can be checked. Ideally #1149 should be merged before this.

Longer term it might be worth us considering a switch from the RTD theme to one that is more actively maintained. Options that work well for relatively small documentation projects are `furo` (https://github.com/pradyunsg/furo) and `sphinx-book-theme` (https://github.com/executablebooks/sphinx-book-theme).

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,0 +1,5 @@
+.classifier:before {
+    font-style: normal;
+    margin: 0.5em;
+    content: ":";
+}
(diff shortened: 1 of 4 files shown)


ianthomas23:
Here is some output from the RTD build showing that the fix works.

Screenshot 2023-01-10 at 12 15 25

martindurant:
Perfect, thank you.

PR #1151, opened 2023-01-10
Validate maxdepth kwarg

### LINKED ISSUE #1135: Correctly handle maxdepth kwarg
As identified in discussion of PR #1128, `maxdepth` kwarg that is used in `walk` and related functions can be improved in two ways:

- Raise as an error if `maxdepth < 0`.
- `maxdepth=0` should be handled differently to `maxdepth=1`.

ianthomas23:
Fixes #1135.

This PR adds checks that kwarg `maxdepth` passed to `walk` and `expand_paths` is at least `1` and raises a `ValueError` if not.

Strictly speaking the check isn't really needed in `expand_paths` as it eventually calls `find` which calls `walk`, but I thought it better to add an explicit check at the top of the function so that the traceback is clear.

I have also added the fix to the `asyn.py` functions `_walk` and `_expand_paths` and corresponding async tests in `test_http.py`. This is my first attempt to write async tests in `fsspec` so I expect they could be improved; any advice gratefully received!

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -577,6 +577,9 @@ class AsyncFileSystem(AbstractFileSystem):
         raise NotImplementedError
 
     async def _walk(self, path, maxdepth=None, **kwargs):
+        if maxdepth is not None and maxdepth < 1:
+            raise ValueError("maxdepth must be at least 1")
+
         path = self._strip_protocol(path)
         full_dirs = {}
         dirs = {}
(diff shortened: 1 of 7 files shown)


fsspec/asyn.py
```diff
@@ -751,6 +755,8 @@ async def _expand_path(self, path, recursive=False, maxdepth=None):
                 if p not in out and (recursive is False or (await self._exists(p))):
                     # should only check once, for the root
                     out.add(p)
+            # reduce depth on each recursion level unless None or 0
+            maxdepth = maxdepth if not maxdepth else maxdepth - 1
```
ianthomas23, line 759: Decrementing `maxdepth` has moved from the beginning to the end of this `else` clause otherwise it is decremented before it is first used.

fsspec/spec.py
```diff
@@ -429,6 +432,8 @@ def walk(self, path, maxdepth=None, topdown=True, **kwargs):
         if maxdepth is not None:
             maxdepth -= 1
             if maxdepth < 1:
+                if not topdown:
+                    yield path, dirs, files
```
ianthomas23, line 436: I have should have put this in PR #1128 but it only came to light when writing all the tests for this PR.

martindurant: It's a _good_ thing that these are coming out, thank you.

martindurant:
Perfect, thank you.

PR #1152, opened 2023-01-10
Order lists in docs alphabetically

ianthomas23:
There are a few lists in the docs, mostly in the API reference. This PR changes the order of the list contents to be alphabetical rather than the order they were added, as I find this easier to scan. But I don't really have a strong view about this.


martindurant:
Seems like a reasonable thing to do, yes.

Issue #1153, opened 2023-01-12
ArrowFile: `seekable()` method


nyamashi:
`ArrowFile.seekable()` seems  to always return `False` even setting `seekable=True` in  https://github.com/fsspec/filesystem_spec/blob/2022.11.0/fsspec/implementations/arrow.py#L152.

 I guess it should be added `"seekable"` in https://github.com/fsspec/filesystem_spec/blob/2022.11.0/fsspec/implementations/arrow.py#L196

```
import os
import fsspec

hdfs_url = 'hdfs://foo/bar/baz/'

fs, data_path = fsspec.core.url_to_fs(hdfs_url)

for p1, p2, p3 in fs.walk(data_path):
    if p3:
        for filename in p3:
            path = os.path.join(p1, filename)
            # print(path)
            f = fs.open(path, model="rb", seekable=True)
            print(f.seekable())
            break 
```
before
```
False
```

with `"seekable"` in https://github.com/fsspec/filesystem_spec/blob/2022.11.0/fsspec/implementations/arrow.py#L196
```
True
```

Could 'seekable' be added?

martindurant:
Yes please!

nyamashi:
@martindurant 
Thanks for your quick response. 😄 
I've added PR https://github.com/fsspec/filesystem_spec/pull/1154. 
Is that enough?

PR #1154, opened 2023-01-13
[add] seekable for ArrowFile

nyamashi:
Fix https://github.com/fsspec/filesystem_spec/issues/1153

ref:
https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem
- `open_*_streams` and `open_input_file` of `pyarrow.fs.HadoopFileSystem` return `pyarrow.NativeFile`.
- `pyarrow.NativeFile` supports `seekable`.

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -200,7 +200,7 @@ class ArrowFSWrapper(AbstractFileSystem):
 
 
 @mirror_from(
-    "stream", ["read", "seek", "tell", "write", "readable", "writable", "close", "size"]
+    "stream", ["read", "seek", "tell", "write", "readable", "writable", "close", "size", "seekable"]
 )
 class ArrowFile(io.IOBase):
     def __init__(self, fs, stream, path, mode, block_size=None, **kwargs):


nyamashi:
Got it! I'll try to add a test in `test_arrow.py`.

nyamashi:
I'd added a test case but I couldn't checked via `pytest fsspec`. 🤔 
However the following code looks working on my local environment.

```
def test_seekable(fs, hdfs_path):
    for seekable in [True, False]:                                                                                                                                                   
        with fs.open(hdfs_path, "rb", seekable=seekable) as file:                                                                                                        
            assert file.seekable() == seekable 
```
```
my_hdfs_url = 'hdfs://foo/bar/baz/'
fs, my_hdfs_path = fsspec.core.url_to_fs(my_hdfs_url)
test_seekable(fs, my_hdfs_path)
```
  
Could you run the test code on your environment? 🙇

f4hy:
This looks like it makes it possible to fix https://github.com/pandas-dev/pandas/issues/50639 but it is unclear how pandas could pass the seekable option.

PR #1158, opened 2023-01-19
Fix reference fs cat path list

ap--:
Hello everyone,

here's a potential fix for the issue revealed in zarr-developers/zarr-python#1324

I added a test to check the order of returned data when using `FSMap.getitems`
Without the fix, it errors with

```python
    def test_mapping_getitems(m):
        m.pipe({"a": b"A", "b": b"B"})
    
        refs = {
            "a": ["a"],
            "b": ["b"],
        }
        h = fsspec.filesystem("memory")
        fs = fsspec.filesystem("reference", fo=refs, fs=h)
        mapping = fs.get_mapper("")
>       assert mapping.getitems(["b", "a"]) == {"a": b"A", "b": b"B"}
E       AssertionError: assert {'a': b'B', 'b': b'A'} == {'a': b'A', 'b': b'B'}
E         Differing items:
E         {'b': b'A'} != {'b': b'B'}
E         {'a': b'B'} != {'a': b'A'}
E         Full diff:
E         - {'a': b'A', 'b': b'B'}
E         + {'a': b'B', 'b': b'A'}

```

Regarding the fix: I'm not sure if this should be solved differently, since the non-merging codepath is probably the more common. Also I haven't tested if `DFReferenceFileSystem.cat_ranges` has the same issue...

Cheers,
Andreas 😃

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -357,7 +357,8 @@ class ReferenceFileSystem(AsyncFileSystem):
             bytes_out = fs.cat_ranges(new_paths, new_starts, new_ends)
             if len(urls2) == len(bytes_out):
                 # we didn't do any merging
-                for p, b in zip(paths2, bytes_out):
+                for url, b in zip(new_paths, bytes_out):
+                    p = paths2[urls2.index(url)]
                     out[p] = b
             else:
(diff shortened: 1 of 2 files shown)


martindurant:
Don't worry about DFReferenceFileSystem, it is strictly experimental.

PR #1163, opened 2023-01-24
Async cp/get/put the same as sync behaviour

ianthomas23:
This PR implements the correct `_cp`, `_get` and `_put` behaviour for `AsyncFileSystem` that PR #1148 implemented for the synchronous `AbstractFileSystem`. The changes are all in `asyn.py` and are almost identical to those made in `spec.py` except where a check for a trailing slash needs to be made earlier on as a subsequent call to `_strip_protocol` also strips the trailing slash.

There are no tests added here, instead I will shortly submit PRs to both `s3fs` and `gcsfs` to add them in line with the similar tests made in PR #1148. The tests I have added to those `s3fs` and `gcsfs` branches all fail when run against the latest commit of `fsspec` and pass when run against this PR. We will need to coordinate the CI/merging of the three PRs before we can be completely sure that everything consistently works together.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -340,7 +340,13 @@ class AsyncFileSystem(AbstractFileSystem):
             on_error = "raise"
 
         paths = await self._expand_path(path1, maxdepth=maxdepth, recursive=recursive)
-        path2 = other_paths(paths, path2)
+        isdir = isinstance(path2, str) and await self._isdir(path2)
+        path2 = other_paths(
+            paths,
+            path2,
+            exists=isdir and isinstance(path1, str) and not path1.endswith("/"),
+            is_dir=isdir,
+        )
         batch_size = batch_size or self.batch_size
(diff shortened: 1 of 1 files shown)


martindurant:
The CIs pull from the other repos main/master branches, but could be set temporarily to the incoming branch. Im not actually worried if CI shows red for a day or more while the different components land, so maybe best not to change the CI settings.

Having made all these changes, it would be interesting to compare with what generic.rsync does for similar situations.

martindurant:
Also, the green ✅ suggests that gcsfs needs more tests, we should have expected something to break!

martindurant:
:) our rsync doesn't really get close to the many options of gnu rsync

martindurant:
The code here looks good, I am happy to merge when you say you are done, and the changes in gcsfs and s3fs can follow.

@hayesgb @TomAugspurger - this may effect adlfs

ianthomas23:
> The code here looks good, I am happy to merge when you say you are done, and the changes in gcsfs and s3fs can follow.

I am done here, and happy for this to be merged.

PR #1166, opened 2023-01-26
Add codespell config and workflow + just few typo fixes

yarikoptic:
If accepted -- no typo (besides declared to be skipped) should sneak in, and @yarikoptic would no longer contribute silly typos fixing PRs ;-)


martindurant:
Thanks for contributing this and sorry it took me so long. I will merge as is, but reserve the right to remove it should it prove too annoying!

PR #1172, opened 2023-02-01
Add Python 3.11 to CI

ianthomas23:
This PR adds Python 3.11 to CI. Tests all pass locally on macOS using Python 3.11. 

I've also updated the classifiers to indicate 3.11 is supported.

The `versioneer.py` that we vendor only claims support up to Python 3.9 but I haven't experienced any problems with it and left it as it is.


martindurant:
(at some point we may need to bin versioneer in favour of something new...)

PR #1173, opened 2023-02-02
Faulty condition for non-merged ranges in ReferenceFS

martindurant:
Fixes https://github.com/fsspec/kerchunk/issues/291


martindurant:
@pl-marasco - seems to fix your case. I also recommend you always use absolute paths rather than relative ones in your scripts.

pl-marasco:
@martindurant positively tested, it definetly solve the issue! thanks as well for the tip.

PR #1178, opened 2023-02-04
Added block cache with pre-loading of the data in the background thread.

VladyslavUsenko:
Hi everyone,

this PR contains an implementation of BackgroundBlockCache that tries to pre-fetch the next block of data in the background thread before we actually need it for reading. This should reduce the time our main thread spends blocked on IO. 

One example use case for this type of cache is video playback (similar to YouTube). We want to pre-fetch some blocks in front to avoid video "freezing" while we have to load new data.

This implementation passes all the tests and works well for me. The only concern I have is thread-safety of `super()._fetch`. Is it safe to run it from a separate thread?

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -234,6 +234,7 @@ Read Buffering
    fsspec.caching.MMapCache
    fsspec.caching.ReadAheadCache
    fsspec.caching.FirstChunkCache
+   fsspec.caching.BackgroundBlockCache
 
 .. autoclass:: fsspec.caching.BlockCache
    :members:
@@ -248,6 +249,9 @@ Read Buffering
    :members:
 
 .. autoclass:: fsspec.caching.FirstChunkCache
+   :members:
+
+.. autoclass:: fsspec.caching.BackgroundBlockCache
    :members:
 
 Utilities
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -1,10 +1,10 @@
+import collections
 import functools
 import io
 import logging
 import math
 import os
 import warnings
-import collections
 from concurrent.futures import ThreadPoolExecutor
 
 logger = logging.getLogger("fsspec")
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -4,6 +4,8 @@ import logging
 import math
 import os
 import warnings
+import collections
+from concurrent.futures import ThreadPoolExecutor
 
 logger = logging.getLogger("fsspec")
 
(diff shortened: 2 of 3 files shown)


martindurant:
Thanks for putting so much work into this!

Are there any beta testers out there who would like to give this a go?

It might be tricky to think up good tests of functionality here. I wonder, what would happen if the calling environment were itself multi-threaded?

VladyslavUsenko:
I've made some patches to make _fetch method thread-safe when calling environment is multi-threaded. Let me know if you think there are any issues with it.

Here is a small snippet that highlights the functionality:
```python
import fsspec
import time
from functools import partial

fs = fsspec.filesystem("http")
# Binary file stored on http server (~150 MB).
file_path = 'https://vision.in.tum.de/tumvi/calibrated/512_16/dataset-calib-cam7_512_16.bag'

def test_read(cache_type, block_size=5*1024*1024):
    with fs.open(file_path, mode='rb', cache_type=cache_type) as fp:
        for chunk in iter(partial(fp.read, block_size), b''):
            # Simulate CPU intensive processing of the chunk with sleep. 
            time.sleep(1)

if __name__ == "__main__":
    import timeit
    print("background:", timeit.timeit('test_read("background")', number=1, setup="from __main__ import test_read"))
    print("block:", timeit.timeit('test_read("block")', number=1, setup="from __main__ import test_read"))
```
On my system I get the following results (in seconds):
```
background: 33.90016986196861
block: 62.28582467697561
```
With "background" cache the processing time is around twice faster because it pre-loads the data in the background while processing the current chunk.

I'm not sure if it is easy to make a unit test out of this sample, but it passes all the tests in `fsspec/tests/test_caches.py` that are used to test other cache implementations.

martindurant:
I added a minimalist test to show that fetches happen in a different thread than where the file object lives. The code is fairly long and complicated, so I would appreciate more eyes on it than just my own, but your numbers above sound good. I wonder, have you also tried with block_size=None (i.e., pure streaming, no separate byte-range requests)?

VladyslavUsenko:
I've added a small fix. Hopefully it fixes the tests.

I have tried running the example with "none" cache. I think it should correspond to pure streaming: 
```
print("none:", timeit.timeit('test_read("none")', number=1, setup="from __main__ import test_read"))
```
Here are the results that I get:
```
none: 62.67970448406413
background: 34.75059202709235
block: 62.23740289919078
```
In the end, there are ~30 seconds of download time and ~30 seconds of processing time (in this case sleep time). If you do not run data loading in a separate thread it will always add up.

martindurant:
Ah, sorry once request: please add this to the reference API in the docs/source/api.rst under "Read buffering"

VladyslavUsenko:
> Ah, sorry once request: please add this to the reference API in the docs/source/api.rst under "Read buffering"

Done

Issue #897, opened 2022-01-27
filesystem.get mixes up filenames (when using GitHub)


FirefoxMetzger:
Disclaimer: I might be going about this the wrong way. If I do, please enlighten me :D

I have the following snippet which (I hoped) would copy 3 files from a remote into a local folder:

```python
fs = fsspec.filesystem("github", org="githubtraining", repo="hellogitworld")
fs.get(
    ["README.txt", "/resources/labels.properties", "/src/test/java/com/github/AppTest.java"],
    ["README.txt", "labels.properties", "AppTest.java")],
)
```

It does copy the content of the three files; however the order is mixed up. I.e., the copying happens as follows (order: remote --> local):

- "README.txt" --> "AppTest.java"
- "/resources/labels.properties" --> "README.txt"
- "/src/test/java/com/github/AppTest.java" --> "labels.properties"

I haven't dug into the source of fsspec, but my hunch is that it has to do with GH having a different order of files remotely than what is listed in the query.

Is this intended behavior? If so, what is the rationale behind it?

martindurant:
I agree with your analysis, probably fsspec is assuming the ordering of files somewhere, but this is not the case for the returned files. It is *not* the intended behaviour. Do you think you have the time to try to find the problem?

FirefoxMetzger:
@martindurant Maybe 😅 I have a deadline on Thursday next week, so no time to investigate until then. If I remember (or you ping me) on the weekend after that, then yes I can look into that.

martindurant:
No rush, I haven't had any other complaints about this. I assume most people access single specific files with the github backend (so no problem with ordering) or, if they want many files, just clone the repo.

merrysailor:
Hello,

I believe the problem is in `AsyncFileSystem._get` and this line:
rpaths = await self._expand_path(rpath, recursive=recursive)

because in `AsyncFileSystem.expand_path` the rpath(s) are added to a set that is subsequently sorted alphanumerically.

Is it possible to address this please? IMHO it is a serious bug in what is very useful functionality, i.e., getting files in bulk from http-like filesystems, not just github, and likely also affects _copy & _cat.

Thank you.

martindurant:
Do I understand you have a fix in mind, or at least a test case that demonstrated the problem? @merrysailor

merrysailor:
Hi,

[...]

PR #1179, opened 2023-02-07
Fix mixup of paths in put function

quassy:
`put` currently sorts the list of input paths so when they are not in order before already, they don't match their output paths anymore and file names get mixed up. This fixes https://github.com/fsspec/gcsfs/issues/468

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -935,7 +935,7 @@ class AbstractFileSystem(metaclass=_Cached):
         if isinstance(lpath, str):
             lpath = make_path_posix(lpath)
         fs = LocalFileSystem()
-        lpaths = fs.expand_path(lpath, recursive=recursive)
+        lpaths = fs.expand_path(lpath, recursive=recursive, sort=False)
         isdir = isinstance(rpath, str) and self.isdir(rpath)
         rpaths = other_paths(
             lpaths,
@@ -992,7 +992,7 @@ class AbstractFileSystem(metaclass=_Cached):
                 if on_error == "raise":
                     raise
 
-    def expand_path(self, path, recursive=False, maxdepth=None):
+    def expand_path(self, path, recursive=False, maxdepth=None, sort=True):
         """Turn one or more globs or directories into a list of all matching paths
         to files or directories."""
         if maxdepth is not None and maxdepth < 1:
@@ -1026,7 +1026,10 @@ class AbstractFileSystem(metaclass=_Cached):
             maxdepth = maxdepth if not maxdepth else maxdepth - 1
         if not out:
             raise FileNotFoundError(path)
-        return list(sorted(out))
+        if sort:
+            return list(sorted(out))
+        else:
+            return list(out)
 
     def mv(self, path1, path2, recursive=False, maxdepth=None, **kwargs):
         """Move file(s) from one location to another"""


quassy:
Note: In ```for lpath, rpath in callback.wrap(zip(lpaths, rpaths)):```, `zip` should probably be called with [`strict=True`](https://peps.python.org/pep-0618/).

ianthomas23:
> @ianthomas23 , do you have time to review this?

Yes, I can take a look tomorrow morning, UK time.

ianthomas23:
@quassy Thanks for looking at this. I think the solution ends up being more complicated, as is often the case. At the end of `expand_path` if you don't sort `out` then your return is in the order that `out` gives you, and `out` is a `set`, so the order is not the same as the original order. A `set` is used because we want lookups to be quick.

To proceed with this I think there are a few options:

1. Use an ordered set, i.e. a set that preserves insertion order. There isn't an ordered set in the python standard library though. There is a package on PyPI but it isn't very actively maintained and we generally try to avoid adding new dependencies to `fsspec`.
2. Use an `OrderedDict` instead of the `set`, as there is one in the standard library. We only really want it for the keys so the values can be anything that may or may not be useful.
3. Keep the `set` but also add a `list` or similar as well, so we can keep the quick lookup but also the order that was passed in.

Personally I think option 2 is the best, although there might be alternatives that I haven't thought of yet. Are you prepared to try out an `OrderedDict` in this PR?

ianthomas23:
Issue #897 is related to this too.

quassy:
@ianthomas23 Thanks as well for the quick response! _Everything is trivial, except..._

I'm not actually sure anymore if expand_path for lpath makes sense at all and what the expected behaviour of put is, when you do something like this below, because the order of glob is not really known:
```py
# src_directory/c.txt
# src_directory/b.txt
put(lpath=["src_directory/*.txt"], rpath=["gs://dst_bucket/c_path/", "gs://dst_bucket/b_path/"])
# at the moment b.txt would be written to c_path

put(lpath=["src_directory/*.txt"], rpath=["gs://dst_bucket/c_path/", "gs://dst_bucket/b_path/", "gs://dst_bucket/a_path/"])
# this would fail as other_paths() requires the two resulting lists to be the same length
```

Some ideas which I'm all not super happy with:
1. Maybe it would be cleaner if `put()` didn't expand paths at all. Rather a library user does that on their own before and then pass two ordered lists (or a list of tuples `(lpath, rpath)`) that remain mostly untouched. Quite bad for backwards compability.
2. Only call `expand_path()` on lpath if the second argument is a string (or a list with 1 element) so order does not matter and it's clear where glob-matched files get written to.
4. To be backwards compatible add a parameter to `put()` called `expand_lpath=True` so one can set it to `False` and skip the expand_path call as well as the unintentional reordering. But this would still lead to unexpected behaviour if you didn't know that expand_path is actually reordering your inputs.

martindurant:
@ianthomas23 , happy to have a live chat about the best way forward here

ianthomas23:
> @ianthomas23 , happy to have a live chat about the best way forward here

Good idea, I have booked a slot in your calendar tomorrow.

martindurant:
@quassy , can you please check if this is still an issue given @ianthomas23 's work in the area?

PR #1180, opened 2023-02-07
add rust impl

martindurant:
Uses https://github.com/martindurant/rfsspec


martindurant:
Going to close this here - any rust implementation should live in rfsspec until it is much more mature

Issue #1165, opened 2023-01-24
Import of `fsspec/implementations/arrow.py` fails due to 3.8+ code


philsc:
We're seeing this error with the new `2023.1.0` release on Python 3.7:
```
ImportError:
 cannot import name 'cached_property' from 'functools' 
(/var/.../python3.7/lib/python3.7/functools.py)
```
This is happening because `cached_property` only exists on Python 3.8 or higher.
https://docs.python.org/3/library/functools.html#functools.cached_property

Looks like this was introduced in #1122.

Solutions I see:
1. Use the pip package on 3.7: https://pypi.org/project/cached-property/
2. Bump minimum required Python version to 3.8.

martindurant:
We should probably remove py37, as so many adjacent OSS repos already have. However, there's not really a way to go back and fix someone installing this version and facing the specific import error. 
Pyarrow does still release for py37, but conda-forge no longer builds these.

PR #1182, opened 2023-02-08
fsspec.implementations.reference: allow import in python 3.7

ap--:
Hello @martindurant,

I noticed this due to a failing CI test running on python3.7.

```python
tiffslide/tiffslide.py:34: in <module>
    from fsspec.implementations.reference import ReferenceFileSystem
E     File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/fsspec/implementations/reference.py", line 7[97](https://github.com/bayer-science-for-a-better-life/tiffslide/actions/runs/4129073532/jobs/7134279388#step:6:98)
E       if x := self.dataframes[pref]["raw"][i]:
E             ^
E   SyntaxError: invalid syntax
```

This PR should restore fsspec's Python3.7 compatibility.

Cheers,
Andreas 😃

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -812,7 +812,8 @@ class DFReferenceFileSystem(AbstractFileSystem):
             if isinstance(inds, int):
                 inds = [inds]
             for i in inds:
-                if x := self.dataframes[pref]["raw"][i]:
+                x = self.dataframes[pref]["raw"][i]
+                if x:
                     thislist.append(x)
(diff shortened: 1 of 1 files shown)


ap--:
I just saw: https://github.com/fsspec/filesystem_spec/issues/1165#issuecomment-1403854578

I'll exclude the last fsspec version for the py3.7 version of my library and then it doesn't have to be fixed here, and fsspec's lower version can be raised.

PR #1183, opened 2023-02-09
Allow numpy-like inputs to async cat/ranges

martindurant:
Fixes https://github.com/fsspec/kerchunk/issues/294

cc @hakon-matland-adsk


martindurant:
cc @jon-ryfetten-adsk

PR #1184, opened 2023-02-10
raise lower bound to python>=3.8

ap--:
Addresses #1165 by bumping minor python version to 3.8 and removes now obsolete checks/workarounds for `python < 3.8`

The only version with minor compatibility issues with python 3.7 is `2023.1.0` which would require this patch: https://github.com/ap--/filesystem_spec/commit/256378afa3bbf350602d553632bda146d39e1008
Not sure if it's worth the effort to release a fixed version `2023.1.1` to pypi though.

### DIFF
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -1,11 +1,4 @@
-try:
-    from importlib.metadata import entry_points
-except ImportError:  # python < 3.8
-    try:
-        from importlib_metadata import entry_points
-    except ImportError:
-        entry_points = None
-
+from importlib.metadata import entry_points
 
 from . import _version, caching
 from .callbacks import Callback
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -6,7 +6,6 @@ import io
 import numbers
 import os
 import re
-import sys
 import threading
 from contextlib import contextmanager
 from glob import has_magic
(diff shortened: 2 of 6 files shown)


martindurant:
Thanks for picking up on this!

martindurant:
As to the broken 3.7 version, once we have a follow up release, we can yank it. So py >= 3.8 will get new ones, and 3.7 will get the last good version before.

ap--:
Sounds good!

https://github.com/ap--/filesystem_spec/tree/python-3.7-2023.1.0 <- this branch branches off 2023.1.0, contains the above mentioned commit, and would be ready for tagging and release.

martindurant:
I suppose we could also have a 2023.1.0post1 :|

ap--:
> I suppose we could also have a 2023.1.0post1 :|

[pep440#post-releases](https://peps.python.org/pep-0440/#post-releases) says:
_The use of post-releases to publish maintenance releases containing actual bug fixes is strongly discouraged. In general, it is better to use a longer release number and increment the final component for each maintenance release._

But I don't know if `2023.1.0.1` (I think tifffile does it this way) or `2023.1.1` would then be better...

ap--:
But I guess it doesn't really matter, and the post versioning would at least fix installs for the few that run py37 and have ==2023.1.0 pinned ? ... 🤷

martindurant:
Thanks for this, and please feel free to ping me more on the 3.7 release early next week

PR #1185, opened 2023-02-12
Missing: raise the lower bound to >=3.8

ap--:
Seems, I forgot to actually raise the bound in #1184 ... :see_no_evil: 
@martindurant


martindurant:
Hah, yes we need that!

PR #1188, opened 2023-02-16
Lazy Parquet References

agoodm:
For some background see the discussion from this [kerchunk issue thread](https://github.com/fsspec/kerchunk/issues/293) and this [pangeo discourse thread](https://discourse.pangeo.io/t/trick-for-improving-kerchunk-performance-for-large-numbers-of-chunks-files/3090) which discusses a method I have proposed to further improve the performance for loading very large numbers of references, in particular the ability to lazily load small numbers of them at a time using parquet row groups. Here I have added a custom mapping class that can be made to work when passed to a `ReferenceFilesystem` but there are a few other ideas from the `DFReferenceFilesystem` implementation that need to be tested incorporated.

Code to generate the compatible parquet dataframes is tracked in this [separate PR](https://github.com/fsspec/kerchunk/pull/298).

(Note: I changed the name of the mapping class from `ParquetReferenceMapper` to `LazyReferenceMapper` since I thought this might be more appropriate)

Some things to work on:
- [x] Support separate dtype columns
- [x] Wrap LazyReferenceMapper inside `ReferenceFilesystem` constructor when user input is a directory
- [ ] ~~Test with an intake catalog~~
- [ ] preffs style support (multiple paths per chunk key)?

### DIFF
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -11,6 +11,8 @@ dependencies:
   - lz4<3.1.3
   - pyftpdlib
   - cloudpickle
+  - fastparquet
+  - pandas
   - pytest
   - pytest-asyncio
   - pytest-benchmark
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -279,6 +279,9 @@ class ReferenceFileSystem(AsyncFileSystem):
         simple_templates=True,
         max_gap=64_000,
         max_block=256_000_000,
+        cache_size=128,
+        engine="fastparquet",
+        categorical_urls=False,
         **kwargs,
     ):
         """
@@ -287,8 +290,10 @@ class ReferenceFileSystem(AsyncFileSystem):
         ----------
         fo : dict or str
             The set of references to use for this instance, with a structure as above.
-            If str, will use fsspec.open, in conjunction with target_options
-            and target_protocol to open and parse JSON at this location.
+            If str referencing a JSON file, will use fsspec.open, in conjunction
+            with target_options and target_protocol to open and parse JSON at this
+            location. If a directory, then assume references are a set of parquet
+            files to be loaded lazily.
         target : str
             For any references having target_url as None, this is the default file
             target to use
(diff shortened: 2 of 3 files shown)


martindurant:
Sorry I lost track of this one. I will be happy to merge along with https://github.com/fsspec/kerchunk/pull/298 and we can come back and make any small speed improvements that may be available after that. I think we can safely delete my parquet DF referenceFS from fsspec.implementaions.reference , or maybe for one release cycle mark it as deprecated in the docstring.

agoodm:
Just a quick interesting but encouraging development I wanted to share here, following our discussion on what would happen if `kerchunk.combine` was used in conjunction with `LazyReferenceMapper` I had some initial concerns about performance since the iteration was hundreds of times slower than iterating over the dict natively. The main reason for this is all the additional overhead in converting string keys to integer chunk indices and then figuring out which parquet file to load from them. So in my last commit I implemented some additional tricks so that separate logic to generate key value pairs gets used when iterating over the entire reference set. It will never be possible to iterate more quickly than an actual dict but if the iteration time is still lower than the combined time to load the original references from JSON while using much less memory, it should be a big win. Here I was able to get some very encouraging numbers:

```
In [25]: mapper = LazyReferenceMapper('refs_noaa_rec/')

In [26]: %%memit
    ...: for k, v in mapper.items():
    ...:     pass
    ...: 
peak memory: 156.10 MiB, increment: 26.98 MiB

In [27]: def load_refs_noaa():
    ...:     with open('refs_noaa.json.zst', 'rb') as f:
    ...:         s = f.read()
    ...:     blob = zstandard.decompress(s)
    ...:     refs = ujson.loads(blob)
    ...:     for k, v in refs.items():
    ...:         pass
  
In [28]: %%time
     ...: for k, v in mapper.items():
     ...:     pass
     ...: 
CPU times: user 2min 7s, sys: 913 ms, total: 2min 8s
Wall time: 2min 11s

In [29]: %memit load_refs_noaa()
peak memory: 42047.29 MiB, increment: 40838.20 MiB

In [30]: %time load_refs_noaa()
CPU times: user 2min 34s, sys: 32.5 s, total: 3min 6s
Wall time: 3min 7s
```

martindurant:
Again, I made a PR https://github.com/agoodm/filesystem_spec/pull/1 rather than put in comments here. I suspect there are still performance gains fairly easily to be had, but if we don't see any problems with things as they are, no need to chase those. My changes are mostly about simplifying the implementation, so that I could understand it.

martindurant:
Somehow the diff here seems to include a whole load of commits that are already in master. I am no git ninja, I don't know why this would happen.

agoodm:
This happened as a result of me rebasing my branch to include these commits so that the diff in yours would become more manageable. I can remove these later once I merge it into this one

agoodm:
Ok, I think I managed to clean up the commits with another rebase. Does this look good to go?

PR #1191, opened 2023-02-20
spec: get_file: don't pollute with files and fds

efiop:
Currently if anything happens in fs.open we will leave an empty file in user's workspace and also won't close the file descriptor. Both things are pretty bad.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -848,25 +848,26 @@ class AbstractFileSystem(metaclass=_Cached):
         """Copy single remote file to local"""
         if isfilelike(lpath):
             outfile = lpath
-        else:
-            if self.isdir(rpath):
-                os.makedirs(lpath, exist_ok=True)
-                return None
+        elif self.isdir(rpath):
+            os.makedirs(lpath, exist_ok=True)
+            return None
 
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -848,25 +848,26 @@ def get_file(
         """Copy single remote file to local"""
         if isfilelike(lpath):
             outfile = lpath
-        else:
-            if self.isdir(rpath):
-                os.makedirs(lpath, exist_ok=True)
-                return None
+        elif self.isdir(rpath):
+            os.makedirs(lpath, exist_ok=True)
+            return None
```
efiop, line 853: These don't seem mutually exclusive, but I've preserved logic that was here before. Not sure what `outfile` argument is even used for, but it is strange that `makedirs`-behaviour (which is also odd by itself) doesn't affect it.

efiop: Also calling `isdir` on every file is very wasteful. Ideally, we should get rid of this at some point in the future, as it seems to be a misplaced piece of logic from `fs.get` rather than something that is intentional here. From user's perspective I would expect `IsADirectoryError` here.

Issue #1130, opened 2022-12-05
Feature Request: single environment variable config for protocol


mauvilsa:
Currently it is possible to provide overrides with environment variables of the style `FSSPEC_{protocol}_{kwargname}=value`. However, this is limited since it is not possible to provide a dict value. A need for this is for example s3 with a custom `endpoint_url`, see https://github.com/fsspec/s3fs/issues/432.

I propose to support a single environment variable `FSSPEC_{protocol}` expecting its value to be json or yaml (parsed with pyyaml). This will make the previous s3 use case possible. It is just an extension so there would be no breaking changes. Current individual variables `FSSPEC_{protocol}_{kwargname}=value` would override the more general `FSSPEC_{protocol}`.

martindurant:
It is reasonable to have environment variables interpreted as JSON. Either they would need to be labelled "json:{...}" or we adopt a try/except to interpret all variables. 
It still leaves the problem where we actually only want to set one nested parameter like endpoint_url, but the user wishes to have freedom to change other ones. Right now, they would need to create the whole set of kwargs, but one could imagine wanting to be more clever about merging sources of arguments.
Likewise, this does not solve for cases where the argument is not a simple type (although JSON does at least allow us to differentiate between strings, ints and floats).

mauvilsa:
If I were to create a pull request implementing this support, how likely is it to be accepted? Any recommendations before thinking about implementing?

martindurant:
Sorry, this fell into the holiday abyss.

Yes, I would be happy to consider a PR, but it may take some iteration to get right.

>  can you please give an examples where the argument is not a simple type

The timeout parameter to HTTP? It should be a `aiohttp.ClientTimeout`.

mauvilsa:
> Yes, I would be happy to consider a PR, but it may take some iteration to get right.

Created #1194.

> The timeout parameter to HTTP? It should be a `aiohttp.ClientTimeout`.

This would be easy to do with jsonargparse, see [class-type-and-sub-classes](https://jsonargparse.readthedocs.io/en/stable/#class-type-and-sub-classes). Not great to add a dependency, but could be implemented as an optional feature only available if jsonargparse is installed.

A downside is that currently in fsspec the parsing of environment variables is unaware of protocols, and there would be a need to restrict which classes are allowed for each protocol. Allowing any class would be a potential security risk. How to restrict the classes could be resolved with a bit of discussion.

mauvilsa:
[...]

PR #1194, opened 2023-02-23
Add support for single json dict environment variable for protocol

mauvilsa:
Implements #1130.

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -371,8 +371,8 @@ Alternatively, you can provide overrides with environment variables of the style
 
 Configuration is determined in the following order, with later items winning:
 
-#. the contents of ini files, and json files in the config directory, sorted
-   alphabetically
+#.  ini and json files in the config directory (``FSSPEC_CONFIG_DIRECTORY`` or ``$HOME/.config/fsspec/``), sorted
+   lexically by filename
 #. ``FSSPEC_{protocol}`` environment variables
 #. ``FSSPEC_{protocol}_{kwargname}`` environment variables
 #. the contents of ``fsspec.config.conf``, which can be edited at runtime
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -365,14 +365,16 @@ For instance:
 Obviously, you should only define default values that are appropriate for
 a given file system implementation. INI files only support string values.
 
-Alternatively, you can provide overrides with environment variables of
-the style ``FSSPEC_{protocol}_{kwargname}=value``.
+Alternatively, you can provide overrides with environment variables of the style
+``FSSPEC_{protocol}=<json_dict_value>`` and
+``FSSPEC_{protocol}_{kwargname}=<string_value>``.
 
 Configuration is determined in the following order, with later items winning:
 
 #. the contents of ini files, and json files in the config directory, sorted
    alphabetically
-#. environment variables
+#. ``FSSPEC_{protocol}`` environment variables
+#. ``FSSPEC_{protocol}_{kwargname}`` environment variables
 #. the contents of ``fsspec.config.conf``, which can be edited at runtime
 #. kwargs explicitly passed, whether with ``fsspec.open``, ``fsspec.filesystem``
    or directly instantiating the implementation class.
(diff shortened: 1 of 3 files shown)


martindurant:
Thanks for working on this!
Will look at this early next week - please ping if you don't hear anything.

fsspec/config.py
```diff
@@ -21,12 +26,24 @@ def set_conf_env(conf_dict, envdict=os.environ):
...
+        if key.startswith("FSSPEC_") and len(key) > 7 and key[7] != "_":
+            if key.count("_") > 1:
+                kwarg_keys.append(key)
                 continue
-            _, proto, kwarg = key.split("_", 2)
-            conf_dict.setdefault(proto.lower(), {})[kwarg.lower()] = envdict[key]
+            try:
+                value = json.loads(envdict[key])
+            except json.decoder.JSONDecodeError:
+                pass
```
timsnyder, line 38: Should we log this exception?  Silently ignoring the error seems like it could be annoying, especially given how strict the JSON parser is.

fsspec/config.py
```diff
@@ -21,12 +26,24 @@ def set_conf_env(conf_dict, envdict=os.environ):
...
+            else:
+                if isinstance(value, dict):
+                    _, proto = key.split("_", 1)
+                    conf_dict.setdefault(proto.lower(), {}).update(value)
```
timsnyder, line 42: If your assumption about the env vars always being a dict at the toplevel is violated, it would also be good to log rather than silently ignore.

timsnyder (commented):
some logging of the silently dropped cases might be helpful.  I know it would help me as a user.

docs/source/features.rst
```diff
@@ -365,14 +365,16 @@ For instance:
...
-Alternatively, you can provide overrides with environment variables of
-the style ``FSSPEC_{protocol}_{kwargname}=value``.
+Alternatively, you can provide overrides with environment variables of the style
+``FSSPEC_{protocol}=<json_dict_value>`` and
+``FSSPEC_{protocol}_{kwargname}=<string_value>``.
 
 Configuration is determined in the following order, with later items winning:
 
 #. the contents of ini files, and json files in the config directory, sorted
    alphabetically
```
timsnyder, line 375: Not really a part of this PR but...

Every time I read this first item, I find myself thinking "where the heck does fsspec think the 'config directory' is"?  And then I have to go search and find https://github.com/fsspec/filesystem_spec/blob/05b886ab23ae9de7d101d0621476ba61d66df29d/fsspec/config.py#L6-L7

Could we lump in this change here:
```suggestion
#.  ini and json files in the config directory (``FSSPEC_CONFIG_DIRECTORY`` or $HOME/.config/fsspec/), sorted
   lexically by filename
```

martindurant: Feel free to repeat, yes

fsspec/config.py
martindurant: Yes, this should be logged. A logger can exist specifically in this module, as there are a number of things that we may want to output about the config. A would even consider upping this to a warning, since the possibility of using this method is new, and we want to make people use valid configs.

fsspec/tests/test_config.py
```diff
@@ -14,7 +14,20 @@ def clean_conf():
...
+    env = {
+        "FSSPEC": "missing_proto",
+        "FSSPEC_": "missing_proto",
+        "FSSPEC__INVALID_KEY": "invalid_proto",
+        "FSSPEC_INVALID1": "not_json_dict",
+        "FSSPEC_INVALID2": '["not_json_dict"]',
+    }
+    cd = {}
+    set_conf_env(conf_dict=cd, envdict=env)
+    assert cd == {}
```
martindurant, line 27: I we make warnings, then we should assert that we have 5 raised here

fsspec/config.py
mauvilsa: It is now a warning.

fsspec/config.py
mauvilsa: Also now a warning.

fsspec/tests/test_config.py
mauvilsa: Done.

martindurant:
Thank you!

PR #1199, opened 2023-03-01
Copy/get/put empty directories correctly

### LINKED ISSUE #1198: Local file system copies empty dirs incorrectly
There are some inconsistencies in the way the local (haven't tested with s3 yet) file system implementation copies folders.  Basically if the source folder is empty, the source folder itself is copied into the destination folder, so instead of being empty, the destination folder becomes `/dest/source`.

```
import pytest
import fsspec

local = fsspec.filesystem("file")

@pytest.fixture
def source_folder():
    # Set up source directory for testing
    parent_temp = tempfile.mkdtemp()
    src_dir = os.path.join(parent_temp, "source", "")
    nested_dir = os.path.join(src_dir, "nested")
    local.mkdir(nested_dir)
    local.touch(os.path.join(src_dir, "original.txt"))
    local.touch(os.path.join(nested_dir, "more.txt"))
    yield src_dir
    shutil.rmtree(parent_temp)

@pytest.fixture
def empty_source_folder():
    # Set up source directory for testing
    parent_temp = tempfile.mkdtemp()
    src_dir = os.path.join(parent_temp, "source", "")
    local.mkdir(src_dir)
    yield src_dir
    shutil.rmtree(parent_temp)

def test_local_provider_get_l(source_folder):
    assert source_folder.endswith("/")
    with tempfile.TemporaryDirectory() as dest_tmpdir:
        dest_tmpdir = dest_tmpdir + "/"
        local.get(source_folder, dest_tmpdir, recursive=True)

        print(f"\nDest non empty {dest_tmpdir}")
        for i, f in enumerate(local.find(dest_tmpdir, withdirs=True)):
            print('  ', i, f)

def test_local_provider_get_empty_l(empty_source_folder):
    assert empty_source_folder.endswith("/")
    with tempfile.TemporaryDirectory() as dest_tmpdir:
        dest_tmpdir = dest_tmpdir + "/"
        local.get(empty_source_folder, dest_tmpdir, recursive=True)

        print(f"\nDest copying empty {dest_tmpdir}")
        for i, f in enumerate(local.find(dest_tmpdir, withdirs=True)):
            print('  ', i, f)
```

This issue was cut from a discussion on https://github.com/fsspec/filesystem_spec/issues/724

ianthomas23:
Fixes #1198.

`cp`, `get` and `put` of an empty directory wasn't correctly respecting the presence of a trailing slash.

Fix is located in `other_paths()` so will correct all sync and async uses. There were separate code paths for `len(paths)==1` and `len(paths)>1` source paths, and the handling of the trailing slash using the `exists` arg was already in the `>1` code path so needed to also be in the `==1` code path. This change has allowed me to combine both code paths into one thus making the function simpler.

With this change the new "empty directory" tests pass. The only existing tests that failed as a result of it were two of the `test_other_paths` which I have updated to include the `exists` argument which has been presumably been added since those tests were first created.

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -910,6 +910,55 @@ def test_cp_get_put_directory_recursive(tmpdir, funcname):
         assert fs.find(target) == [make_path_posix(os.path.join(target, "file"))]
 
 
+@pytest.mark.parametrize("funcname", ["cp", "get", "put"])
+def test_cp_get_put_empty_directory(tmpdir, funcname):
+    # https://github.com/fsspec/filesystem_spec/issues/1198
+    # cp/get/put of empty directory.
+    fs = LocalFileSystem(auto_mkdir=True)
+    empty = os.path.join(str(tmpdir), "empty")
+    fs.mkdir(empty)
+
+    target = os.path.join(str(tmpdir), "target")
+    fs.mkdir(target)
+
+    if funcname == "cp":
+        func = fs.cp
+    elif funcname == "get":
+        func = fs.get
+    elif funcname == "put":
+        func = fs.put
+
+    # cp/get/put without slash, target directory exists
+    assert fs.isdir(target)
+    func(empty, target)
+    assert fs.find(target, withdirs=True) == [
+        make_path_posix(os.path.join(target, "empty"))
+    ]
+
+    fs.rm(target + "/empty", recursive=True)
+
+    # cp/get/put with slash, target directory exists
+    assert fs.isdir(target)
+    func(empty + "/", target)
+    assert fs.find(target, withdirs=True) == []
+
+    fs.rm(target, recursive=True)
+
+    # cp/get/put without slash, target directory doesn't exist
+    assert not fs.isdir(target)
+    func(empty, target)
+    assert fs.isdir(target)
+    assert fs.find(target, withdirs=True) == []
+
+    fs.rm(target, recursive=True)
+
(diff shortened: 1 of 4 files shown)


martindurant:
> combine both code paths into one thus making the function simpler. 

This I like!

martindurant:
The failure on py3.10 is a temporary network thing, can rerun.

The gcsfs failures seem to be a real consequence - do we need changes in that repo?

ianthomas23:
> The gcsfs failures seem to be a real consequence - do we need changes in that repo?

Probably, I'll take a look. I checked that this PR didn't cause any problems in `s3fs` but not `gcsfs`!

ianthomas23:
@martindurant The `gcsfs` tests need to stay as they are, and I need to change this fix to accommodate them. The problem is actually what we were talking about offline earlier today, a call to `gcsfs.isdir(path)` always returns `False` for directories that are empty even though they may effectively be directories in that you can put files in them.  The fix in `fsspec` is to be more tolerant and change the `isdir` test for target directories in the various `cp`, `get` and `put` functions to something more like
```python
fs.isdir(target_path) or target_path.endswith("/")
```
There is similar `endswith("/")` code in the same functions when checking the `source_path` so it seems a reasonable thing to do.

martindurant:
> fs.isdir(target_path) or target_path.endswith("/")

Does it make sense for order of the checks be reversed for the sake of performance?

martindurant:
Perfect, thank you

PR #1210, opened 2023-03-10
set location by webhdfs server

dkim010:
The location has query parameters when using WebHDFS.
[The query parameters is encrypted by apache knox](https://knox.apache.org/books/knox-2-0-0/user-guide.html#WebHDFS+URL+Mapping).
Therefore, the location should not be manipulated like `location.replace("CREATE", "APPEND")`.
The webhdfs location should be refered by WebHDFS server.

### DIFF
--- a/fsspec/implementations/webhdfs.py
+++ b/fsspec/implementations/webhdfs.py
@@ -420,7 +420,9 @@ class WebHDFile(AbstractBufferedFile):
                 location, headers={"content-type": "application/octet-stream"}
             )
             out2.raise_for_status()
-        self.location = location.replace("CREATE", "APPEND")
+            # after creating empty file, change location to append to
+            out2 = self.fs._call("APPEND", "POST", self.path, redirect=False, **kwargs)
+            self.location = self.fs._apply_proxy(out2.headers["Location"])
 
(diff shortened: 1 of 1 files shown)


martindurant:
OK, I stared at this a little while before realising you are right. I don't suppose we can test this in CI, since we don't knox. Can you confirm that the change is working for you? What error did you get without it?

dkim010:
Without this change, I got the error like `HTTPError: 500 Server Error: Internal Server Error for url: {knox server url with gateway}/webhdfs/v1/tmp/test.parquet/part.0.parquet?_=AAAACAAAABAAAABwwbuPoogq3K69nF5VTG1_EEfZsIqFGtsH2QopCbeuTzQKxPJzEDP03-U9L8kFGVNQ-XdqvxbwRJqa070w61K-CSOE50ICNykkdoicNVNPu1IzRuASi95aNBCum9qRylJqagkNSwnNmqGc1bL_7a1hhWyaa_B1mvdYazVBCcikL8_NZiICKxQ05A `.
It shows that the query parameters are encrypted, as mentioned the document.
I confirmed that the change works correctly, because I already the patch in my project with the private pypi.

PR #1211, opened 2023-03-11
Fix BackgroundBlockCache. Do not clean up the fetch variables if we do not need the block for a current read.

VladyslavUsenko:
Sorry, but it looks like I've introduced a bug when trying to make the cache thread safe if `fetch` is called from different threads.

The current version removes the background fetch future if it does not need the result for the current read. It results in constantly creating new fetch tasks. This PR fixes the issue, such that we clean up background fetch futures, only if we are done or we have a local copy and have to wait for results. If we do not need the result for current read, it will just keep the current future and check if fetching is done on the next call.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -661,6 +661,9 @@ class BackgroundBlockCache(BaseCache):
                     self._fetch_block_cached.add_key(
                         self._fetch_future.result(), self._fetch_future_block_number
                     )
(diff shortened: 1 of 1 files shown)


martindurant:
Does the current testing touch both places in the code?

VladyslavUsenko:
Yes. The code is functional even in the previous version, but it keeps re-creating a background fetch thread if we do not need it for the current read (and it did not finish yet).

PR #1216, opened 2023-03-14
Abstract test harness

ianthomas23:
This is a proof of concept of an abstract test harness in `fsspec` that can be used to define, in a fairly generic way, the set of tests that every implementation should pass. The tests only need to be written once in `fsspec` and each of the implementations can inherit and run these tests with minimal code.

There are just 3 tests included so far, one each of `cp`, `get` and `put`, to test remote to remote, remote to local and local to remote transfers. I have `s3fs` and `gcsfs` code to work with this, and all these run and pass on my local machine.

The tests are defined in `fspec/test/abstract`, one class per `fsspec` function to be tested and each class contains one (it will be more!) test. There is an `AbstractFixtures` class that provides local filesystem functionality.

There are two implementations, `local` and `memory`. The code to inherit and hence to run the tests is really simple, just a class that inherits from each of the `Abstract` test classes and also from a mixin class that derives from `AbstractFixtures`. The mixin class, e.g. `MemoryFixtures`, provides the filesystem to be used by the tests and is simple for `local` and `memory` but will be much more complicated for `s3fs` and `gcsfs` for example. The test class inheritance and use of mixin for fixtures wasn't the first approach I tried but it works well with the `pytest` approach of dynamically finding tests and fixtures.

At this stage the plan is just to run it through CI to see what fails, I expect Windows tests will produce some problems.

### DIFF
--- a/fsspec/implementations/tests/local/local_fixtures.py
+++ b/fsspec/implementations/tests/local/local_fixtures.py
@@ -1,0 +1,16 @@
+import pytest
+
+from fsspec.implementations.local import LocalFileSystem
+from fsspec.tests.abstract import AbstractFixtures
+
+
+class LocalFixtures(AbstractFixtures):
+    @staticmethod
+    @pytest.fixture
+    def fs():
+        return LocalFileSystem(auto_mkdir=True)
+
+    @staticmethod
+    @pytest.fixture
+    def fs_path(tmpdir):
+        return str(tmpdir)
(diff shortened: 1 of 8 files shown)


fsspec/implementations/tests/local/local_fixtures.py
```diff
@@ -0,0 +1,16 @@
...
+class LocalFixtures(AbstractFixtures):
+    @staticmethod
+    @pytest.fixture
+    def fs():
+        return LocalFileSystem(auto_mkdir=True)
+
+    @staticmethod
+    @pytest.fixture
+    def fs_path(tmpdir):
+        return str(tmpdir)
```
martindurant, line 16: You probably want this posix-ised, since we explicitly use "/" elsewhere in the tests, and windows will only accept posix paths if they are wholly posix.

PR #1217, opened 2023-03-15
Bug fixes on AbstractAsyncStreamedFile

### LINKED ISSUE #1204: Bug report - AbstractAsyncStreamedFile - asynchronous writes are not working!
Hi, There seems to be a bug in `fsspec/asyn.py` - `AbstractAsyncStreamedFile`. 

I'm attempting to use `AbstractAsyncStreamedFile` to read/write data asynchronously but unfortunately i couldn't perform write operation as there seems to be a bug when calling `_upload_chunk` from `flush` method in `AbstractAsyncStreamedFile`. it doesn't seem to be awaiting. 

Error: 

```
RuntimeWarning: coroutine 'StreamFile._upload_chunk' was never awaited
  if self._upload_chunk(final=force) is not False:
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Snippet highlighting the problem. 
```python
        if self._upload_chunk(final=force) is not False: 
            self.offset += self.buffer.seek(0, 2)
            self.buffer = io.BytesIO()
```
_upload_chunk should be awaited.  Could anyone please cross check this and let me know if this intentional for some reason. I'm happy to make a PR too!

vminfant:
`AbstractAsyncStreamedFile`-  `_upload_chunk` from `flush` method is now modified to be awaited. 

Also, `_initiate_upload` method is now made non mandatory to be implemented by Async-backends (same behaviour as `AbstractBufferedFile`).

Fixes #1204

### DIFF
--- a/fsspec/tests/test_async.py
+++ b/fsspec/tests/test_async.py
@@ -193,5 +193,5 @@ async def test_async_streamed_file_write():
 async def test_async_streamed_file_read():
     test_fs = DummyAsyncFS()
     streamed_file = await test_fs.open_async("misc/foo.txt", mode="rb")
-    await streamed_file.read()
+    assert (await streamed_file.read(3)) + (await streamed_file.read()) == b"foo-bar"
     await streamed_file.close()
--- a/fsspec/tests/test_async.py
+++ b/fsspec/tests/test_async.py
@@ -1,5 +1,6 @@
 import asyncio
 import inspect
+import io
 import os
 import time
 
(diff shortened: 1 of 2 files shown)


fsspec/tests/test_async.py
```diff
@@ -145,3 +145,53 @@ async def go():
...
+            block_size,
+            autocommit,
+            cache_options=cache_options,
+            **kwargs,
+        )
+
+
+class DummyAsyncStreamedFile(fsspec.asyn.AbstractAsyncStreamedFile):
+    async def _fetch_range(self, start, end):
+        return "foo-bar"[start:end]
```
martindurant, line 178: I suppose if this should emulate a real streaming file, start!=self.loc is an error, right?

fsspec/tests/test_async.py
```diff
@@ -145,3 +145,53 @@ async def go():
...
+
+    async def _upload_chunk(self, final=False):
+        return final
+
+
+@pytest.mark.asyncio
+async def test_async_streamed_file_write():
+    test_fs = DummyAsyncFS()
+    streamed_file = await test_fs.open_async("misc/foo.txt", mode="wb")
+    await streamed_file.write("foo-bar".encode("utf8"))
```
martindurant, line 188: This is actually a no-op, right? Or should we check at least that `streamed_file.loc` changed.

fsspec/tests/test_async.py
```diff
@@ -145,3 +145,53 @@ async def go():
...
+    await streamed_file.close()
+
+
+@pytest.mark.asyncio
+async def test_async_streamed_file_read():
+    test_fs = DummyAsyncFS()
+    streamed_file = await test_fs.open_async("misc/foo.txt", mode="rb")
+    await streamed_file.read()
```
martindurant, line 196:
```suggestion
    assert (await streamed_file.read(3)) + (await streamed_file.read()) == b"foo-bar"
```

the point being that we want to ensure not just that await must happen, but that the output is what we expect; and with two different variants of read().

martindurant (commented):
Thanks for putting this up.

I am not sure if the write functionality is actually being tested here; _upload_chunk would need to actually do something for that.

fsspec/tests/test_async.py
vminfant: Yes. that is correct - but we will never have `start != self.loc` as `self.loc` is being used as a `start` from read method.

fsspec/tests/test_async.py
vminfant: I have expanded the test `test_async_streamed_file_write` to assert both self.loc as well the data being streamed out.

vminfant:
@martindurant .. Thank you for taking the time to review. Sorry about the mishap - my intention was just to ensure the async methods work as intended but didnt pay much attention to functionality itself. I have expanded the tests now. please have a look and let me know if any feedbacks! thanks.

martindurant:
Perfect, thank you.

PR #1218, opened 2023-03-16
mv: no-op when paths are the same

DavidKatz-il:
fixes: https://github.com/fsspec/s3fs/issues/711

**Previous:** If you run `mv('path', 'path')`, the `mv` function will call `cp` and then `rm`, resulting in the removal of the original 'path'.
**Revised:** If you run `mv('path', 'path')`, the `mv` function will not perform any action as the source and destination paths are the same.

### DIFF
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -79,6 +79,13 @@ def test_mv_recursive(m):
     assert not m.exists("src")
 
 
+def test_mv_same_paths(m):
+    m.mkdir("src")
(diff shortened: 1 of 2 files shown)


martindurant:
Any chance of adding a test in one of the backends?

DavidKatz-il:
Where should we add this test (which test file)?

martindurant:
I suggest fsspec.implementations.tests.test_memory and use a test fixture `m` (an instance of MemoryFileSystem) like all the other tests there.

PR #1222, opened 2023-03-27
Handle local file URI without authority

SaravananSathyanandhaQC:
The `fsspec` Local implementation currently doesn't handle file URIs that skip the authority properly, as per https://www.rfc-editor.org/rfc/rfc8089#appendix-A - ref: https://github.com/fsspec/universal_pathlib/issues/81

(The pytest CI seems to hang on my fork, not certain why)

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -631,7 +631,8 @@ def test_strip_protocol_expanduser():
 def test_strip_protocol_no_authority():
     path = "file:\\foo\\bar" if WIN else "file:/foo/bar"
     stripped = LocalFileSystem._strip_protocol(path)
-    assert stripped == "/foo/bar"
+    assert "file:" not in stripped
+    assert stripped.endswith("/foo/bar")
 
 
 def test_mkdir_twice_faile(tmpdir):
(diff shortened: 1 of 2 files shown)


fsspec/implementations/tests/test_local.py
```diff
@@ -628,6 +628,13 @@ def test_strip_protocol_expanduser():
     assert not LocalFileSystem._strip_protocol("./").endswith("/")
 
 
+def test_strip_protocol_no_authority():
+    path = "file:\\foo\\bar" if WIN else "file:/foo/bar"
+    stripped = LocalFileSystem._strip_protocol(path)
+    assert "file:" not in stripped
+    assert stripped.endswith("/foo/bar")
```
martindurant, line 635: Why not `== "/foo/bar"`, is there any other possibility?

SaravananSathyanandhaQC: Because on Windows the absolute path includes a drive letter so becomes `C:/foo/bar`

SaravananSathyanandhaQC: See failed windows run https://github.com/SaravananSathyanandhaQC/filesystem_spec/actions/runs/4530930619/jobs/7980467557

martindurant: Got it, thank you.

martindurant:
Seems like CI is having problems generally... we will investigate.

martindurant:
Now github itself is having issues, so we wait :)
https://www.githubstatus.com/

SaravananSathyanandhaQC:
The green tick of joy doesn't show here for some reason, but it's showing on my fork

PR #1225, opened 2023-03-29
Fix doc build warning about missing newline after list in docstring

ianthomas23:
Latest doc build (https://readthedocs.org/projects/filesystem-spec/builds/19933772/) has a warning:
```
<snip>/reference.py:docstring of fsspec.implementations.reference.ReferenceFileSystem.__init__:44: WARNING: Definition list ends without a blank line; unexpected unindent.
```

This PR removes the warning, although it is a little strange having an extra line of whitespace in a docstring just to keep the doc build happy.

At some point we should revisit the idea of having the RTD build appear in CI so that it is easy to check that the doc build succeeds on each PR. But it is not a high priority of course.

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -332,6 +332,7 @@ class ReferenceFileSystem(AsyncFileSystem):
                 - a dict of protocol:filesystem, where each value is either a filesystem
                   instance, or a dict of kwargs that can be used to create in
                   instance for the given protocol
+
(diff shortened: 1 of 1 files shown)


ianthomas23:
I've confirmed RTD is happy with this: https://readthedocs.org/projects/filesystem-spec/builds/

martindurant:
Thanks for the fix!
One would think there ought to be a tool out there to automatically deal with this kind of thing...

PR #1227, opened 2023-03-29
RF: add a helper add_cache to avoid duplication (.name) and allow for formalized addition of caches

yarikoptic:
Aiming to possibly to expand with another cache without (initially) contributing to fsspec, thought to define helper.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -777,7 +777,7 @@ def register_cache(cls, clobber=False):
 
     Parameters
     ----------
-    overload: bool, optional
+    clobber: bool, optional
         If overload is set to True (default is False) - allow to overload existing
         entry.
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -766,10 +766,13 @@ class BackgroundBlockCache(BaseCache):
             return b"".join(out)
 
 
-caches = {}
+caches = {
+    # one custom case
+    None: BaseCache,
+}
 
 
-def add_cache(cls, overload=False):
+def register_cache(cls, clobber=False):
     """'Register' cache implementation.
 
     Parameters
@@ -783,28 +786,20 @@ def add_cache(cls, overload=False):
     ValueError
     """
     name = cls.name
-    if not overload and name in caches:
+    if not clobber and name in caches:
         raise ValueError(f"Cache with name {name!r} is already known: {caches[name]}")
     caches[name] = cls
-    # custom handling for None
-    if name == "none":
-        if not overload:
-            if None in caches:
-                raise ValueError(f"Cache for None is already known: {caches[None]}")
-        caches[None] = cls
 
 
-[
-    add_cache(c)
-    for c in (
-        BaseCache,
-        MMapCache,
-        BytesCache,
-        ReadAheadCache,
-        BlockCache,
-        FirstChunkCache,
-        AllBytes,
-        KnownPartsOfAFile,
-        BackgroundBlockCache,
-    )
-]
+for c in (
+    BaseCache,
+    MMapCache,
+    BytesCache,
+    ReadAheadCache,
+    BlockCache,
+    FirstChunkCache,
+    AllBytes,
+    KnownPartsOfAFile,
+    BackgroundBlockCache,
+):
+    register_cache(c)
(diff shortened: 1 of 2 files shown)


fsspec/caching.py
```diff
@@ -766,15 +766,45 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
-    "mmap": MMapCache,
-    "bytes": BytesCache,
-    "readahead": ReadAheadCache,
-    "block": BlockCache,
-    "first": FirstChunkCache,
-    "all": AllBytes,
-    "parts": KnownPartsOfAFile,
-    "background": BackgroundBlockCache,
-}
+caches = {}
```
martindurant, line 769: I think it would be good to keep `None` here rather than have the extra logic in `add_cache` for this case.

fsspec/caching.py
```diff
@@ -766,15 +766,45 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
+
+
+def add_cache(cls, overload=False):
```
martindurant, line 772: We use "overwrite" or "clobber" in other places.

fsspec/caching.py
```diff
@@ -766,15 +766,45 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
+    caches[name] = cls
+    # custom handling for None
+    if name == "none":
+        if not overload:
+            if None in caches:
+                raise ValueError(f"Cache for None is already known: {caches[None]}")
+        caches[None] = cls
+
+
+[
```
martindurant, line 797: This is not really a list, or at least not one we keep - so how about a simple
```
for c in [ ... ]:
    add_cache(c)
```

martindurant (commented):
Useful, thanks.

fsspec/caching.py
yarikoptic:
```
❯ git grep overwrite=
fsspec/implementations/dbfs.py:    def _create_handle(self, path, overwrite=True):

❯ git grep clobber=
docs/source/changelog.rst:- Set default clobber=True on impl register
fsspec/registry.py:def register_implementation(name, cls, clobber=True, errtxt=None):
```

so I will for consistency
- name it `register_cache` 
- change to `clobber=True` for consistency, but I think it would be better to default to `False` in both cases. WDYT?

yarikoptic: pushed meanwhile all above and otherwise recommended changes

martindurant: For the DBFS, overwrite I think is the right choice. For register_implementation, I could be persuaded, but perhaps it shouldn't happen in this PR.

fsspec/caching.py
```diff
@@ -767,14 +767,39 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
-    "readahead": ReadAheadCache,
-    "block": BlockCache,
-    "first": FirstChunkCache,
-    "all": AllBytes,
-    "parts": KnownPartsOfAFile,
-    "background": BackgroundBlockCache,
 }
+
+
+def register_cache(cls, clobber=True):
```
martindurant, line 775: Ah, if you meant this new value of clobber's default value then yes, I agree that False is better

yarikoptic: ok, dropped the last commit, no clobber by default here.  Indeed I think it would be better to not clobber on implementations as well, separate PR it is - let's see what that would break ;)

fsspec/caching.py
```diff
@@ -767,14 +767,39 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
-    "background": BackgroundBlockCache,
 }
+
+
+def register_cache(cls, clobber=False):
+    """'Register' cache implementation.
+
+    Parameters
+    ----------
+    overload: bool, optional
```
martindurant, line 780:
```suggestion
    clobber: bool, optional
```

fsspec/caching.py
```diff
@@ -767,14 +767,39 @@ def _read_cache(self, start, end, start_block_number, end_block_number):
...
+        If overload is set to True (default is False) - allow to overload existing
```
martindurant, line 781:
```suggestion
        If set to True (default is False) - allow to overwrite existing
```

PR #1235, opened 2023-04-03
Establish tool.pytest.ini_options in pyproject.toml with pytest markers

yarikoptic:
Otherwise running tests for me resulted in the flood on the screen with bunch of warnings like

	❯ python -m pytest -s -v -k test_register_cache
	======================================================================================================================= test session starts =======================================================================================================================
	platform linux -- Python 3.11.2, pytest-7.2.2, pluggy-1.0.0 -- /home/yoh/venvs/dev3.11/bin/python
	cachedir: .pytest_cache
	rootdir: /home/yoh/proj/misc/filesystem_spec, configfile: pyproject.toml
	plugins: xdist-3.2.1, cov-4.0.0, vcr-1.0.2
	collected 1109 items / 1108 deselected / 3 skipped / 1 selected

	fsspec/tests/test_caches.py::test_register_cache PASSED

	======================================================================================================================== warnings summary =========================================================================================================================
	fsspec/implementations/tests/test_dirfs.py:99
	  /home/yoh/proj/misc/filesystem_spec/fsspec/implementations/tests/test_dirfs.py:99: PytestUnknownMarkWarning: Unknown pytest.mark.asyncio - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
		@pytest.mark.asyncio
	...  many more ...

and AFAIK there is no easy way in pytest to define markers in  code (@jwodder could correct me if I am wrong).

Related unanswered question: https://github.com/pytest-dev/pytest/discussions/9913

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,3 +21,10 @@ exclude = '''
   | versioneer.py
 )
 '''
+
+[tool.pytest.ini_options]
+# custom markers, need to be defined to avoid warnings
+markers = [
+    "asyncio",
+    "vcr",
+]


martindurant:
I suppose there isn't a simple way to show this works in CI, where we don't get the warnings anyway, but I take your word for it and nothing broke :)

yarikoptic:
> I suppose there isn't a simple way to show this works in CI,

well, you can configure CI to error out on warnings.  I think it is worthwhile, we do that in dandi-cli and then whitelist the ones we consider benign or already reported etc.  See  https://github.com/dandi/dandi-cli/blob/master/tox.ini#L51 .  I will send a quick PR to test that...

martindurant:
I think there are too many potential warnings from the various backends, I wouldn't bother with this; in any case, we do NOT expect these warnings to trigger in CI, since we know that the extra packages are installed.

yarikoptic:
> I think there are too many potential warnings from the various backends, I wouldn't bother with this

FWIW -- once again, IMHO it is very handy generally to "triage" any new warning to be proactive and not reactive. E.g. to ensure that your code is ready for new code which otherwise issues DeprecationWarnings, or that some new issues in underlying backends code do not go unnoticed.  Otherwise you end up playing a catch up game which to some degree is more "intense" since requires rapid releases etc. FWIW sent https://github.com/fsspec/filesystem_spec/pull/1238

PR #1236, opened 2023-04-04
Writable lazy references dict

martindurant:
@agoodm : this is an attempt o start in principle one way to tackle combining references to parquet. The idea is, that kerchunk.MultiarrToZarr can have a LazyRefernceMapper instead of a dict for its `out` attribute, and that this will reuse some of the key-location logic to fill dicts. When a dict is full, it gets written to parquet and saved; else the calling code calls flush() when finished to write whatever has been accumulated and not cleared. 

(Perhaps it's possible to have a mapper which references both unchanged original chunks of references and updated sections simultaneously, to allow for "append"-like logic; at this point it feels like the set of parquet files we are pointing to could themselves live in a reference file-system...)

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -90,6 +90,7 @@ class LazyReferenceMapper(collections.abc.MutableMapping):
         self,
         root,
         fs=None,
+        out_root=None,
         cache_size=128,
     ):
         """
@@ -113,11 +114,8 @@ class LazyReferenceMapper(collections.abc.MutableMapping):
         met = json.loads(self._items[".zmetadata"])
         self.record_size = met["record_size"]
         self.zmetadata = met["metadata"]
-        if self.root:
-            self.url = self.root + "/{field}/refs.{record}.parq"
-        else:
-            self.url = "{field}/refs.{record}.parq"
-        self.writecounts = collections.defaultdict(0)
+        self.url = self.root + "/{field}/refs.{record}.parq"
+        self.out_root = out_root or self.root
 
         # Define function to open and decompress refs
         @lru_cache(maxsize=cache_size)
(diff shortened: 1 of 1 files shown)


martindurant:
I added a few docstrings and streamlining around file listing for myself to better understand how the methods of the lazy mapper were used.

fsspec/implementations/reference.py
```diff
@@ -213,27 +295,63 @@ def items(self):
...
-            self._items[key] = None
+            if field and not is_meta:
+                record, _, _ = self._key_to_record(key)
+                subdict = self._items.setdefault((field, record), {})
+                subdict[chunk] = None
+                if len(subdict) == self.record_size:
+                    self.write(field, record)
+            else:
+                # metadata or top-level
+                self._items[key] = None
```
agoodm, line 329: Should consolidate the logic here.

fsspec/implementations/reference.py
```diff
@@ -213,27 +295,63 @@ def items(self):
         return RefsItemsView(self)
 
     def __getitem__(self, key):
-        if key in self._items:
-            val = self._items[key]
-            if val is None:
-                raise KeyError
-        if key in self.zmetadata:
-            # spec requires bytes even if we already decoded the metadata
-            return json.dumps(self.zmetadata[key]).encode()
```
agoodm, line 222: Why was this portion deleted? I presume there will be logic added back later.

agoodm (commented):
Thanks, I have left a few suggestions and questions. I may have a bit more to add later once the portion for writing out the parquet files is incorporated.

fsspec/implementations/reference.py
martindurant: The same logic is in _load_one_key

fsspec/implementations/reference.py
```diff
@@ -213,27 +295,63 @@ def items(self):
...
+                subdict = self._items.setdefault((field, record), {})
+                subdict[chunk] = None
+                if len(subdict) == self.record_size:
+                    self.write(field, record)
+            else:
+                # metadata or top-level
+                self._items[key] = None
+
+    def write(self, field, record, base_url=None, storage_options=None):
+        # TODO: if the dict is incomplete, also load records and merge in
```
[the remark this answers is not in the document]
martindurant: Ah, indeed - if we already have an fs, we don't need storage options at all. I do want to leave open, that we might ead from one place and write to another; but MultiZarrToZarr doesn't actually need that, since it only writes.

martindurant:
> once the portion for writing out the parquet files is incorporated.

This is exactly what the code in the inner loop of kerchunk.df.refs_to_dataframe does, but in the code here we are iterating with a completely different pattern.

agoodm:
> > once the portion for writing out the parquet files is incorporated.
> 
> This is exactly what the code in the inner loop of kerchunk.df.refs_to_dataframe does, but in the code here we are iterating with a completely different pattern.

Just to make sure my current understanding is correct, the main difference is that rather than where `refs_to_dataframe` currently iterates the outer loop by variable and then processes all references, the outer loop is instead by each dataset to be combined (and possibly filesystem if the combined datasets are on different filesystems)?, eg for two datasets with:

```
# ds1
foo/1
bar/1

# ds2
foo/1
bar/1
```

Would combine in the order

```
foo/1
bar/1
foo/2
bar/2
```

instead of

```
foo/1
foo/2
bar/1
bar/2
```

martindurant:
Yes, that's right - and so in the original version, you would be sure to fill up one batch of references before moving onto the next one, or getting to the end of the refs for that variable.

martindurant:
@agoodm , I wrote code that is likely similar to what might be final. This is completely untested, just posting so you can see the state of things.

agoodm:
Apologies for responding a little late to this, been a bit swamped with other stuff at the moment. I'll try to take a closer look and test these changes in a couple of days.

martindurant:
Merging now to iterate later, so it can be in the release

PR #1237, opened 2023-04-04
Disable clobbering of implementations by default

yarikoptic:
I think nothing good can come out from clobbing by default. The necessity to clobber=True  in process_entries  was not well understood by me, but seems to stem from some artificial situation causing re-import of the fsspec

Follow up to #1227 harmonizing default behavior to have `clobber=False` in both similar functionalities.

### DIFF
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -51,7 +51,13 @@ def process_entries():
             for spec in specs:
                 err_msg = f"Unable to load filesystem from {spec}"
                 register_implementation(
-                    spec.name, spec.value.replace(":", "."), errtxt=err_msg
+                    spec.name,
+                    spec.value.replace(":", "."),
+                    errtxt=err_msg,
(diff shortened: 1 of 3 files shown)


martindurant:
> seems to stem from some artificial situation causing re-import of the fsspec

Do you have a reference for this?

martindurant:
The `clear_registry` fixture should deal with that. So this test would fail if we blanket make clobber False for registering? To be clear, this is not a reimport, but evaluation of an EntryPoint (which *can* cause import, but doesn't need to). 

In short, I think everything is fine, unless you have other concerns.

martindurant:
This should NOT happen. It seems that register is being called twice, perhaps because two instances are being created simultaneously. I bet if you run `fun()` once before submitting to the thread pool, it works OK.

martindurant:
Maybe we should guard against this anyway, since doing `open` simultaneously in threads is a reasonable workflow, especially for dask.

PR #1238, opened 2023-04-04
Announce all warnings to be a warning during tests

yarikoptic:
A blind change, untested to work etc - let's see what CI says.

ref: https://github.com/fsspec/filesystem_spec/pull/1235#issuecomment-1496418779

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,3 +28,6 @@ markers = [
     "asyncio",
     "vcr",
 ]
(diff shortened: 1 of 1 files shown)


martindurant:
Yeah, but look at how many there are! We are more likely to write a few tests where we manually in the test make sure that specific warnings are not raised. There are so many legitimate warning scenarios.

PR #1242, opened 2023-04-06
Pass `ls` kwargs through `expand_path`

mariosasko:
To allow passing custom kwargs to `ls` when calling `expand_path`.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -995,9 +995,12 @@ class AbstractFileSystem(metaclass=_Cached):
                 if on_error == "raise":
                     raise
 
-    def expand_path(self, path, recursive=False, maxdepth=None):
+    def expand_path(self, path, recursive=False, maxdepth=None, **kwargs):
         """Turn one or more globs or directories into a list of all matching paths
-        to files or directories."""
+        to files or directories.
+
+        kwargs are passed to ``ls``.
+        """
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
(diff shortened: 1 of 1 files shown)


fsspec/spec.py
```diff
@@ -1013,13 +1016,18 @@ def expand_path(self, path, recursive=False, maxdepth=None):
                     if recursive:
                         out |= set(
                             self.expand_path(
-                                list(bit), recursive=recursive, maxdepth=maxdepth
+                                list(bit),
+                                recursive=recursive,
+                                maxdepth=maxdepth,
+                                **kwargs,
```
martindurant, line 1022: What about `glob`, a few lines above?

martindurant (commented):
Actually we pass the kwargs to find(), right? Only in the default implementation is this using ls, some backends have more specific code.

mariosasko:
Yes, `glob` passes the `ls` kwargs to `find`.

fsspec/spec.py
mariosasko: Yes, these kwargs should also be passed to `glob`. It's fixed now!

fsspec/spec.py
```diff
@@ -995,9 +995,12 @@ def copy(self, path1, path2, recursive=False, on_error=None, **kwargs):
                 if on_error == "raise":
                     raise
 
-    def expand_path(self, path, recursive=False, maxdepth=None):
+    def expand_path(self, path, recursive=False, maxdepth=None, **kwargs):
         """Turn one or more globs or directories into a list of all matching paths
-        to files or directories."""
+        to files or directories.
+
+        kwargs are passed to ``ls``.
```
martindurant, line 1002:
```suggestion
        kwargs are passed to ``glob`` or ``find``.
```

Issue #1234, opened 2023-04-03
Copying a list of files from different directories gives (possibly) wrong result


ianthomas23:
Another discrepancy that has arisen from https://github.com/fsspec/filesystem_spec/pull/1226 and related work.

```python
cp(["source/file1", "source/subdir/subfile2"], "target/")
```
gives `target/file1` and `target/subdir/subfile2`, i.e. it keeps the relative directory structure of the source files.

The command-line equivalent gives
```
$ cp source/file1 source/subdir/subfile2 target/
$ tree target
target
├── file1
└── subfile2
```
so that source files are all copied into the `target` directory.

For consistency I lean towards doing the same as command-line copy, but I suspect that this was a specific design decision and a substantial amount of work has gone into making the behaviour as it currently is.

Issue #1233, opened 2023-04-03
Copying a glob to a directory creates extra directory level


ianthomas23:
Another discrepancy that has arisen from #1226 and related work.

```python
cp("source/somedir/*", "target/")
```

creates a new directory `target/somedir` and places the files in that. According to command-line copy, the `somedir` should not be created and the files should be copied directly into `target`.

martindurant:
This should be identical to the case that the source is a list of files.

Issue #1232, opened 2023-04-03
Non-recursive directory copy incorrect


ianthomas23:
A non-recursive directory copy, which is a slightly silly thing to attempt but is supported, gives the wrong result.

```python
cp("source/somedir/", "target/", recursive=False)
```

creates an empty directory `target/somedir`. But command-line `cp`, which is what we are basing our behaviour on, does nothing at all here.

martindurant:
It's not _very_ wrong :)  In fact, with the trailing "/", it does look like "put the source item INTO the target", surprised that CLI cp is a no-op.

ianthomas23:
Command-line copy gives you
```
$ cp source/subdir/ target/
cp: source/subdir/ is a directory (not copied).
```
regardless of trailing slashes on source or target.

I agree that it is not very wrong, ours feels like better behaviour really. But I would very much like to be able to say "we follow command-line copy behaviour" and not have a single exception to this.

Issue #1231, opened 2023-04-03
cp/get/put ignore maxdepth kwarg


ianthomas23:
The `maxdepth` kwarg is ignored in `cp`, `get` and `put`. It is needed when recursively copying directories e.g.

```python
cp("somedir/", "target/", recursive=True, maxdepth=2)
```

`maxdepth` is already handled by `expand_path` which is the function that is called by `cp`, `get` and `put`, so the fix may be as simple as passing the `maxdepth` kwarg through to `expand_path`.

Note that `maxdepth` is correctly handled in `walk` and `find`.

PR #1247, opened 2023-04-17
Docs and tests for ~10 most common cp/get/put use cases

ianthomas23:
This is a continuation of #1226, adding documentation and tests for the 10 most common use cases for `copy`, `get` and `put`. Tests that should succeed but do not are commented out, I have opened issues for these (#1231, #1232, #1233, #1234).

I have also modified some of the existing documentation where a trailing slash is optional for a target directory to encourage users to include it.

I would like to merge this before fixing any of the issues as then the new tests will be available and we can be sure that the individual fixes do not break any of the other tests.

### DIFF
--- a/docs/source/copying.rst
+++ b/docs/source/copying.rst
@@ -121,8 +121,11 @@ Forward slashes are used for directory separators throughout.
 
     .. warning::
 
-       ``recursive=False`` is not correct.
-       ``maxdepth`` is not yet implemented for copying functions.
+       ``recursive=False`` is not correct
+       (`issue 1232 <https://github.com/fsspec/filesystem_spec/issues/1232>`_).
+
+       ``maxdepth`` is not yet implemented for copying functions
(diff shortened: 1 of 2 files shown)


ianthomas23:
New built docs are at https://filesystem-spec--1247.org.readthedocs.build/en/1247/copying.html

docs/source/copying.rst
```diff
@@ -115,5 +117,207 @@ Forward slashes are used for directory separators throughout.
...
+        ├── 📄 file2
+        └── 📄 subfile1
+
+    All of the files are copied to the target directory regardless of their relative paths in the
+    source filesystem. The trailing slash on the ``target`` is optional but recommended as it
+    explicitly indicates that it is a directory.
+
+    .. warning::
+
+        This is not correct currently.
```
martindurant, line 299: Since you made issues, we could and a link to these, so that users can see we are working on it, or exctly in what manner they might be "wrong"

ianthomas23: Yes, good idea. Done!

PR #1249, opened 2023-04-20
Clobber default change - follow up

martindurant:
See #1237


wradstok:
Hi, would it be possible to create a new release on pypi with this fix? We are intermittently running into the issue that this PR fixes (register abfss multiple times) and this would save us the workaround 😄

PR #1253, opened 2023-05-03
Fix del self.zip giving AttributeError

mauvilsa:
The deletion of the self.zip should only be performed if that attribute exists. Without this there are errors for example like the following:

```
    File "/home/runner/work/jsonargparse/jsonargparse/.tox/py-all-extras/lib/python3.8/site-packages/fsspec/implementations/zip.py", line 76, in __del__
      del self.zip
  AttributeError: zip
```

Extracted from https://github.com/omni-us/jsonargparse/actions/runs/4868536023/jobs/8682110610


martindurant:
I like short fixes :)

PR #1254, opened 2023-05-03
Do nothing for non-recursive copy of empty directory

### LINKED ISSUE #1232: Non-recursive directory copy incorrect

ianthomas23:
Fixes #1232.

This is for a non-recursive copy of an empty directory such as
```python
cp("source/somedir/", "target/", recursive=False)
```
which is now a no-op to agree with command-line copy.

There were two possible approaches. The one I've opted for is to leave `expand_path()` as it is, returning the empty directory, and checking this separately in `cp`, `get` and `put` to identify the "do nothing" scenario and return early.

The other approach would have been to add a new kwarg to `expand_path()` to override the current `withdirs=True`. This would allow an empty return from the function which would have needed extra handling so I considered this a worse solution than the one I have adopted. I am writing this here because I suspect in the long run this solution may be required to support corner cases of `cp` passing a list of individual files and empty directories, which probably nobody is doing but must eventually be supported. Very low priority though.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -281,7 +281,7 @@ def trailing_sep(path):
     # TODO: if all incoming paths were posix-compliant then separator would
     # always be a forward slash, simplifying this function.
     # See https://github.com/fsspec/filesystem_spec/pull/1250
-    return path.endswith(os.sep) or os.altsep and path.endswith(os.altsep)
+    return path.endswith(os.sep) or (os.altsep is not None and path.endswith(os.altsep))
 
 
 def trailing_sep_maybe_asterisk(path):
@@ -294,10 +294,8 @@ def trailing_sep_maybe_asterisk(path):
     # TODO: if all incoming paths were posix-compliant then separator would
     # always be a forward slash, simplifying this function.
     # See https://github.com/fsspec/filesystem_spec/pull/1250
-    return (
-        path.endswith((os.sep, os.sep + "*"))
-        or os.altsep
-        and path.endswith((os.altsep, os.altsep + "*"))
+    return path.endswith((os.sep, os.sep + "*")) or (
+        os.altsep is not None and path.endswith((os.altsep, os.altsep + "*"))
     )
 
 
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -939,32 +939,24 @@ def test_cp_get_put_empty_directory(tmpdir, funcname):
     # cp/get/put without slash, target directory exists
     assert fs.isdir(target)
     func(empty, target)
-    assert fs.find(target, withdirs=True) == [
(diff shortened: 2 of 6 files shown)


fsspec/implementations/local.py
```diff
@@ -281,7 +281,7 @@ def trailing_sep(path):
     # TODO: if all incoming paths were posix-compliant then separator would
     # always be a forward slash, simplifying this function.
     # See https://github.com/fsspec/filesystem_spec/pull/1250
-    return path.endswith(os.sep) or os.altsep and path.endswith(os.altsep)
+    return path.endswith(os.sep) or (os.altsep is not None and path.endswith(os.altsep))
```
ianthomas23, line 284: This wasn't quite correct before but now is.

fsspec/implementations/tests/test_local.py
```diff
@@ -939,32 +939,24 @@ def test_cp_get_put_empty_directory(tmpdir, funcname):
     # cp/get/put without slash, target directory exists
     assert fs.isdir(target)
     func(empty, target)
-    assert fs.find(target, withdirs=True) == [
-        make_path_posix(os.path.join(target, "empty"))
-    ]
-
-    fs.rm(target + "/empty", recursive=True)
+    assert fs.find(target, withdirs=True) == []
```
ianthomas23, line 942: This test is effectively duplicated in the abstract test harness, so it can eventually be removed. But I have fixed it here for the time being.

fsspec/spec.py
```diff
@@ -886,13 +886,21 @@ def get(self, rpath, lpath, recursive=False, callback=_DEFAULT_CALLBACK, **kwarg
...
+        rpaths = self.expand_path(rpath, recursive=recursive)
+        if (
+            len(rpaths) == 1
+            and not recursive
+            and (trailing_sep(rpaths[0]) or self.isdir(rpaths[0]))
+        ):
+            # Non-recursive copy of directory does nothing.
+            return
+        source_is_str = isinstance(rpath, str)
+
```
ianthomas23, line 898: I have reordered the code here (and the other 2 functions below) to separate out the handling of `source` and `target` paths, with a line of whitespace between, for increased clarity.

ianthomas23:
`s3fs-pytest` test failure is the same one we are seeing in the `s3fs` repo itself (e.g. https://github.com/fsspec/s3fs/actions/runs/4852093306/jobs/8650821223?pr=731).

martindurant:
>  test failure is the same one we are seeing in the s3fs repo

OK.

martindurant:
Thorough code, thank you.

PR #1255, opened 2023-05-04
Fix copying of single glob to directory

### LINKED ISSUE #1233: Copying a glob to a directory creates extra directory level

ianthomas23:
Fixes #1233.

The headline problem here of not creating an extra directory was already fixed by #1250. The other problem of not copying empty directories that the glob identifies is a generalisation of #1254.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -886,15 +886,13 @@ class AbstractFileSystem(metaclass=_Cached):
             trailing_sep_maybe_asterisk,
         )
 
-        rpaths = self.expand_path(rpath, recursive=recursive)
-        if (
-            len(rpaths) == 1
-            and not recursive
-            and (trailing_sep(rpaths[0]) or self.isdir(rpaths[0]))
-        ):
-            # Non-recursive copy of directory does nothing.
-            return
         source_is_str = isinstance(rpath, str)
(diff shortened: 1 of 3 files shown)


fsspec/spec.py
```diff
@@ -886,15 +886,13 @@ def get(self, rpath, lpath, recursive=False, callback=_DEFAULT_CALLBACK, **kwarg
...
-            and not recursive
-            and (trailing_sep(rpaths[0]) or self.isdir(rpaths[0]))
-        ):
-            # Non-recursive copy of directory does nothing.
-            return
         source_is_str = isinstance(rpath, str)
+        rpaths = self.expand_path(rpath, recursive=recursive)
+        if source_is_str and not recursive:
+            # Non-recursive glob does not copy directories
+            rpaths = [p for p in rpaths if not (trailing_sep(p) or self.isdir(p))]
```
ianthomas23, line 893: Alternative use of `filter` with a `lambda` here is slightly more code so I have opted for the easier to understand list comprehension.

Issue #1205, opened 2023-03-06
`AbstractArchiveFileSystem.ls` has `detail=True` by default


lhoestq:
This was introduced at https://github.com/fsspec/filesystem_spec/commit/0f3ecd8e629043646ab19b1a2b00d895f0553a81 and is inconsistent with the rest of the filesystems which have `detail=False`. I guess this bug was introduced inadvertently ?

cc @martindurant

martindurant:
Actually, AbstractFileSystem (the base class) has detail=True, so it was an effort to be consistent with that. Other filesystems should migrate to detail=True.

PR #1259, opened 2023-05-10
Implement maxdepth for recursive cp/get/put

### LINKED ISSUE #1231: cp/get/put ignore maxdepth kwarg

ianthomas23:
Fixes #1231.

I have explicitly added the `maxdepth` as an argument to `cp`, `get` and `put` as it is already explicit in `expand_path`, `find` and `glob`. But the alternative of leaving it within the `kwargs` and using `get` to read it from there would be fine too.

### DIFF
--- a/fsspec/tests/abstract/copy.py
+++ b/fsspec/tests/abstract/copy.py
@@ -102,15 +102,9 @@ class AbstractCopyTests:
                 assert fs.isfile(fs_join(target, "subfile2"))
                 assert fs.isdir(fs_join(target, "nesteddir"))
                 assert fs.isfile(fs_join(target, "nesteddir", "nestedfile"))
+                assert not fs.exists(fs_join(target, "subdir"))
 
-                fs.rm(
-                    [
-                        fs_join(target, "subfile1"),
-                        fs_join(target, "subfile2"),
-                        fs_join(target, "nesteddir"),
-                    ],
-                    recursive=True,
-                )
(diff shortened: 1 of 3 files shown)


fsspec/tests/abstract/copy.py
```diff
@@ -102,15 +102,9 @@ def test_copy_directory_to_existing_directory(
...
 
-                fs.rm(
-                    [
-                        fs_join(target, "subfile1"),
-                        fs_join(target, "subfile2"),
-                        fs_join(target, "nesteddir"),
-                    ],
-                    recursive=True,
-                )
+                fs.rm(fs.ls(target, detail=False), recursive=True)
```
ianthomas23, line 107: Care needed using `ls` as some implementations have default `detail` of `True` and some `False`. cc https://github.com/fsspec/filesystem_spec/issues/1205

martindurant: being explicit is always good

martindurant:
I like including the kwarg, it makes top-level sense for these operations.

PR #1260, opened 2023-05-10
Update versions of github actions

ianthomas23:
We are seeing warnings in CI such as:
```
[3.10-pytest](https://github.com/fsspec/filesystem_spec/actions/runs/4938014254/jobs/8827540326)
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```
These are fixed by upgrading `actions/checkout@v2` to `actions/checkout@v3`. I've also updated other actions to their latest version at the same time.

If this works, need to do the same in `s3fs` and so on.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -64,9 +64,9 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
-      - uses: pre-commit/action@v3.0.0
+      - uses: actions/checkout@main
+      - uses: actions/setup-python@main
+      - uses: pre-commit/action@main
 
   downstream:
     name: downstream
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


martindurant:
Is there some `@main` variant to just use the latest?

ianthomas23:
Oh, I didn't know that was possible. I'll take a look tomorrow.

martindurant:
I'm not sure if all action repos support it

ianthomas23:
I've tried using `@main` for all 3 actions that I changed for just the `lint` job, and they work fine. From logs:
```
2023-05-11T09:35:12.2359001Z Download action repository 'actions/checkout@main' (SHA:f095bcc56b7c2baf48f3ac70d6d6782f4f553222)
2023-05-11T09:35:12.8934537Z Download action repository 'actions/setup-python@main' (SHA:db9987b4c1f10f0404fa60ee629f675fafbd6763)
2023-05-11T09:35:13.4757509Z Download action repository 'pre-commit/action@main' (SHA:5f528da5c95691c4cf42ff76a4d10854b62cbb82)
```
So for these 3 actions it is using `HEAD` of the `main` branch.

I am not sure that this is a good idea compared to pinning to a particular version of each action, as we'd just be taking whatever state the `main` branch is in rather than a coherent set of changes that would comprise a tagged release. But I am not sure :shrug:.

martindurant:
I am OK with main until it becomes a problem.

PR #1264, opened 2023-05-12
Add boxfs to registry

tehunter:
Add [`boxfs`](https://github.com/IBM/boxfs) to registry, allowing users to connect to file storage hosted by box.com


martindurant:
Thank you

PR #1267, opened 2023-05-16
Complete sync and async copy/get/put tests

ianthomas23:
This adds a full set of `AbstractGetTests` and `AbstractPutTests` that are equivalent to the existing `AbstractCopyTests`, and ports the recent changes from synchronous copy/get/put code (#1250, #1254, #1255, #1259) to async. This completes the top 10 most common cp/get/put use cases documented by #713.

The abstract test suite all passes locally for me using `local` and `memory` filesystems (the latter skips 4 tests, see below) and also `s3fs` (using the branch that is fsspec/s3fs#713).

Noteworthy implementation details:

- Filesystem scope for tests is per test class, so there is one setup-and-teardown combination for each of the `AbstractCopyTests`, `AbstractGetTests` and `AbstractPutTests`. This doesn't make any real difference for `local` and `memory` tests, but makes the `s3fs` tests much faster than individual test scope. The scope could be made once per test session, but that would involve a bit of a refactor so is not planned yet.
- Setting up the `source` and `target` directories for these tests uses fixtures that automatically clean up after themselves at the end of each test, ensuring that the filesystem is empty ready for the next test.

Work to do in future PRs soon:

- Four of the `memory` `get` tests are skipped as currently there is a bespoke code path that does not auto create a local directory when required. Needs a fix to `MemoryFileSystem`.
- I haven't fully dealt with all combinations of trailing slashes on `source` and `target` directories yet.
- `gcsfs` doesn't pass all the new tests yet.

I propose the following order of operations:

1. Review and merge this when CI happy.
2. I'll update fsspec/s3fs#713 to check that `s3fs` is happy with the new changes, then review and merge of that PR.
3. A new release of `fsspec` is advisable then, as one of my recent PRs broke `async` `get` in some situations and this is fixed in this PR.
4. I'll fix the above 3 "future PRs" issues.

Longer term there are other improvements that are not urgent:

1. Cover some of the less-common `cp/get/put` use cases.
2. Test more of implementations that ship in `fsspec`.
3. Document how to use the abstract test harness in downstream projects.
5. Move tests from original test scripts to the new abstract test classes.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -13,6 +13,12 @@ from typing import Iterable
 
 from .callbacks import _DEFAULT_CALLBACK
 from .exceptions import FSTimeoutError
+from .implementations.local import (
+    LocalFileSystem,
+    make_path_posix,
+    trailing_sep,
+    trailing_sep_maybe_asterisk,
+)
 from .spec import AbstractBufferedFile, AbstractFileSystem
 from .utils import is_exception, other_paths
 
@@ -331,8 +337,6 @@ class AsyncFileSystem(AbstractFileSystem):
         batch_size=None,
         **kwargs,
     ):
-        from .implementations.local import trailing_sep, trailing_sep_maybe_asterisk
-
         if on_error is None and recursive:
             on_error = "ignore"
         elif on_error is None:
@@ -492,13 +496,6 @@ class AsyncFileSystem(AbstractFileSystem):
         constructor, or for all instances by setting the "gather_batch_size" key
         in ``fsspec.config.conf``, falling back to 1/8th of the system limit .
         """
-        from .implementations.local import (
-            LocalFileSystem,
-            make_path_posix,
-            trailing_sep,
-            trailing_sep_maybe_asterisk,
-        )
-
         source_is_str = isinstance(lpath, str)
         if source_is_str:
             lpath = make_path_posix(lpath)
@@ -565,13 +562,6 @@ class AsyncFileSystem(AbstractFileSystem):
         constructor, or for all instances by setting the "gather_batch_size" key
         in ``fsspec.config.conf``, falling back to 1/8th of the system limit .
         """
-        from .implementations.local import (
-            LocalFileSystem,
-            make_path_posix,
-            trailing_sep,
-            trailing_sep_maybe_asterisk,
-        )
-
         source_is_str = isinstance(rpath, str)
         # First check for rpath trailing slash as _strip_protocol removes it.
         source_not_trailing_sep = source_is_str and not trailing_sep_maybe_asterisk(
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -331,20 +331,30 @@ class AsyncFileSystem(AbstractFileSystem):
         batch_size=None,
         **kwargs,
     ):
+        from .implementations.local import trailing_sep, trailing_sep_maybe_asterisk
+
         if on_error is None and recursive:
             on_error = "ignore"
         elif on_error is None:
             on_error = "raise"
 
+        source_is_str = isinstance(path1, str)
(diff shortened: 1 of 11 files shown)


fsspec/implementations/tests/test_memory.py
```diff
@@ -23,19 +23,6 @@ def test_strip(m):
     assert m._strip_protocol("/b/c/") == "/b/c"
 
 
-def test_put_single(m, tmpdir):
```
ianthomas23, line 26: This test is no longer correct, but rather than change the test in line with current functionality I have deleted it as it is equivalent to the "copy file to directory" test.

fsspec/asyn.py
```diff
@@ -535,21 +565,38 @@ async def _get(
         constructor, or for all instances by setting the "gather_batch_size" key
         in ``fsspec.config.conf``, falling back to 1/8th of the system limit .
         """
-        from fsspec.implementations.local import LocalFileSystem, make_path_posix
+        from .implementations.local import (
```
martindurant, line 568: Do we get circular imports if we move all these from the methods to the top of the module?

fsspec/tests/abstract/__init__.py
```diff
@@ -20,55 +19,97 @@ def fs_join():
...
-        subdir = fs_join(source, "subdir")
-        nesteddir = fs_join(subdir, "nesteddir")
-        fs.makedirs(nesteddir)
-        fs.touch(fs_join(source, "file1"))
-        fs.touch(fs_join(source, "file2"))
-        fs.touch(fs_join(subdir, "subfile1"))
-        fs.touch(fs_join(subdir, "subfile2"))
-        fs.touch(fs_join(nesteddir, "nestedfile"))
-        return source
+        source = self._scenario_cp(fs, fs_join, fs_path)
```
martindurant, line 29: Little surprised here, because it looks like we need cp to be working correctly before we get to running the tests. Wasn't touch() more explicit?

martindurant: Looking further, I see you just moved the touch() calls. But isn't _scenario_cp misnamed, then? In general, for any of these that a downstream developer is expected to implement (if any), it would be worthwhile describing all the arguments of all the methods better. Perhaps general documentation in https://filesystem-spec.readthedocs.io/en/latest/developer.html and/or https://filesystem-spec.readthedocs.io/en/latest/copying.html would help.

martindurant (commented):
Basically all good, just a couple of questions for you (one of which escaped as a standalone comment)

fsspec/asyn.py
ianthomas23: Moving the imports to the top works fine locally, I've pushed a commit to check in CI.

fsspec/tests/abstract/__init__.py
martindurant: how about bulk_operations_scenario, perhaps suffixed with 0 if there might eventually be more of them?

ianthomas23:
I think `s3fs` test failure is unrelated as the other recent PR has the same error. CI passed OK 5 days ago. Differences in environment between passing and failing:

| package | passed | failing |
| --- | --- | --- |
| c-ares | 1.19.0 | 1.19.1 |
| coverage | 7.2.5 | 7.2.6 |
| grpcio | 1.54.2 | 1.55.0 |
| libgrpc | 1.54.2 | 1.55.0 |
| libprotobuf | 3.21.12 | 4.23.1 |
| moto | 4.1.9 | 4.1.10 |
| protobuf | 4.21.12 | 4.23.1 |
| requests | 2.29.0 | 2.31.0 |   
| typing-extensions | 4.5.0 | 4.6.0 | 
| websocket-client | 1.5.1 | 1.5.2 |

martindurant:
Likely moto...

Issue #625, opened 2021-04-29
Type annotations?


evinism:
Are you interested in type annotations? E.g. if I added a bunch of type annotations to the project, is that something that you'd see being easily merged?

martindurant:
Personally, I wouldn't say that I am "interested", but I am also not opposed. Call that ambivalent maybe.
I suspect it may be a great deal of work, and I am uncertain if, in the long run, it would entail more or less work for further development of the library. I am prepared to be convinced, and perhaps more voices should speak up here.

martindurant:
cc @ianthomas23

ianthomas23:
Thanks @aucampia, you present a strong case and a potential roadmap for starting this work. It is also evident that many people would like type annotations. No-one should be dissuaded by any lack of interest shown here, if a PR was submitted starting this work with some commitment to continue it, I would expect it to be merged without too much hassle. Or to put it another way, given that the python ecosystem is going in this direction I think it is inevitable that we will have to adopt type annotations eventually, it is just a question of when.

The problem really is that even if a large number of people request this then it makes no difference, what is needed is someone to actually do the work. It isn't a priority for established contributors, most of whom are inclined to fix problems that are impeding them, or add new functionality that they need. Maintainers with a wider view of the project are mostly working on consistency and reliability across the code base. Infrastructure issues like this often do not get near enough to the top of anybody's "todo" list to actually be actioned. There is a space here for a new contributor with type annotation experience to fill and if that doesn't happen probably an existing maintainer will eventually get around to it but I wouldn't expect it to be soon.

aucampia:
> The problem really is that even if a large number of people request this then it makes no difference, what is needed is someone to actually do the work.

I agree with this and I understand it, it is significantly important to have fsspec be useful and defect free. I hope someone like @evinism finds the time to make this contribution. And hopefully whoever does have the time to do this will take care to do it with incrementally with minimal disruption.

edgarrmondragon:
Would it be alright to get a list of modules, then as a first step configure mypy to ignore all of them and start tackling individually in separate PRs. I'd be happy to help with that as much as possible, but I wonder if that's too big of a burden on the maintainers.

martindurant:
Having a mypy config which ignores all files would be fine with me, if it allows others a more obvious route to adding types to the codebase.

[...]

PR #1268, opened 2023-05-23
Setup a minimal mypy config

alanhdu:
This is a first step to https://github.com/fsspec/filesystem_spec/issues/625: we setups a new CI step with `mypy` and do the minimal amount of work to get it passing.

Note that by default `mypy` does not type-check any function that doesn't already have type annotations in the function signature, so for now this isn't really type-checking any function internals. Still, I think having the CI setup will be a good first step.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -68,6 +68,22 @@ jobs:
       - uses: actions/setup-python@main
       - uses: pre-commit/action@main
 
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup conda
+        uses: mamba-org/provision-with-micromamba@main
+        with:
+          environment-file: ci/environment-typecheck.yml
+
+      - name: mypy
+        shell: bash -l {0}
+        run: |
+          mypy fsspec
+
   downstream:
     name: downstream
     runs-on: ubuntu-latest
--- a/ci/environment-typecheck.yml
+++ b/ci/environment-typecheck.yml
@@ -1,0 +1,13 @@
+name: test_env
+channels:
+  - conda-forge
+dependencies:
+  - mypy=1.3
+  - pyarrow
+  - python=3.8
+  - pip
+  - pip:
+    - types-paramiko
+    - types-requests
+    - types-tqdm
+    - types-ujson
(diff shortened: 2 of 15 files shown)


ianthomas23:
Thanks for starting this @alanhdu.

If you set up `pre-commit` hooks locally you can automatically fix the linting errors which are all about import order.

In the end we'd probably run `mypy` within `pre-commit` but it seems fine initially to run it in a separate CI run as you have done so that the environment can be carefully controlled.

I have some knowledge of `mypy` but I am not an expert in it. I probably know enough to be dangerous but not enough to be useful! So please bear with me if I ask some trivial questions.

It seems to me that as we are starting the type annotations from scratch it is a good time to agree a sensible approach for how we should proceed. I am seeing annotations like
```python
var: Union[str, Dict]
```
but I am more comfortable looking at annotations like
```python
var: str | dict
```
In other words using `|` instead of `Union` and the non-capitalised types for built-in types that do not need to be explicitly imported rather than the capitalised types that do. Is there a particular reason for opting for the former rather than the latter?

alanhdu:
> In other words using | instead of Union and the non-capitalised types for built-in types that do not need to be explicitly imported rather than the capitalised types that do. Is there a particular reason for opting for the former rather than the latter?

It's a good question! I think the answer is probably not.

The issue with the `str | dict` syntax is that it was only introduced in Python 3.10. For Python 3.8 and 3.9, you have to use add a `from __future__ import annotations` statement everywhere, but that doesn't always play nicely with run-time uses of type annotations (things like pydantic or attrs). I haven't followed the full saga (https://docs.python.org/3/library/__future__.html#id1 is probably the right breadcrumb to follow), but I got used to avoiding the `__future__` import at work since we do a lot based on the runtime type hints.

But since fsspec *doesn't* do anything with the type hints at runtime, so AFAIK there's no reason not to use the `__future__` import.

alanhdu:
Ok --- pushed two commits: first commit should fix the isort problems, and the 2nd commit shows that things look like if we use `from __future__ import annotations`.

martindurant:
> and the non-capitalised types for built-in types

I think it was even earlier python versions (3.6? 3.7?) that required this, especially for the case where you wanted to specialise a type like dict.

hansharhoff:
It was 3.10:

https://peps.python.org/pep-0604/

alanhdu:
Rebased onto master.

PR #1280, opened 2023-05-29
bugfix: prevent RuntimeError from catching NotImplementedError

mukhery:
The existing `except RuntimeError` will catch the `NotImplementedError` raised inside the try block to prevent infinite loops. This fix explicitly catches and re-raises the `NotImplementedError` to resolve.

### DIFF
--- a/fsspec/tests/test_async.py
+++ b/fsspec/tests/test_async.py
@@ -53,7 +53,12 @@ class _DummyAsyncKlass:
         await asyncio.sleep(1)
         return True
 
+    async def _bad_multiple_sync(self):
+        fsspec.asyn.sync_wrapper(_DummyAsyncKlass._dummy_async_func)(self)
+        return True
+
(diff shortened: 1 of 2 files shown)


martindurant:
Sorry, totally missed this. Can we possibly hit this in a test?

mukhery:
> Sorry, totally missed this. Can we possibly hit this in a test?

yeah, I'll add one later today or tomorrow

fsspec/tests/test_async.py
```diff
@@ -77,6 +82,12 @@ def test_sync_wrapper_treat_timeout_0_as_none():
     assert test_obj.dummy_func(timeout=0)
 
 
+def test_sync_wrapper_bad_multiple_sync():
+    test_obj = _DummyAsyncKlass()
+    with pytest.raises(NotImplementedError):
+        test_obj.bad_multiple_sync_func(timeout=5)
```
mukhery, line 88: @martindurant test added

PR #1282, opened 2023-05-30
Create parent directories in get_file

ianthomas23:
This is a fix when calling `get` with a file to get it into a new (i.e. doesn't exist yet) directory we need to create the parent directory to put the file in. The issue was discovered as part of the new `cp/get/put` test harness (https://github.com/fsspec/filesystem_spec/pull/1267). With this fix all `memory` filesystem tests pass whereas before this the 4 "get file/directory/glob into a new directory" tests all failed as the parent directory wasn't auto created.

### DIFF
--- a/fsspec/implementations/tests/memory/memory_test.py
+++ b/fsspec/implementations/tests/memory/memory_test.py
@@ -1,5 +1,3 @@
-import pytest
-
 import fsspec.tests.abstract as abstract
 from fsspec.implementations.tests.memory.memory_fixtures import MemoryFixtures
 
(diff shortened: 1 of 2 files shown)


ianthomas23:
Windows test failing as a result of the usual different path separators. I will have to investigate further.

martindurant:
> I will have to investigate further.

We should hold the PR, then? 

Maybe we should replace os.makedirs with LocalFileSystem.makedirs, which should do path normalisation ?

ianthomas23:
Yes, hold the PR until I have looked again.

ianthomas23:
OK, this is good now using `LocalFileSystem.makedirs` rather than `os.makedirs`.

PR #1284, opened 2023-06-01
Add typehints for `fsspec.caching` and `fsspec.utils`

alanhdu:
This add typehints to `fsspec.caching` and `fsspec.utils` (as best as I can understand them), and then turns on `check_untyped_defs` for them as well (so that every function in those files will be type-checked, note just those with type annotations).

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -93,7 +93,7 @@ class MMapCache(BaseCache):
         self.location = location
         self.cache = self._makefile()
 
-    def _makefile(self) -> bytearray | mmap.mmap:
+    def _makefile(self) -> mmap.mmap:
         import mmap
         import tempfile
 
@@ -166,7 +166,7 @@ class ReadAheadCache(BaseCache):
         self.start = 0
         self.end = 0
 
-    def _fetch(self, start, end: int | None) -> bytes:
+    def _fetch(self, start: int | None, end: int | None) -> bytes:
         if start is None:
             start = 0
         if end is None or end > self.size:
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -93,7 +93,7 @@ class MMapCache(BaseCache):
         self.location = location
         self.cache = self._makefile()
 
-    def _makefile(self) -> mmap.mmap:
+    def _makefile(self) -> mmap.mmap | bytearray:
         import mmap
         import tempfile
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -144,7 +144,7 @@ class MMapCache(BaseCache):
         del state["cache"]
         return state
 
-    def __setstate__(self, state) -> None:
+    def __setstate__(self, state: dict[str, Any]) -> None:
         # Restore instance attributes
         self.__dict__.update(state)
         self.cache = self._makefile()
@@ -278,7 +278,7 @@ class BlockCache(BaseCache):
         del state["_fetch_block_cached"]
         return state
 
-    def __setstate__(self, state) -> None:
+    def __setstate__(self, state: dict[str, Any]) -> None:
         self.__dict__.update(state)
         self._fetch_block_cached = functools.lru_cache(state["maxblocks"])(
             self._fetch_block
@@ -471,10 +471,6 @@ class AllBytes(BaseCache):
         self.data = data
 
     def _fetch(self, start: int | None, stop: int | None) -> bytes:
-        if start is None:
-            start = 0
-        if stop is None:
-            stop = len(self.data)
         return self.data[start:stop]
 
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import collections
 import functools
 import logging
@@ -5,9 +7,33 @@ import math
 import os
 import threading
 import warnings
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Generic,
+    NamedTuple,
+    OrderedDict,
+    TypeVar,
+)
+
+if TYPE_CHECKING:
+    import mmap
+
+    from typing_extensions import ParamSpec
+
+    P = ParamSpec("P")
+else:
+    P = TypeVar("P")
+
+T = TypeVar("T")
+
 
 logger = logging.getLogger("fsspec")
+
+Fetcher = Callable[[int, int], bytes]  # Maps (start, end) to bytes
 
 
 class BaseCache:
@@ -26,14 +52,14 @@ class BaseCache:
         How big this file is
     """
 
-    name = "none"
+    name: ClassVar[str] = "none"
 
-    def __init__(self, blocksize, fetcher, size):
+    def __init__(self, blocksize: int, fetcher: Fetcher, size: int) -> None:
         self.blocksize = blocksize
         self.fetcher = fetcher
         self.size = size
 
-    def _fetch(self, start, stop):
+    def _fetch(self, start: int | None, stop: int | None) -> bytes:
         if start is None:
             start = 0
         if stop is None:
@@ -54,13 +80,20 @@ class MMapCache(BaseCache):
 
     name = "mmap"
 
-    def __init__(self, blocksize, fetcher, size, location=None, blocks=None):
+    def __init__(
+        self,
+        blocksize: int,
+        fetcher: Fetcher,
+        size: int,
+        location: str | None = None,
+        blocks: set[int] | None = None,
+    ) -> None:
         super().__init__(blocksize, fetcher, size)
         self.blocks = set() if blocks is None else blocks
         self.location = location
         self.cache = self._makefile()
 
-    def _makefile(self):
+    def _makefile(self) -> bytearray | mmap.mmap:
         import mmap
         import tempfile
 
@@ -82,7 +115,7 @@ class MMapCache(BaseCache):
 
         return mmap.mmap(fd.fileno(), self.size)
 
-    def _fetch(self, start, end):
+    def _fetch(self, start: int | None, end: int | None) -> bytes:
         logger.debug(f"MMap cache fetching {start}-{end}")
         if start is None:
             start = 0
@@ -105,13 +138,13 @@ class MMapCache(BaseCache):
 
         return self.cache[start:end]
 
-    def __getstate__(self):
+    def __getstate__(self) -> dict[str, Any]:
         state = self.__dict__.copy()
         # Remove the unpicklable entries.
         del state["cache"]
         return state
 
-    def __setstate__(self, state):
+    def __setstate__(self, state) -> None:
         # Restore instance attributes
         self.__dict__.update(state)
         self.cache = self._makefile()
@@ -127,13 +160,13 @@ class ReadAheadCache(BaseCache):
 
     name = "readahead"
 
-    def __init__(self, blocksize, fetcher, size):
+    def __init__(self, blocksize: int, fetcher: Fetcher, size: int) -> None:
         super().__init__(blocksize, fetcher, size)
         self.cache = b""
         self.start = 0
         self.end = 0
 
-    def _fetch(self, start, end):
+    def _fetch(self, start, end: int | None) -> bytes:
         if start is None:
             start = 0
         if end is None or end > self.size:
@@ -168,11 +201,11 @@ class FirstChunkCache(BaseCache):
 
     name = "first"
 
-    def __init__(self, blocksize, fetcher, size):
+    def __init__(self, blocksize: int, fetcher: Fetcher, size: int) -> None:
         super().__init__(blocksize, fetcher, size)
-        self.cache = None
+        self.cache: bytes | None = None
 
-    def _fetch(self, start, end):
+    def _fetch(self, start: int | None, end: int | None) -> bytes:
         start = start or 0
         end = end or self.size
         if start < self.blocksize:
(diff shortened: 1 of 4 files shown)


fsspec/utils.py
```diff
@@ -266,11 +295,14 @@ def read_block(f, offset, length, delimiter=None, split_before=False):
         length = end - start
 
     f.seek(offset)
-    b = f.read(length)
+    if length is not None:
+        b = f.read(length)
+    else:
+        b = f.read()
```
alanhdu, line 301: I assume we needed to do this (otherwise `if delimiter is None and length is None` I'm pretty sure this crashes), but happy to revert this.

fsspec/utils.py
```diff
@@ -500,7 +554,7 @@ def merge_offset_ranges(paths, starts, ends, max_gap=0, max_block=None, sort=Tru
     if not isinstance(starts, list):
         starts = [starts] * len(paths)
     if not isinstance(ends, list):
-        ends = [starts] * len(paths)
+        ends = [ends] * len(paths)
```
alanhdu, line 557: I assume this was just a bug? I can't figure out what this could would do otherwise.

fsspec/caching.py
```diff
@@ -406,23 +445,33 @@ def _fetch(self, start, end):
...
+        data: bytes | None = None,
+    ) -> None:
+        super().__init__(blocksize, fetcher, size)  # type: ignore[arg-type]
         if data is None:
             data = self.fetcher(0, self.size)
         self.data = data
 
-    def _fetch(self, start, end):
-        return self.data[start:end]
+    def _fetch(self, start: int | None, stop: int | None) -> bytes:
```
alanhdu, line 473: I added support for `start = stop = None`, but hapyp to revert that and ust add a `# type: ignore` instead (complains that there's a different type signature than the base class).

fsspec/caching.py
```diff
@@ -378,23 +417,27 @@ def _fetch(self, start, end):
...
-                pass
-            elif end - self.end > self.blocksize:
-                self.cache = self.fetcher(start, bend)
-                self.start = start
-            else:
-                new = self.fetcher(self.end, bend)
-                self.cache = self.cache + new
+        else:
+            assert self.start is not None
+            assert self.end is not None
```
alanhdu, line 422: I needed to add this so that `mypy` wouldn't complain about `self.start` or `self.end` being None. Happy to restructure the code if you think there's a better way to do this.

martindurant: *THIS* annoys me about mypy. Asserts have no place outside of tests. The comparisons above would fail for the case of None.

martindurant:
I am waiting on this one hoping that anyone with more types experience than me can have a look. I don't see any glaring problems.

fsspec/caching.py
```diff
@@ -510,33 +576,37 @@ def _fetch(self, start, stop):
         return out + super()._fetch(start, stop)
 
 
-class UpdatableLRU:
+class UpdatableLRU(Generic[P, T]):
```
martindurant, line 579: Can we describe what P and T do here?

fsspec/caching.py
```diff
@@ -510,33 +576,37 @@ def _fetch(self, start, stop):
...
     """
     Custom implementation of LRU cache that allows updating keys
 
     Used by BackgroudBlockCache
     """
 
-    CacheInfo = collections.namedtuple(
-        "CacheInfo", ["hits", "misses", "maxsize", "currsize"]
-    )
+    class CacheInfo(NamedTuple):
```
martindurant, line 586: It feels like this is a different construct, but it turns out to work the same. I didn't know you could express it this way.

fsspec/caching.py
```diff
@@ -510,33 +576,37 @@ def _fetch(self, start, stop):
...
+        self._cache: OrderedDict[Any, T] = collections.OrderedDict()
         self._func = func
         self._max_size = max_size
         self._hits = 0
         self._misses = 0
         self._lock = threading.Lock()
 
-    def __call__(self, *args):
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
+        if kwargs:
```
martindurant, line 601: Why reject kwargs if actually we pass them on to _func, below?

fsspec/utils.py
```diff
@@ -10,12 +10,32 @@
...
+    Callable,
+    Iterable,
+    Iterator,
+    Sequence,
+    TypeVar,
+)
 from urllib.parse import urlsplit
 
+if TYPE_CHECKING:
+    from typing_extensions import TypeGuard
```
martindurant, line 26: What's this for?

fsspec/utils.py
```diff
@@ -150,13 +173,13 @@ def build_name_function(max_int):
...
 
-    def name_function(i):
+    def name_function(i: int) -> str:
         return str(i).zfill(pad_length)
 
     return name_function
 
 
-def seek_delimiter(file, delimiter, blocksize):
+def seek_delimiter(file: IO[bytes], delimiter: bytes, blocksize: int) -> bool:
```
martindurant, line 182: `file` is "file-like" (i.e., has seek/read), doesn't need to actually be a subclass of anything in particular. Is that what this says?

fsspec/utils.py
martindurant: Let's not change the logic in this PR. I am not immediately sure about what you suggest.

fsspec/utils.py
```diff
@@ -314,7 +347,7 @@ def stringify_path(filepath):
     """
     if isinstance(filepath, str):
         return filepath
-    elif hasattr(filepath, "__fspath__"):
+    elif hasattr(filepath, "__fspath__") or isinstance(filepath, os.PathLike):
```
martindurant, line 350: The isinstance is actually the same check, but slower.
```
class PathLike(abc.ABC):
    @classmethod
    def __subclasshook__(cls, subclass):
        if cls is PathLike:
            return _check_methods(subclass, '__fspath__')
        return NotImplemented
```

fsspec/utils.py
```diff
@@ -388,25 +429,25 @@ def other_paths(paths, path2, is_dir=None, exists=False, flatten=False):
     return path2
 
 
-def is_exception(obj):
+def is_exception(obj: Any) -> bool:
     return isinstance(obj, BaseException)
 
 
-def isfilelike(f):
+def isfilelike(f: Any) -> TypeGuard[IO[bytes]]:
```
martindurant, line 436: Return is bool, no?

fsspec/utils.py
```diff
@@ -482,11 +530,18 @@ def wrapper(cls):
 
 
 @contextmanager
-def nullcontext(obj):
+def nullcontext(obj: T) -> Iterator[T]:
```
martindurant, line 533: This is a generator before applying the decorator, but not a generator afterwards (it becomes a context). So, I'm not sure which state should be declared here.

fsspec/utils.py
martindurant: @rjzamora  ?

martindurant (commented):
Most of these are questions to educate myself.

fsspec/utils.py
alanhdu: Basically! What we want is a Protocol (https://mypy.readthedocs.io/en/stable/protocols.html), which is a way to formalize duck-typing without requiring subclassing. `IO[bytes]` is a built-in protocol for "file-like objects that return bytes" [0], so it works for anything fulfilling the interface.

[0] In practice though "file-like" is actually ill-defined in practice (e.g. do you want a `ReadOnlyFileLike` to statically disallow `write`?), so we might want to eventually make custom protocols for exactly the subset we need.

fsspec/utils.py
alanhdu: It's a bool, but `TypeGuard` is a special annotation that says `if isfilelike(x):` then you can treat `x` like a `IO[bytes]` from then on (for flow-sensitive typing)

https://peps.python.org/pep-0647/ has the full details

fsspec/utils.py
alanhdu: Yeah it's a little confusing. This is the correct because `contextmanager` has a type annotation that turns the iterator into a context manager for you.

fsspec/caching.py
alanhdu: This is just for backwards compatibility with the previous code (which doesn't take `kwargs`). Happy to also allow `kwargs` to be passed if you'd prefer!

Unfortunately, `ParamSpec` is quite limited and there's no way to say `args-only` for ParamSpec AFAICT -- we could also juts do `*args: Any` without type-checking the # of kind of args.

fsspec/caching.py
alanhdu: Sure!

`T` is a type variable that represents the output of the function. `P` is a special `ParamSpec` that represents the function's signature (`*args` and `**kwargs`). That lets us re-use the exact signature of what function that we're caching here

fsspec/caching.py
```diff
@@ -472,7 +533,12 @@ def __init__(self, blocksize, fetcher, size, data={}, strict=True, **_):
         else:
             self.data = data
 
-    def _fetch(self, start, stop):
+    def _fetch(self, start: int | None, stop: int | None) -> bytes:
+        if start is None:
```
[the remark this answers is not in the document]
alanhdu: Sort of. I'm pretty confident that passing in `start=None` would be cause a crash previously (since the `loc <= start < loc1` would error out), so in that sense it's not just to appease mypy.

But making making `start: int | None` (instead of just saying `start: int`) *is* to appease mypy to obey subtyping rules -- (basically: subtypes must have "at least as broad" type signatures as the base class, otherwise if I had `cache: BaseCache; cache._fethc(None, None)` this woudl crash if `cache` was a `KnownPartsOfAFile`).

Mypy's FAQ is decent here https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides.

fsspec/utils.py
alanhdu: Sure -- can revert and add a `type: ignore` instead.

fsspec/utils.py
martindurant: Aha, got it! So this avoids the need for something like `assert isinstance(x, io.FileIO)` elsewhere.

fsspec/utils.py
rjzamora: Certainly looks like a bug to me. That said, I don't think `starts` or `ends` ever correspond to anything other than `list` anyway. Perhaps we can just drop the `int` support since the result would have been "wrong" before anyway?

fsspec/utils.py
```diff
@@ -321,16 +351,18 @@ def stringify_path(filepath):
     elif hasattr(filepath, "path"):
         return filepath.path
     else:
-        return filepath
+        return filepath   # type: ignore[return-value]
```
martindurant, line 354:
```suggestion
        return filepath  # type: ignore[return-value]
```

fsspec/utils.py
```diff
@@ -500,7 +552,7 @@ def merge_offset_ranges(paths, starts, ends, max_gap=0, max_block=None, sort=Tru
     if not isinstance(starts, list):
         starts = [starts] * len(paths)
     if not isinstance(ends, list):
-        ends = [starts] * len(paths)
+        ends = [ends] * len(paths)
     if len(starts) != len(paths) or len(ends) != len(paths):
         raise ValueError
```
Skylion007, line 557: This seems like a bug, shouldn't we actually instatiate the error, and ideally print an error message about why it's thrown?

martindurant: A message would be a reasonable thing to add here, but it's not necessary to instantiate if there is no message - python will do it automatically.
```
>>> raise ValueError
ValueError
```

Skylion007 (approved):
I'm impressed, the typing in this PR is actually well/properly done. @martindurant since you wanted someone with more typing experience to take a look, this is actually pretty done well and except for minor comments all the typing annotations look correct.

fsspec/utils.py
```diff
@@ -266,11 +295,11 @@ def read_block(f, offset, length, delimiter=None, split_before=False):
         length = end - start
 
     f.seek(offset)
-    b = f.read(length)
+    b = f.read(length)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
Skylion007: Ma even go further and raise a ValueError is length is None and delimiter is None that they cannot both be None at the same time.

martindurant:
Thanks for having a look, @Skylion007 . @alanhdu , do you expect to have time to respond and fix the conflicts that have appeared since you submitted this?

fsspec/caching.py
calpaterson: I would suggest `start = typing.cast(int, self.start)` and then using `start`.  Same for end.

This is also bunkum code which exists solely for the typechecker but at least the intention is a little clearer

martindurant:
This work has gone stale - is there any chance of reviving it?

fsspec/utils.py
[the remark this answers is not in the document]
alanhdu: I added an `assert` here, but I wonder if we should to what I originally suggested of:
```
if length is None:
    b = f.read()
else:
    b = f.read(length)
```

The docs say that if `length is None` we want to read until the end of the string, which I believe this will.

>>> unless f.read can actually take None and it's just a type stubbing problem?

Arguably it is? I did just test with an actual file and confirm that `open(file).read(None)` gives reasonable results.

alanhdu:
@martindurant @skylion007 Sorry for the delay here. I rebased and things should look  be ok (modulo the discussion on https://github.com/fsspec/filesystem_spec/pull/1284#discussion_r1316062635).

I'm ok adding the `assert` (or raise a more informative `ValueError)`, but we could keep it as `# type: ignore` if we want to guarantee no change in the runtime behavior.

alanhdu:
Fixed!

PR #1286, opened 2023-06-06
bugfix: Strip file protocol when calling LocalFileSystem.rm_file

ajarvelin:
These changes fix calling `LocalFileSystem.rm_file(self, path)` with paths prefixed with `file://` protocol. The issue with the original code is that it does not call `LocalFileSystem._strip_protocol(self, path)` before passing the path to `os.remove(path)`.


martindurant:
@ianthomas23 : looks OK, but if you happen to have a minute to double check

ianthomas23 (approved):
This looks fine to me.

PR #1290, opened 2023-06-13
Fix conditional for async file systems

janjagusch:
We were noticing the following error when working with the async version of `gcsfs`:

https://github.com/fsspec/filesystem_spec/blob/f511ed2d79e2222fa002a16c5681d2864cf71c3d/fsspec/asyn.py#L80-L81

We believe that's because the `loop` attribute is always set to `None` when `asynchronous=True`. We assume that's a bug and that it should be the other way around.

If you agree, we would appreciate a patch release. 🙏

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -305,7 +305,7 @@ class AsyncFileSystem(AbstractFileSystem):
     def __init__(self, *args, asynchronous=False, loop=None, batch_size=None, **kwargs):
         self.asynchronous = asynchronous
         self._pid = os.getpid()
-        if not asynchronous:
+        if asynchronous:
             self._loop = loop or get_loop()
         else:
             self._loop = None


martindurant:
Would you mind showing how you hit this problem, please?

janjagusch:
Thanks for your prompt response, @martindurant. I believe the error lies on our side.
For context, we use `fsspec` in [Quetz](https://github.com/mamba-org/quetz), the open-source Conda server.

[We recently moved to the async versions of fsspec](https://github.com/mamba-org/quetz/pull/626) as we had performance problems with large concurrent file uploads.

However, it seems like we forgot to provide a loop to the constructor:

https://github.com/mamba-org/quetz/blob/a189de84b5bf2628f13a4fb1aa014b8cbfb28bd0/quetz/pkgstores.py#L605-L611

I initially thought this was a problem on the `fsspec` side, but was proven wrong by the [documentation](https://filesystem-spec.readthedocs.io/en/latest/async.html#using-from-async):

> For the non-asynchronous case, fsspec will normally create an asyncio event loop on a specific thread. However, the calling application may prefer IO processes to run on a loop that is already around and running (in another thread).

So, we should solve this problem by explicitly providing a loop when instantiating the `fsspec` objects. Does this sound like a reasonable approach to you, @martindurant?

martindurant:
Passing loop= is never a bad idea. In the future for async mode, it should not be necessary, but I am not surprised if there are still some edges of the code that need it. 

I did look very briefly at the linked PR code, and I don't actually see things like `await fs.pipe_file(fn, data)`, but only file-like operations. The file-like API is necessarily blocking/sync, since that's how python defines it. The experimental `fs.open_async` alternative is not yet ready for production use unfortunately.

PR #1291, opened 2023-06-15
Correct install GCS example in docs

### LINKED ISSUE #936: docs: text says S3, but example is for GCS
Text: "For example to be able to access data in S3, you can use the optional pip install syntax below, or install the specific package required"

but then the pip install example below it is for installing the Google Cloud Storage (gcs) library.

ianthomas23:
Fixes #936.

Minor docs change to correct the words about installing extra dependencies to match the example code.


ianthomas23:
Also added `afile` to codespell's list of ignore words as CI was failing on this. I figured this didn't warrant a new PR.

PR #1293, opened 2023-06-16
Smb non standard port

ondave:
The current samba implementation fails for non-standard ports. This is because the port argument in smbclient does not get propagated to the individual smbclient methods. These changes add the necessary port arguments to fix this issue.


martindurant:
Thanks for going as far as to add tests :)

PR #1295, opened 2023-06-16
Make support for bz2 compression optional

tsibley:
Python may be compiled without bz2 support, in which case import of the standard library module fails with a traceback like:

        …/lib/python3.7/site-packages/fsspec/__init__.py:12: in <module>
            from .compression import available_compressions
        …/lib/python3.7/site-packages/fsspec/compression.py:2: in <module>
            from bz2 import BZ2File
        …/lib/python3.7/bz2.py:19: in <module>
            from _bz2 import BZ2Compressor, BZ2Decompressor
    ModuleNotFoundError: No module named '_bz2'

While rare, it does happen, so make the bz2 registration optional.

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -1,5 +1,4 @@
 """Helper functions for a standard streaming compression API"""
-from bz2 import BZ2File
 from zipfile import ZipFile
 
 import fsspec.utils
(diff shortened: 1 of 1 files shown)


martindurant:
I hadn't realised _bz2 was not guranteeed. Thanks for posting this, it'll go in when green.

tsibley:
Thanks!

PR #1296, opened 2023-06-20
Add CacheMapper to map from remote URL to local cached basename

ianthomas23:
This is the start of work to make the caching system more configurable by splitting up different areas of responsibility of `CachingFileSystem` into separate class hierarchies that will eventually be pluggable.

It starts here by separating out the mapping from remote URL to local cached basename into a new class hierarchy based on `AbstractCacheMapper`. There is no change in functionality here, the two concrete derived classes cover the `same_names=True` and `same_names=False` options of `CachingFileSystem`. It will easy in future to add new `CacheMapper` classes or to modify the existing ones with extra attributes, e.g. number of directories to consider as well as the basename.

Implementation details:

1. `CacheMapper` is the best name I can think of but I am happy to change it if someone has a better idea.
2. `CachingFileSystem` no longer stores the `same_names` attribute, it is just used to create the `CacheMapper` which is stored. Removal of the `same_names` attribute could be considered an API change as it was technically public.
3. I have kept the `CachingFileSystem.hash_name` function even though it now defers to the `CacheMapper`. This is just for backwards compatibility in case downstream libraries are using it. The local `fsspec` tests all pass if it is removed.

In future, to support more pluggable options we will need to be able to pass `CacheMapper` instances to the `CachingFileSystem` constructor, but we will probably also have to keep the `same_names` arg for backward compatibility.

(Edited: typo)

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import hashlib
 import inspect
 import logging
 import os
@@ -9,13 +8,14 @@ import pickle
 import tempfile
 import time
 from shutil import rmtree
-from typing import ClassVar
+from typing import Any, ClassVar
 
 from fsspec import AbstractFileSystem, filesystem
 from fsspec.callbacks import _DEFAULT_CALLBACK
 from fsspec.compression import compr
 from fsspec.core import BaseCache, MMapCache
 from fsspec.exceptions import BlocksizeMismatchError
+from fsspec.implementations.cache_mapper import create_cache_mapper
 from fsspec.spec import AbstractBufferedFile
 from fsspec.utils import infer_compression
 
@@ -115,9 +115,7 @@ class CachingFileSystem(AbstractFileSystem):
         self.check_files = check_files
         self.expiry = expiry_time
         self.compression = compression
-        # TODO: same_names should allow for variable prefix, not only
-        #  to keep the basename
-        self.same_names = same_names
+        self._mapper = create_cache_mapper(same_names)
         self.target_protocol = (
             target_protocol
             if isinstance(target_protocol, str)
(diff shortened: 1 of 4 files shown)


martindurant:
Sorry for not touching this, coming to it soon

ianthomas23:
> Sorry for not touching this, coming to it soon

Thanks, that would be useful. There is quite a lot of refactoring and new functionality to be added on top of this, if you think it is going in the right direction.

fsspec/implementations/cached.py
```diff
@@ -255,11 +253,11 @@ def clear_expired_cache(self, expiry_time=None):
...
         for path, detail in self.cached_files[-1].copy().items():
             if time.time() - detail["time"] > expiry_time:
-                if self.same_names:
-                    basename = os.path.basename(detail["original"])
-                    fn = os.path.join(self.storage[-1], basename)
-                else:
-                    fn = os.path.join(self.storage[-1], detail["fn"])
+                fn = getattr(detail, "fn", "")
+                if not fn:
+                    # fn should always be set, but be defensive here.
```
martindurant, line 258: Feel like if we have a condition we expect never to hit, it's probably an exception or warning. I suppose this line is not covered by tests?

martindurant (approved):
I can find nothing wrong with this at all, the one tiniest question. Can merge anyway.

Also, would be happy to remove the eq/hash stuff - does it look like it has any use? Maybe no need to do that right now.

fsspec/implementations/cached.py
ianthomas23: I've changed this to raise an exception, and added an explicit test that the exception is raised in `test_cached.py::test_clear_expired` as that already contains the machinery to modify the cached metadata.

ianthomas23:
> Also, would be happy to remove the eq/hash stuff - does it look like it has any use? Maybe no need to do that right now.

We can safely remove it here, we'd just need to remove 2 lines of object and hash equality tests in `test_chained_equivalent`. But I'd be concerned that someone downstream might be using the public `hash_name` function. If you think it is safe to remove, maybe that should be in a separate PR so it is easy for us to revert if the need should arise?

ianthomas23:
I'll merge this tomorrow morning (UK time), leaving time for any further comments before then.

PR #1302, opened 2023-07-04
expose a preconfigured `dirfs` instance on mapper objects

### LINKED ISSUE #1269: using filesystem methods with mapper keys
The mapper interface is really convenient, especially when working with nested urls like
```
zip://directory::https://example.host/path/to/file.zip
```
However, getting open file objects or querying file metadata is not as easy and requires explicit path / url manipulations. For these cases I have been successfully using the internal `_key_to_str` and `_str_to_key` methods together with the mapper's `fs` instance, but since they are marked as private (and not really documented) I don't feel too comfortable with this.

Would it be possible to expose those two methods as public API, or add a filesystem wrapper that allows working with relative paths instead of absolute urls? For the latter, I was thinking about something like this:
```python
m.relative.open("relative/file/path", **options)
m.relative.exists("relative/file/path")
```
would translate to
```python
m.fs.open(m._key_to_str("relative/file/path"), **options)
m.fs.exists(m._key_to_str("relative/file/path"))
```
where `m` is the mapper object.

Otherwise, am I missing something obvious? Is there a better way to do this?

keewis:
- [x] closes #1269

I'm pretty happy with the way `dirfs` instances help working with mapper objects, but constructing one that uses the same parameters as the mapper is a bit more work than I would wish:

``` python
m = fsspec.get_mapper(url, ...)
dirfs = fsspec.filesystem("dirfs", path=url, fs=m.fs)
dirfs.open(...)
```

As suggested in #1269, this exposes a `dirfs` instance as a cached property, which means that anyone who doesn't need the `dirfs` instance doesn't have to wait the (very) little time it takes to construct one. With that, the above snipped would become:

``` python
m = fsspec.get_mapper(url, ...)
m.dirfs.open(...)
```
I'm not particularly attached to the name of the property, so I'm happy to change that.

Also, I didn't discuss this change before implementing it, so feel free to close if you don't think growing the mapper API makes sense in this case.

### DIFF
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -62,6 +62,7 @@ class FSMap(MutableMapping):
 
     @cached_property
     def dirfs(self):
+        """dirfs instance that can be used with the same keys as the mapper"""
         from .implementations.dirfs import DirFileSystem
 
         return DirFileSystem(path=self._root_key_to_str, fs=self.fs)
--- a/fsspec/mapping.py
+++ b/fsspec/mapping.py
@@ -2,6 +2,7 @@ import array
 import posixpath
 import warnings
 from collections.abc import MutableMapping
+from functools import cached_property
 
 from .core import url_to_fs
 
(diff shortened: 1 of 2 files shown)


martindurant:
I am happy with this. I only request a short docstring on the attribute explaining what it does, and to include it in the API doc for FSMap.

PR #1303, opened 2023-07-04
Adding support for onerror parameter in walk

### LINKED ISSUE #1297: Adding support for onerror parameter in walk
The current `walk()` implementation is missing the `onerror` parameter.

[See os.walk](https://docs.python.org/3/library/os.html#os.walk)

```
If optional argument onerror is specified, it should be a function;

it will be called with one argument, an OSError instance.

It can report the error to continue with the walk, or raise the exception to abort the walk
```

I can do the PR if approved.

epizut:
Fixes #1297, I was not sure how to test the async version since I haven't found any suitable existing test to start with.

The current `walk()` implementation is missing the `onerror` parameter.

[See os.walk](https://docs.python.org/3/library/os.html#os.walk)

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -641,7 +641,7 @@ class AsyncFileSystem(AbstractFileSystem):
     async def _ls(self, path, detail=True, **kwargs):
         raise NotImplementedError
 
-    async def _walk(self, path, maxdepth=None, **kwargs):
+    async def _walk(self, path, maxdepth=None, onerror=None, **kwargs):
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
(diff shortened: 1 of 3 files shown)


martindurant:
I quite like `onerror` being a function, but in the existing API, it's a string like "omit", "raise"...  Can you please revisit other usage and try to achieve consistency?

For async, the tricky part can be to figure out when to consider errors. Most of the coroutines will return error instances, so we can check for that. You will need to look over in s3fs for probably the most complete application.

epizut:
Sorry, I miss the existing `on_error`, I planned to use a function to mimic `os.walk()`.
To comply with others' on_error usages, I set the default to `raise`, unfortunatly it is a behavior change because the current implementation ignores every exception. I will be happy to default on `omit` if needed.

Issue #838, opened 2021-11-26
FileNotFoundError raised if filename contains braces []


saabeilin:
For example, uploading a file containing braces, leads to raising `FileNotFoundError` when `expand_path` calls `glob`

saabeilin:
Steps to reproduce:
 1. Create a file with braces in the filename
 2. Make sure you can do `ls <thisFile>` in your shell
 3. Execute:

```
from fsspec.implementations.local import LocalFileSystem
fs = LocalFileSystem()
fs.expand_path("/home/user/path [with braces]")
```

martindurant:
I suppose that's getting caught up by the regex which is used for glob patterns. Does using escapes work?
```
fs.expand_path("/home/user/path \\[with braces\\]")
```

saabeilin:
- that filename comes from some other call like `ls` or something like that
- the braces `[]` are valid characters in path right? At least `touch []` works without escaping :)
- however, the main question is why this `expand_path` is used for, say, uploading a single file.

martindurant:
> why this expand_path is used for, say, uploading a single file

We do not know whether a given path resolves to one or more files without trying to expand. If you are certain you are working with one file, you can call the one-file variant directly and skip that step
```
fs.put_file(local, remote)
```

martindurant:
> the braces [] are valid characters in path right?

I think this is backend-specific

saabeilin:
Hm. If I want to recursively upload a directory with braces?

martindurant:
You will probably need to do some string processing yourself, sorry. Probably `find()` works for you, or maybe you need to escape the special characters, and then pass the generated list of filenames to `put`, and I would expect that to work. Else, you could duplicate the code of `put` to prevent using the expand code.

I'm not sure I can think of a way to cope with your situation for all cases. They very easiest thing you can do, is to rename the files to contain only regular characters!

saabeilin:
I agree the limitation for file names is backend-specific, but this expansion behaviour is implemented in `AbstractFileSystem`. I would expect the `.upload()` will just, hm, try to upload (for instance - recursively), and should throw some exception if the remote backend FS is rejecting the call.

I would add that this path expansion in this call is absolutely unexpected behaviour, since the "ls"-like call from another library returns this unescaped and does file operations with unescaped filename.

saabeilin:
(sorry) one more comment, the call `fsspec.filesystem("file").ls("/tmp")` returns the names unescaped as well:
```
[...]

PR #1306, opened 2023-07-07
Match escaped square brackets in paths literally not as regexes

ianthomas23:
This is an experiment looking at issue #838 which fails to find file with names containing square brackets. As discussed in that issue square brackets are usually interpreted as regexes and we'd rather not add an extra kwarg throughout the API to specify whether regex or literal interpretation should apply.

So one option is to escape the square brackets using backslashes. That doesn't work on the `master` branch, but after this one-line addition it does work. Test script:
```python
import fsspec

# Create test files
fs.makedirs("issue838")
fs.touch("issue838/file[a].txt")
fs.touch("issue838/other.txt")

print("ls", fs.ls("issue838/"))
print("find", fs.find("issue838/"))

print("expand wildcard", fs.expand_path("issue838/file*"))
print("expand regex", fs.expand_path("issue838/[fo]*"))

# Remaining examples all fail on master branch.

# The following 3 are equivalent
print("expand literal", fs.expand_path("issue838/*\[*"))
print("expand literal", fs.expand_path("issue838/*\\[*"))
print("expand literal", fs.expand_path(r"issue838/*\[*"))

print("expand literal", fs.expand_path("issue838/*\]*"))
```
which using this PR gives
```
ls ['/tmp/issue838/file[a].txt', '/tmp/issue838/other.txt']
find ['/tmp/issue838/file[a].txt', '/tmp/issue838/other.txt']
expand wildcard ['/tmp/issue838/file[a].txt']
expand regex ['/tmp/issue838/file[a].txt', '/tmp/fsspec_temp/issue838/other.txt']
expand literal ['/tmp/issue838/file[a].txt']
expand literal ['/tmp/issue838/file[a].txt']
expand literal ['/tmp/issue838/file[a].txt']
expand literal ['/tmp/issue838/file[a].txt']
```
This isn't complete, it doesn't deal with anything other than square brackets and it causes some existing tests to fail.

Alternative implementations are possible. There are quite a lot of lines of string and regex processing around the change, and these could no doubt be modified to include the new behaviour but I would be wary of doing so and causing other failures elsewhere.

There is a certain irony that the one line change to force the behaviour to not be a regex is itself a regex in which the important characters have to be escaped to be interpreted correctly!

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -603,6 +603,11 @@ class AbstractFileSystem(metaclass=_Cached):
         )
         pattern = re.sub("[*]{2}", "=PLACEHOLDER=", pattern)
         pattern = re.sub("[*]", "[^/]*", pattern)
+
+        # Replace \\[ with \[ and \\] with \], which ensures that user-specified
+        # escaped square brackets are matched literally and not as regexes.
+        pattern = re.sub(r"\\\\([\[\]])", r"\\\1", pattern)
+
         pattern = re.compile(pattern.replace("=PLACEHOLDER=", ".*"))
         out = {
             p: allpaths[p]


ianthomas23:
I am closing this as I don't think it is a good solution.

PR #1307, opened 2023-07-09
fix(dirfs): allow protocol prefixed paths

Mause:


martindurant:
I would accept a PR to add dirfs into fsspec.registry.known_implementations , so that it could be instantiated with `fsspec.filesystem("dirfs", ...)` and other top-level functions that rely on it.

keewis:
isn't that already the case? Except that it's called `"dir"` instead of `"dirfs"`: https://github.com/fsspec/filesystem_spec/blob/0071252c2c5363f3599dd76db42468eb0ee6d791/fsspec/registry.py#L198

martindurant:
Oops, indeed :)

Mause:
> isn't that already the case? Except that it's called `"dir"` instead of `"dirfs"`:
> 
> https://github.com/fsspec/filesystem_spec/blob/0071252c2c5363f3599dd76db42468eb0ee6d791/fsspec/registry.py#L198

Should the protocol be `dir` then, to match?

martindurant:
D'oh, yes! (I seem to have taken my eyes off the ball on this one, sorry)

PR #1308, opened 2023-07-10
Warn if s3fs is ancient

martindurant:
Fixes https://github.com/dask/dask/issues/10397

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -233,10 +233,8 @@ def get_filesystem_class(protocol):
     return cls
 
 
-s3_msg = """Your installed version of s3fs is very old. This can happen
-when you have an explicit dependence on botocore (e.g., from boto3),
-which forces the solver to pick a version of s3fs which did not yet
-depend on aiobotocore.
+s3_msg = """Your installed version of s3fs is very old and known to cause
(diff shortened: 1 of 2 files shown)


martindurant:
@fjetter , is this wording to your satisfaction?

fsspec/registry.py
```diff
@@ -233,20 +233,36 @@ def get_filesystem_class(protocol):
...
 
 
+s3_msg = """Your installed version of s3fs is very old. This can happen
+when you have an explicit dependence on botocore (e.g., from boto3),
+which forces the solver to pick a version of s3fs which did not yet
+depend on aiobotocore.
+
+To fix, you should specify a lower version bound on s3fs, or
+update the current installation.
+"""
```
fjetter, line 243:
```suggestion
s3_msg = """Your installed version of s3fs is very old and known to cause
severe performance issues, see also https://github.com/dask/dask/issues/10276

To fix, you should specify a lower version bound on s3fs, or
update the current installation.
"""
```

I think it's important that we tell users that this is a severe performance problem. Many users are fine with using outdated libraries as long as they do not encounter an issue and they should not need to connect the dots themselves if they see something being slow.

Otherwise, I'm happy with the message but I would probably just link to a github issue with a more thorough explanation instead of making the warning text longer.

PR #1309, opened 2023-07-10
Fix MemoryFileSystem.touch by just using the default super impl

FHTMitchell:
The current behaviour of `.exists` doesn't do smart checking for directories.

```py
import fsspec
fs, _ = fsspec.core.url_to_fs('memory://')
fs.touch('/a/b')
fs.exists('/a/b')
Out[5]: True
fs.exists('/a')
Out[6]: False
fs.isdir('/a')
Out[7]: True
```

We can simply use the super implementation which just checks `.info` doesn't error. 

Also added some tests :)

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -137,10 +137,6 @@ class MemoryFileSystem(AbstractFileSystem):
         else:
             raise FileNotFoundError(path)
 
-    def exists(self, path, **kwargs):
-        path = self._strip_protocol(path)
-        return path in self.store or path in self.pseudo_dirs
-
     def info(self, path, **kwargs):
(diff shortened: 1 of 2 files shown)


fsspec/implementations/memory.py
```diff
@@ -191,11 +187,14 @@ def _open(
...
             else:
                 raise FileNotFoundError(path)
-        if mode == "wb":
+        elif mode == "wb":
             m = MemoryFile(self, path, kwargs.get("data"))
             if not self._intrans:
                 m.commit()
             return m
+        else:
+            name = self.__class__.__name__
```
martindurant, line 196: Do you expect this implementation to have subclasses?

martindurant (approved):

PR #1311, opened 2023-07-14
fix(dirfs): correct protocol of DirFileSystem

Mause:
Per discussion on https://github.com/fsspec/filesystem_spec/pull/1307


Mause:
@martindurant I assume this is what was meant?

PR #1318, opened 2023-07-26
Support cache mapper that is basename plus fixed number of parent directories

ianthomas23:
This adds support for a cache mapper that is a fixed number of parent directories as well as the basename. This will be useful when using multiple directories containing files with the same names but you still want the cached filenames to be clearly identifiable rather than using hashes.

I have chosen to add a new `directory_levels` attribute to `BasenameCacheMapper` rather than create a new class.

Example use:
```python
from fsspec import filesystem
from fsspec.implementations.cache_mapper import BasenameCacheMapper

fs = filesystem("filecache", target_protocol='file', cache_mapper=BasenameCacheMapper(2))
```

I have made some design decisions for this that are up for discussion:

1. I've kept the `same_names` kwarg to `CachingFileSystem.__init__` for backward compatibility as well as adding the new `cache_mapper` kwarg. You cannot specify both. This seemed a better approach than the alternative of allowing values for `same_names` other than boolean and allows us to easily support other `CacheMapper` classes in the future.
2. For the cached filenames I am not using usual path separators as I wanted to use the same separators on all OSes and avoid any problems in inferring that slashes are indeed directory separators and looking in the wrong place. I am using a somewhat arbitrary string of `_^_` so far but not tied to this.

### DIFF
--- a/fsspec/implementations/cache_mapper.py
+++ b/fsspec/implementations/cache_mapper.py
@@ -45,7 +45,7 @@ class BasenameCacheMapper(AbstractCacheMapper):
         self.directory_levels = directory_levels
 
         # Separator for directories when encoded as strings.
-        self._separator = "_^_"
+        self._separator = "_@_"
 
     def __call__(self, path: str) -> str:
         dirname, basename = os.path.split(path)
--- a/fsspec/implementations/cache_mapper.py
+++ b/fsspec/implementations/cache_mapper.py
@@ -48,14 +48,11 @@ class BasenameCacheMapper(AbstractCacheMapper):
         self._separator = "_@_"
 
     def __call__(self, path: str) -> str:
-        dirname, basename = os.path.split(path)
-
-        if self.directory_levels > 0:
-            dirs = dirname.split(os.sep)[-self.directory_levels :]
-            dirs.append(basename)
-            basename = self._separator.join(dirs)
-
-        return basename
+        prefix, *bits = path.rsplit(os.sep, self.directory_levels + 1)
+        if bits:
+            return self._separator.join(bits)
+        else:
+            return prefix  # No separator found, simple filename
 
     def __eq__(self, other: Any) -> bool:
         return super().__eq__(other) and self.directory_levels == other.directory_levels
--- a/fsspec/implementations/cache_mapper.py
+++ b/fsspec/implementations/cache_mapper.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import abc
 import hashlib
-import os
 from typing import TYPE_CHECKING
+
+from fsspec.implementations.local import make_path_posix
 
 if TYPE_CHECKING:
     from typing import Any
@@ -48,7 +49,8 @@ class BasenameCacheMapper(AbstractCacheMapper):
         self._separator = "_@_"
 
     def __call__(self, path: str) -> str:
-        prefix, *bits = path.rsplit(os.sep, self.directory_levels + 1)
+        path = make_path_posix(path)
+        prefix, *bits = path.rsplit("/", self.directory_levels + 1)
         if bits:
             return self._separator.join(bits)
         else:
--- a/fsspec/implementations/cache_mapper.py
+++ b/fsspec/implementations/cache_mapper.py
@@ -30,14 +30,38 @@ class AbstractCacheMapper(abc.ABC):
 
 
 class BasenameCacheMapper(AbstractCacheMapper):
-    """Cache mapper that uses the basename of the remote URL.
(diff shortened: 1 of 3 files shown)


ianthomas23:
I've added type annotations to the lines I have changed so now `mypy` is interested in the `cached.py` file and complaining about the
```python
self._strip_protocol = _strip_protocol
```
line.

I've added a `# ignore: type[method-assign]` to get rid of the error. I am not sure if we have a policy on this yet, I could always put the ignore in the `setup.cfg` instead.

martindurant:
> I am not sure if we have a policy on this yet

There is no policy, typing is very limited in fsspec.

> a somewhat arbitrary string of `_^_` so far

So long as it isn't a glob re special, I don't mind. I think re only uses it within `[..]`.

fsspec/implementations/cache_mapper.py
```diff
@@ -30,14 +30,38 @@ def __hash__(self) -> int:
...
     def __call__(self, path: str) -> str:
-        return os.path.basename(path)
+        dirname, basename = os.path.split(path)
+
+        if self.directory_levels > 0:
+            dirs = dirname.split(os.sep)[-self.directory_levels :]
+            dirs.append(basename)
+            basename = self._separator.join(dirs)
+
+        return basename
```
martindurant, line 58: The use of `os.sep` here probaby broke the windows test?

Also, I think it might be simpler:

```suggestion
        prefix, *bits = dirname.rsplit(separator, self.directory_labels + 1)
        return self._separator.join(bits)
```

fsspec/implementations/cache_mapper.py
```diff
@@ -30,14 +30,38 @@ def __hash__(self) -> int:
...
+
+    def __eq__(self, other: Any) -> bool:
+        return super().__eq__(other) and self.directory_levels == other.directory_levels
+
+    def __hash__(self) -> int:
+        return super().__hash__() ^ hash(self.directory_levels)
```
martindurant, line 64:
```suggestion
```

fsspec/implementations/cached.py
```diff
@@ -128,7 +150,7 @@ def _strip_protocol(path):
             # acts as a method, since each instance has a difference target
             return self.fs._strip_protocol(type(self)._strip_protocol(path))
 
-        self._strip_protocol = _strip_protocol
+        self._strip_protocol = _strip_protocol  # type: ignore[method-assign]
```
martindurant, line 153: I don't know how to make mypy happy usually.

```suggestion
        self._strip_protocol: Callable = _strip_protocol  # works ?
```

ianthomas23:
> > a somewhat arbitrary string of `_^_` so far
> 
> So long as it isn't a glob re special, I don't mind. I think re only uses it within `[..]`.

I've changed it to `_@_` just in case there is any re confusion over use of `^`

fsspec/implementations/cache_mapper.py
ianthomas23: Yes, your simpler approach works.

fsspec/implementations/cache_mapper.py
ianthomas23: I haven't made the suggested change yet. Doing so doesn't break any old tests, only the new one I added to explicitly check what I added.

I only added the `__eq__` and `__hash__` functionality because of the same functions on `CachingFileSystem` used to depend on the `same_names` attribute so now depend on the `_mapper` attribute, and for equivalent behaviour `BasenameCacheMapper(1)` must always be equal to any other `BasenameCacheMapper(1)` and different to any `BasenameCacheMapper(2)`. But if we don't need this equality/inequality then it can all go. But then I would ask if this is consistent with the existence of `CachingFileSystem.__eq__`, i.e. what is the purpose of the latter?

ianthomas23:
To get this working on Windows I've used an explicit `make_path_posix` call. We can't just rely on `os.sep` as sometimes forward slashes are OK on Windows (e.g. `MemoryFileSystem`) so this seems the easiest solution.

fsspec/implementations/cache_mapper.py
martindurant: I think al the eq/hash should be dropped if they are only used for tests or not at all. For the former, we can always look at the value of the attributes, after all; plus every filesystem instance has a `._fs_token_` attribute too.

martindurant:
Everything is resolved here, except I leave it up to you whether to remove the eq/hash stuff now or not.

ianthomas23:
> Everything is resolved here, except I leave it up to you whether to remove the eq/hash stuff now or not.

OK, let's hold off the eq/hash changes to a separate PR as I need to think about it a bit more.

PR #1320, opened 2023-07-29
Verify Content-Encoding when querying Content-Length

### LINKED ISSUE #1319: Truncated response when downloading "Content-Type: gzip" file
Hi, all. I have a minimal example for a truncated response, and am wondering if I am using fsspec incorrectly or if this is a bug:
```python
import fsspec
with fsspec.open("https://s3.amazonaws.com/bossdb-open-data/neurodata/templier/Wafer1/C1_EM/info") as f:
  resp = f.read()
  print(len(resp))  # Should be 1123 Bytes, but returns 298, cutting of the json file
```

* The JSON file is gzipped on S3, Content-Encoding is set to gzip
* `requests` and `aiohttp` both correctly download and decode the JSON file
* 298 Bytes is the size of the _compressed_ object (and matches the Content-Length indicated by the server).

I believe the problem is this assumption here: https://github.com/fsspec/filesystem_spec/blob/9a1e624022f3ad39071de5b17bafa23214b8662b/fsspec/implementations/http.py#L819-L820
Not sure if S3 is at fault for ignoring the `Accept-Encoding` header or fsspec for ignoring the returned `Content-Encoding: gzip`, but the latter might be easier to fix.

I'll open a PR that fixes the issue, but I am no expert on HTTP specs or fsspec, so open to hear your thoughts regarding correct usage of fsspec for this use case.

nkemnitz:
Fixes #1319.


fsspec/implementations/http.py
```diff
@@ -841,7 +840,10 @@ async def _file_info(url, session, size_policy="head", **kwargs):
         #                 or 'Accept-Ranges': 'none' (not 'bytes')
         #  to mean streaming only, no random access => return None
         if "Content-Length" in r.headers:
-            info["size"] = int(r.headers["Content-Length"])
+            # Some servers may choose to ignore Accept-Encoding and return
+            # compressed content, in which case the returned size is unreliable.
+            if r.headers.get("Content-Encoding", "identity") == "identity":
+                info["size"] = int(r.headers["Content-Length"])
```
martindurant, line 846: Otherwise is remains `None` and can only be streamed? That is probably right, but could use some tests to make sure we are getting the right behaviour.

nkemnitz:
Added a test that will fail on the current main branch and succeed with this PR.

fsspec/implementations/tests/test_http.py
```diff
@@ -280,6 +280,16 @@ def test_content_length_zero(server):
         assert f.read() == data
 
 
+def test_content_encoding_gzip(server):
+    h = fsspec.filesystem(
+        "http", headers={"give_length": "true", "gzip_encoding": "true"}
+    )
+    url = server + "/index/realfile"
+
+    with h.open(url, "rb") as f:
```
martindurant, line 289: `f` is a streaming type file here, with `None` for the size? Should probably assert that.

nkemnitz: Ah, not super familiar with fsspec, yet - Only tried to mimic the test above this one. Added assertions for type and size, now.

PR #1326, opened 2023-08-03
Separate CacheMetadata class for CachingFileSystem

ianthomas23:
This PR moves handling of cache metadata out of the `CachingFileSystem` and into a new class `CacheMetadata`. The idea is for the metadata handling occur in one place so that we can easily modify and/or extend it in future PRs, e.g. to change the storage format or location.

Essentially `cached_files` used to be a property of `CacheMetadata` but is now a property of the new class. Both `CachingFileSystem` and `CacheMetadata` need access to the `storage` locations, so this is shared between them.

There is no functional change here. The only changes to the tests are due to the change in where the private `cached_files` are stored. There is one failing test that I haven't yet fixed, it is `test_metadata_save_blocked` and is something to do with the hot-swapping of the `open` function that I don't fully understand.

### DIFF
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -31,7 +31,7 @@ class CacheMetadata:
         ----------
         storage: list[str]
             Directories containing cached files, must be at least one. Metadata
-            is stored in the last of these directores by convention.
+            is stored in the last of these directories by convention.
         """
         if not storage:
             raise ValueError("CacheMetadata expects at least one storage location")
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-import contextlib
 import os
 import pickle
-import tempfile
 import time
 from typing import TYPE_CHECKING
+
+from fsspec.utils import atomic_write
 
 if TYPE_CHECKING:
     from typing import Any, Dict, Iterator, Literal
@@ -204,24 +204,3 @@ class CacheMetadata:
     def update_file(self, path: str, detail: Detail) -> None:
         """Update metadata for specific file in memory, do not save"""
         self.cached_files[-1][path] = detail
-
-
-@contextlib.contextmanager
-def atomic_write(path: str, mode: str = "wb"):
-    """
-    A context manager that opens a temporary file next to `path` and, on exit,
-    replaces `path` with the temporary file, thereby updating `path`
-    atomically.
-    """
-    fd, fn = tempfile.mkstemp(
-        dir=os.path.dirname(path), prefix=os.path.basename(path) + "-"
-    )
-    try:
-        with open(fd, mode) as fp:
-            yield fp
-    except BaseException:
-        with contextlib.suppress(FileNotFoundError):
-            os.unlink(fn)
-        raise
-    else:
-        os.replace(fn, path)
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -143,19 +143,25 @@ class CacheMetadata:
         if c["blocks"] is not True and len(c["blocks"]) * f.blocksize >= f.size:
             c["blocks"] = True
 
-    def pop_file(self, path: str) -> None:
+    def pop_file(self, path: str) -> str | None:
+        """Remove metadata of cached file.
+
+        If path is in the cache, return the filename of the cached file,
+        otherwise return ``None``.  Caller is responsible for deleting the
+        cached file.
+        """
         details = self.check_file(path, None)
         if not details:
-            return
+            return None
         _, fn = details
         if fn.startswith(self._storage[-1]):
-            os.remove(fn)
             self.cached_files[-1].pop(path)
             self.save()
         else:
             raise PermissionError(
                 "Can only delete cached file in last, writable cache location"
             )
+        return fn
 
     def save(self) -> None:
         """Save metadata to disk"""
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -90,12 +90,14 @@ class CacheMetadata:
                 return detail, fn
         return False
 
-    def clear_expired(self, expiry_time: int) -> Iterator[str]:
+    def clear_expired(self, expiry_time: int) -> tuple[list[str], bool]:
         """Remove expired metadata from the cache.
 
-        Yields names of files corresponding to expired metadata. Caller is
-        responsible for deleting these files.
+        Returns names of files corresponding to expired metadata and a boolean
+        flag indicating whether the writable cache is empty. Caller is
+        responsible for deleting the expired files.
         """
+        expired_files = []
         for path, detail in self.cached_files[-1].copy().items():
             if time.time() - detail["time"] > expiry_time:
                 fn = detail.get("fn", "")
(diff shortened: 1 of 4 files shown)


fsspec/implementations/cached.py
```diff
@@ -206,25 +162,17 @@ def _check_cache(self):
...
-                    continue
-            if self.expiry:
-                if time.time() - detail["time"] > self.expiry:
-                    continue
-            fn = os.path.join(storage, detail["fn"])
-            if os.path.exists(fn):
-                return detail, fn
-        return False
+
+        def ignore(detail: Detail) -> bool:
```
ianthomas23, line 167: I am passing in this function as a callback to the `_metadata.check_file` function which needs access to information it doesn't have. There are other possible alternatives to this approach.

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+        responsible for deleting these files.
+        """
+        for path, detail in self.cached_files[-1].copy().items():
+            if time.time() - detail["time"] > expiry_time:
+                fn = detail.get("fn", "")
+                if not fn:
+                    raise RuntimeError(
+                        f"Cache metadata does not contain 'fn' for {path}"
+                    )
+                fn = os.path.join(self._storage[-1], fn)
```
martindurant, line 96: Should be "/", i.e., are we assuming the storage locations are fsspec-style URLs ?

martindurant: (also occurs elsewhere)

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+                yield fn
+                self.cached_files[-1].pop(path)
+
+        if self.cached_files[-1]:
+            cache_path = os.path.join(self._storage[-1], "cache")
+            with atomic_write(cache_path) as fc:
+                pickle.dump(self.cached_files[-1], fc)
+
+    def close_file(self, f: Any, path: str) -> None:
```
martindurant, line 105: This method could have a more precise name

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+
+        The actual closing of the file is the responsibility of the caller.
+        """
+        # File must be writeble, so in self.cached_files[-1]
+        c = self.cached_files[-1][path]
+        if c["blocks"] is not True and len(c["blocks"]) * f.blocksize >= f.size:
+            c["blocks"] = True
+
+    def empty(self) -> bool:
+        """Return ``True`` if metadata of the writable storage is empty"""
```
martindurant, line 116: Why only of the writable store?

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+                    loaded_cached_files = pickle.load(f)
+                    for c in loaded_cached_files.values():
+                        if isinstance(c["blocks"], list):
+                            c["blocks"] = set(c["blocks"])
+                    cached_files.append(loaded_cached_files)
+            else:
+                cached_files.append({})
+        self.cached_files = cached_files or [{}]
+
+    def pop_file(self, path: str) -> None:
```
martindurant, line 135: Docstring: remove local copy of file and its metadata

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+                with open(fn, "rb") as f:
+                    cached_files = pickle.load(f)
+                for k, c in cached_files.items():
+                    if k in cache:
+                        if c["blocks"] is True or cache[k]["blocks"] is True:
+                            c["blocks"] = True
+                        else:
+                            # self.cached_files[*][*]["blocks"] must continue to
+                            # point to the same set object so that updates
+                            # performed by MMapCache are propagated back to
```
martindurant, line 165: Shouldn't they already be the same object?

fsspec/implementations/cache_metadata.py
```diff
@@ -0,0 +1,210 @@
...
+                pickle.dump(cache, f)
+            self.cached_files[-1] = cached_files
+
+    def update_file(self, path: str, detail: Detail) -> None:
+        """Update metadata for specific file in memory, do not save"""
+        self.cached_files[-1][path] = detail
+
+
+@contextlib.contextmanager
+def atomic_write(path: str, mode: str = "wb"):
```
martindurant, line 193: Maybe belongs in utils

fsspec/implementations/cached.py
martindurant: I think it would be fine to have the callback signature allow/require the cache instance rather than making a dynamic function closure.

martindurant (commented):
Looks good.

I commented on the use of os.* functions, but maybe thoughts about allowing generic fsspec URLs should be deferred.

fsspec/implementations/cached.py
ianthomas23: OK, I've done that.

fsspec/implementations/cache_metadata.py
ianthomas23: It looks like historically we have always used `os.path.join` for such manipulations. Our docs and tests seem to have a mixture of fsspec-style URLs and OS-specific paths.

I'll try standardising on forward slashes and see what breaks on Windows.

ianthomas23: Just noticed that you thought we could defer this change, so I'll leave this until I've fixed everything else.

fsspec/implementations/cache_metadata.py
ianthomas23: `on_close_cached_file` is a better name.

fsspec/implementations/cache_metadata.py
ianthomas23: We only need to know if the writable store is empty following a call to `CacheMetadata.clear_cache()` as if it is clear then the caller (`CachingFileSystem.clear_expired_cache()`) deletes all of the writable cache and resets the cache access time, etc.

It is clumsy. I could just make the name better. But probably better still is to return a boolean from `CacheMetadata.clear_cache()` to tell `CachingFileSystem.clear_expired_cache()` whether the writable cache is empty. Then we wouldn't have an `empty()` method hanging around waiting to be used incorrectly.

fsspec/implementations/cache_metadata.py
ianthomas23: Done. Also moved the deletion of the cached file up to `CachedFileSystem` as that is consistent with the class responsibilities.

fsspec/implementations/cache_metadata.py
ianthomas23: I have moved this comment and surrounding code verbatim from current `master`: https://github.com/fsspec/filesystem_spec/blob/45a6aec7da1407243f9767c6ab0cff40efee72eb/fsspec/implementations/cached.py#L169-L172

I haven't yet invested any time into the block caching to have an opinion about this yet.

fsspec/implementations/cache_metadata.py
ianthomas23: Yes.

fsspec/implementations/cache_metadata.py
martindurant: I like the plan you describe

martindurant:
The one failing test is indeed related to caching, but I can't tell if it's really failing or perhaps just the text of the error changed so the pyrest match got broken.

fsspec/implementations/cache_metadata.py
ianthomas23: Done.

ianthomas23:
The failure is a real one in `fsspec/implementations/tests/test_cached.py::test_metadata_save_blocked`. Somehow I have contrived to break the dynamic swapping of the `open` function which raises a `NameError` to simulate an interpreter shutdown. I am still trying to understand the cunning `CachingFileSystem.__getattribute__` to work out whether I have fundamentally broken something important or I just need a slightly different test.

Issue #1322, opened 2023-08-02
LocalFileSystem glob removes trailing slash


john-jam:
According to the [documentation of AbstractFileSystem](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L541:L542), `glob` should act as `ls` when the path ends with `/`.

But when using the `LocalFileSystem` method, [the trailing slash is removed](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/local.py#L69) when `_strip_protocol` is called.

Example:

```python
from fsspec.implementations.local import LocalFileSystem

lfs = LocalFileSystem(auto_mkdir=True)

lfs.touch(path="/tmp/data/a.txt")
print(lfs.glob(path="/tmp/data/"))
```

Output:

```bash
['/tmp/data']
```

Expected output:

```bash
['/tmp/data', '/tmp/data/a.txt']
```

Issue #1315, opened 2023-07-25
Recursive ** in glob is 1+ directories instead of 0+


grantgustafson:
Using `glob` with `**` matches one or more directories instead of zero or more:

```py
In [1]: import fsspec
   ...: from fsspec.implementations.memory import MemoryFileSystem
   ...: 
   ...: fs = MemoryFileSystem()
   ...: fs.touch("a/b.txt")
   ...: fs.touch("a/c/d.txt")

In [2]: fs.glob("**/*")
Out[2]: ['/a/b.txt', '/a/c', '/a/c/d.txt']

In [3]: fs.glob("a/**/*")
Out[3]: ['/a/c/d.txt']
```

martindurant:
Can you please state what you think the output of the two calls ought to be?

grantgustafson:
My mistake,
I believe both should be:
```
['/a/b.txt', '/a/c', '/a/c/d.txt']
```

ianthomas23:
Closing this as no action in necessary.

PR #1329, opened 2023-08-06
Better double asterisks `**` support

### LINKED ISSUE #1321: Avoid destination directory creation when using glob double asterisk(**)
I would like to avoid creating the last folder of the source path a the destination with glob double asterisk(**). When using them in the `copy` method on the `LocalFileSystem` (and other remote fs implementation as well), the destination folder (the last path) is created in several unwanted situations.

### Example

**Code:**

```python

fs = LocalFileSystem(auto_mkdir=True)

# List source files
print(f"Source files: {fs.glob('/tmp/data/**.txt')}")

# Copy to a location with trailing slash
fs.copy(path1="/tmp/data/**.txt", path2="/tmp/out/")
print(f"Test1: {fs.glob('/tmp/out/**.txt')}")
fs.rm(path="/tmp/out", recursive=True)

# Copy to a location without trailing slash
fs.copy(path1="/tmp/data/**.txt", path2="/tmp/out")
print(f"Test2: {fs.glob('/tmp/out/**.txt')}")
fs.rm(path="/tmp/out", recursive=True)

# Copy twice to a location without trailing slash
fs.copy(path1="/tmp/data/**.txt", path2="/tmp/out")
fs.copy(path1="/tmp/data/**.txt", path2="/tmp/out")
print(f"Test3: {fs.glob('/tmp/out/**.txt')}")
fs.rm(path="/tmp/out", recursive=True)
```

**Output:**
```bash
Source files: ['/tmp/data/a/b/f3.txt', '/tmp/data/a/f2.txt', '/tmp/data/f1.txt']
Test1: ['/tmp/out/data/a/b/f3.txt', '/tmp/out/data/a/f2.txt', '/tmp/out/data/f1.txt']
Test2: ['/tmp/out/a/b/f3.txt', '/tmp/out/a/f2.txt', '/tmp/out/f1.txt']
Test3: ['/tmp/out/a/b/f3.txt', '/tmp/out/a/f2.txt', '/tmp/out/data/a/b/f3.txt', '/tmp/out/data/a/f2.txt', '/tmp/out/data/f1.txt', '/tmp/out/f1.txt']
```

When a trailing slash is added to the destination (Test1), the folder `data` is created. When the trailing slash is not added and we copy only once (Test2), the folder `data` is not created (expected results). But when we do twice the copy operation, the folder `data` is created once again.

**Expect:** `['/tmp/out/a/b/f3.txt', '/tmp/out/a/f2.txt', '/tmp/out/f1.txt']` even when we call copy twice.

john-jam:
## Description

This PR:
- [x] makes the `glob` method compliant with posix rules and add better support for double asterisks `**` in globs.
- [x] supports `maxdepth` for the `glob` method.
- [x] fixes the `copy`, `get` and `put` methods to not create unwanted directories on some scenarios.

I added some tests to check the posix rules against the python `glob.glob` implementation and the bash `globstar` support. Both do not support double asterisks with other prefix/suffix than `/` so for example `**.json` is the same as `*.json` and not the same as `**/*.json`. But since this repository already supported that well, this feature is kept and patterns like `ab**cd` are still working properly (returning paths like `ab/ef/cd`).

New rules:
- the root dir is returned when using `find`. It is because of the posix `**` rule compliance. For example, it should return an empty string among the other paths when using `**` or return `ab` among the other paths when using `ab/**`.
- when using a trailing slash in globs, it returns only directories.
- the double asterisks `**` matches 0+ directories.
- the `maxdepth` option is applied on the first `**` when using a glob.

## References
- Closes #1321, #1315, #1322
- Posix rules for glob: https://man7.org/linux/man-pages/man7/glob.7.html


john-jam:
@martindurant @ianthomas23 I can't run all the tests on my machine locally (errors with dask and ftp). Could you trigger a CI run?

martindurant:
You are failing on the range of things that we typically see break whenever you change anything about how paths work :|. In particular, you will find it difficult to reconcile what happens on windows without having a system of your own to run tests on.

fsspec/implementations/local.py
```diff
@@ -65,10 +65,6 @@ def ls(self, path, detail=False, **kwargs):
         else:
             return [posixpath.join(path, f) for f in os.listdir(path)]
 
-    def glob(self, path, **kwargs):
-        path = self._strip_protocol(path)
```
martindurant, line 69: Removing _strip_protocol from the flow here is likely the cause of all the windows errors, because for LocalFileSystem, it also converts windows-style paths to posix.

martindurant:
@ianthomas23 , do you think you will have time to go through this?

ianthomas23:
> @ianthomas23 , do you think you will have time to go through this?

Yes I will, but later in the week.

fsspec/implementations/local.py
john-jam: This is what caused this issue: https://github.com/fsspec/filesystem_spec/issues/1322
How about using `make_posix_path` helper here then?

john-jam:
> You are failing on the range of things that we typically see break whenever you change anything about how paths work

Yes, there's a lot of updates to force `glob` compliant with `posix` rules but maybe some of them are not wanted? For example the following posix rules could be reverted if you think it's not worth it:
- the root dir is returned when using double asterisks **. 
- when using a trailing slash in globs, it returns only directories.

.github/workflows/main.yaml
```diff
@@ -157,7 +157,7 @@ jobs:
 
       - name: Clone
         shell: bash -l {0}
-        run: git clone https://github.com/fsspec/${{ matrix.FRIEND }}
+        run: git clone -b better-double-asterisk-support https://github.com/john-jam/${{ matrix.FRIEND }}
```
john-jam, line 160: Note for later: remove this (decide to merge the friends repo branches or not)

john-jam:
@martindurant @ianthomas23 I created PRs on `gcsfs` and `s3fs` to add some fixes and derived tests in `gcsfs`.

fsspec/implementations/local.py
john-jam: I fixed the windows tests by using `make_path_posix` on the expected outputs when required.

.github/workflows/main.yaml
ianthomas23: Noted. I'd normally be inclined to merge this first, then follow up with the `s3fs` and `gcsfs` PRs in quick succession.

fsspec/tests/abstract/__init__.py
```diff
@@ -61,7 +85,7 @@ def local_target(self, local_fs, local_join, local_path):
         if local_fs.exists(target):
             local_fs.rm(target, recursive=True)
 
-    def _bulk_operations_scenario_0(self, some_fs, some_join, some_path):
+    def _bulk_operations_scenario_0(self, some_fs, some_join, some_path, glob=False):
```
ianthomas23, line 88: The initial idea here was that there would be one and only one scenario for all the `copy`-like tests and the tests would be identical to the examples in the docs so that it is easy to demonstrate that the tests are appropriate and easy to follow for future contributors. Having two variants of the scenario 0 breaks this, which makes me slightly uncomfortable, but I can see that it is better this way than using the more complicated `glob` scenario for all the `copy` tests. So I am OK to go ahead with this, but can you add a new diagram in the docstring below to illustrate what the `glob=True` scenario consists of?

fsspec/tests/conftest.py
```diff
@@ -17,6 +17,186 @@
 ).read()
 win = os.name == "nt"
 
+GLOB_TESTS = [
```
ianthomas23, line 20: Can you add a comment here explaining what the 4 tuple elements are?

fsspec/tests/test_spec.py
```diff
@@ -611,3 +1039,248 @@ def check_events(lpaths, rpaths):
...
+        .replace("(", "\\(")
+        .replace(")", "\\)")
+        .replace("|", "\\|")
+    )
+    bash_output = subprocess.run(
+        ["bash", "-c", f"shopt -s globstar && stat -c %N {bash_abs_path}"],
+        capture_output=True,
+    )
+    bash_output = bash_output.stdout.decode("utf-8").replace("'", "").split("\n")
+    assert _clean_paths(bash_output, local_fake_dir) == _clean_paths(expected)
```
ianthomas23, line 1104:
```suggestion
    try:
        bash_output = subprocess.run(
            ["bash", "-c", f"shopt -s globstar && stat -c %N {bash_abs_path}"],
            capture_output=True,
            check=True,
        )

        bash_output = bash_output.stdout.decode("utf-8").replace("'", "").split("\n")
        assert _clean_paths(bash_output, local_fake_dir) == _clean_paths(expected)
    except subprocess.CalledProcessError as e:
        # If bash or stat don't support the required options, skip this part
        # of the test.
        pass
```
We need to be careful here on macOS as the default versions of `bash` and `stat` that it ships with do not support `shopt` and the `-c` flag respectively. To avoid potential new contributors being put off by tests failing because of this, I think it best to skip this part of the test if the `subprocess.run` call fails.

ianthomas23 (changes requested):
Thanks, this is both excellent and comprehensive. It fixes a number of corner cases that we hadn't managed to address yet, and I suspect we'll be able to close quite a few open issues afterwards.

I've requested a few changes, then I would be happy to approve it (after reverting the workflow change).

ianthomas23:
The `mypy` error is because we don't have quite the correct exclusion set up in `setup.cfg`. If you change the final line from
```python
exclude = (test_.*|conftest)\.py$
```
to
```python
exclude = (test.*|conftest)\.py$
```
that test should pass.

fsspec/tests/test_spec.py
john-jam: You're right, I split the posix tests between glob and bash tests and allow skipping bash tests if one of the following is true:
- the tests run on windows
- bash is not found
- globstar option is not found

BTW, those tests were useful for me while developing, to understand how the double asterisk works in bash and python glob. But we could also remove it completely since both of them don't actually test the `fsspec` library behavior.

fsspec/tests/conftest.py
john-jam: I added the "argnames" directly inside this object alongside the `argvalues" (both argument names of the `pytest.mark.parameterize` method) instead so it is clearer what the elements are.

fsspec/tests/abstract/__init__.py
john-jam: You're totally right. I reverted my changes of removing the `test_*_glob_to_*_directory` old tests and the `glob: bool` option here. This way it is clearer that globs supports the scenario in the documentation. For the remaining edge cases glob tests, I moved them into a new `test_*_glob_edge_cases` and add a new fixture to create the files for those edge cases.

In the process, I realized that the following assertion was wrong and fixed the files/folder removal in some tests with for loops:
```python
# wrong assertion since it test ` fs.ls(target) == []` in one case and `[dummy]` in the other which is always True
assert fs.ls(target) == [] if supports_empty_directories else [dummy]

# replaced by the following which test ` fs.ls(target) == []` in one case and `fs.ls(target) == [dummy]` in the other
assert fs.ls(target) == ([] if supports_empty_directories else [dummy])
```

john-jam:
> I've requested a few changes, then I would be happy to approve it (after reverting the workflow change).

@ianthomas23 Thanks for your review! I let you resolve the discussions if my updates work for you and trigger a new CI run to check if everything is ok now. Then, I'll revert my workflow change to merge this PR. Regarding the `gcsfs` and `s3fs` repos, I'll mark those PRs soon as ready for review. Should I keep my derived tests in `gcsfs` or do you want to merge your branch instead?

.github/workflows/main.yaml
john-jam: Note: reverted

fsspec/tests/abstract/__init__.py
ianthomas23: Thanks.

ianthomas23 (approved):
This is good to merge now as far as I am concerned. I'll give @martindurant the final say.

When this is merged we can re-run the CI of the `s3fs` and `gcsfs` PRs, and I will review them properly. They look fine at initial glance. Ideally we would want to merge both of those quite quickly after this so that everything is consistent before the next releases.

I've closed my `gcsfs` abstract tests PR fsspec/gcsfs#535 as fsspec/gcsfs#572 is definitely a better solution.

john-jam:
@ianthomas23 I changed the posix tests with python glob and bash glob to use relative paths instead of absolute paths.
This allowed me to detect an incoherence with the `**` use case. `fsspec` was returning an empty string `""` for the current directory but the python glob and bash glob, both don't return an empty string `""` for the current dir when using relative paths.

When using a prefix though (e.g. `dir/**`), the three implementations return the root folder (e.g. `dir`).

I fixed it in [this commit](https://github.com/fsspec/filesystem_spec/pull/1329/commits/85b7ae0596a29fdb052311db2f2e09a18cb352f6#diff-9e28a72564624d8305a58b77c04c514d15e016b61899fd507b52b30388ef6ab6R815).

@martindurant Please let me know if there's any work still to be done in this PR for it to be merged soon?

john-jam:
@ianthomas23 @martindurant I found another edge case with `gcsfs`. I created two PRs ([here](https://github.com/john-jam/filesystem_spec/pull/1) and [here](https://github.com/john-jam/gcsfs/pull/1)) but need to merge this PR before. I'll change the PR targets accordingly once this one is merged.

martindurant:
Merging now, and then we can deal with the small leftovers.

john-jam:
@martindurant @ianthomas23 Thanks for the merge of this PR and the ones in `s3fs` and `gcsfs`!

Do you know approx. when will you release a new version of those 3 repos? And also if you plan to include [those changes](https://github.com/fsspec/gcsfs/pull/576) in `gcsfs`?

martindurant:
Releases should be out this week, so long as lingering PRs are cleaned up.

Issue #1332, opened 2023-08-15
Exclusive file creation mode="x" does not work well with `get_fs_token_paths`


GenevieveBuckley:
fsspec does not correctly expand paths when file mode "x" is enabled. In these cases, we'd expect the same result as when file mode "w" was used.

Here is a simplified minimal, verifiable example:
```python
In [1]: from fsspec import get_fs_token_paths

In [2]: get_fs_token_paths("filepath", "w", num=2, expand=True)
Out[2]:
(<fsspec.implementations.local.LocalFileSystem at 0x11141ff10>,
 '2f8b33add9d85ea4bd7b108ef4ef68f6',
 ['/Users/genevieb/Documents/GitHub/filepath/0.part',
  '/Users/genevieb/Documents/GitHub/filepath/1.part'])

In [3]: get_fs_token_paths("filepath", "x", num=2, expand=True)
Out[3]:
(<fsspec.implementations.local.LocalFileSystem at 0x11141ff10>,
 '2f8b33add9d85ea4bd7b108ef4ef68f6',
 ['/Users/genevieb/Documents/GitHub/filepath'])
```

This was originally raised in this dask issue https://github.com/dask/dask/issues/10442, copied here for context.
> **Case 1: Using `mode="x"`**
> 
> ```python
> def test_to_csv_with_exclusive_mode():
>     df0 = pd.DataFrame(
>         {"x": ["a", "b", "c", "d"], "y": [1, 2, 3, 4]}
>     )
>     df = dd.from_pandas(df0, npartitions=2)
>     with tmpdir() as dir:
>         df.to_csv(dir, index=False, mode="x")
>         result = dd.read_csv(os.path.join(dir, "*")).compute()
>     assert_eq(result, df0, check_index=False)
> ```
> 
> raises `FileExistsError`.

<details><summary>Full Traceback</summary>
<p>

```
_________________ test_to_csv_with_exclusive_mode _________________

    def test_to_csv_with_single_file_and_exclusive_mode():
        df0 = pd.DataFrame(
            {"x": ["a", "b", "c", "d"], "y": [1, 2, 3, 4]}
        )
        df = dd.from_pandas(df0, npartitions=2)
        with tmpdir() as dir:
            csv_path = os.path.join(dir, "test.csv")
>           df.to_csv(dir, index=False, mode="x")#, single_file=True)

dask/dataframe/io/tests/test_csv.py:1550: 
[...]

PR #1333, opened 2023-08-15
Allow file mode="x" with get_fs_token_paths

GenevieveBuckley:
Closes https://github.com/fsspec/filesystem_spec/issues/1332
Related: https://github.com/dask/dask/issues/10442

fsspec does not always behave correctly when file mode="x" is used. This PR addresses one particular instance that is causing flow-on problems for dask.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -635,6 +635,8 @@ def get_fs_token_paths(
     else:
         if "w" in mode and expand:
             paths = _expand_paths(paths, name_function, num)
+        elif "x" in mode and expand:
+            paths = _expand_paths(paths, name_function, num)
         elif "*" in paths:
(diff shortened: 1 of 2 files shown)


martindurant:
I am fine with this, but ... mode x isn't actually enforced on remote filesystems. On local, the mode is passed on to the OS so you get the expected behaviour. Is this a problem?

GenevieveBuckley:
It's a problem for dask users using the default local cluster. Here is Hendrik explaining how he can't save a dataframe to csv using mode="x", where mode="w" works well: https://github.com/dask/dask/issues/10442

martindurant:
Yes, but my point is that this will probably fail anyway on, for example, gcs. Either way, expand should do the right thing, so this change is worthwhile.

PR #1336, opened 2023-08-19
Call cat_ranges in blockcache for async filesystems

monken:
The CachingFileSystem can benefit from calling `cat_ranges` for asynchronous filesystems to concurrently download blocks. This speeds up filling the cache significantly for filesystems like HTTP and S3FS.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -37,6 +37,7 @@ T = TypeVar("T")
 logger = logging.getLogger("fsspec")
 
 Fetcher = Callable[[int, int], bytes]  # Maps (start, end) to bytes
+MultiFetcher = Callable[list[[int, int]], bytes]  # Maps [(start, end)] to bytes
 
 
 class BaseCache:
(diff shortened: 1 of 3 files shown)


fsspec/caching.py
```diff
@@ -93,16 +114,30 @@ def _fetch(self, start, end):
...
             send = min(sstart + self.blocksize, self.size)
-            logger.debug(f"MMap get block #{i} ({sstart}-{send}")
-            self.cache[sstart:send] = self.fetcher(sstart, send)
+            ranges.append((sstart, send))
             self.blocks.add(i)
 
+        if not ranges:
+            return self.cache[start:end]
+
+        if self.multi_fetcher:
```
martindurant, line 130: Since .cat_ranges always exists and falls back to getting each block serially, should we just always call that?

monken: In `caching.py`, I wanted to maintain backwards compatibility. If someone uses the `MMapCache` class directly, their code would break if we require the `multi_fetcher` style interface.

martindurant: OK, but let's agree to eventually deprecate this

fsspec/implementations/cached.py
```diff
@@ -400,7 +400,21 @@ def _open(
...
+        def _fetch_ranges(ranges):
+            return self.fs.cat_ranges(
+                [path] * len(ranges),
+                [r[0] for r in ranges],
+                [r[1] for r in ranges],
+                **kwargs,
+            )
+
+        multi_fetcher = (
+            None if not self.fs.async_impl or self.compression else _fetch_ranges
```
martindurant, line 413: It might still be useful for a sync filesystem, since cat_ranges might combine some adjacent reads.

martindurant:
Sorry I let this one slide - I still think it's useful. I have just the one comment above.

martindurant:
Can you please solve the conflict, and I can have another look?

monken:
@martindurant sorry that I abandoned this for so long. I fixed the merge conflicts and updated the tests. Please let me know if you still seem this valuable.

PR #1344, opened 2023-08-28
Add lakefs-spec as known implementation

AdrianoKF:
The [lakefs-spec](https://github.com/appliedAI-Initiative/lakefs-spec) package provides access to lakeFS data lakes through an fsspec filesystem.

This commit adds a link to the list of known implementations in the API docs and in `fsspec.registry.known_implementations`.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -208,7 +208,8 @@ Other Known Implementations
 - `dvc`_ to access DVC/Git repository as a filesystem
 - `gcsfs`_ for Google Cloud Storage
 - `gdrive`_ to access Google Drive and shares (experimental)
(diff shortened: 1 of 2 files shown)


martindurant:
I am happy to include this. However, are you sure you want the protocol name "lakefs" rather than just "lake"?

AdrianoKF:
Awesome, thanks for the quick response! As a matter of fact, lakeFS is internally using `lakefs://` URIs already to identify files within a given repository, so it makes sense to keep the protocol name for interoperability reasons.

Side question: `setup.py` has a bunch of extra dependencies defined, but I assume those are for built-in protocols, right? If not, I am happy to add `lakefs` as another extra there.

martindurant:
> lakeFS is internally using lakefs:// 

OK

> I assume those are for built-in protocols

Correct.

PR #1345, opened 2023-08-28
Working async rsync

martindurant:
Fixes async-async and sync-async copying for generic/rsync by using local files to buffer whole remote files.

If one of the filesystems actually *is* the local filesystem, this would mean duplication of space and operations for now :|


martindurant:
Tested manually...

PR #1349, opened 2023-09-02
Avoid expanding paths when source and destination are lists

### LINKED ISSUE #1347: Sorted paths in expand_path produces bad mapping between source and destination files
When using lists as source and destination, the mapping between the two during copy/get/put operations is not accurate.

For example:
```python
from fsspec.implementations.memory import MemoryFileSystem
from hashlib import md5

def get_hash(value: str) -> str:
    return md5(value.encode("utf-8")).hexdigest()

fs = MemoryFileSystem()

paths_from = []
paths_to = []
for i in range(10):
    hashed_i = get_hash(value=str(i))
    path_from = f"in/{hashed_i}.txt"
    path_to = f"out/{hashed_i}.txt"
    fs.pipe(path=path_from, value=f"{i}".encode("utf-8"))
    paths_from.append(path_from)
    paths_to.append(path_to)

fs.copy(path1=paths_from, path2=paths_to)

for i in range(10):
    hashed_i = get_hash(value=str(i))
    file_content = fs.cat(f"out/{hashed_i}.txt").decode("utf-8")
    if file_content != str(i):
        print(f"file content {file_content} | expected {str(i)}")
```

This code outputs:
```bash
file content 6 | expected 0
file content 9 | expected 1
file content 7 | expected 2
file content 4 | expected 3
file content 1 | expected 4
file content 2 | expected 5
file content 8 | expected 6
file content 0 | expected 7
file content 5 | expected 8
file content 3 | expected 9
```

This is because the `path1` list gets sorted in the `expand_path` method but the `path2` remains the original one. So we can't expect an accurate mapping when calling copy/get/put operations like this: 
```python
fs.copy(
    path1=["in/myfile1", "in/myfile2"],
    path2=["out/myfile1", "out/myfile2"],
)
```

If it's an expected behavior, please let me know how to do batch copy when we want to control the destination path of each file.
If not, I think I can create a PR this week end.

john-jam:
As described in #1347, this PR makes sure that the order between the source and destination are kept when lists are provided.

Closes #1347

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -346,7 +346,7 @@ class AsyncFileSystem(AbstractFileSystem):
 
         if isinstance(path1, list) and isinstance(path2, list):
             # No need to expand paths when both source and destination
-            # are provided as string
+            # are provided as lists
             paths1 = path1
             paths2 = path2
         else:
@@ -516,7 +516,7 @@ class AsyncFileSystem(AbstractFileSystem):
         """
         if isinstance(lpath, list) and isinstance(rpath, list):
             # No need to expand paths when both source and destination
-            # are provided as string
+            # are provided as lists
             rpaths = rpath
             lpaths = lpath
         else:
@@ -593,7 +593,7 @@ class AsyncFileSystem(AbstractFileSystem):
         """
         if isinstance(lpath, list) and isinstance(rpath, list):
             # No need to expand paths when both source and destination
-            # are provided as string
+            # are provided as lists
             rpaths = rpath
             lpaths = lpath
         else:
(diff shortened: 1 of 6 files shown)


martindurant:
Thank you!

john-jam:
@martindurant thanks for the fsspec release!
Any chance that `gcsfs` and `s3fs` will be also released?

martindurant:
I have been waiting for aiobotocore 2.6.0 on conda-forge ( https://github.com/conda-forge/aiobotocore-feedstock/issues/68 ) , but I'll release with 2.5.4 later today if there is no movement.

PR #1352, opened 2023-09-04
Apply ruff C4 autofixes for perf improvement

Skylion007:
Apply C4 autofixes from ruff / flake8-comprehensions. Replaces various comprehensions with equivalent ones that are more efficient and removes unnecessary double casts to lists etc. This should yield minor perf improvements and make it more pythonic. If you want to have these fixes automated enabled for future PRs consider adding the ruff pre-commit hook.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -833,7 +833,7 @@ class AsyncFileSystem(AbstractFileSystem):
 
     async def _find(self, path, maxdepth=None, withdirs=False, **kwargs):
         path = self._strip_protocol(path)
-        out = dict()
+        out = {}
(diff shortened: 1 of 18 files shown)


martindurant:
Some of these changes are fine, but some are not. The contract of certain methods specify to return a `list`, not an iterable, so `list(sorted(...))` is a reasonable thing unless we can ensure absolutely that the only ever use of it is in an iteration and not passed to the user. I don't believe the majority of the cases here meet that.

martindurant:
Hah, quite right you are, thanks. I suppose the sort algorithm must be working internally on a list in-place - makes sense.

PR #1353, opened 2023-09-07
Save cache metadata in JSON format

### LINKED ISSUE #825: Memory leak in `CachingFileSystem` (possibly caused by `pickle.load` ?)
I've been using `fsspec` caching support:

```python
urlpath = 'gs://mybucket/image.jpg'
fsspec.open(f'filecache::{urlpath}', filecache={'cache_storage': '/tmp/files', 'expiry_time': 604800})
```
I noticed that after opening many files (10k +) my application's memory consumption goes to the roof - only a restart causes memory deallocation.
After some profiling with `tracemalloc` these are the 2 top consumers:

```
#1: /usr/local/lib/python3.8/site-packages/fsspec/implementations/cached.py:156: 1443275.5 KiB
cached_files = pickle.load(f) "
#2: /usr/local/lib/python3.8/site-packages/fsspec/implementations/cached.py:137: 700547.5 KiB
loaded_cached_files = pickle.load(f)
```

If `filecache` is not used, memory consumption is back to normal.
I don't know much about the `pickle` module. Could it be causing the memory leak ?

ianthomas23:
This changes the file format used for storing cached metadata from pickle to JSON. It eliminates the problem that pickled files written using one version of Python might not be readable using a different version of python, and an extra benefit is that such files are now human readable so easier to understand/debug.

All loading and saving of cache metadata is now performed in specific low-level functions `CacheMetadata._load` and `CacheMetadata._save`.

We still support the reading of cached metadata from pickled files so that pre-existing caches should still be usable. Such files will be saved as JSON the next time they are saved.

The implementation uses `ujson` if it is installed, falling back to `json`. This is the same approach used in `ReferenceFileSystem`.

The only slight complexity here is that we have to test both JSON and pickle formats in CI, so I have added a private boolean attribute `CacheMetadata._force_save_pickle`. This should only be used in tests, not by end users. If set to `True` subsequent saves are in pickle format so we can simulate old pickled metadata to confirm it can still be read and is converted to JSON on next save.

Fixes #825.

### DIFF
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -41,6 +41,16 @@ class CacheMetadata:
         self._storage = storage
         self.cached_files: list[Detail] = [{}]
 
+    def _load(self, fn: str) -> Detail:
+        """Low-level function to load metadata from specific file"""
+        with open(fn, "rb") as f:
+            return pickle.load(f)
+
+    def _save(self, metadata_to_save: Detail, fn: str) -> None:
+        """Low-level function to save metadata to specific file"""
+        with atomic_write(fn) as f:
+            pickle.dump(metadata_to_save, f)
+
     def _scan_locations(
         self, writable_only: bool = False
     ) -> Iterator[tuple[str, str, bool]]:
@@ -111,8 +121,7 @@ class CacheMetadata:
 
         if self.cached_files[-1]:
             cache_path = os.path.join(self._storage[-1], "cache")
-            with atomic_write(cache_path) as fc:
-                pickle.dump(self.cached_files[-1], fc)
+            self._save(self.cached_files[-1], cache_path)
 
         writable_cache_empty = not self.cached_files[-1]
         return expired_files, writable_cache_empty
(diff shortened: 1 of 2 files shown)


fsspec/implementations/cache_metadata.py
```diff
@@ -41,6 +49,28 @@ def __init__(self, storage: list[str]):
...
+        # Private attribute to force saving of metadata in pickle format rather than
+        # JSON for use in tests to confirm can read both pickle and JSON formats.
+        self._force_save_pickle = False
+
+    def _load(self, fn: str) -> Detail:
+        """Low-level function to load metadata from specific file"""
+        try:
+            with open(fn, "r") as f:
+                return json.load(f)
+        except Exception:
```
Skylion007, line 61: Exception should be way more specific.

martindurant: IOError or JSON decode error. The latter will depend on the JSON library used, I don't think they nicely call it the same thing - but probably both subclass ValueError.

ianthomas23: The `Exception` was purposely generic. If anything fails with the JSON load it tried the pickle load. A corner case failure in the first will be repeated in the second.

It is not as simple as IOError or JSON decode errror. It is easy to obtain a UnicodeDecodeError for example, and I expect there are others that I haven't explicitly tested for.

But I don't have a strong opinion either way. I have changed it to a `ValueError` as suggested.

Issue #1359, opened 2023-09-12
fsspec.gui.FileSelector, doesn't work with remote protocols? 


okz:
FileSelector implementation suggests remote protocol support is implemented, but I haven't been able to figure out how to use it? 

Maybe the implementation has a bug with storage options? 

```
fsspec.gui.FileSelector("az://.", **so)
>> TypeError: FileSelector.__init__() got an unexpected keyword argument 'account_name'

fsspec.gui.FileSelector("file://.", storage_options=so)
>>
TypeError: FileSelector.__init__() got an unexpected keyword argument 'storage_options'

```

martindurant:
I see the documentation on this is somewhat lacking. The GUI allows you to supply the initial value of the "kwargs" text widget, which must be a string, but evaluate to a python dict. So something like
```
fsspec.gui.FileSelector("az://.", kwargs="""{"account_name": "myname"}""")
```

okz:
Ah thanks Martin, I understand why but definitely not obvious. 

However there is a bug after all:

The text widget is initialised with an empty dictionary, and since `storage_options` use the widget value then onwards,  an exception is raised immediately. 

Fixed the initialisation here: 
```
        --self.kwargs = pn.widgets.TextInput(name="kwargs", value="{}", align="center")
      ++self.kwargs = pn.widgets.TextInput(name="kwargs", value=self.init_kwargs, align="center")
```





storage_options, was

martindurant:
That does seem right. If you make a PR, I will merge it. This can also be tested easily.

ianthomas23:
This was fixed by #1360 so closing as completed.

PR #1360, opened 2023-09-12
kwargs widget should be initialized with parameter

okz:
fixes 
[fsspec.gui.FileSelector, doesn't work with remote protocols?](https://github.com/fsspec/filesystem_spec/issues/1359)

### DIFF
--- a/fsspec/gui.py
+++ b/fsspec/gui.py
@@ -262,7 +262,9 @@ class FileSelector(SigSlot):
             name="protocol",
             align="center",
         )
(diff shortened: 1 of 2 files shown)


martindurant:
Would you like to write a test for this? It would involve making an instance with a value, and then asserting that the GUI's filesystem instance attribute has the right option set. A good candidate would be '{"auto_mkdir": True}' with the "file" protocol, which sets the `auto_mkdir` attribute on the instance (and is False by default).

martindurant:
(please runt he precommit hooks for linting - you are getting a little formatting edits and trailing whitespace)

PR #1363, opened 2023-09-18
fix: register_implementation() does not work as expected when clobber=True

moriyoshi:
This is obviously a bug.


moriyoshi:
No, it was not. But the extra test should help.

martindurant:
I am happy to accept a PR that only extends a test

PR #1365, opened 2023-09-18
Revert auto_mkdir default in open() and special case local

martindurant:
Fixes https://github.com/fsspec/s3fs/issues/790
Fixes https://github.com/dask/dask/issues/10515

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -210,6 +210,7 @@ def open_files(
     num=1,
     protocol=None,
     newline=None,
+    auto_mkdir=True,
     expand=True,
     **kwargs,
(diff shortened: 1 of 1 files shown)


fsspec/core.py
```diff
@@ -284,6 +288,11 @@ def open_files(
         protocol=protocol,
         expand=expand,
     )
+    if fs.protocol == "file":
+        fs.auto_mkdir = auto_mkdir
```
martindurant, line 292: I do not see an easier way to do this than this fragile line, because the creation of the fs is hidden within the unchaining logic inside the call above.

PR #1366, opened 2023-09-19
Use setup-micromamba github action

ianthomas23:
The github action `mamba-org/provision-with-micromamba` is deprecated and the recommendation is to use `mamba-org/setup-micromamba` instead (https://github.com/mamba-org/provision-with-micromamba#migration-to-setup-micromamba). This PR follows that recommendation, and also adds a `python-version` to the `lint` github action to suppress another warning.


ianthomas23:
This has removed all the CI warnings.

martindurant:
+1

PR #1370, opened 2023-09-21
In Python 3, classes inherit object implicitly

DimitriPapadopoulos:
No need to inherit (object) explicitly.


DimitriPapadopoulos:
All examples in the Python 3 documentation omit the `object` inheritance.

Also _pyupgrade_ removes `(object)`.

PR #1374, opened 2023-09-21
Change legacy string formatting to f-strings

DimitriPapadopoulos:


martindurant:
Sorry, it will take me a little time to check all of these little changes.
Was this automatically generated?

fsspec/implementations/jupyter.py
```diff
@@ -83,7 +83,7 @@ def pipe_file(self, path, value, **_):
             "format": "base64",
             "type": "file",
         }
-        self.session.put(self.url + "/" + path, json=json)
+        self.session.put(self.url + f"/{path}", json=json)
```
martindurant, line 86:
```suggestion
        self.session.put(f"{self.url}/{path}", json=json)
```

fsspec/implementations/jupyter.py
```diff
@@ -96,11 +96,11 @@ def mkdir(self, path, create_parents=True, **kwargs):
             "content": None,
             "type": "directory",
         }
-        self.session.put(self.url + "/" + path, json=json)
+        self.session.put(self.url + f"/{path}", json=json)
```
martindurant, line 99:
```suggestion
        self.session.put(f"{self.url}/{path}", json=json)
```

fsspec/implementations/jupyter.py
```diff
@@ -96,11 +96,11 @@ def mkdir(self, path, create_parents=True, **kwargs):
...
 
     def _rm(self, path):
         path = self._strip_protocol(path)
-        self.session.delete(self.url + "/" + path)
+        self.session.delete(self.url + f"/{path}")
```
martindurant, line 103:
```suggestion
        self.session.delete(f"{self.url}/{path}")
```

fsspec/implementations/local.py
```diff
@@ -240,7 +240,7 @@ def make_path_posix(path, sep=os.sep):
             return path
         if path.startswith("./"):
             path = path[2:]
-        return os.getcwd() + "/" + path
+        return os.getcwd() + f"/{path}"
```
martindurant, line 243:
```suggestion
        return f"{os.getcwd()}/{path}"
```

fsspec/implementations/local.py
```diff
@@ -251,7 +251,7 @@ def make_path_posix(path, sep=os.sep):
             # abspath made some more '\\' separators
             return make_path_posix(osp.abspath(path))
         else:
-            return os.getcwd() + "/" + path
+            return os.getcwd() + f"/{path}"
```
martindurant, line 254:
```suggestion
            return f"{os.getcwd()}/{path}"
```

fsspec/implementations/sftp.py
```diff
@@ -77,18 +77,18 @@ def mkdir(self, path, create_parents=False, mode=511):
...
 
         for part in parts:
-            path += "/" + part
+            path += f"/{part}"
             if not self.exists(path):
                 self.ftp.mkdir(path, mode)
 
     def rmdir(self, path):
-        logger.debug("Removing folder %s" % path)
+        logger.debug(f"Removing folder {path}")
```
martindurant, line 91: In fact, for logging it it recommended to use % notation and a list of string arguments (NOT % format) so that the string doesn't need to be evaluated at all if it's not to be rendered.
```suggestion
        logger.debug(f"Removing folder %s", path)
```

Where the argument is a string as opposed to something harder to evaluate, it makes negligible difference...

martindurant: There are a few other cases of this.

martindurant (commented):
I am more than halfway through, just a few very small things so far,

martindurant (approved):

martindurant:
That was weird, had to rerun many test runs before they passed. Perhaps dockerhub was having a bad day...

This is going in now.

Issue #510, opened 2020-12-22
feature: size limit on caching filesystem


majidaldo:
It would be nice to set a limit on the size of the caching filesystems. This would truly enable 'set and forget' kind of work instead of having to manually go into the cache location to clear things out.

martindurant:
I can imagine such a feature, and wouldn't be too hard to implement. You would clear old files by timestamp, oldest first. On simplecache, it would mean having to list files occasionally; on other ones to keep the metadata up to date.

JulioHC00:
Hi, is there any solution available now for limiting the cache size?

RoelantStegmann:
I agree that this would be very nice. I would like to add that in simple caching (so not block caching), it should also not even attempt to cache a file that is in itself larger than the cache size.

PR #1377, opened 2023-09-27
Implement cache size for CachingFileSystem

ianthomas23:
This PR implements retrieving the cache size of `CachingFileSystem`. The latest size is stored as a private attribute which is invalidated on any write to the cache and only recalculated if necessary when `cache_size()` is called. If multiple storage locations are specified it only considers the last one, i.e. the writable one.

This is a necessary precursor to providing a cache size limit (#510).

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -128,6 +128,11 @@ class CachingFileSystem(AbstractFileSystem):
         self.expiry = expiry_time
         self.compression = compression
 
+        # Size of cache in bytes. If None then the size is unknown and will be
+        # recalculated the next time cache_size() is called. On writes to the
+        # cache this is reset to None.
+        self._cache_size = None
+
         if same_names is not None and cache_mapper is not None:
             raise ValueError(
                 "Cannot specify both same_names and cache_mapper in "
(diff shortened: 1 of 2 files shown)


fsspec/implementations/cached.py
```diff
@@ -202,6 +219,7 @@ def clear_cache(self):
         """
         rmtree(self.storage[-1])
         self.load_cache()
+        self._cache_size = None
```
martindurant, line 222: Should be 0 ?

fsspec/implementations/cached.py
```diff
@@ -684,7 +707,7 @@ def _check_file(self, path):
                 return fn
 
     def save_cache(self):
-        pass
+        self._cache_size = None
```
martindurant, line 710: Because of the size of the metadata file?

fsspec/implementations/cached.py
ianthomas23: Unfortunately not, as the `self.load_cache()` call in the line above creates the cache directory which has OS-specific non-zero size.

fsspec/implementations/cached.py
ianthomas23: Actually this isn't necessary, the tests pass without it.

ianthomas23: I have removed this now.

ianthomas23:
I've temporarily removed one of the checks for `filecache` on Windows as when you delete a cached file the parent directory gets larger not smaller 🤷 

The remaining test that is failing on Windows is a `blockcache` one that does not allow deletion of a file that it claims is still in use. I don't yet know if this is something wrong with Windows or fsspec.

martindurant:
> does not allow deletion of a file that it claims is still in use. I don't yet know if this is something wrong with Windows or fsspec.

Windows does like to keep locks on files...

ianthomas23:
If I remove the problematic asserts related to file locking for `blockcache` on Windows then there isn't really anything left worth testing, so I've just skipped that test on Windows. This isn't ideal, but I would rather have the functionality available on posix systems than not at all.

martindurant:
> This isn't ideal, but I would rather have the functionality available on posix systems than not at all.

Agreed - but I expect it works fine in practice, and at most results in warnings when closing the interpreter. "Sparse" files are not a thing on windows anyway, so you only get part of the benefit of blockcache.

PR #1378, opened 2023-09-27
Adds LazyReferenceMapper to api.rst

norlandrhagen:


docs/source/api.rst
```diff
@@ -124,6 +124,7 @@ Built-in Implementations
    fsspec.implementations.local.LocalFileSystem
    fsspec.implementations.memory.MemoryFileSystem
    fsspec.implementations.reference.ReferenceFileSystem
+   fsspec.implementations.reference.LazyReferenceMapper
```
martindurant, line 127: I suppose we should update the class's docs too, if you have the time and inclination

norlandrhagen: Happy to! It looks like `categorical_threshold` is missing a docstring. What else would you like to add?

image

martindurant: The basic description of what the class is for; it doesn't otherwise really belong in the list of "implementations" (but we have nowhere better to put it). 

>     """Interface to read parquet store as if it were a standard kerchunk
>     references dict."""

martindurant:
ping

docs/source/api.rst
norlandrhagen: Sounds good! Just to clarify. 

1. Should `LazyReferenceMapper` be added to the [Base Classes](https://filesystem-spec.readthedocs.io/en/latest/api.html#base-classes) API docs?
2. Add/expand the description for `LazyReferenceMapper`

martindurant: 1. I am not sure where the doc fits - certainly not Base Classes. It could stay here as a companion to ReferenceFS, or maybe in the Utilities section?

2. Yes please, as much as you can. This makes 1. less important, as people won't accidentally try to use it as a FS by itself if it clearly says what it's for.

martindurant: Same meaning as in kerchunk.df.refs_to_dataframe

```
    categorical_threshold : int
        Encode urls as pandas.Categorical to reduce memory footprint if the ratio
        of the number of unique urls to total number of refs for each variable
        is greater than or equal to this number. (default 10)
```

PR #1382, opened 2023-09-29
Make `glob` consistent with `glob.glob`

### LINKED ISSUE #1380: `.glob("**/filename")` returns incorrect results
Reproducer (requires **`fsspec>=2023.9.0`**):
```python
>>> import fsspec
>>> print(fsspec.filesystem("github", org="gpanders", repo="vim-medieval").glob("**/eval[-._ 0-9]*"))
['autoload/medieval.vim', 'doc/medieval.txt', 'ftplugin/markdown/medieval.vim']
```
`**/` is converted to `.*` in the `AbstactFileSystem.glob`, so the result contains filenames that don't start with `eval`, making it incorrect.

This bug can only be reproduced on filesystems that don't resolve the glob path internally, as prefixing the path with `<base_path>/` leads to `/**` taking [precedence](https://github.com/fsspec/filesystem_spec/blob/89626e89022dc9ed62aa7b0fb8dba9427fa892b3/fsspec/spec.py#L626) over `**/` in the "glob to regex" conversion, which stops `**/` from being replaced by `.*`.

Maybe the most robust solution would be to align the implementation with `glob.glob` (behaves as expected in this instance) to stop such inconsistencies from happening in the future. The non-posix behavior is error-prone, so dropping it should be fine, no? I'm happy to submit a PR if this sounds good to you (based on https://github.com/python/cpython/pull/106703).

mariosasko:
Makes `glob(path)` consistent with `glob.glob(path, recursive=True, include_hidden=True)`.

This means `**` would only be allowed between the path separators. Although breaking, I think this change is good as the current handling of `**` is somewhat unexpected (and not particularly useful?).

Also fixes #1380.


mariosasko:
Should be easy to fix the failing tests:
* `gcsfs`: replace the line `assert fn in gcs.glob(TEST_BUCKET + "/**1")` in `test_gcs_glob` with `assert fn in gcs.glob(TEST_BUCKET + "/**/*1")` 
* downstream (`dask`) : replace the line `root = "http://localhost:8999/"` in `test_open_glob` with `root = "http://localhost:8999"`

(I can open PRs there after this one gets approved.)

mariosasko:
> So I understand, is this the only change in the behaviour of glob that we expect from your change?!

Yes, that's correct.

martindurant:
@mariosasko , do you have an opinion on https://github.com/fsspec/filesystem_spec/issues/1394 ?

mariosasko:
Commented on the issue :)

martindurant:
Due to the test breakages downstream, I will release before merging, and then merge maybe on Monday, so that users won't see this until we've tidied up the other libraries too.

mariosasko:
Sounds good!

martindurant:
Will you be available to make the PR to dask? I will merge when you have time to do it, so that we don't cause their CI to break for too long. gcsfs is less urgent, and I can do that anyway.

mariosasko:
PR opened in dask :)

martindurant:
Do you intend to fix gcsfs, or will I need to deal with it?

mariosasko:
I linked a `gcsfs` PR with the fix.

LZHgrla:
@mariosasko 
Hi! Great work!

I think this PR has been included in the latest release (2023.12.0)!
Maybe we can delete the pin version of `datasets`?

efiop:
For the record: turns out this also broke adlfs but we didn't have daily tests to notice (which were also broken for another reason 🤦‍♂️ ), so I've added those and fixed the tests.

martindurant:
> turns out this also broke adlfs

Do you mean that things are now in a good place?

efiop:
@martindurant Yes, all good now, no action needed. Here's the main PR fixing that https://github.com/fsspec/adlfs/pull/448 and a few more (https://github.com/fsspec/adlfs/pull/447, https://github.com/fsspec/adlfs/pull/446, https://github.com/fsspec/adlfs/pull/445) related to the other reasons for adlfs tests failing.

PR #1386, opened 2023-10-04
Improve some parentheses

DimitriPapadopoulos:


fsspec/utils.py
```diff
@@ -507,14 +507,14 @@ def merge_offset_ranges(paths, starts, ends, max_gap=0, max_block=None, sort=Tru
     starts = [s or 0 for s in starts]
     # Sort by paths and then ranges if `sort=True`
     if sort:
-        paths, starts, ends = [
+        paths, starts, ends = (
```
martindurant, line 510: A few lines below, we use slicing and len(paths), so I'm not sure this one works. 

This might mean this code is not covered?

DimitriPapadopoulos: I haven't changed `paths` into a tuple.

Here `paths` is just an item in the top-level tuple/list. This top-level tuple has no name/handle, it is immediately **unpacked** to `paths, starts, ends`. So nothing prevents `paths`, `starts`, or `ends` from being a list, or whatever type it actually is.

martindurant: Oops, OK

martindurant: I suppose we could have as easily written out three lines instead, then - but never mind that

PR #1387, opened 2023-10-04
`.super(C, self)` → `.super()`

DimitriPapadopoulos:
In Python 3, you generally don't need the arguments for `super()` any more – they are inserted automatically ([PEP 3135](https://www.python.org/dev/peps/pep-3135/)).


Skylion007:
This rule can be enabled codebase wide with a ruff pyupgrade check.

Issue #922, opened 2022-03-15
Entrypoint support for abfss in adlfs prevent this protocol to be overwritten by another package 


pierre-godard:
Hi!

Sorry if this issue is misplaced and should go to the adlfs package instead.
I already created an issue there, but after thinking of it, maybe it is more relevant to post it here.

You'll find all the details in this issue: fsspec/adlfs#302

I'm not sure if I should copy/paste the whole issue here or if the link is enough, let me know!

martindurant:
(the link is fine)

It wasn't really anticipated that there would be multiple implementations of a given protocol, although "ssh" has, and although Intake deals with a similar plugin problem. I think that in your case, you might consider including a call to `fsspec.registry.register_implementation` in your code and require users to import your package if they want the non-default backend.

For the more general issue:
- I don't know how entrypoints decides on the order of hits, but suspect it might just be the native filesystem order of directories. That means that the specific installation steps would matter, and no general rule like "first" or "last" will suffice.
- really, this sounds like a configuration thing. I think it would make sense to optionally add to known_implementations via the config, and that these take priority (and calls to register_implementation are highest priority).

KeerthiYandaOS:
@martindurant Looks like the implementation (register_implementation) invoked "last" is taking precedence over the previous one. 
As the entry point list is in below order:
```
[EntryPoint(name='abfs', value='mypackage.AzureBlobFileSystem', group='fsspec.specs'),
 EntryPoint(name='abfss', value='mypackage.AzureBlobFileSystem', group='fsspec.specs'),
 EntryPoint(name='abfss', value='adlfs.AzureBlobFileSystem', group='fsspec.specs')]
```
As the `adlfs.AzureBlobFileSystem` is getting registered at the end, the request for abfss protocol is reaching the adlfs.AzureBlobFileSystem implementation instead of the mypackage.AzureBlobFileSystem

 @pierre-godard were you able to find any alternate solution? it will be helpful if you can share your findings.

Unfortunately, I can't use the register_implementation as it will be lost once the package is re-installed in my scenario.

KeerthiYandaOS:
Created a pull request in fsspec to avoid the entry-point overriding: https://github.com/fsspec/filesystem_spec/pull/1389

PR #1389, opened 2023-10-10
Avoid the Entrypoint overriding if its already registered

KeerthiYandaOS:
When there are multiple entry points for the same protocol the latest entry point comes at the top of the list like below:
```
[EntryPoint(name='abfs', value='mypackage.AzureBlobFileSystem', group='fsspec.specs'),
 EntryPoint(name='abfss', value='mypackage.AzureBlobFileSystem', group='fsspec.specs'),
 EntryPoint(name='abfss', value='adlfs.AzureBlobFileSystem', group='fsspec.specs')]
```
But the process_entries registers the entry points in the order they are in the list, so instead of the top entry, the "last" entry is taking precedence over others. So adding logic to skip the override if a registry is already present for the protocol.

Resolves issue: https://github.com/fsspec/filesystem_spec/issues/922

### DIFF
--- a/fsspec/__init__.py
+++ b/fsspec/__init__.py
@@ -48,10 +48,15 @@ def process_entries():
                 specs = eps.select(group="fsspec.specs")
             else:
                 specs = eps.get("fsspec.specs", [])
+            registered_names = {}
             for spec in specs:
                 err_msg = f"Unable to load filesystem from {spec}"
+                name = spec.name
+                if name in registered_names:
+                    continue
+                registered_names[name] = True
                 register_implementation(
-                    spec.name,
+                    name,
                     spec.value.replace(":", "."),
                     errtxt=err_msg,
                     # We take our implementations as the ones to overload with if


martindurant:
How do you know whether the "first" or "last" is the correct entry to process? Does entrypoints even maintain a consistent ordering?

KeerthiYandaOS:
In the latest Python version (3.10) packages and entry points are discovered in the order they are installed. For the packages that consume or override the entry-point for abfss, I believe adlfs is the required package. So custom package that overrides the abfss protocol will be installed after the adlfs package. Which results in the order of entry points from the latest package installation on the top.

Yes, as entry-points is the dictionary, its order is preserved/guaranteed from Python 3.7v onwards: https://docs.python.org/3/whatsnew/3.7.html#pep-468-preserving-the-order-of-dict-items

KeerthiYandaOS:
`What about installations via requirements.txt or conda, do we know how they would get ordered?`
For requirements.txt, package managers like `pip` follow the dependency resolution mechanism to determine the order in which packages should be installed. This mechanism takes into account the package dependencies and ensures that packages are installed in a way that satisfies all dependencies. Similarly, conda also installs the packages based on the dependency tree and installs all the dependent packages before the actual packages.

In the current scenario, packages having default protocol implementations are the base and required packages for other custom packages that override the implementation. So, if there are multiple implementations for the same protocol, then the implementation received at the top of the list (from the package installed at the end) will be the user-defined implementation (in the custom package) which needs to be prioritized.

martindurant:
OK, I'm going to allow this in, in time for release.
A _better_ solution would be a config system which allows users to allow/disallow certain entrypoints, and we should implement that too.

PR #1396, opened 2023-10-20
Add typehints to `fsspec.mapping` and `fsspec.dircache`

alanhdu:
Continuation of https://github.com/fsspec/filesystem_spec/pull/1284, but add type hints to `fsspec.mapping` and `fsspec.dircache`.


fsspec/dircache.py
```diff
@@ -1,9 +1,17 @@
...
 
 
-class DirCache(MutableMapping):
+class DirEntry(TypedDict):
+    name: str
+    size: int
+    type: str
+
+
+class DirCache(MutableMapping[str, List[DirEntry]]):
```
alanhdu, line 14: This class doesn't ever specify the key and value types, so theoretically I could make this a generic type. The docstrings imply this is the right type though?

fsspec/dircache.py
```diff
@@ -1,9 +1,17 @@
...
+from typing import MutableMapping, List, TypedDict, Iterator, Any
+
 from functools import lru_cache
 
 
-class DirCache(MutableMapping):
+class DirEntry(TypedDict):
+    name: str
+    size: int
+    type: str
```
alanhdu, line 11: See https://peps.python.org/pep-0589/ for more details.

fsspec/dircache.py
```diff
@@ -45,36 +53,36 @@ def __init__(
...
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._cache)
 
-    def __contains__(self, item):
+    def __contains__(self, item: object) -> bool:
         try:
-            self[item]
+            self[item]  # type: ignore[index]
```
alanhdu, line 80: This is a little unfortunate -- the `MutableMappign` typehints say that `__contrains__` should take all `object`s, not just the key type, hence the type ignore here.

fsspec/mapping.py
```diff
@@ -61,28 +84,44 @@ def __init__(self, root, fs, check=False, create=False, missing_exceptions=None)
...
+
+    @overload
+    def getitems(
+        self, keys: Iterable[str], on_error: Literal["return"]
+    ) -> dict[str, bytes | Exception]:
+        pass
+
+    def getitems(
+        self,
+        keys: Iterable[str],
```
alanhdu, line 115: I could keep this as a `list` if we want to add that restriction.

fsspec/mapping.py
```diff
@@ -61,28 +84,44 @@ def __init__(self, root, fs, check=False, create=False, missing_exceptions=None)
...
+    def clear(self) -> None:
         """Remove all keys below root - empties out mapping"""
         try:
             self.fs.rm(self.root, True)
             self.fs.mkdir(self.root)
         except:  # noqa: E722
             pass
 
-    def getitems(self, keys, on_error="raise"):
+    @overload
```
alanhdu, line 101: This is a bit of ugliness to make the output type of `getitems` dynamic based on `on_error`

fsspec/dircache.py
martindurant: More fields are typically present, these are the required ones.

fsspec/mapping.py
martindurant: Can we not just use `str` and ignore the overload? I don't think this kind of construct is worth it.

martindurant (commented):
Almost everything is fine here.

fsspec/dircache.py
alanhdu: Yeah, I don't love it either.. it's not great.

fsspec/dircache.py
alanhdu: Hm.. I can just make it a `Dict[str, Any]` for now then? I think technically `TypedDict` supports having extra keys, but I've run into a bunch of weird limitations if you actually try to use it like that.

martindurant: Yes please, simple is good

fsspec/dircache.py
Skylion007: You can around this by adding an assert `isinstance` here.

PR #1401, opened 2023-10-24
Update docstring for `fsspec.asyn.sync_wrapper`

lobis:
When reading the documentation I found what I think is a typo. It should say **sync** instead of **async** as this method is used to wrap an async function so it can be called in a blocking context.


martindurant:
Actually, I think it's just out of date. There was previously an effort to have the same methods be callable from async or sync contexts, but this proved too hard to maintain. I would remote the "sync or" words now: "blocking" is probably the best description.

PR #1402, opened 2023-10-24
fix the incorrect list of ends for _cat_ranges

AlbertXingZhang:
I think the list of "ends" should come from the argument "ends" if it is not iterable for the function "_cat_ranges".

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -475,7 +475,7 @@ class AsyncFileSystem(AbstractFileSystem):
             A list of of filepaths on this filesystems
         starts, ends: int or list
             Bytes limits of the read. If using a single int, the same value will be
-            used to read all the specified files.    
+            used to read all the specified files.
         """
         # TODO: on_error
         if max_gap is not None:
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -467,6 +467,16 @@ class AsyncFileSystem(AbstractFileSystem):
         on_error="return",
         **kwargs,
     ):
+        """Get the contents of multiple files
+
+        Parameters
+        ----------
+        paths: list
+            A list of of filepaths on this filesystems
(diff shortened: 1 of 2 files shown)


martindurant:
(also in `fsspec.spec`)

martindurant:
Whilst you are right, the inputs are expected to be lists. This functions should have a docstring saying so, and tests making sure that it works. Obviously, this has been somewhat an experimental method, but it should be graduated to full use.

AlbertXingZhang:
Hi @martindurant, thanks for pointing out another place I missed. I also added it to the changes and wrote some docstrings for them. Please let me know if I need to make more changes or feel free to help revise them. Thanks.

fsspec/asyn.py
```diff
@@ -467,6 +467,16 @@ async def _cat_ranges(
         on_error="return",
         **kwargs,
     ):
+        """Get the contents of multiple files
```
martindurant, line 470:
```suggestion
        """Get the contents of byte ranges from one or more files
```

fsspec/spec.py
```diff
@@ -828,14 +828,24 @@ def pipe(self, path, value=None, **kwargs):
     def cat_ranges(
         self, paths, starts, ends, max_gap=None, on_error="return", **kwargs
     ):
+        """Get the contents of multiple files
```
martindurant, line 831:
```suggestion
        """Get the contents of byte ranges from one or more files
```

AlbertXingZhang:
Hi @martindurant, thanks for reviewing. I also trimmed the trailing whitespace complained by lint. I'm sorry I didn't notice it before.

martindurant:
Thank you @AlbertXingZhang

AlbertXingZhang:
Thanks for merging, @martindurant🎉

Issue #1410, opened 2023-10-31
Spurious var directory created


horta:
Create file `main.py`:

```python3
from fsspec.implementations.cached import WholeFileCacheFileSystem
from fsspec.implementations.zip import ZipFileSystem

fs = WholeFileCacheFileSystem(fs=ZipFileSystem("dir.zip"))
with fs.open("/dir/file.txt", "rb") as file:
    pass
```

And enter:

```sh
mkdir dir
touch dir/file.txt
zip -r dir.zip dir
rm -rf dir

pip install fsspec==2023.10.0

test -d var || echo "Not found var directory."
python main.py
test -d var && echo "Found var directory."
```

The above will print:

```
Not found var directory.
Found var directory.
```

under MacOS.

martindurant:
Can you please check the directory your filesystem things it should be writing to, `fs.storage`? Without any arguments, it should pick a temporary directory. Do you perhaps have some fsspec config set (`fsspec.config.conf`) ?

horta:
Thank you for the reply!

I get

```
❱ python main.py
['/var/folders/d0/jxsscwdj2csbjj2ntd928_w80000gn/T/tmp3bmq_f5f']
❱ python main.py
['/var/folders/d0/jxsscwdj2csbjj2ntd928_w80000gn/T/tmpbrv0cwji']
❱
```

from

```
from fsspec.implementations.cached import WholeFileCacheFileSystem
from fsspec.implementations.zip import ZipFileSystem

fs = WholeFileCacheFileSystem(fs=ZipFileSystem("dir.zip"))
print(fs.storage)
```

Also no config:

```python
>>> print(fsspec.config.conf)
{}
```

horta:
Environment variables:

```
❱ env | grep TMP
TMPDIR=/var/folders/d0/jxsscwdj2csbjj2ntd928_w80000gn/T/
❱ env | grep TEMP
```

horta:
Let me know if you want me to check some breakpoints that you think it might reveal the problem.

martindurant:
So which of those contains the cached files, /var/... or ./var/ ?

horta:
You mean regarding `python main.py`? I just ran it twice and it generated two random temporary directories.

martindurant:
I mean any files cached by the filesystem; I suppose in your example, you didn't even get as far as reading anything.

horta:
I think I found the neaby line with the error:

```python
[...]

PR #1413, opened 2023-11-06
Test bug reported in #1410

horta:
Spurious directory created using cache.


martindurant:
Thanks for setting this up

martindurant:
This fixes your test - do you find it reasonable? I haven't yet checked if it breaks anything else.

horta:
Thanks for the fix! Seems reasonable and the tests are passing...

PR #1418, opened 2023-11-09
Fix open_local returning list for pathlib.Path

### LINKED ISSUE #1400: Inconsistent return type in open_local for string and pathlib.Path
If `open_local` is given a string URL for a local file without globs a string is returned. If the same URL is given as a `pathlib.Path` a list is returned.

```python
import fsspec
from pathlib import Path

url = "/path/to/file"

fsspec.open_local(url)  # -> returns string

p = Path(url)
fsspec.open_local(p)  # -> returns list
```

Granted the `open_local` documentation says that arguments to `url` should be a string or a list of strings but since `open_files` can also handle `pathlib.Path` which is a useful feature I think this should be adapted (also documentation of return value could be better because even while it is implied and maybe even logical it is more transparent if the documentation makes it clear that a string or a list of strings can be returned).

Anyway I think the relevant line is: https://github.com/fsspec/filesystem_spec/blob/e20f626b87b5bb87d223495a56aefd768272a7ca/fsspec/core.py#L492

BENR0:
Input of single `pathlib.Path` to url argument now returns string instead of list similar to input of single str.

- [x] closes #1400
- [x] Tests added
- [x] Fully documented

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -474,7 +474,7 @@ def open(
 
 def open_local(
     url: str | list[str] | Path | list[Path],
-    mode: str,
+    mode: str = "rb",
     **storage_options: dict,
 ) -> str | list[str]:
     """Open file(s) which can be resolved to local
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -1,10 +1,11 @@
+from __future__ import annotations
+
 import io
 import logging
 import os
 import re
 from glob import has_magic
 from pathlib import Path
-from typing import Any, Literal, Optional
 
 # for backwards compat, we export cache things from here too
 from .caching import (  # noqa: F401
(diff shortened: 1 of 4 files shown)


fsspec/core.py
```diff
@@ -469,7 +471,11 @@ def open(
     return out[0]
 
 
-def open_local(url, mode="rb", **storage_options):
+def open_local(
+    url: str | list[str] | Path | list[Path],
+    mode: Optional[Literal["rb"]] = "rb",
```
martindurant, line 476: This annotation is wrong, I think, it implies mode can be ["rb", None] and nothing else. 
```suggestion
    mode: str = "rb",
```

The actual legal possibilities depend on the target filesystem, but include ["rb", "r", "rt", "w", "wb", "wt"]

martindurant (commented):
Thanks for adding comprehensive tests

martindurant:
I think you may need
```
from __future__ import annotations
```
to get those type things to import. Or just remove them.

PR #1422, opened 2023-11-10
Pass `kwargs` to `info` and `exists` calls in `glob`

mariosasko:
Same as `find`, `info` and `exists` propagate the `kwargs` values to `ls`, so we should also pass the `kwargs` values in `glob` to them to be consistent.

(We have a use case for this in the `HfFileSystem`: https://github.com/huggingface/huggingface_hub/pull/1815)

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -672,9 +672,9 @@ class AsyncFileSystem(AbstractFileSystem):
             [self._size(p) for p in paths], batch_size=batch_size
         )
 
-    async def _exists(self, path):
+    async def _exists(self, path, **kwargs):
         try:
-            await self._info(path)
+            await self._info(path, **kwargs)
             return True
         except FileNotFoundError:
             return False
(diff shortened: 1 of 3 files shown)


fsspec/asyn.py
```diff
@@ -760,11 +760,11 @@ async def _glob(self, path, maxdepth=None, **kwargs):
         detail = kwargs.pop("detail", False)
 
         if not has_magic(path):
-            if await self._exists(path):
+            if await self._exists(path, **kwargs):
```
martindurant, line 763: Can you please also fix these same lines in spec.py:AbstractFileSystem?

mariosasko: They are fixed already: https://github.com/fsspec/filesystem_spec/blob/3c80a530c18f71518239b4215d4adcff69917a69/fsspec/spec.py#L577

martindurant: Sorry, I must have been looking at an old version...

Issue #1416, opened 2023-11-08
[FR] Set transaction class on file system


nicholasjng:
I have an application of transactions at hand where I want to extend the functionality of fsspec's builtin `Transaction`. In particular, I want to write my own stack unwind routine, and I want to yield the transaction object on entry into the context manager like so:

```python
from fsspec.transaction import Transaction

class MyTransaction(Transaction):
    def __enter__(self):
        self.start()
        return self  # enables the use of a transaction in the context.

    def foo(self, *args, **kwargs):
         ...  # do whatever.

fs = MyFileSystem()
with fs.transaction as tx:
    ... # write files, etc.
    tx.foo()
```

Since the `AbstractFileSystem.transaction` member is a base `Transaction`, I have to override the `transaction` member and `start_transaction` method on the file system, even though I only want to change the transaction type.

So Q1: Would it be possible to make a file system class aware of its associated transaction type (maybe through a `transaction_type` property, default being `fsspec.transactions.Transaction`) to avoid having to overload these methods?

And Q2: Would it be feasible to `return self` in the base `Transaction`'s context entry, to be able to use the object?

Thanks in advance!

martindurant:
Yes, I think it's totally reasonable that Transaction should be subclassable and assigned to filesystem instances (or classes - but I prefer instances). 

There are three major limitations in the current implementation anyway:
- only open() is intercepted, so other write operations like pipe() are still immediate (see discussion linked above)
- in-transaction files don't show up in list or read operations
- on transaction finish, files are committed sequentially, even though a bulk/async operation may be available.

The possible interaction with simplecache is interesting too - that files could be written locally and then uploaded in bulk when the transaction finishes. This is like a totally different conception of a transaction.

nicholasjng:
Thanks! The cache case sounds interesting. I'm actually interested in conducting versioning operations (think `git tag`s, `git merge`s etc.) in the file list unwinding, so at least I would require sequential consistency there (I am basically treating it as a FIFO stack right now), but I see the potential of async there, too.

I can send in a PR with at least the transaction type part if you're interested. My rough idea is this:

```python
from fsspec.transaction import Transaction

class MyTransaction(Transaction):
    ...

class MyFS(AbstractFileSystem)
    _transaction_type = MyTransaction  # or transaction_type=MyTransaction in __init__ for instance-level

    @property
    def transaction_type() -> type[Transaction]:
        return self._transaction_type

    @property
    def transaction() -> Transaction:
[...]

PR #1424, opened 2023-11-10
Give file system classes a transaction type, switch to deque

nicholasjng:
This means that subclasses of `AbstractFileSystem` need not override the `transaction` property and `start_transaction` function anymore to be able to use custom transaction classes.

Instead, they can just override the newly created `transaction_type` property to achieve the same.

Also changes the base `Transaction`'s file queue to a `collections.deque` to pick up thread safety on `pop` and `append` events.

Finally, returning `self` in the transaction's context entry means that transaction objects can now be access in the enclosing transaction context.

For reviewers:
The `files` member shouldn't need to be flushed anymore on `start`, so this could just be removed?
Also, I am not sure on how to make the transaction type an instance variable, as the keyword arguments all seem to go to the `DirCache`. (Should I just use `kwargs.pop("transaction_type", Transaction)`?)

Addresses part of #1416.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -110,6 +110,7 @@ class AbstractFileSystem(metaclass=_Cached):
     async_impl = False
     mirror_sync_methods = False
     root_marker = ""  # For some FSs, may require leading '/' or other character
+    transaction_type = Transaction
 
     #: Extra *class attributes* that should be considered when hashing.
     _extra_tokenize_attributes = ()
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -228,6 +228,10 @@ def current(cls):
             return cls._cache[cls._latest]
         return cls()
 
+    @property
+    def transaction_type(self):
+        return Transaction
```
martindurant, line 233: How about a class attribute? Then it could be overridden either by any subclass having a different attribute or by setting the attribute on any instance.

nicholasjng: Changed, see [0271589](https://github.com/fsspec/filesystem_spec/pull/1424/commits/0271589c1f767098155e74f7a4fd5bdd1d1d90e4).

Issue #1425, opened 2023-11-13
Error when processing file mode string


lobis:
In the `memory` filesystem there is an error if you try to open it with `r+b` mode saying this mode is not supported.

However this mode is apparently supported (https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/memory.py#L178).

What I think is happening is that the `mode=r+b`(or `mode=rb+`) argument is being processed and transformed internally to `r+b` and then the comparison to `rb+` fails.

(I assume that it should work for any order for the characters in the mode string: `rb+` or `r+b`).

Reproducer:

```
import fsspec
uri = "memory://path/file"
open_file = fsspec.open(uri, mode="rb+")  # r+b doesn't work either
assert open_file.mode == "rb+"
file = open_file.open() # raises ValueError
```

martindurant:
I honestly don't know if "r+b" and "rb+" should be considered equivalent or not, and which of these is canonical. In fact, any mode that is not "w" will have the same effect. In general, other filesystems only tend to suppose "r", "w" and sometimes "a". The special case of _local_ files support the full range of options.

lobis:
> I honestly don't know if "r+b" and "rb+" should be considered equivalent or not, and which of these is canonical. In fact, any mode that is not "w" will have the same effect. In general, other filesystems only tend to suppose "r", "w" and sometimes "a". The special case of _local_ files support the full range of options.

Not sure if I get your point. What I am saying here is that neither `r+b` or `rb+` is working because of some internal processing of the string into `r+b` then trying to compare it with `rb+`. From my experience I think `r+b` and `rb+` should both work in the same way but there may be some subtly difference I am not aware of. Anyway I think `rb+` mode at the very least should work right?

martindurant:
> Anyway I think rb+ mode at the very least should work right?

If you like! In fact, this is what you get with "r" too :)
```
In [1]: import fsspec
In [2]: fs = fsspec.filesystem("memory")
In [3]: f = fs.open("myfile", "wb")
In [4]: f.write(b"hello")
In [5]: f = fs.open("myfile", "rb")
In [6]: f.write(b"oi")
In [7]: f.seek(0)
In [8]: f.read(5)
Out[8]: b'oillo'
```

PR #1426, opened 2023-11-13
Fix inconsistent file mode formatting and comparison (`rb+` changed to `r+b`)

lobis:
Triggered by https://github.com/fsspec/filesystem_spec/issues/1425

The current processing of the file mode string causes the `b` character to always be on the end of the mode e.g. `rb+` becomes `r+b`.

However in some instances (such as in the memory file system), the valid file modes listed are not compatible with this format, for instance `rb+` is listed. This comparison will never take place given how the mode string is processed.

This PR replaces all instances of `rb+` to `r+b` so that this works now.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -111,7 +111,7 @@ class MMapCache(BaseCache):
             fd.write(b"1")
             fd.flush()
         else:
-            fd = open(self.location, "rb+")
+            fd = open(self.location, "r+b")
 
         return mmap.mmap(fd.fileno(), self.size)
 
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -62,14 +62,14 @@ class OpenFile:
     """
 
     def __init__(
-        self,
-        fs,
-        path,
-        mode="rb",
-        compression=None,
-        encoding=None,
-        errors=None,
-        newline=None,
+            self,
+            fs,
+            path,
+            mode="rb",
+            compression=None,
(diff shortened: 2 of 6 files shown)


lobis:
@martindurant after starting this PR I realised that there are different places where the mode is specified (not everything goes through the OpenFile.__enter__) (This is why the new test fails). I could normalize the mode everywhere but it's a bit messy. 

Do you agree on the need to normalize (sort) the file mode string? I think it's dangerous to do string comparisons inside the file system against some reference modes ("wb", "rb+", etc.) when these modes can be in different order (but meaning the same thing). Currently I think the only failing case is when using "+" but still I think it should be addressed (memory filesystem cannot be opened with "rb+" or "r+b" because of the reordering taking place).

martindurant:
Maybe you can sprinkle `set()` or `sorted()` where exact equality is tried? In most cases, `in` is used instead.

lobis:
At the end I chose not to modify how mode is processed: `mode = self.mode.replace("t", "").replace("b", "") + "b"`.

Instead I replaced all instances of `rb+` to `r+b`. Now it works fine.

I think it makes sense anyway that the file mode should always have `b` at the end since this is a python thing.

lobis:
@martindurant this is now ready, please let me know if you any have feedback.

martindurant:
[python](https://docs.python.org/3/library/functions.html#open) seems to agree

PR #1428, opened 2023-11-14
WebHDFS Impl: Allow '=' in URLs, which is a valid character in HDFS path

markhatch:
Currently `=` is percent encoded.
= is a valid character in a path, especially with dynamically partitioned Hive tables.
Default arg for `safe` includes `/`, so I just included the `=` character.

Before (does not work):
`GET /webhdfs/v1/user/hive/warehouse/some_hive_table/datestamp%3D2023-11-11?user.name=test&op=LISTSTATUS HTTP/1.1" 200 None`

After:
`GET /webhdfs/v1/user/hive/warehouse/some_hive_table/datestamp=2023-11-11?user.name=test&op=LISTSTATUS HTTP/1.1" 200 None`

### DIFF
--- a/fsspec/implementations/tests/test_webhdfs.py
+++ b/fsspec/implementations/tests/test_webhdfs.py
@@ -135,3 +135,63 @@ def test_webhdfs_cp_file(hdfs_cluster):
     assert fs.exists(src)
     assert fs.exists(dst)
     assert fs.cat(src) == fs.cat(dst)
+
+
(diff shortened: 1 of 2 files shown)


martindurant:
Makes sense. Should we make a test to roundtrip such a filename, to show that it's working?

markhatch:
> Makes sense. Should we make a test to roundtrip such a filename, to show that it's working?

Definitely, add some real simple ones. I can add more if it's appropriate.

PR #1436, opened 2023-11-22
bugfix: ReferenceFileSystem.cat() scrambles and omits reference mappings

### LINKED ISSUE #1435: bug in `ReferenceFileSystem.cat()` scrambles and omits reference mappings
This is an edge case, happens when `ReferenceFileSystem()`'s references do not have a protocol specified and we pass `.cat()` some missing reference(s) and set  `oe_error='return` or `oe_error='omit`. See the reproducer below

### Reproducer
```python
from fsspec.implementations.reference import ReferenceFileSystem

my_fs = ReferenceFileSystem(
    fo={
        'ref1': 'first reference', 
        'ref2': 'some other reference'}
)

result = my_fs.cat(['nonexistent', 'ref1', 'ref2'], on_error='return')
```

This produces:
```bash
{'nonexistent': b'first reference', 'ref1': b'some other reference'}
```
However it should produce:
```bash
{'nonexistent': FileNotFoundError('nonexistent'), 'ref1': b'first reference', 'ref2': b'some other reference'}
```

### Downstream implications 

This currently breaks Zarr's spec's uninitialized chunk behavior in xarray, resulting in `KeyError` wherever `.cat()` is used to find chunks under the hood. 
For example when evaluating a variable of  `xarray.Dataset()` by calling `ds.my_var.values`
I can confirm this is the case on kerchunked zarrs from NetCDFs. As well as 'native' zarrs that I created via the zarr lib and open
as a key-value store without writing to disk

### Proposed solution
Checking for valid path references in `.cat()` and extending the unit tests to cover this edge case.

PR is inbound with updated test and bugfix.

arongergely:
This PR fixes #1435 by distinguishing valid path references regardless of the protocol.

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -804,15 +804,15 @@ class ReferenceFileSystem(AsyncFileSystem):
                 # found and on_error is "raise"
                 try:
                     u, s, e = self._cat_common(p)
-                    urls.append(u)
-                    starts.append(s)
-                    ends.append(e)
                 except FileNotFoundError as err:
                     if on_error == "raise":
                         raise
                     if on_error != "omit":
                         out[p] = err
                 else:
+                    urls.append(u)
+                    starts.append(s)
+                    ends.append(e)
                     valid_paths.append(p)
 
             # process references into form for merging
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -798,7 +798,7 @@ class ReferenceFileSystem(AsyncFileSystem):
         out = {}
         for proto, paths in proto_dict.items():
             fs = self.fss[proto]
-            urls, starts, ends = [], [], []
+            urls, starts, ends, valid_paths = [], [], [], []
             for p in paths:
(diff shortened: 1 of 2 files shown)


martindurant:
That was fast! :)

arongergely:
@martindurant I spent 2 days troubleshooting my kerchunk+xarray code. So this reaction time is not a true indication of effort :laughing:

fsspec/implementations/reference.py
```diff
@@ -812,14 +812,16 @@ def cat(self, path, recursive=False, on_error="raise", **kwargs):
                         raise
                     if on_error != "omit":
                         out[p] = err
+                else:
+                    valid_paths.append(p)
```
martindurant, line 816: Perhaps it would be clearer if this were inside the `try` block, to show that there must be the same number of urls, starts, ends, valid_paths when we are done. Alternatively, all four come here in the `else` block, since it's only the `_cat_common` line which can error.

arongergely: that makes sense, I will adjust it. All under `else:` seems the cleanest to me

PR #1452, opened 2023-12-07
fix: correctly handle local files with colons (`:`) in name

### LINKED ISSUE #1447: Local files with colons in name no longer readable after `2023.12.0` version
Newer `fsspec` versions [2023.12.1](https://github.com/fsspec/filesystem_spec/releases/tag/2023.12.1) and [2023.12.0](https://github.com/fsspec/filesystem_spec/releases/tag/2023.12.0) break the ability to read local files with `:` in the file name.

This is the behaviour in version `2023.10.0`
```python
import fsspec
fsspec.core.url_to_fs("file-with:colons.txt")

(<fsspec.implementations.local.LocalFileSystem object at 0x101b13610>, '/Users/lobis/git/uproot/file-with:colons.txt')
```

On a newer version:
```python
import fsspec
fsspec.core.url_to_fs("file-with:colons.txt")

  File "/Users/lobis/anaconda3/envs/uproot/lib/python3.11/site-packages/fsspec/core.py", line 383, in url_to_fs
    chain = _un_chain(url, kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lobis/anaconda3/envs/uproot/lib/python3.11/site-packages/fsspec/core.py", line 332, in _un_chain
    cls = get_filesystem_class(protocol)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lobis/anaconda3/envs/uproot/lib/python3.11/site-packages/fsspec/registry.py", line 233, in get_filesystem_class
    raise ValueError(f"Protocol not known: {protocol}")
ValueError: Protocol not known: file-with

```

It appears that previously the protocol delimiter was `://` but now has been updated to `:`.

This is currently breaking the uproot `fsspec` source for local files since we rely on being able to parse file names with `:` https://github.com/scikit-hep/uproot5/issues/1054.

I believe files with a `:` in the filename are still valid filenames (unlike files with `://`) so I assume this change was not intentional. Please let me know otherwise so I will modify how we use `fsspec` to read local files.

Thanks!

lobis:
Fixes #1447

### DIFF
--- a/fsspec/generic.py
+++ b/fsspec/generic.py
@@ -136,8 +136,7 @@ def rsync(
         fs.cp(source_files, target_files, **kwargs)
     logger.debug(f"{len(to_delete)} files to delete")
     if delete_missing:
-        for file in to_delete:
-            fs.rm(file)
+        fs.rm(to_delete)
 
 
 class GenericFileSystem(AsyncFileSystem):
--- a/fsspec/generic.py
+++ b/fsspec/generic.py
@@ -254,12 +254,11 @@ class GenericFileSystem(AsyncFileSystem):
         urls = url
         if isinstance(urls, str):
             urls = [urls]
-        for url in urls:
-            fs = _resolve_fs(url, self.method)
-            if fs.async_impl:
-                await fs._rm(url, **kwargs)
-            else:
-                fs.rm(url, **kwargs)
+        fs = _resolve_fs(url[0], self.method)
+        if fs.async_impl:
+            await fs._rm(url, **kwargs)
+        else:
+            fs.rm(url, **kwargs)
 
     async def _makedirs(self, path, exist_ok=False):
         logger.debug("Make dir %s", path)
--- a/fsspec/generic.py
+++ b/fsspec/generic.py
@@ -251,11 +251,15 @@ class GenericFileSystem(AsyncFileSystem):
             return fs.pipe_file(path, value, **kwargs)
 
     async def _rm(self, url, **kwargs):
-        fs = _resolve_fs(url, self.method)
-        if fs.async_impl:
-            await fs._rm(url, **kwargs)
-        else:
-            fs.rm(url, **kwargs)
+        urls = url
+        if isinstance(urls, str):
+            urls = [urls]
+        for url in urls:
+            fs = _resolve_fs(url, self.method)
+            if fs.async_impl:
+                await fs._rm(url, **kwargs)
+            else:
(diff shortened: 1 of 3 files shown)


fsspec/implementations/tests/test_local.py
```diff
@@ -56,6 +56,8 @@ def filetexts(d, open=open, mode="t"):
     try:
         os.chdir(dirname)
         for filename, text in d.items():
+            if dirname := os.path.dirname(filename):
```
martindurant, line 59: We still run on py3.8 - was this already allowed by then?

martindurant:
Looks like fsspec.generic.py:_resolve_fs should not pass a list to split_protocol

lobis:
Apparently windows does not support `:` in paths (makes sense I guess). I'll skip the test for windows.

lobis:
I'll double check in our end that this fixes things and I'll mark this PR as ready.

fsspec/generic.py
```diff
@@ -136,7 +136,8 @@ def rsync(
         fs.cp(source_files, target_files, **kwargs)
     logger.debug(f"{len(to_delete)} files to delete")
     if delete_missing:
-        fs.rm(to_delete)
+        for file in to_delete:
+            fs.rm(file)
```
martindurant, line 140: No, we definitely don't want this - all rm() methods should be able to delete in batch and that's the right call here. 

(There is a separate rm_file method that requires a single path. )

The problem, instead, is in attempting to resolve the filesystem protocol from a list.

martindurant, line 140:
```suggestion
        fs.rm(to_delete)
```

fsspec/generic.py
```diff
@@ -250,11 +251,15 @@ async def _pipe_file(
...
-            fs.rm(url, **kwargs)
+        urls = url
+        if isinstance(urls, str):
+            urls = [urls]
+        for url in urls:
+            fs = _resolve_fs(url, self.method)
+            if fs.async_impl:
+                await fs._rm(url, **kwargs)
+            else:
+                fs.rm(url, **kwargs)
```
martindurant, line 262:
```suggestion
        fs = _resolve_fs(url[0], self.method)
        if fs.async_impl:
            await fs._rm(url, **kwargs)
        else:
            fs.rm(url, **kwargs)
```

lgray:
Could a release come before Dec 13th?

lgray:
@martindurant ping!

Issue #1075, opened 2022-10-18
feature: scoped callbacks


skshetry:
Currently, it's hard in fsspec to deregister/close callbacks, especially branched/nested ones after one `put_file`/`get_file` operation is done. In dvc, we are trying to do batched uploads for better performance, which means using `get`/`put` instead and letting filesystems batch as they see fit. We may be uploading thousands of files, so that'd mean we'd have thousands of callbacks/progress bars. And currently there is no easy way to deregister/close the callbacks once the single operation is finished (we don't want to be left with 100s of finished/stuck progress bars).

1. You have to depend upon logic like `self.value >= self.size` to detect if the callback can be closed, which is fragile.
2. You have to trust the filesystem to call `set_size` and `relative_update`/`absolute_update`. This may not always happen or be possible. Sometimes, fs may not provide `size` or may be expensive to fetch/compute. Sometime `relative_update`/`absolute_update` may be unavailable (eg: when we use clone/reflink files).
3. The callback will persist even in errors, where you want to upload as much as possible before giving the control back. This may already be the case in AsyncFileSystem where we use `asyncio.gather()`, the scheduled tasks still run until we cancel them.

So, my proposal is to make callback a `contextmanager`.

```python
class Callback:
    def __enter__(self):
        return self

    def __exit__(self, *exc_args):
        return self.close()

    def close(self):
        pass
```

The `fsspec.callbacks.Callback` will provide a `close`/`__enter__`/`__exit__` methods. And by default, `__exit__` will call `close` so that we can still close the callbacks without using as a contextmanager. The callback can get access to the exception if it extends `__exit__` (which should be a rare usecase).

The `branch()` also needs to return the instance so that we can use it as in a `with` statement, instead of just adding to the `kwargs`.

### Usage
```python
with TqdmCallback() as callback:
    fs.put_file("file", "file", callback=callback)
```

#### With branches
```python
def put(self, lpath, rpath, **kwargs):
    callback.set_size(len(rpaths))
    for lpath, rpath in callback.wrap(zip(lpaths, rpaths)):
        with callback.branch(lpath, rpath, kwargs):
            self.put_file(lpath, rpath, **kwargs)
```

martindurant:
I appreciate the pain of having many callback instances, and those potentially persisting following an exception. 

[...]

PR #1460, opened 2023-12-12
callbacks: introduce scoped callbacks

skshetry:
1. This PR introduces a `close()` API that gives an opportunity for callbacks to close themselves, either on error or on completions. We use this to extend `Callback`s into a context-manager by using that `close()` api on `__exit__()`.
2. This PR also introduces `branched()` API that returns an instance of a callback. Since the callback is a context-manager, this gives an opportunity to close child callbacks after they complete. This is very similar to `branch()` API, but unlike `branch`, it should not mutate passed `kwargs`, and always return an instance of a callback. The default return is _DEFAULT_CALLBACK or NoopCallback if left unimplemented.

   For backwards compatibility, this will still call `branch()` API, try to read from the kwargs it has set, pop the `callback` kwarg and return it. But anyone implementing `Callback` does not need to implement `branch` API and should prefer `branched()` instead from now on, and the older API should be considered deprecated.

Closes fsspec/filesystem_spec#1075.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -568,7 +568,7 @@ class AsyncFileSystem(AbstractFileSystem):
         coros = []
         callback.set_size(len(file_pairs))
         for lfile, rfile in file_pairs:
-            put_file = callback.wrap_and_branch_coro(self._put_file)
+            put_file = callback.branch_coro(self._put_file)
             coros.append(put_file(lfile, rfile, **kwargs))
 
         return await _run_coros_in_chunks(
@@ -645,7 +645,7 @@ class AsyncFileSystem(AbstractFileSystem):
         coros = []
         callback.set_size(len(lpaths))
         for lpath, rpath in zip(lpaths, rpaths):
-            get_file = callback.wrap_and_branch_coro(self._get_file)
+            get_file = callback.branch_coro(self._get_file)
             coros.append(get_file(rpath, lpath, **kwargs))
         return await _run_coros_in_chunks(
             coros, batch_size=batch_size, callback=callback
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -568,8 +568,8 @@ class AsyncFileSystem(AbstractFileSystem):
         coros = []
         callback.set_size(len(file_pairs))
         for lfile, rfile in file_pairs:
-            callback.branch(lfile, rfile, kwargs)
-            coros.append(self._put_file(lfile, rfile, **kwargs))
+            put_file = callback.wrap_and_branch_coro(self._put_file)
+            coros.append(put_file(lfile, rfile, **kwargs))
 
         return await _run_coros_in_chunks(
             coros, batch_size=batch_size, callback=callback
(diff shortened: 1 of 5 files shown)


skshetry:
ping @martindurant

fsspec/asyn.py
```diff
@@ -568,8 +568,8 @@ async def _put(
         coros = []
         callback.set_size(len(file_pairs))
         for lfile, rfile in file_pairs:
-            callback.branch(lfile, rfile, kwargs)
-            coros.append(self._put_file(lfile, rfile, **kwargs))
+            put_file = callback.branch_coro(self._put_file)
```
skshetry, line 571: I don't like adding `branch_coro` but I am not sure how to support scoped/structured callbacks in async without one either, since we are working task/futures.

skshetry:
> Could we please have some tests that call the new API? I see you made some changes in the batch operations, and the existing tests pass, so that's good.

I have added the tests.

fsspec/asyn.py
martindurant: Yeah, this is not well explorer territory. I would have thought that all calls into the callbacks machinery can by sync, since they will be fast and so safe to call from sync or async contexts.

martindurant:
@efiop , any thoughts on this?

efiop (approved):
lgtm

fsspec/callbacks.py
```diff
@@ -25,6 +28,56 @@ def __init__(self, size=None, value=0, hooks=None, **kwargs):
...
+            arguments passed to child method, e.g., put_file.
+
+        Returns
+        -------
+        callback: Callback
+            A callback instance to be passed to the child method
+        """
+        self.branch(path_1, path_2, kwargs)
+        # mutate kwargs so that we can force the caller to pass "callback=" explicitly
+        return kwargs.pop("callback", _DEFAULT_CALLBACK)
```
skshetry, line 67: Note that this is a default implementation that re-uses `callback.branch()` implementation for compatibility. Newer implementation should just implement `branched` themselves and never call `super().branched(...)`.

fsspec/callbacks.py
```diff
@@ -25,6 +28,56 @@ def __init__(self, size=None, value=0, hooks=None, **kwargs):
...
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc_args):
+        self.close()
+
+    def close(self):
+        """Close callback."""
+
+    def branched(self, path_1, path_2, kwargs):
```
skshetry, line 40: Here, I matched the function signature with `branch`, but I am open to any suggestions here (if there should be `kwargs` or `**kwargs` or if it should be there at all).

fsspec/callbacks.py
martindurant: Rather than a comment here, tis should be documented in comments/docstrings for other devs

skshetry: Thanks, I have added that in 2cc3c55.

fsspec/callbacks.py
martindurant: I have no better suggestion.

skshetry: I am trying to think of what could be the use case of `kwargs`. So far, it feels like an unnecessary required argument to pass.

I think following will be an example of how someone will use a `branched` callback. 

```python
with callback.branched(path_1, path_2, {}) as child:
    pass
```

The kwargs feel unnecessary to me. Either it should be `kwargs=None` by default or should be removed. I am leaning towards removing it though (can be added if we get some requests).

Issue #1313, opened 2023-07-19
Kwargs (such as overwrite=True) does not work with CachingFileSystem methods, such as open


SoptikHa2:
I'm trying to use azure blob storage under the CachingFileSystem wrapper. However, I need to pass `overwrite=True` to the `open` call. Caching FS does not remember it and does not pass it to `az` filesystem when the cache is commited.

The minimal breaking example looks like this:

```py
# Run azurite
#
# docker-compose.yml:
#
# services:
#   blob_storage:
#     image: mcr.microsoft.com/azure-storage/azurite
#     command: azurite-blob --blobHost 0.0.0.0 --blobPort 10000
#     ports:
#       - 10000:10000
import fsspec

CONN_STRING = "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"

fs = fsspec.filesystem("filecache", target_protocol="az", expiry_time=720,
     target_options={'connection_string': CONN_STRING})
#fs = fsspec.filesystem("az", connection_string=CONN_STRING)

fs.rm("test-container")
fs.mkdir("test-container") # ok

assert fs.ls("test-container") == [ ] # ok

with fs.open("test-container/foo", mode="wb") as file_handle:
    file_handle.write(b"foo") # ok

assert fs.ls("test-container") == [ "test-container/foo" ] # ok

# BREAKS: overwrite=True is ignored
with fs.open("test-container/foo", mode="wb", overwrite=True) as file_handle:
    file_handle.write(b"bar") # fails
```

I believe the root cause is the same as in #633.

I think, that the cause is that although `open` receives `overwrite=True` in `kwargs` [here](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/cached.py#L304), it does not propagate it further when cache is [commited](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/cached.py#L826).

martindurant:
It sounds like you have the right solution here, to capture and pass on the kwargs at commit time. Would you like to make this a PR? To test, you can probably just us e a mock to see that the argument gets passed.

SoptikHa2:
[...]

Issue #633, opened 2021-05-11
`filecache` and `simplecache` both ignore `version_id` for versioned S3 filesystems


AndreiBarsan:
Caching S3 data with e.g., `simplecache` works fine. However, S3FS also supports [versioned buckets](https://s3fs.readthedocs.io/en/latest/#bucket-version-awareness).

When a `version_id` is provided to `open`, using a cache leads to incorrect behavior, as the cache seems to drop the version argument and always requests, and subsequently caches, the latest version.

The code below should reproduce the issue, as long as you provide it with the bucket name for a versioned bucket. Removing the caching fixes the issue.

Possible fixes:

- make caching aware of S3 version IDs (may need to "hack" the local disk cache paths to include the version IDs)
- make caches raise an error when they detect an attempt to wrap a versioned bucket, in order to protect the user from this surprising behavior

Please let me know if I can provide any additional information or code!

```python
import random
import time

import fsspec
import numpy as np
import s3fs


def eval_version_caching():
    bucket = "MY_TEST_BUCKET"

    fs = s3fs.S3FileSystem(version_aware=True)
    print(fs.ls(bucket))

    version_dir = f"tmp/dbg-version-eval-{random.randint(0, 100000):010d}"
    data_path = f"{bucket}/{version_dir}/data.npy"

    # 8 * 2 * 2 Mb, roughly = 32 Mb
    some_data = np.ones((2000, 2000), dtype=np.float64) * 1.0
    some_other_data = np.ones((2000, 2000), dtype=np.float64) * 42.0

    with fs.open(data_path, "wb") as f_out:
        np.save(f_out, some_data)

    print("Saved first version... Waiting to flush.")
    time.sleep(1.0)

    with fs.open(data_path, "wb") as f_out:
        np.save(f_out, some_other_data)

    print("Saved second version... Waiting to flush.")
    time.sleep(1.0)

    versions = fs.object_version_info(data_path)
    assert 2 == len(versions)

    # Versions are listed in reverse chronological order. In our case, we call the first version "a", and the second
    # version "b".
    id_rev_a = versions[1]["VersionId"]
    id_rev_b = versions[0]["VersionId"]
    print(id_rev_a)
    print(id_rev_b)

    # The cached FS always returns the latest version, no matter what version_id we provide.
    cached_fs = fsspec.filesystem(
        "filecache", target_protocol="s3", target_options={"version_aware": True}, cache_storage="/tmp/s3_file_cache"
    )
[...]

PR #1462, opened 2023-12-12
fix: cached fs kwargs not being passed into underlying fs (#1313)

SoptikHa2:
Fixes https://github.com/fsspec/filesystem_spec/issues/1313, https://github.com/fsspec/filesystem_spec/issues/633

The issue was cached fs not passing kwargs to the underlying fs when a file was commited, so certain options (such as `overwrite`) were ignored.

This fixes the issue by passing kwargs to the `LocalTempFile` constructor, which later uses them in `commit`.

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -524,7 +524,7 @@ class WholeFileCacheFileSystem(CachingFileSystem):
     protocol = "filecache"
     local_file = True
 
-    def open_many(self, open_files):
+    def open_many(self, open_files, **kwargs):
         paths = [of.path for of in open_files]
         if "r" in open_files.mode:
             self._mkcache()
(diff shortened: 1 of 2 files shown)


fsspec/implementations/cached.py
```diff
@@ -650,7 +651,13 @@ def _open(self, path, mode="rb", **kwargs):
         path = self._strip_protocol(path)
         if "r" not in mode:
             fn = self._make_local_details(path)
-            return LocalTempFile(self, path, mode=mode, fn=fn)
+            user_specified_kwargs = {
+                k: v
+                for k, v in kwargs.items()
+                # those kwargs were added by open(), we don't want them
+                if k not in ["autocommit", "block_size", "cache_options"]
```
martindurant, line 658: I wonder if there's any way to guarantee that this list doesn't go stale.

SoptikHa2: No idea as of now. At least, the test should catch addition of new kwargs (it caught it when I forgot to add those three), but it doesn't handle removal.

Regardless, I'd like to find a better way if there is one.

Issue #1464, opened 2023-12-12
ArrowFSWrapper should not use "/" root_marker for all filesystems


b-phi:
For the arrow S3FileSystem for example, using `root_market = "/"` throws an error when trying to upload a local file. Love the arrow integration, please let me know if I'm misusing something here. 

```python
from pyarrow.fs import S3FileSystem
from fsspec.implementations.arrow import ArrowFSWrapper

fs = ArrowFSWrapper(S3FileSystem())
fs.upload('foo.txt', 'bucket/folder/foo.txt')
```

>Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../fsspec/spec.py", line 1530, in upload
    return self.put(lpath, rpath, recursive=recursive, **kwargs)
  File ".../fsspec/spec.py", line 1057, in put
    self.put_file(lpath, rpath, **kwargs)
  File ".../fsspec/spec.py", line 984, in put_file
    self.mkdirs(self._parent(os.fspath(rpath)), exist_ok=True)
  File ".../fsspec/spec.py", line 1498, in mkdirs
    return self.makedirs(path, exist_ok=exist_ok)
  File ".../fsspec/implementations/arrow.py", line 22, in wrapper
    return func(*args, **kwargs)
  File ".../fsspec/implementations/arrow.py", line 193, in makedirs
    self.fs.create_dir(path, recursive=True)
  File "pyarrow/_fs.pyx", line 603, in pyarrow._fs.FileSystem.create_dir
  File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Path cannot start with a separator ('/bucket/folder')

If I set `root_market` to `''` manually, the upload succeeds
```python
fs.__class__.root_marker = ''
fs.upload('foo.txt', 'bucket/folder/foo.txt')
```

martindurant:
root_marker is a class attribute and (I think) only used in class methods, so I'm not sure how we can make it dependent on the instance being wrapped. 

I don't suppose the following would have solved your problem
```
fs.root_marker = ""
```

I am open to suggestions!

PR #1467, opened 2023-12-13
ArrowFSWrapper should not use "/" root_marker for all filesystems

b-phi:
Perhaps a bit too hacky to be merged, but Fixes https://github.com/fsspec/filesystem_spec/issues/1464 , in theory. Basic idea is to check if the provided arrow filesystem supports a `/` path, then create an `ArrowFSWrapper` class dynamically based on the result.

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -62,6 +62,20 @@ class ArrowFSWrapper(AbstractFileSystem):
         return "hdfs_" + tokenize(self.fs.host, self.fs.port)
 
     @classmethod
+    def from_fs(cls, fs, **kwargs):
+        override = {"local": "file"}
+        type_name = override.get(fs.type_name, fs.type_name)
+
+        try:
+            fs.from_uri(f"{type_name}:///")
+            root_marker = "/"
+        except BaseException:
+            root_marker = ""
+
+        wrapper = type(cls.__name__, (cls,), {"root_marker": root_marker})
+        return wrapper(fs, **kwargs)
+
+    @classmethod
     def _strip_protocol(cls, path):
         ops = infer_storage_options(path)
         path = ops["path"]
--- a/fsspec/implementations/tests/test_arrow.py
+++ b/fsspec/implementations/tests/test_arrow.py
@@ -241,3 +241,17 @@ def test_seekable(fs, remote_dir):
     with fs.open(remote_dir + "/a.txt", "rb", seekable=False) as file:
         with pytest.raises(OSError):
             file.seek(5)
+
+
+@pytest.mark.parametrize(
+    ["filesystem", "root_marker"],
+    [
+        (pyarrow_fs.LocalFileSystem(), "/"),
+        (pyarrow_fs.S3FileSystem(), ""),
+        (pyarrow_fs.GcsFileSystem(), ""),
+    ],
+)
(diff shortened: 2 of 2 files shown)


fsspec/implementations/arrow.py
```diff
@@ -61,6 +61,20 @@ def protocol(self):
...
+        override = {"local": "file"}
+        type_name = override.get(fs.type_name, fs.type_name)
+
+        try:
+            fs.from_uri(f"{type_name}:///")
+            root_marker = "/"
+        except BaseException:
+            root_marker = ""
+
+        wrapper = type(cls.__name__, (cls,), {"root_marker": root_marker})
```
martindurant, line 75: So this is effectively a metaclass?

b-phi: There's no metaclasses involved, but it works similarly yes. For such a simple case, metaclasses might be overkill, but maybe a top-level function that isn't attached to `ArrowFSWrapper` instead? 

```python
def wrap_arrow_fs(fs: pyarrow.fs.FileSystem, **kwargs) -> ArrowFSWrapper:
    # build class dynamically based on fs
```

Because it's a class attribute, I don't think we can get around creating different classes per arrow filesystem. An alternative to the dynamic class construction would be to have wrapper subclasses for each `root_marker` variant and a lookup table (dict) based on the provided arrow filesystem.

fsspec/implementations/arrow.py
```diff
@@ -61,6 +61,20 @@ def protocol(self):
     def fsid(self):
         return "hdfs_" + tokenize(self.fs.host, self.fs.port)
 
+    @classmethod
+    def from_fs(cls, fs, **kwargs):
+        override = {"local": "file"}
+        type_name = override.get(fs.type_name, fs.type_name)
+
+        try:
+            fs.from_uri(f"{type_name}:///")
```
martindurant, line 70: This is the foolproof way to know whether the first character can/should be "/"?

b-phi: It definitely works for (3/4) Local, S3, and GCS. I'm not sure about Hadoop, I'm never used it and am not sure how to test locally. 

https://arrow.apache.org/docs/python/filesystems.html#filesystem-interface

martindurant: hadoop has a derived class anyway, HadoopFileSystem. We should make sure it has "/", and it doesn't need to go through the check in this PR.

martindurant:
How about https://github.com/fsspec/filesystem_spec/commit/b1298609f437ea57c5a8c1fae18282969c484c7e , does this doe everything, but without `type()` ?

b-phi:
> How about [b129860](https://github.com/fsspec/filesystem_spec/commit/b1298609f437ea57c5a8c1fae18282969c484c7e) , does this doe everything, but without `type()` ?

Hey @martindurant yes that commit looks great. Dropping the `@classmethod` decorator is much cleaner, thanks!

martindurant:
> Dropping the @classmethod decorator is much cleaner, thanks!


I don't immediately see any downside, except maybe we should guard against being called as `ArrowFSWrapper._strip_protocol` by testing the type of `self`. If it is being used as a classmethod, we'll have to try to analyse the path for what backend it is each time.

martindurant:
This one got lost... should we finish it off?

PR #1472, opened 2023-12-18
Add session_cert and session_verify parameters to WebHDFS class (enables mTLS)

markhatch:
Provides mechanism to supply certificates in the WebHDFS session.

### DIFF
--- a/fsspec/implementations/webhdfs.py
+++ b/fsspec/implementations/webhdfs.py
@@ -54,6 +54,8 @@ class WebHDFS(AbstractFileSystem):
         kerb_kwargs=None,
         data_proxy=None,
         use_https=False,
+        session_cert=None,
+        session_verify=True,
         **kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


martindurant:
I feel like this could be done simpler with optional kwargs to the existing webHDFS rather than making a new class. Since fsspec keys URLs by their protocol prefixes, this would ensure that "webhdfs" would serve both uses, but allow for the extra arguments to be defined in config.

markhatch:
> I feel like this could be done simpler with optional kwargs to the existing webHDFS rather than making a new class. Since fsspec keys URLs by their protocol prefixes, this would ensure that "webhdfs" would serve both uses, but allow for the extra arguments to be defined in config.

Sounds good, I'll take a stab at that!

martindurant:
(please ping me when you want another look)

markhatch:
Had to add #noqa to the url line on 105 to pass pre-commit, error didn't make sense to me (and I didn't change the line)

```
fsspec/implementations/webhdfs.py:105:56: E231 missing whitespace after ':'
fsspec/implementations/webhdfs.py:105:65: E231 missing whitespace after ':'
```

markhatch:
@martindurant think the changes are ready. Initially thought I could just pass `session_kwargs` through, but guess that requests.Session doesn't have init kwargs so used explicit parameters instead.

Issue #1471, opened 2023-12-16
Reading data from `raw.githubusercontent.com` hangs


nenb:
## What

When using the GitHub `fsspec` implementation, an attempt to `open(...)` data at `raw.githubusercontent.com` hangs indefinitely on my local machine.

## Why

My DNS server returns both IPv4 and IPv6 addresses for `raw.githubusercontent.com`:
```bash
> nslookup raw.githubusercontent.com <IP_OF_MY_DNS_SERVER>

...

Name:	raw.githubusercontent.com
Address: 185.199.109.133
Name:	raw.githubusercontent.com
Address: 185.199.108.133
Name:	raw.githubusercontent.com
Address: 185.199.111.133
Name:	raw.githubusercontent.com
Address: 185.199.110.133
Name:	raw.githubusercontent.com
Address: 2606:50c0:8002::154
Name:	raw.githubusercontent.com
Address: 2606:50c0:8000::154
Name:	raw.githubusercontent.com
Address: 2606:50c0:8003::154
Name:	raw.githubusercontent.com
Address: 2606:50c0:8001::154
```

*(**Note:** `urllib3`/`requests` gives an identical response for me.)*

The IPv4 addresses are fine. However, I'm not able to establish a connection with the IPv6 address eg
```bash
❯ ping -w 30 2606:50c0:8001::154
PING 2606:50c0:8001::154(2606:50c0:8001::154) 56 data bytes

--- 2606:50c0:8001::154 ping statistics ---
30 packets transmitted, 0 received, 100% packet loss, time 29673ms
```

This creates a problem when a program attempts to first connect to an IPv6 address rather than IPv4 (ie program hangs waiting to establish a connection). Some (likely unresolved) issues from GitHub community that appear relevant:
[Issue 1](https://github.com/orgs/community/discussions/13893)
[Issue 2](https://github.com/orgs/community/discussions/32501)

## Proposed fix

Introduce a `timeout` keyword on the filesystem instance [here](https://github.com/fsspec/filesystem_spec/blob/5df4b0b30dc011f9d6eceded5a078e92b2b5c11d/fsspec/implementations/github.py#L40-L53). This will allow a program to move forward in the event no response. This is also the recommended approach from [`requests`](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts). If this sounds okay, let me know and I would be happy to open a small PR.

Thanks for a cool project!

martindurant:
[...]

PR #1473, opened 2023-12-18
Add timeout kwarg to GitHub implementation

nenb:
Addresses #1471.

### DIFF
--- a/fsspec/implementations/github.py
+++ b/fsspec/implementations/github.py
@@ -36,8 +36,11 @@ class GithubFileSystem(AbstractFileSystem):
     url = "https://api.github.com/repos/{org}/{repo}/git/trees/{sha}"
     rurl = "https://raw.githubusercontent.com/{org}/{repo}/{sha}/{path}"
     protocol = "github"
+    timeout = (60, 60)  # connect, read timeouts
 
-    def __init__(self, org, repo, sha=None, username=None, token=None, **kwargs):
+    def __init__(
+        self, org, repo, sha=None, username=None, token=None, timeout=None, **kwargs
+    ):
         super().__init__(**kwargs)
         self.org = org
         self.repo = repo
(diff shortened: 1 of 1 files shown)


fsspec/implementations/github.py
```diff
@@ -73,13 +78,16 @@ def repos(cls, org_or_user, is_org=True):
...
+            Connection and read timeouts, in seconds
 
         Returns
         -------
         List of string
         """
         r = requests.get(
-            f"https://api.github.com/{['users', 'orgs'][is_org]}/{org_or_user}/repos"
+            f"https://api.github.com/{['users', 'orgs'][is_org]}/{org_or_user}/repos",
+            timeout=timeout,
```
martindurant, line 90: Why not reuse `self.timeout` here?

nenb: This is a `classmethod` and so `self.timeout` won't necessarily exist, right?

martindurant: I see what you mean. There should probably be a default value as a class attribute.

nenb: I can do that. Just to be clear: I will set the `timeout` as a class attribute rather than as an instance attribute. This will mean (amongst other things) that if a GitHub filesystem instance changes the timeout value, this will also change the timeout for all other GitHub filesystem instances. Is this okay?

martindurant: > a class attribute rather than as an instance attribute

Both! If the user provides a non-None value, set it on the instance, so that `fs.timeout` will get the instance one if it exists but fall back to the class one otherwise.

nenb: Done!

fsspec/implementations/github.py
```diff
@@ -89,6 +97,7 @@ def tags(self):
         """Names of tags in the repo"""
         r = requests.get(
             f"https://api.github.com/repos/{self.org}/{self.repo}/tags",
+            self.timeout,
```
martindurant, line 100: Shouldn't this and the ones below all be `timeout=`, rather than a positional argument?

nenb: Ugh, yes, sorry about that. It should be fixed now.

PR #1475, opened 2023-12-20
Fix non-str case in `AbstractFileSystem.expand_path()`

nicholasjng:
If the input to `expand_path` is not a string, it needs to be iterable, since the if-branch for the container case tries to loop over the contents.

Previously, the code would fall through to the iterable case even for perfectly legal inputs to `AbstractFileSystem._strip_protocol()`, like `pathlib.Path`s, causing a crash with a less-than-ideal error message.

All of these inputs are now legalized by a changed `isinstance` check on the scalar case, which now contains all types legal in `_strip_protocol` and by proxy `stringify_path`.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import io
 import logging
 import os
-import pathlib
 import threading
 import warnings
 import weakref
@@ -1135,7 +1134,7 @@ class AbstractFileSystem(metaclass=_Cached):
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
-        if isinstance(path, (str, os.PathLike, pathlib.Path)):
+        if isinstance(path, (str, os.PathLike)):
             out = self.expand_path([path], recursive, maxdepth)
         else:
             out = set()
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import io
 import logging
 import os
+import pathlib
 import threading
 import warnings
 import weakref
@@ -1134,7 +1135,7 @@ class AbstractFileSystem(metaclass=_Cached):
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
-        if isinstance(path, str):
+        if isinstance(path, (str, os.PathLike, pathlib.Path)):
             out = self.expand_path([path], recursive, maxdepth)
         else:
             out = set()
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -5,6 +5,7 @@ import pickle
 import subprocess
 import sys
 from collections import defaultdict
+from pathlib import Path
 
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


fsspec/spec.py
```diff
@@ -1134,7 +1135,7 @@ def expand_path(self, path, recursive=False, maxdepth=None, **kwargs):
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
-        if isinstance(path, str):
+        if isinstance(path, (str, os.PathLike[str], pathlib.Path)):
```
martindurant, line 1138:
```suggestion
        if isinstance(path, (str, os.PathLike, pathlib.Path)):
```

martindurant: The [str] form is not allowed on older pythons, causing CI fails. Also, is there any justification of PathLike alongside Path?

martindurant (commented):
Please add a test using a Path to hit this line.

fsspec/spec.py
nicholasjng: I just took this over from `stringify_path`, which contains legacy logic to deal with Pythons were Path is not an instance of `PathLike`.

nicholasjng:
Added a test on a `pathlib.Path` input, let me know if that's sufficient. 
Also, what's the verdict on the input typing? The distinction between Path and PathLike is only relevant for Pythons pre 3.6, if I read this correctly: https://docs.python.org/3/library/os.html#os.PathLike

martindurant:
> The distinction between Path and PathLike is only relevant for Pythons pre 3.6

Then drop it. We don't support <3.8

nicholasjng:
Done. (I kept `os.PathLike` since that's the abstract interface of Path.)

martindurant:
The perils of paths...
```
FAILED fsspec/tests/test_spec.py::test_expand_path[test_paths18-False-None-expected18] - AssertionError: assert ['misc\\foo.txt'] == ['misc/foo.txt']
  At index 0 diff: 'misc\\foo.txt' != 'misc/foo.txt'
  Full diff:
  - ['misc/foo.txt']
  ?       ^
  + ['misc\\foo.txt']
  ?       ^^
```
(only on windows)

fsspec/tests/test_spec.py
```diff
@@ -647,6 +649,12 @@ def _open(
             None,
             DummyTestFS.get_test_paths() + [DummyTestFS.root_marker],
         ),
+        [
+            (Path("misc/foo.txt"),),
+            False,
+            None,
+            ["misc/foo.txt" if platform.system() != "Windows" else "misc\\foo.txt"],
```
martindurant, line 656:
```suggestion
            [f"misc{os.sep}foo.txt"],
```
maybe?

...but shouldn't we always be returning posix paths always?

nicholasjng: I don't know all of the details, but I assumed that a `LocalFileSystem` on Windows would not be happy with POSIX-only paths.

martindurant:
OK, I'll allow it for now and put it on the to-do list to figure out where a windows separator is coming from. Also on the list is to find out why SMB times out sometimes.

martindurant:
> That separator comes from stringify_path, which calls str on pathlib.Paths, and that in turn uses the given OS separator.

OK, that's fine, then. When used by `LocalFileSystem`, it will be made posix in strip_protocol anyway.

nicholasjng:
Thanks!

Issue #945, opened 2022-04-19
LocalFileSystem inconsistent with Windows paths


kasuteru:
```python
from fsspec.implementations.local import LocalFileSystem
fs = LocalFileSystem()
fs.ls("C:/")
```

This lists the content of os.getcwd() instead of C:/. The same for similar commands like
```
fs.ls("file://C:/")
```

pretty sure this is a bug?

martindurant:
I suppose the Windows path is actually `r"C:\"` ?

kasuteru:
No, that doesn't work either. Your version raises an error because even in raw paths, the backslash cannot be at the end. And `fs.ls("C:\\")` still outputs the contents of cwd instead of C:.

kasuteru:
Any update on this?

martindurant:
Any windows path expert care to have a look?

efiop:
@kasuteru Seems like you are pretty deep into it already. Maybe you could try debugging it further? That would be the fastest way to resolve this.

kasuteru:
Not an expert, but I traced it, and the problem seems to be the following: 

1.) os.listdir("C:") actually returns the contents of os.getcwd(). Not a problem in itself, but:
2.) This line is causing the problem:

https://github.com/fsspec/filesystem_spec/blob/5d6c87e562396f352645df7cd555a2b4b6eec84a/fsspec/implementations/local.py#L196

It strips away the "/", meaning what ends up getting passed to os.listdir() is "C:" (see 1.)

kasuteru:
If I implemented this from scratch, I would probably go with pathlib.Path instead of strings, since there a lot of special cases regarding file systems and strings (which pathlib handles implicitely by using WindowsPath and UnixPath). Since the code already exists, I am trying to think of a way to handle this edge case without breaking all the other functionality...

martindurant:
pathlib is oppressively slow compared to strings. We endeavour to convert all paths to posix at the earliest opportunity, so we only have to deal with that from then on.

kasuteru:
It looks like all occurences of `rstrip("/")` in https://github.com/fsspec/filesystem_spec/blob/5d6c87e562396f352645df7cd555a2b4b6eec84a/fsspec/implementations/local.py are unnecessary except the one in _strip_protocol? Since the very last command of `_strip_protocol` is already `rstrip`, and `rstrip` is only ever called right after `_strip_protocol`?

kasuteru:
[...]

PR #1477, opened 2023-12-21
LocalFileSystem fix _strip_protocol for root directory

fleming79:
This modifies LocalFileSystem to fix https://github.com/fsspec/filesystem_spec/issues/945.

All local tests pass on Windows.


nicholasjng:
I'd like to see this go into the `AbstractFileSystem._strip_protocol` base class method as well - I've had trouble with it in the case of lakeFS, where stripping away trailing slashes has some implications on URI parsing and API call results.

martindurant:
> I'd like to see this go into the AbstractFileSystem._strip_protocol base class 

The trouble here is, that for real filesystems (local, ftp, ssh, ...), files cannot e d with "/", even though directory names are sometimes printed with the extra character. So for those, stripping is always the right thing to do. Similarly, in s3 and gcs, "/" is the "separation" character used with the remote API and also has a special place, if not exactly the same. So I think I would for now recommend that implementations that need it should provide their own _strip_protocol rather than changing the default one.

nicholasjng:
> The trouble here is, that for real filesystems (local, ftp, ssh, ...), files cannot e d with "/", even though directory names are sometimes printed with the extra character. So for those, stripping is always the right thing to do. Similarly, in s3 and gcs, "/" is the "separation" character used with the remote API and also has a special place, if not exactly the same. So I think I would for now recommend that implementations that need it should provide their own _strip_protocol rather than changing the default one.

I see your point, but on the other hand, trailing slashes are also the defining hint of directories in many implementations - in a different world, in those implementations, all file-ops could raise `IsADirectoryError`s if they are not applied recursively.

In my case, I am working with URIs similiar to S3, of the structure `lakefs://<bucket>/<version>/<resource>`, where an empty `resource` marks the root of the bucket (i.e. / as root marker). I cannot just simply return the root marker there since the bucket and version are still in front, so the `filename or cls.root_marker` check will not do what I want.

But since the subclass implementation works fine in our case, I see no harm in keeping that in place for now.

fleming79:
My apologies if the updates have been noisy. I'm a novice when it comes to git and pull requests. 

The other history has been removed when I switched to using a nested function so that remove_trailing_slash is always respected. 

I note there is an error for [test_fsmap_error_on_protocol_keys](https://github.com/fsspec/filesystem_spec/blob/069e107fa056166fd4e7db75e909ebac85f372fd/fsspec/tests/test_mapping.py#L205). 

```
def test_fsmap_access_with_suffix(tmp_path):
    tmp_path.joinpath("b").mkdir()
    tmp_path.joinpath("b", "a").write_bytes(b"data")
    m = fsspec.get_mapper(f"file://{tmp_path}")

    with pytest.raises(KeyError):
        _ = m["b/"]

    assert m["b/a/"] == b"data"
```
I haven't used `FSMap` before but wonder if the line in the test that is failing:

``` assert m["b/a/"] == b"data"` ```

should be without the trailing separator?

``` assert m["b/a"] == b"data"'```

File: fsspec/implementations/local.py [hunk 1]
```diff
@@ -232,10 +248,20 @@ def chmod(self, path, mode):
...
+            make_path_posix(p, remove_trailing_slash=remove_trailing_slash)
+            for p in path
+        )
+    path = _make_path_posix(path, sep)
+    if os.name == "nt" and len(path) == 3 and path[1] == ":":
+        return path  # windows root
+    return path.rstrip("/") if remove_trailing_slash else path
+
+
+def _make_path_posix(path, sep=os.sep):
```
martindurant, line 264: Why split up the function? The windows-root check is a rare one and should ideally be checked after more common things. This would be a second case within "# windows full path"?

fsspec/implementations/local.py
```diff
@@ -232,10 +248,20 @@ def chmod(self, path, mode):
...
     if isinstance(path, (list, set, tuple)):
-        return type(path)(make_path_posix(p) for p in path)
+        return type(path)(
+            make_path_posix(p, remove_trailing_slash=remove_trailing_slash)
+            for p in path
+        )
+    path = _make_path_posix(path, sep)
+    if os.name == "nt" and len(path) == 3 and path[1] == ":":
+        return path  # windows root
+    return path.rstrip("/") if remove_trailing_slash else path
```
martindurant, line 261: Presumably this i what breaks the mapper test; the key in question I think would have hit "if path.startswith("/"):" previously.

fsspec/implementations/local.py
fleming79: My reasoning was that the original `make_path_posix` function has several return statements. *Wrapping* the original function enables detection of special edge cases but also adds *support* for stripping the trailing suffix where appropriate.

fsspec/implementations/local.py
fleming79: >> The trouble here is, that for real filesystems (local, ftp, ssh, ...), files cannot e d with "/", even though directory names are sometimes printed with the extra character. 

The filename ends with a trailing '/'. Is it supposed to?



image

I should mention that I also added that check in open, however the issue still occurs if the check is disabled.

The issue can be resolved by stripping the trailing suffix in` mapping.py->__getitem__`.   Doing doesn't appear to cause any regression.

fleming79:
`make_path_posix` has been divided into a *native* posix version and *nt* (windows) version. It has also been ordered with assumptions made about the most common types of paths. 
A test, `test_parent` for testing the results of  the method: `_parent` has also been added. Do you think _parent of a root should return the root, or should it raise an error?

fsspec/mapping.py
```diff
@@ -37,7 +37,10 @@ class FSMap(MutableMapping):
 
     def __init__(self, root, fs, check=False, create=False, missing_exceptions=None):
         self.fs = fs
-        self.root = fs._strip_protocol(root).rstrip("/")
+        try:
+            self.root = fs._strip_protocol(root, remove_trailing_slash=True)
+        except TypeError:
```
martindurant, line 42: This is unfortunate; in what case does `rstrip` fail? In this case, I think we might be explicitly joining on the separator anyway.

fsspec/implementations/local.py
```diff
@@ -205,25 +207,36 @@ def modified(self, path):
...
+            return path_
 
     @classmethod
-    def _strip_protocol(cls, path):
+    def _strip_protocol(cls, path, sep=os.sep, remove_trailing_slash=False):
         path = stringify_path(path)
         if path.startswith("file://"):
             path = path[7:]
+            if sep == "\\":
+                path = path.lstrip("/")
```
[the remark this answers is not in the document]
fleming79: The proper handling is done by `LocalFileSystem._strip_protocol`. 

``` python
    @classmethod
    def _strip_protocol(cls, path, sep=os.sep, remove_trailing_slash=False):
        path = stringify_path(path)
        if path.startswith("file://"):
            path = path[7:]
            if sep == "\\":
                path = path.lstrip("/")
        elif path.startswith("file:"):
            path = path[5:]
        elif path.startswith("local://"):
            path = path[8:]
        elif path.startswith("local:"):
            path = path[6:]
        return make_path_posix(path, sep, remove_trailing_slash)
```

Currently there is a test which does this check for the NT version.

``` python
assert make_path_posix("file://C:\\path\\file", sep="\\") == "C:/path/file"
```

Maybe the test can be removed and the related code removed from `make_path_posix`?

fsspec/mapping.py
fleming79: On Windows if the root is something like `c:/`, `rstrip` will cause a root of `c:` which is treated like '.', the current working directory instead of the root.
On posix `rstrip` will make `/` change to ` `.

fsspec/implementations/local.py
```diff
@@ -205,25 +207,36 @@ def modified(self, path):
...
+            return path.rsplit("/", 1)[0] or "/"
         else:
-            return cls.root_marker
+            # NT
+            path_ = path.rsplit("/", 1)[0]
+            if len(path_) <= 3:
+                if path_[1:2] == ":":
+                    # nt root (something like c:/)
+                    return path_[0] + ":/"
+                raise NotImplementedError(path)
```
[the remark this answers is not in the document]
fleming79: I've removed the error, but think there might be other cases in _parent that will need handling, such as network drives.

fsspec/implementations/local.py
```diff
@@ -179,6 +179,8 @@ def unstrip_protocol(self, name):
 
     def _open(self, path, mode="rb", block_size=None, **kwargs):
         path = self._strip_protocol(path)
+        if path.endswith("/"):
```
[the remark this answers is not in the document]
fleming79: Okay - has been removed.

fsspec/implementations/local.py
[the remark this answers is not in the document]
fleming79: `_strip_protocol` has been updated.

martindurant:
py38 does not support removeprefix
```
AttributeError: 'str' object has no attribute 'removeprefix'
```

fsspec/implementations/tests/test_local.py
```diff
@@ -472,33 +472,48 @@ def test_make_path_posix():
...
-    assert "./" not in pp and ".\\" not in pp
+    cd = make_path_posix(cwd)
+    assert pp == cd + "/path"
+    # Userpath
+    userpath = make_path_posix("~/path")
+    assert userpath.endswith("/path")
+
+
+def test_parent():
+    assert LocalFileSystem._parent("/file or folder", sep="/") == "/"
```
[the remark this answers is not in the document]
fleming79: If you are referring to the test on line 512, that test is simply to check that `_parent` applied to `"/file or folder"` returns `"/"`. 

'sep' is specified here to perform the same test regardless of the OS the test is run on.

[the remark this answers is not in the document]
martindurant: My point is, that I would think this should behave the same independent of OS for this particular call, even without specifying `sep`

fsspec/implementations/local.py
```diff
@@ -205,25 +212,50 @@ def modified(self, path):
...
+            # NT
+            path_ = path.rsplit("/", 1)[0]
+            if len(path_) <= 3:
+                if path_[1:2] == ":":
+                    # nt root (something like c:/)
+                    return path_[0] + ":/"
+            # More cases may be required here
+            return path_
+
+    if sys.version_info < (3, 9):
```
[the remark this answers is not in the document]
fleming79: Okay, will use the 3.8 version.

fsspec/implementations/tests/test_local.py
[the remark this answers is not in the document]
fleming79: The test for that line would fail on NT if  `sep` was omitted. It would return something like: `c:/`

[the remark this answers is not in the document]
martindurant: OK, I'll allow it, although my gut says that operations on posix paths should always work; but I suppose on windows there just is no root which would be the parent here.

martindurant:
```
_strip_protocol("file:///path")
```
gives a relative path on WIN (or "/path" when specifying sep="/"). What should be the behaviour of this, do you think - is it just poorly specified on WIN? 

On posix doing
```
_strip_protocol(path, sep="\\")
```
causes RecursionError (with or without a prefix) *unless* path starts with "X:" or "\\\\".

fleming79:
The latest changes remove `sep` accordingly.

fsspec/implementations/tests/test_jupyter.py
```diff
@@ -38,6 +38,7 @@ def jupyter(tmpdir):
         P.terminate()
 
 
+@pytest.mark.skipif(WIN, reason="Subprocess gets stuck in a loop")
```
martindurant, line 41: I didn't notice this before. Any idea what's going on?

fleming79: I found the problem - Something to do with the temp dir passed to Jupyterlab in the launch call showing up as an invalid path. Surrounding the path with "" fixed it. The test now passes when I run it on Windows.

```
cmd = f'jupyter notebook --notebook-dir="{tmpdir}" --no-browser --port=5566'
```

martindurant:
please merge from master to pick up the constraint on pytest that should fix CI

fleming79:
Did you mean something other than to sync the fork?

martindurant:
That's exactly what I meant - sorry I didn't get a notification

fleming79:
No problem. 
I fixed the Ruff lint issue, but there were also errors with 'dbfs'. I'm not sure how they could be related to this PR.

martindurant:
@nils-braun , @a24lorie , any idea why DBFS should start failing, but only for some python versions? It is also happening on other PRs. 

@fleming79 , you may with to take the commit from https://github.com/fsspec/filesystem_spec/pull/1533 , which should at least squash the intermittent SMB failures.

a24lorie:
> @nils-braun , @a24lorie , any idea why DBFS should start failing, but only for some python versions? It is also happening on other PRs.

I have found this to be an issue with the pytest-vcr and urllib3 version 2.x in the python 3.10 and python 3.11. 
Downgrading urllib3 to version 1.26.18 in Python 3.10 removes this error

martindurant:
Please update your fork following https://github.com/fsspec/filesystem_spec/pull/1533

fleming79:
I synced the fork with GitHub.

fsspec/implementations/local.py
```diff
@@ -16,6 +15,12 @@
 logger = logging.getLogger("fsspec.local")
 
 
+def _remove_prefix(text: str, prefix: str):
```
martindurant, line 18: Pretty soon we can use https://docs.python.org/3/library/stdtypes.html#str.removeprefix when we drop py 3.8

fsspec/implementations/local.py
```diff
@@ -206,24 +211,32 @@ def modified(self, path):
...
-    def _strip_protocol(cls, path):
+    def _strip_protocol(cls, path, remove_trailing_slash=False):
         path = stringify_path(path)
-        if path.startswith("file://"):
-            path = path[7:]
-        elif path.startswith("file:"):
-            path = path[5:]
-        elif path.startswith("local://"):
-            path = path[8:]
+        if path.startswith("file:"):
```
martindurant, line 231: Should iterate over self.protocol (which can be a single string or a tuple, `("file", "local")` being the default for LocalFileSystem) to account for aliasing or subclasses.

fsspec/implementations/local.py
```diff
@@ -206,24 +211,32 @@ def modified(self, path):
...
+            path = _remove_prefix(_remove_prefix(path, "file://"), "file:")
```
martindurant, line 232: It is faster to use regex
```
pattern = re.compile("file:[//]?")
%timeit pattern.sub("", "file://my/path/file")
180ns
%timeit _remove_prefix(_remove_prefix("file://my/path/file", "file://"), "file:")
260ns
```

and you _could_ do all possible patterns in one
```
pattern = re.compile(f"({'|'.join(protocol)}):[//]?")
pattern.sub("", path)
```
(you want to compile just once, and then the sub part is faster)

fsspec/implementations/tests/test_jupyter.py
martindurant: This line was to be removed according to the thread above

fsspec/implementations/tests/test_local.py
```diff
@@ -472,33 +472,48 @@ def test_make_path_posix():
...
         assert make_path_posix("/a/posix/path") == "/a/posix/path"
         assert make_path_posix("/posix") == "/posix"
     assert make_path_posix("relpath") == posixpath.join(make_path_posix(cwd), "relpath")
     assert make_path_posix("rel/path") == posixpath.join(
         make_path_posix(cwd), "rel/path"
     )
+    # NT style
     if WIN:
         assert make_path_posix("C:\\path") == "C:/path"
-        assert make_path_posix("file://C:\\path\\file") == "C:/path/file"
```
martindurant, line 484: I'm not sure why this line disappeared?

fsspec/mapping.py
martindurant: This is the only place that the special case for LocalFileSystem leaks, and I don't like it :|. I think it may be better to let through strange behaviour if the user thinks they want to may their whole C drive (and gets only cwd files instead).

martindurant (commented):
Sorry this has been taking so long...

I wonder, since we now are essentially calling _strip_protocol from within LocalFileSystem, would it be easy to split off the optional remove_trailing_slash= into a separate method, to maintain the previous signature? I am still not sure what the best default is, since in most places we think of paths not having terminal sep characters even when they refer to directories.

fsspec/implementations/local.py
fleming79: The first `timeit` isn't taking into account compiling the pattern. Would the pattern be compiled during loading as a global?

fsspec/implementations/tests/test_jupyter.py
fleming79: Quite right - I made the changes but forgot to push them. It is done now.

fsspec/implementations/tests/test_local.py
fleming79: The call now fails because the path starts with a protocol. Instead it should pass through _strip_protocol.

I've moved the test to `test_strip_protocol_no_authority` using `_strip_protocol` instead of `make_path_posix`.

fsspec/mapping.py
fleming79: I've changed this back.

fsspec/implementations/local.py
fleming79: There is an *if* case for both `file` and `local`. In both cases the protocol will be stripped regardless of what `self.protocol` is set.

fsspec/implementations/local.py
fleming79: I have no objections if you can re-write it as a regex.

fsspec/implementations/tests/test_local.py
martindurant: ok

fsspec/implementations/local.py
martindurant: Right, but I am thinking that there might be a subclass of LocalFileSystem that has a different protocol(s). Maybe that's not important - it would simply need to override this method

fsspec/implementations/local.py
martindurant: > Would the pattern be compiled during loading as a global?

Exactly what I was thinking. 

However, I am finding that for the case where the path passed is already stripped, your code is actually (just) faster, so no change needed.

fsspec/implementations/local.py
fleming79: I image reading a property and checking types will slow things down.

fsspec/mapping.py
fleming79: But without `.rstrip("/")`

fleming79:
The fork has been synced, with no other changes.

martindurant:
:)

PR #1479, opened 2023-12-22
Allow files in `LocalFileSystem.ls()`

### LINKED ISSUE #1478: `LocalFileSystem.ls()` does not work on filenames
Repro:

```
fs = LocalFileSystem()

fs.touch("a.json")
fs.ls("a.json")
```

I expected this to return a single-entry list with the stat object for `a.json`, but instead it raises a `NotADirectoryError` because `os.scandir()` does not allow file-like inputs:

https://github.com/fsspec/filesystem_spec/blob/069e107fa056166fd4e7db75e909ebac85f372fd/fsspec/implementations/local.py#L63-L64

In contrast, a subsequent `ls -la a.json` behaves as expected, showing the single file stat result.

nicholasjng:
Implemented as a 2x2 switch-esque statement on `detail` and `isdir(path)`.

Adds a simple test on a single-file directory, which is assumed to be equal to calling `ls` on the file itself.

Fixes #1478.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -3,7 +3,6 @@ import io
 import logging
 import os
 import os.path as osp
-import posixpath
 import re
 import shutil
 import stat
@@ -59,11 +58,16 @@ class LocalFileSystem(AbstractFileSystem):
 
     def ls(self, path, detail=False, **kwargs):
         path = self._strip_protocol(path)
-        if detail:
+        info = self.info(path)
+        if info["type"] == "directory":
             with os.scandir(path) as it:
-                return [self.info(f) for f in it]
+                infos = [self.info(f) for f in it]
         else:
-            return [posixpath.join(path, f) for f in os.listdir(path)]
+            infos = [info]
+
+        if not detail:
+            return [i["name"] for i in infos]
+        return infos
 
     def info(self, path, **kwargs):
         if isinstance(path, os.DirEntry):
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -367,6 +367,16 @@ def test_directories(tmpdir):
     assert fs.ls(fs.root_marker)
 
 
+def test_ls_on_file(tmpdir):
+    tmpdir = make_path_posix(str(tmpdir))
+    fs = LocalFileSystem()
+    resource = tmpdir + "/a.json"
+    fs.touch(resource)
+    assert fs.exists(resource)
+    assert fs.ls(tmpdir) == fs.ls(resource)
+    assert fs.ls(resource, detail=True)[0] == fs.info(resource)
+
+
 @pytest.mark.parametrize("file_protocol", ["", "file://"])
 def test_file_ops(tmpdir, file_protocol):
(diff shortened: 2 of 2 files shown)


fsspec/implementations/local.py
```diff
@@ -59,11 +59,16 @@ def rmdir(self, path):
...
+        if detail and isdir:
             with os.scandir(path) as it:
                 return [self.info(f) for f in it]
-        else:
+        elif detail and not isdir:
+            return [self.info(path)]
+        elif not detail and isdir:
             return [posixpath.join(path, f) for f in os.listdir(path)]
+        else:
+            return [path]
```
martindurant, line 71: Do we get here if the path doesn't exist at all?

nicholasjng: Yes, that's the `not detail and not isdir` branch, which is triggered also if the file does not exist. Should this be a `FileNotFoundError`, maybe in front of the statement?

martindurant: Yes, I would say ls(not-a-file) should give FileNotFOundError

fsspec/implementations/tests/test_local.py
```diff
@@ -367,6 +367,15 @@ def test_directories(tmpdir):
     assert fs.ls(fs.root_marker)
 
 
+def test_ls_on_file(tmpdir):
+    tmpdir = make_path_posix(str(tmpdir))
+    fs = LocalFileSystem()
+    resource = tmpdir + "/a.json"
+    fs.touch(resource)
+    assert fs.exists(resource)
+    assert fs.ls(tmpdir) == fs.ls(resource)
```
martindurant, line 376: and 
```suggestion
    assert fs.ls(tmpdir) == fs.ls(resource)
    fs.ls(resource, detail=True)[0] == fs.info(resource)
```

fsspec/implementations/local.py
```diff
@@ -59,11 +59,19 @@ def rmdir(self, path):
 
     def ls(self, path, detail=False, **kwargs):
         path = self._strip_protocol(path)
-        if detail:
+        if not os.path.exists(path):
+            raise FileNotFoundError(path)
```
nicholasjng, line 63: I'm guessing I should also add `errno.ENOENT` here to get a proper `No such file or directory` message.

martindurant: I don't think that's necessary.

martindurant:
After some thought, it would be a better idea not to postentially call, exists(), isdir(), scandir()/info() one after the other. Perhaps info() once at the start (which would raise if no such file or directory), followed by scandir() if needed?

nicholasjng:
> After some thought, it would be a better idea not to postentially call, exists(), isdir(), scandir()/info() one after the other. Perhaps info() once at the start (which would raise if no such file or directory), followed by scandir() if needed?

Sure, give me a second.

JordiMForgeFlow:
Hi @martindurant 

Would it be possible to know when the next release is planned including this fix? Thank you.

martindurant:
There was one planned for this month, but it looks like we'll miss that. You may expect one by the end of this week.

PR #1483, opened 2024-01-02
Add mimetype and redirect url to HTTP info dict

joouha:
This will allow discovery of a file's mime-type and redirect URL without unnecessary additional HTTP requests.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -848,7 +848,7 @@ async def _file_info(url, session, size_policy="head", **kwargs):
             info["size"] = int(r.headers["Content-Range"].split("/")[1])
 
         if "Content-Type" in r.headers:
-            info["mime"] = r.headers["Content-Type"].partition(";")[0]
+            info["mimetype"] = r.headers["Content-Type"].partition(";")[0]
 
         info["url"] = str(r.url)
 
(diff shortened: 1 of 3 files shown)


fsspec/implementations/http.py
```diff
@@ -847,6 +847,11 @@ async def _file_info(url, session, size_policy="head", **kwargs):
         elif "Content-Range" in r.headers:
             info["size"] = int(r.headers["Content-Range"].split("/")[1])
 
+        if "Content-Type" in r.headers:
+            info["mime"] = r.headers["Content-Type"].partition(";")[0]
```
martindurant, line 851: It is possible this header doesn't exist

martindurant (commented):
Please add a test showing this being passed through.

fsspec/implementations/http.py
joouha: This is presumably taken care of by the `if` statement.
Would you prefer `mime` was set to `None` if the header was missing?

martindurant: Ah indeed. First day back...
By the way, what does the ";" do, how do you expect this to format?

joouha: The [`Content-Type`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) HTTP header can take the following form:

```
Content-Type: text/html; charset=utf-8
Content-Type: multipart/form-data; boundary=something
```

so partitioning on the first `;` selects the mimetype from the value

martindurant: OK. Is there any need for verification on that? I suppose we can just pass on the string and any client might do that.

joouha: I agree it makes sense to just pass on the string and leave verification to the client

joouha:
I've added tests

Issue #1482, opened 2024-01-02
SFTPFileSystem default value create_parents


bartvaneswhiffle:
Hi guys,
The [API documentation on mkdir](https://filesystem-spec.readthedocs.io/en/latest/api.html) seems not to correspond with SFTPFileSystem implementation of mkdir; the default value of create_parents=True is by default False for the SFTPFileSystem implementation.

Function implentation: 
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/sftp.html#SFTPFileSystem

bartvaneswhiffle:
You are right, should be good now!

PR #1484, opened 2024-01-02
Fix create_parents SFTP

bartvaneswhiffle:
Fixes https://github.com/fsspec/filesystem_spec/issues/1482

### DIFF
--- a/fsspec/implementations/tests/test_sftp.py
+++ b/fsspec/implementations/tests/test_sftp.py
@@ -205,7 +205,7 @@ def test_mkdir_create_parent(ssh, path):
     f = fsspec.get_filesystem_class("sftp")(**ssh)
 
     with pytest.raises(FileNotFoundError):
-        f.mkdir(path)
+        f.mkdir(path, create_parents=False)
 
     f.mkdir(path)
     assert f.exists(path)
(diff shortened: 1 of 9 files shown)


martindurant:
Please see the test logs. The point is, the default is now different, so if you wanted to test the old behaviour, you would now need to explicitly pass `False`

fsspec/implementations/tests/test_sftp.py
```diff
@@ -221,10 +221,10 @@ def test_mkdir_create_parent(ssh, path):
 def test_makedirs_exist_ok(ssh, path):
     f = fsspec.get_filesystem_class("sftp")(**ssh)
 
-    f.makedirs(path)
+    f.makedirs(path, exist_ok=False)
 
     with pytest.raises(FileExistsError, match=path):
-        f.makedirs(path, exist_ok=False)
+        f.makedirs(path)
```
martindurant, line 227: Probably better to keep the original line here, since that's explicitly what this function is testing.

bartvaneswhiffle: Sure, line 224 can stay the same right?

martindurant: Yes, although explicit in every call is fine (in case we ever change the default in makedirs).

PR #1486, opened 2024-01-02
add replication argument to HadoopFileSystem class

### LINKED ISSUE #1485: Unable to Configure Hadoop FileSystem Replication Argument
**Hi everyone,**

**I'd like to report an issue with the `replication` argument in PyArrow's Hadoop FileSystem implementation.**

**The problem:**

- The default `replication` value of 3 in PyArrow overrides the value set in the HDFS site file, preventing external configuration.
- This can lead to incorrect replication factors, storage usage issues, and failures in clusters with a replication factor of 2.

**Proposed solution:**

- Add the `replication` argument to the constructor of the `HadoopFileSystem` class with a default value (e.g., 3).
    - This would allow users to explicitly set the replication factor when creating a new FileSystem object.

**I'm happy to provide more details, code, or error messages as needed. Please let me know if there's any specific information that would be helpful.**

I'm eager to contribute and writing the pull request.

samssh:
This PR fixes #1485 with add replication argument to HadoopFileSystem class.

### DIFF
--- a/fsspec/implementations/arrow.py
+++ b/fsspec/implementations/arrow.py
@@ -255,6 +255,7 @@ class HadoopFileSystem(ArrowFSWrapper):
         port=0,
         user=None,
         kerb_ticket=None,
+        replication=3,
         extra_conf=None,
         **kwargs,
     ):
@@ -280,6 +281,7 @@ class HadoopFileSystem(ArrowFSWrapper):
             port=port,
             user=user,
             kerb_ticket=kerb_ticket,
+            replication=replication,
             extra_conf=extra_conf,
         )
         super().__init__(fs=fs, **kwargs)
@@ -294,4 +296,6 @@ class HadoopFileSystem(ArrowFSWrapper):
             out["user"] = ops["username"]
         if ops.get("port", None):
             out["port"] = ops["port"]
+        if ops.get("replication", None):
+            out["replication"] = ops["replication"]
         return out


fsspec/implementations/arrow.py
```diff
@@ -294,4 +296,6 @@ def _get_kwargs_from_urls(path):
             out["user"] = ops["username"]
         if ops.get("port", None):
             out["port"] = ops["port"]
+        if ops.get("replication", None):
+            out["replication"] = ops["replication"]
```
martindurant, line 300: How would you specify the replication in a URL, what would that mean?

fsspec/implementations/arrow.py
```diff
@@ -255,6 +255,7 @@ def __init__(
         port=0,
         user=None,
         kerb_ticket=None,
+        replication=3,
```
martindurant, line 258: Since it's a new kwarg, we had better have a docstring entry for it. Is it possible to set per-file at open(, mode="w") time?

fsspec/implementations/arrow.py
samssh: refer to this [documentation](https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem.from_uri) some thing like this:

`hdfs://localhost:8020/?user=test&replication=1`

fsspec/implementations/arrow.py
samssh: I add docstring for this parameter.

As far as I have searched, in the pyarrow implementation, this argument is only configurable by the constructor.

fsspec/implementations/arrow.py
martindurant: If you put that URL through infer_storage_options, "user" and "replication" and bundled in the `query` field.

martindurant: This is still outstanding. In effect, I don't think this has any effect, since `ops` will never include they key - but it would be nice to write a test for it.

samssh: I apologize for the delay. I didn't understand the behavior of the infer_storage_options function correctly. I have now fixed that and I have written a test for it.

Issue #1488, opened 2024-01-04
429 Client Error: Too Many Requests for url azuredatabricks.net/api/2.0/dbfs/mkdirs


a24lorie:
I'm using the  "fsspec.implementations.dbfs import DatabricksFileSystem" with pyarrow to write a parquet dataset on the DatabricksFilesystem DBFS, but when using the DatabricksFilesystem implementation with the write_to_dataset method from pyarrow  I'm getting the following error:

429 Client Error: Too Many Requests for url: https://adb-<databricks_instance>.azuredatabricks.net/api/2.0/dbfs/mkdirs

The code used to write to DBFS is the following:

"""
import pyarrow as pa
import pyarrow.dataset as ds
import pyarrow.parquet as pq
from fsspec.implementations.dbfs import DatabricksFileSystem

base_path = "/FileStore/write"
test_df = pd.read_csv("../data/diabetes/csv/nopart/diabetes.csv")

filesystem = DatabricksFileSystem(
            instance=<databricks_instance>,
            token=<databricks_token>
 )

pq.write_to_dataset(arr_table, filesystem=filesystem, compression='none',
                    existing_data_behavior='error', partition_cols=["Pregnancies"],
                    root_path=f"{base_path}/parquet/part",
                    use_threads= False)
"""

And the full stack trace is:

"""
Traceback (most recent call last):
  File "<workdir>/venv/lib/python3.9/site-packages/requests/models.py", line 971, in json
    return complexjson.loads(self.text, **kwargs)
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
[...]

PR #1491, opened 2024-01-07
Added decorator for retry_request and handle err 429

a24lorie:
Added test for pyarrow integration using the DBFS implementation


a24lorie:
This PR is a quick fix for handling HTTP error 429 with DBFS filesystem implementation by adding a decorator to the _send_to_api function. It does not change the base code used to interact with the DBFS. I have also added some integration tests that use the pyarrow implementation to read and write parquet files from DBFS. 
The current patch solves the issue #1488 reported but the current code still lacks the capability to read from a partitioned directory structure using the hive partition method.

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
         """
         return DatabricksFile(self, path, mode=mode, block_size=block_size, **kwargs)
 
+    def retry_request(func, retries=5):
```
martindurant: This is a function rather than a method, as it and `is_retiable` I think would be better as standalone functions.

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+        RETRIABLE_EXCEPTIONS = (
+            requests.exceptions.ChunkedEncodingError,
+            requests.exceptions.ConnectionError,
+            requests.exceptions.ReadTimeout,
+            requests.exceptions.Timeout,
+            requests.exceptions.ProxyError,
+            requests.exceptions.SSLError,
+            requests.exceptions.ContentDecodingError,
+
+            aiohttp.client_exceptions.ClientError,
```
martindurant, line 238: How does aiohttp enter into this, since all the calls in this module are via requests?

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+            """Returns True if this exception is retriable."""
+            errs = list(range(500, 505)) + [
+                408,  # Request Timeout
+                429  # Too Many Requests
+            ]
+            errs += [str(e) for e in errs]
+            if type(exception) is requests.exceptions.HTTPError:
+                return exception.response.status_code in errs
+
+            return isinstance(exception, RETRIABLE_EXCEPTIONS)
```
martindurant, line 253: the `isinstance` check should come first, since things like `.response` don't exist on other classes of exception.

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+    def retry_request(func, retries=5):
+
+        RETRIABLE_EXCEPTIONS = (
+            requests.exceptions.ChunkedEncodingError,
+            requests.exceptions.ConnectionError,
+            requests.exceptions.ReadTimeout,
+            requests.exceptions.Timeout,
+            requests.exceptions.ProxyError,
+            requests.exceptions.SSLError,
+            requests.exceptions.ContentDecodingError,
```
martindurant, line 237: Do these capture things like socket timeout and DNS lookup timeout? Maybe not a great need to be too comprehensive at this point.

Would be requests.exceptions.RequestException (which includes all of the above) be too general?

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+        def retry_wrapper(*args, **kwargs):
+            attempts = 0
+            while attempts < retries:
+                try:
+                    return func(*args, **kwargs)
+                except (
+                        DatabricksException,
+                        requests.exceptions.RequestException,
+                        requests.exceptions.HTTPError,
+                        Exception
```
martindurant: `Exception` includes all of the other classes.

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+                ) as exc:
+                    if is_retriable(exc):
+                        if type(exc) is requests.exceptions.HTTPError:
+                            time.sleep(int(exc.response.headers["Retry-After"]))
```
martindurant, line 272: This header is not necessarily present. Furthermore, it is not necessarily an int value. In other places, it's typical to wait something like
```
base * multiplier ** attempts
```
(perhaps with a reasonable maximum wait value of 0.5s or so)

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+                        attempts += 1
+                        # logger.debug(f"{func.__name__} retrying after exception: {exc}")
+                        continue
+                    if (isinstance(exc, aiohttp.client_exceptions.ClientResponseError)
```
martindurant: This branch is no different from the next

fsspec/implementations/dbfs.py
```diff
@@ -219,6 +226,60 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
...
+            ]
+            errs += [str(e) for e in errs]
+            if type(exception) is requests.exceptions.HTTPError:
+                return exception.response.status_code in errs
+
+            return isinstance(exception, RETRIABLE_EXCEPTIONS)
+
+        def retry_wrapper(*args, **kwargs):
+            attempts = 0
+            while attempts < retries:
```
martindurant, line 257: If we run out of retries, we'll return with `None`, but should instead reraise the last exception.

fsspec/implementations/dbfs.py
```diff
@@ -381,18 +442,18 @@ class DatabricksFile(AbstractBufferedFile):
...
-        self,
-        fs,
-        path,
-        mode="rb",
-        block_size="default",
-        autocommit=True,
-        cache_type="readahead",
-        cache_options=None,
-        **kwargs,
+            self,
```
martindurant, line 445: This indenting is by `black`? It doesn't look right.

fsspec/implementations/tests/data/diabetes.csv
```diff
@@ -0,0 +1,769 @@
+Pregnancies,Glucose,BloodPressure,SkinThickness,Insulin,BMI,DiabetesPedigreeFunction,Age,Outcome
+6,148,72,35,0,33.6,0.627,50,1
```
martindurant, line 2: Why include this big file? A small sample in code would be fine for the sake of testing.

fsspec/implementations/tests/test_dbfs.py
```diff
@@ -150,10 +150,145 @@ def test_dbfs_read_range(dbfsFS):
...
+    content = b"This is a test\n" * (1 * 2 ** 18) + b"For this is the end\n"
     with dbfsFS.open("/FileStore/large_file.txt", "wb") as f:
         f.write(content)
     assert "/FileStore/large_file.txt" in dbfsFS.ls("/FileStore", detail=False)
     assert dbfsFS.cat_file("/FileStore/large_file.txt", start=8) == content[8:]
     dbfsFS.rm("/FileStore/large_file.txt")
     assert "/FileStore/large_file.txt" not in dbfsFS.ls("/FileStore/", detail=False)
+
+
+@pytest.mark.vcr()
```
martindurant, line 162: The recordings were not included.

fsspec/implementations/tests/test_dbfs.py
```diff
@@ -150,10 +150,145 @@ def test_dbfs_read_range(dbfsFS):
...
+def test_dbfs_write_pyarrow_non_partitioned(dbfsFS):
+    import pandas as pd
+    import pyarrow as pa
+    import pyarrow.parquet as pq
```
martindurant, line 166: This will fail if these are missing in the environment. Can we do without them?

fsspec/implementations/dbfs.py
```diff
@@ -226,7 +225,7 @@ def _open(self, path, mode="rb", block_size="default", **kwargs):
         """
         return DatabricksFile(self, path, mode=mode, block_size=block_size, **kwargs)
 
-    def retry_request(func, retries=5):
+    def retry_request(func, retries=3):
```
martindurant, line 228: The number of retries should ideally be configurable in the FS init

fsspec/implementations/dbfs.py
```diff
@@ -236,15 +235,16 @@ def retry_request(func, retries=5):
...
             """Returns True if this exception is retriable."""
             errs = list(range(500, 505)) + [
-                408,  # Request Timeout
-                429  # Too Many Requests
+                # Request Timeout
+                408,
+                # Too Many Requests
+                429,
             ]
             errs += [str(e) for e in errs]
```
martindurant, line 249: This gets remade every time, would be better to just write out the strings

fsspec/implementations/dbfs.py
```diff
@@ -236,15 +235,16 @@ def retry_request(func, retries=5):
...
             if type(exception) is requests.exceptions.HTTPError:
```
martindurant, line 250: isinstance is better

fsspec/implementations/tests/test_dbfs.py
```diff
@@ -184,46 +184,47 @@ def test_dbfs_write_pyarrow_non_partitioned(dbfsFS):
...
-            "/FileStore/pyarrow/diabetes/Pregnancies=13",
-            "/FileStore/pyarrow/diabetes/Pregnancies=14",
-            "/FileStore/pyarrow/diabetes/Pregnancies=15",
-            "/FileStore/pyarrow/diabetes/Pregnancies=17"
-        ])
-    ) == set()
-
-    dbfsFS.rm("/FileStore/pyarrow", recursive=True)
-    assert "/FileStore/pyarrow" not in dbfsFS.ls("/FileStore/", detail=False)
+# @pytest.mark.vcr()
```
martindurant, line 187: Is this deliberate? If the test is no good, we should remove it

fsspec/implementations/dbfs.py
a24lorie: I have declared the retry_request as a class method because this decorator has been created solely for the DatabricksFilesystem class, if you think is better as a standalone function maybe we should create a new python file for the decorator and the retry.py on the gcsfs

Same with the is_retirable function

fsspec/implementations/dbfs.py
a24lorie: I will remove it, copy-pasted and forgot to remove it

fsspec/implementations/dbfs.py
a24lorie: agree!

fsspec/implementations/dbfs.py
a24lorie: Already fixed!!

fsspec/implementations/tests/test_dbfs.py
a24lorie: I have recorded the tests and uploaded the new cassets

fsspec/implementations/tests/test_dbfs.py
a24lorie: I have commented them, is it better to remove them?

martindurant: yes please

fsspec/implementations/dbfs.py
a24lorie: I think they only capture Http requests timeouts, all of the above (and more) inherit from requests.exceptions.RequestException, should we retry on all request exceptions (I just kept the ones from the code U share)? I found the issue only with the 429 error, it doesn't make sense to retry on all the request expectations, we should agree which ones to retry

martindurant: I'm not sure! It probably does *not* make sense to retry everything, some errors are permanent. I am happy to be conservative and add extra cases to the list later, if they become necessary.

fsspec/implementations/tests/data/diabetes.csv
a24lorie: I was trying to replicate the error I was having in a real case scenario to stress the BDFS API writing concurrently many records, doing this triggered the 429 error, I could remove the file and only put a couple of records but it will not replicate the initial issue

fsspec/implementations/dbfs.py
a24lorie: OK, lets keep the ones already there as they are the same included in the gcsfs implementation (without the aiohttp.client_exceptions.ClientError)

fsspec/implementations/tests/data/diabetes.csv
martindurant: How about 
```
import concurrent.futures
fs = ...
ex = concurrent.futures.ThreadPoolExecutor()
ex.map(lambda i: fs.pipe(f"/path/{i}", b"data"), range(100))
```
You could use a mock to test that the retry is actually getting called; OR allow logging during retry (not a bad idea anyway) and check for logging statements coming out.

martindurant: Actually, we're using VCR, so you'll never get a response except the one you go the original time. I have no idea, then, how to get back a 429 without mocking

fsspec/implementations/tests/test_dbfs.py
```diff
@@ -23,6 +23,10 @@
 import os
 from urllib.parse import urlparse
 
+import numpy
+import pyarrow as pa
+import pyarrow.dataset as ds
```
martindurant, line 28: As discussed, this will fail for some. They should be
```
pa = pytest.importorskip("pyarrow")
```
inside the functions that require them.

fsspec/implementations/dbfs.py
```diff
@@ -42,13 +44,17 @@ def __init__(self, instance, token, **kwargs):
         """
         self.instance = instance
         self.token = token
-
         self.session = requests.Session()
+        self.retries = Retry(total=10,
```
martindurant, line 48: Well that's convenient - does this do everything you wanted?

fsspec/implementations/tests/test_dbfs.py
```diff
@@ -150,10 +176,57 @@ def test_dbfs_read_range(dbfsFS):
...
+    content = b"This is a test\n" * (1 * 2 ** 18) + b"For this is the end\n"
     with dbfsFS.open("/FileStore/large_file.txt", "wb") as f:
         f.write(content)
     assert "/FileStore/large_file.txt" in dbfsFS.ls("/FileStore", detail=False)
     assert dbfsFS.cat_file("/FileStore/large_file.txt", start=8) == content[8:]
     dbfsFS.rm("/FileStore/large_file.txt")
     assert "/FileStore/large_file.txt" not in dbfsFS.ls("/FileStore/", detail=False)
+
+
+# @pytest.mark.vcr()
```
martindurant, line 188: Why was this commented? This is causing the test failure, as the test is trying to connect to a real dbfs.

AlfredoLorie:
Is there anything else required to merge this feature?

fsspec/implementations/dbfs.py
martindurant, line 48: In the future we'll make this configurable, but I won't hold the PR any more

PR #1494, opened 2024-01-09
Raise FileNotFoundError when no such file or directory

QuocDuong1306:


JordiMForgeFlow (approved):

fsspec/implementations/ftp.py
```diff
@@ -170,13 +170,15 @@ def cat_file(self, path, start=None, end=None, **kwargs):
...
-        )
+        try:
+            self.ftp.retrbinary(
+                f"RETR {path}",
+                blocksize=self.blocksize,
+                rest=start,
+                callback=cb,
+            )
+        except (Error, error_perm):
+            raise FileNotFoundError(path)
```
simahawk, line 181:
```suggestion
        except (Error, error_perm) as orig_exc:
            raise FileNotFoundError(path) from orig_exc
```

martindurant:
I can merge this when it passes, but we could have done the same for get_file, and a test would be nice too.

PR #1495, opened 2024-01-10
callbacks: expose DEFAULT_CALLBACK

skshetry:
Since this is used in function signatures in `put`/`get`/`get_file`/`put_file`, which could be used in other implementations, there is no point in hiding `DEFAULT_CALLBACK`.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -11,7 +11,7 @@ from contextlib import contextmanager
 from glob import has_magic
 from typing import TYPE_CHECKING, Iterable
 
-from .callbacks import _DEFAULT_CALLBACK
+from .callbacks import DEFAULT_CALLBACK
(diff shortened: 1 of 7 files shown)


martindurant:
This is fine, but I would have argued that `_` does not mean "hidden" but "not meant to be called from outside the package".

skshetry:
> This is fine, but I would have argued that `_` does not mean "hidden" but "not meant to be called from outside the package".

The `_DEFAULT_CALLBACK` leaves a bad taste IMO, and I don't see a strong reason to discourage it's use. We have a lot of wrapper functions for cross-filesystem transfer operations that use `DEFAULT_CALLBACK` to match the `fsspec`.

I don't have a strong opinion. I'll leave the decision up to you.

PR #1496, opened 2024-01-10
callbacks: replace kwargs positional argument with arbitrary kwargs in `branched`

skshetry:
The following would be an example of someone using a `branched` callback. 

```python
with callback.branched(path_1, path_2, {}) as child:
    pass
```

The kwargs feel unnecessary to me. So I am proposing to remove that. It can be added later if someone requests it.

Also see https://github.com/fsspec/filesystem_spec/pull/1460#discussion_r1447531484.

### DIFF
--- a/fsspec/callbacks.py
+++ b/fsspec/callbacks.py
@@ -37,7 +37,7 @@ class Callback:
     def close(self):
         """Close callback."""
 
-    def branched(self, path_1, path_2, kwargs):
+    def branched(self, path_1, path_2, **kwargs):
         """
         Return callback for child transfers
 
@@ -58,8 +58,8 @@ class Callback:
             Child's source path
         path_2: str
             Child's destination path
-        kwargs: dict
-            arguments passed to child method, e.g., put_file.
+        **kwargs:
+            Arbitrary keyword arguments
 
         Returns
         -------
@@ -77,7 +77,7 @@ class Callback:
 
         @wraps(fn)
         async def func(path1, path2: str, **kwargs):
-            with self.branched(path1, path2, kwargs) as child:
+            with self.branched(path1, path2, **kwargs) as child:
                 return await fn(path1, path2, callback=child, **kwargs)
 
         return func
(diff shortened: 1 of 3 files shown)


martindurant:
Since it's brand new, I suppose it's OK, but I mildly thing `=None` or `**kwrags` is better.

After all this, we should probably update text at https://filesystem-spec.readthedocs.io/en/latest/features.html#callbacks for users and somewhere better describe the process to write new callbacks.

skshetry:
But on isolation, `**kwargs` may be more ergonomic and extensible.

fsspec/callbacks.py
```diff
@@ -37,7 +37,7 @@ def __exit__(self, *exc_args):
     def close(self):
         """Close callback."""
 
-    def branched(self, path_1, path_2, kwargs):
+    def branched(self, path_1, path_2):
```
martindurant, line 40:
```suggestion
    def branched(self, path_1, path_2, **kwargs):
```

fsspec/callbacks.py
```diff
@@ -66,6 +66,7 @@ def branched(self, path_1, path_2, kwargs):
         callback: Callback
             A callback instance to be passed to the child method
         """
+        kwargs = {}
```
martindurant, line 69:
```suggestion
```

fsspec/callbacks.py
```diff
@@ -58,8 +58,8 @@ def branched(self, path_1, path_2, kwargs):
             Child's source path
         path_2: str
             Child's destination path
-        kwargs: dict
-            arguments passed to child method, e.g., put_file.
+        **kwargs:
+            Arbitrary keyword arguments
```
skshetry, line 62: Before, this could be used to influence kwargs passed to `put_file`/`get_file`. I am not sure what the use case for that was, except of adding `callback` kwarg.

So this is deviating with `branch` in that regard.

martindurant: So you vote to remove and add back later if there is a use case?

skshetry: I don't see a use case for mutating kwargs in `branched()`. So, I'd just go with `**kwargs` (i.e. this PR).

I don't have any strong opinion against `kwargs=None` as it is similar but less ergonomic. But it does provide a way of mutating `kwargs` (but I don't see a use case for it).

We could also drop the third argument entirely, but that is not very flexible for future changes (I know I started with that, but 🤷🏽).

PR #1497, opened 2024-01-11
TqdmCallback: add support for custom tqdm subclasses

skshetry:
This will allow customizing the tqdm progressbar used for the callback, and extending (since all `tqdm_kwargs` is passed to the underlying implementation.)

E.g., in DVC, we have custom logic to align all the _progress_ _bars_ of `tqdm,` even if they have different description lengths.

This PR will enable us to use `TqdmCallback` directly in our code, without writing our own implementation.

---

```python
from tqdm import tqdm

class TqdmExtraFormat(tqdm):
    """Provides a `total_time` format parameter"""
    @property
    def format_dict(self):
        d = super(TqdmExtraFormat, self).format_dict
        total_time = d["elapsed"] * (d["total"] or 0) / max(d["n"], 1)
        d.update(total_time=self.format_interval(total_time) + " in total")
        return d

for i in TqdmExtraFormat(
      range(9), ascii=" .oO0",
      bar_format="{total_time}: {percentage:.0f}%|{bar}{r_bar}"):
    if i == 4:
        break
```

For example, there is `format_dict` that allows you to define parameters which will be used in `bar_format`. In above example, you can show an addition info: `total_time` with the extended subclass.

### DIFF
--- a/fsspec/callbacks.py
+++ b/fsspec/callbacks.py
@@ -268,6 +268,28 @@ class TqdmCallback(Callback):
             recursive=True,
             callback=TqdmCallback(tqdm_kwargs={"desc": "Your tqdm description"}),
         )
+
+    You can also customize the progress bar by passing a subclass of `tqdm`.
+
+    .. code-block:: python
+
+        class TqdmFormat(tqdm):
+            '''Provides a `total_time` format parameter'''
+            @property
+            def format_dict(self):
+                d = super().format_dict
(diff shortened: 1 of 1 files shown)


martindurant:
Would it be appropriate to provide an example snippet of usage in the prose documentation where we already mention tqdm? Alternatively, you could include the exact class above in the code with docstrings (linked in the API page) explaining that this is a way to customise the progressbar functionality.

skshetry:
I have updated the docstring with an example.

PR #1499, opened 2024-01-13
Fix logger in memory.py

Skylion007:
Loggers should never be directly instantiated as this bypasses all configuration and also will not be added to the logger tree. See ruff rule: https://docs.astral.sh/ruff/rules/direct-logger-instantiation/


Skylion007:
The failures seem unrelated?

martindurant:
This seems like correcting for an eager auto-correct :)

martindurant:
The failure was an intermittent timeout starting the SMB service

PR #1502, opened 2024-01-15
Add ruff pre-commit hook

### LINKED ISSUE #1501: Add `ruff` linter as pre-commit hook
Ruff is a linter that replaces autoflake, isort, black, and many other common pre-commit tools. It implements most of the rules in FLAKE8 currently and main of the flake8 plugins that you may want to enable for this project. It's also 10-100X faster than running flake8. The most useful feature however is that it can autofix many simple rule violations and suggest fixes for rules where it is not sure. 

I'd recommend add it as an additional pre-commit hook and gradually adding rules to it: https://docs.astral.sh/ruff/rules/ . This is how I caught #1499 for instance. You may be interested in the https://docs.astral.sh/ruff/rules/#flake8-logging-format-g rules for instance to ensure the logger doesn't add any accidental overhead. The PERF, C4, ASYNC, and other rules may also be useful to use for the project.

Skylion007:
Closes #1501 

I mostly only enabled rules that had autofixes available (and were therefore easy to enable, did not change large portions of the codebase, and are rather uncontroversial). Many of these changes are semantically identical but run more efficiently in Python (set generator -> set comprehension). Other changes make the typing more correct with the flake8-PYI linter and so on.

Good rules to enable in future PRs:
* Flake8-bugbear
* Flake8-pytest
* pyupgrade (must set minimum supported Python version)

### DIFF
--- a/fsspec/implementations/cache_mapper.py
+++ b/fsspec/implementations/cache_mapper.py
@@ -2,12 +2,8 @@ from __future__ import annotations
 
 import abc
 import hashlib
-from typing import TYPE_CHECKING
 
 from fsspec.implementations.local import make_path_posix
-
-if TYPE_CHECKING:
-    from typing import Any
 
 
 class AbstractCacheMapper(abc.ABC):
@@ -19,7 +15,7 @@ class AbstractCacheMapper(abc.ABC):
     def __call__(self, path: str) -> str:
         ...
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
         # Identity only depends on class. When derived classes have attributes
         # they will need to be included.
         return isinstance(other, type(self))
@@ -56,7 +52,7 @@ class BasenameCacheMapper(AbstractCacheMapper):
         else:
             return prefix  # No separator found, simple filename
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
         return super().__eq__(other) and self.directory_levels == other.directory_levels
 
     def __hash__(self) -> int:
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -251,7 +251,7 @@ class HTTPFileSystem(AsyncFileSystem):
             if isfilelike(lpath):
                 outfile = lpath
             else:
-                outfile = open(lpath, "wb")
+                outfile = open(lpath, "wb")  # noqa: ASYNC101
 
             try:
                 chunk = True
@@ -279,7 +279,7 @@ class HTTPFileSystem(AsyncFileSystem):
                 context = nullcontext(lpath)
                 use_seek = False  # might not support seeking
             else:
-                context = open(lpath, "rb")
+                context = open(lpath, "rb")  # noqa: ASYNC101
                 use_seek = True
 
             with context as f:
@@ -801,7 +801,7 @@ async def get_range(session, url, start, end, file=None, **kwargs):
     async with r:
         out = await r.read()
     if file:
-        with open(file, "r+b") as f:
+        with open(file, "r+b") as f:  # noqa: ASYNC101
             f.seek(start)
             f.write(out)
     else:
--- a/fsspec/implementations/tests/test_cached.py
+++ b/fsspec/implementations/tests/test_cached.py
@@ -1022,7 +1022,7 @@ def test_multi_cache(protocol):
 def test_multi_cat(protocol, ftp_writable):
     host, port, user, pw = ftp_writable
     fs = FTPFileSystem(host, port, user, pw)
-    for fn in {"/file0", "/file1"}:
+    for fn in ("/file0", "/file1"):
         with fs.open(fn, "wb") as f:
             f.write(b"hello")
 
--- a/fsspec/implementations/tests/test_tar.py
+++ b/fsspec/implementations/tests/test_tar.py
@@ -42,7 +42,7 @@ def test_info():
             assert lhs == expected
 
         # Iterate over all files.
-        for f in archive_data.keys():
+        for f in archive_data:
             lhs = fs.info(f)
 
             # Probe some specific fields of Tar archives.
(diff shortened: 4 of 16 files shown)


Skylion007:
@martindurant Okay, I enabled a bunch of new rules and added a ton of fixes.

Skylion007:
This PR also only enables the linter, not the formatter.

fsspec/implementations/cache_mapper.py
```diff
@@ -19,7 +15,7 @@ class AbstractCacheMapper(abc.ABC):
     def __call__(self, path: str) -> str:
         ...
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
```
martindurant, line 18: So lon we've been told to the the opposite, no?

fsspec/implementations/http.py
```diff
@@ -251,7 +251,7 @@ async def _get_file(
             if isfilelike(lpath):
                 outfile = lpath
             else:
-                outfile = open(lpath, "wb")
+                outfile = open(lpath, "wb")  # noqa: ASYNC101
```
martindurant, line 254: What would be the problem here?

fsspec/implementations/tests/test_cached.py
```diff
@@ -1022,7 +1022,7 @@ def test_multi_cache(protocol):
 def test_multi_cat(protocol, ftp_writable):
     host, port, user, pw = ftp_writable
     fs = FTPFileSystem(host, port, user, pw)
-    for fn in {"/file0", "/file1"}:
+    for fn in ("/file0", "/file1"):
```
martindurant, line 1025: why is a tuple better than a set? It totally doesn't matter in this case, but `contains` should be more efficient for sets, if the set in question is made only once.

fsspec/implementations/tests/test_tar.py
```diff
@@ -42,7 +42,7 @@ def test_info():
             assert lhs == expected
 
         # Iterate over all files.
-        for f, v in archive_data.items():
+        for f in archive_data.keys():
```
martindurant, line 45:
```suggestion
        for f in archive_data:
```

fsspec/implementations/tests/test_zip.py
```diff
@@ -12,7 +12,7 @@ def test_info():
         fs = fsspec.filesystem("zip", fo=z)
 
         # Iterate over all files.
-        for f, v in archive_data.items():
+        for f in archive_data.keys():
```
martindurant, line 15:
```suggestion
        for f in archive_data:
```

fsspec/tests/test_fuse.py
```diff
@@ -123,6 +123,7 @@ def test_chmod(mount_local):
         ["cp", str(mount_dir / "text"), str(mount_dir / "new")],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
+        check=False,
```
martindurant, line 126: Why the extra arg

martindurant (commented):
Some questions - maybe you know why ruff is preferring to make these changes

fsspec/implementations/tests/test_cached.py
Skylion007: This `in` isn't a contains though, it's a for loop. Tuple is more efficient to construct and iterate over than a set, I don't feel strongly about this one, happy to disable. It definitely is one of the more controversial rules, happy to disable.

fsspec/implementations/http.py
Skylion007: https://docs.astral.sh/ruff/rules/open-sleep-or-subprocess-in-async-function/ Not sure there is a good way around it without relying the aiofiles dependency though so I just suppressed it

fsspec/implementations/tests/test_tar.py
Skylion007: @martindurant If you prefer that I have a separate rule that does this automatically repo wide.

fsspec/tests/test_fuse.py
Skylion007: https://docs.astral.sh/ruff/rules/subprocess-run-without-check/ Makes you think about whether you should enable to reraise an exception if the process fails (instead of having it fail silently). Happy to disable that if you prefer though.

fsspec/implementations/http.py
martindurant: What a strange rule! Just about any normal function call blocks for some time, and `open()` ought to be fast. I don't think there's a big reason to switch to `aiofiles` right now, maybe later :)

fsspec/implementations/tests/test_cached.py
martindurant: It's fine, just wondering for my curiosity.

fsspec/implementations/tests/test_tar.py
martindurant: Using `.keys()` I thought had generally disappeared from python since 3.0; at least, if the object in question is really a dict.

fsspec/tests/test_fuse.py
martindurant: I suppose `check=True` is reasonable, then. In a test like this, we would otherwise have the failure a couple of lines later. I'm not really sure why we're using subprocess at all in this case.

fsspec/implementations/http.py
Skylion007: Yeah, I just left it because it seemed useful and we can noqa the nonsense suggestions like this. 👍

fsspec/tests/test_fuse.py
Skylion007: @martindurant Check=True made the test fail so I am just leaving it as is for now.

Skylion007:
@martindurant the one failure looks like a flake.

martindurant:
I have yet to determine why SMB fails to start sometimes

martindurant:
@Skylion007 - merging this now. I may ping you if future PRs show odd ruff failures I don't understand :).

PR #1503, opened 2024-01-16
Simple fix for FTP LIST command returning leading 'total x'

Bizarious:


fsspec/implementations/ftp.py
```diff
@@ -364,15 +364,17 @@ def _mlsd2(ftp, path="."):
     minfo = []
     ftp.dir(path, lines.append)
     for line in lines:
-        line = line.split()
+        split_line = line.split()
+        if len(split_line) < 9:
+            continue
```
martindurant, line 369: Would appreciate a comment here
```suggestion
            # not an entry, maybe something like "total: xxx"
            continue
```

Issue #1500, opened 2024-01-15
Expected behavior for "directory to new directory" depends on trailing slash


Github-dm-CDE:
In [Copying files and directories](https://filesystem-spec.readthedocs.io/en/latest/copying.html#single-source-to-single-target), 1f case "directory to new directory", it is stated that the trailing slashes in both `source` and `target` are optional. However, if `source/subdir` and `target/newdir/` are being used, the resulting target structure is the following:

```
📁 target
 └── 📁 newdir
     └── 📁 subdir
          ├── 📄 subfile1
          ├── 📄 subfile2
          └── 📁 nesteddir
              └── 📄 nestedfile
```

Basically the `subdir` folder itself is taken too, not only its content.
This can be reproduced e.g. out of the box with `LocalFileSystem()` and the `copy()` method.

Is such output expected?

martindurant:
Yes, this sounds exactly as designed, and what you would expect with a posix CLI. It also depends on whether the target directory already exists or not.

CM000n:
Is then the behaviour specified in the documentation for single source to single target case [1f. Directory to new directoy](https://filesystem-spec.readthedocs.io/en/latest/copying.html#single-source-to-single-target) incorrect? Here the documentation specifies an expected output of:

`cp("source/subdir/", "target/newdir/", recursive=True)`

```
📁 target
 └── 📁 newdir
     ├── 📄 subfile1
     └── 📄 subfile2
         └── 📁 nesteddir
             └── 📄 nestedfile
```

martindurant:
> "source/subdir/"

This path contains a terminal "/", so that is probably the difference. It would be best to make the cases with and without, versus expectations, test cases to be fixed. This is obviously not simple territory.

xebab:
The difference is definitely the trailing slash in the source path.

The thing is that the documentation explicitly states
> Trailing slashes on both source and target are optional and do not affect the result.

So the first thing to clarify would be whether there is a problem in the documentation or in the implementation.

PR #1508, opened 2024-01-18
improve test coverage

xebab:
The unit tests for get()/put()/copy() did not cover all combinations of input parameters.

### DIFF
--- a/fsspec/tests/abstract/copy.py
+++ b/fsspec/tests/abstract/copy.py
@@ -120,7 +120,7 @@ class AbstractCopyTests:
             fs.touch(dummy)
         assert fs.isdir(target)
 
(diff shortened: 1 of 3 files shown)


xebab:
These newly added tests demonstrate the case in issue #1500

martindurant:
Have you had a chance to look in the code to see how to fix this?

xebab:
You may want to consider if you want to fix this in the code at all. Because fixing the code to follow the current specification may siliently break a lot of people's applications as files will suddenly end up in a different location. Not using semantic versioning only makes this worse.

That said, it may be a better option to just update the documentation to describe the current behavior, and to fix the unit tests to match.

PR #1512, opened 2024-01-24
Fix ls with detail=False when do not support supports_empty_directories

LuQQiu:
The given dummy in the test is a string type
while the[ ls() by default is detail=True](https://github.com/fsspec/filesystem_spec/blob/0b06d92070db7b48c3499b9137c8fb6e80eee3e6/fsspec/spec.py#L313), which provides a dictionary instead of a string.
This line throws AssertionError when `supports_empty_directories=False` and the underlying filesystem does not overwrite the default `detail` to be `False`


martindurant:
Did you find a failing case without this fix?

LuQQiu:
@martindurant It's interesting that this code path is not in the existing tests that i could find.
Two conditions are needed:
-     
```
@pytest.fixture
    def supports_empty_directories(self):
        return False
```
-  def ls(self, path, detail=True, **kwargs):

Tests like local or memory filesystem test do not need to overwrite ` supports_empty_directories`
Tests like S3/Azure filesystem test set the `detail=False` by default in ls related methods https://github.com/fsspec/s3fs/blob/34a32198188164fd48d4d1abcb267f033d1d1ce1/s3fs/core.py#L966

So my hyphothesis is this line of code never got executed

In my alluxiofs tests which tests against S3 without Alluxio, it will trigger this path because alluxiofs by default ls(detailed=True) following https://github.com/fsspec/filesystem_spec/blob/0c909ee6641381fd8913729b8d803a13a985e5cb/fsspec/spec.py#L313C4-L313C47

martindurant:
(please merge from master to get tests to run - there was a dependencies versions issue)

PR #1514, opened 2024-01-26
Infer .lzma files

mrdaulet:
Currently .lzma files are not inferred correctly and instead are opened as uncompressed stream of data. This PR fixes the problem. There is no issue in the tracker related to this, as far as I can see.


martindurant:
Looks like it was originally a bad copy/paste :|

PR #1519, opened 2024-01-28
[WIP] spec: introduce path utils

### LINKED ISSUE #747: os.path equivalent for fsspec
One of the things that have come up when trying to integrate fsspec into tensorboard (https://github.com/tensorflow/tensorboard/pull/5248) is that there aren't any standard path operations as part of fsspec as far as I can tell. When dealing with more complex chained filesystems the rules get pretty complex for users to implement.

Ideally fsspec would provide the common operations such as:

* join
* basename
* dirname

I.e.
```py
>>> path = "simplecache::zip://foo::s3://bucket/path.zip"
>>> fsspec.path.join(path, "bar") 
"simplecache::zip://foo/bar::s3://bucket/path.zip"
>>> fsspec.path.dirname(path)
"foo"
>>> fsspec.path.basename(path)
"simplecache::zip://::s3://bucket/path.zip"
```

With the chained filesystems this gets really complex and I'm not 100% sure how to implement this in all cases/filesystems so feedback would be appreciated here

efiop:
Draft based on what we have in dvc, but I'm unhappy with how we try to work with both posixpath and ntpath in the base filesystem, so this is due for a simplification.

The number of helpers might seem excessive (e.g. isin/relparts/etc), but this is something that we found super useful and common when working on filesystems and I think it might make sense to keep them around to be able to use those in fs code.

Fixes #747

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -2,14 +2,16 @@ from __future__ import annotations
 
 import io
 import logging
+import ntpath
 import os
+import posixpath
 import threading
 import warnings
 import weakref
 from errno import ESPIPE
 from glob import has_magic
 from hashlib import sha256
-from typing import ClassVar
+from typing import ClassVar, Iterable, Iterator, Optional, Sequence, Tuple
 
 from .callbacks import DEFAULT_CALLBACK
 from .config import apply_config, conf
@@ -105,6 +107,7 @@ class AbstractFileSystem(metaclass=_Cached):
     _cached = False
     blocksize = 2**22
     sep = "/"
+    flavour = posixpath
     protocol: ClassVar[str | tuple[str, ...]] = "abstract"
     _latest = None
     async_impl = False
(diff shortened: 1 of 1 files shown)


skshetry:
I think we should start with a small set of APIs. `commonpath`, `join_path` and `split_path` seems fundamental to me. 

`get_cwd()` /`chdir`/`abspath`/`isabs`/`relpath`/`relparts`/`normpath`, etc. probably does not belong in fsspec, as the paths in `fsspec` are always absolute(?).

Other APIs can be derived from those 3 APIs, or `os.path.*` functions can be used on last part of the paths.

efiop:
> I think we should start with a small set of APIs. commonpath, join_path and split_path seems fundamental to me.

I agree. This is a draft, so will likely lose those unnecessary helpers in the end.

> get_cwd() /chdir/abspath/isabs/relpath/relparts/normpath, etc. probably does not belong in fsspec, as the paths in fsspec are always absolute(?).

It is more like "undefined" rather than always absolute, with the prime example of localfs which is cwd aware even though in an implicit way. I think we in dvc found a pretty elegant way to handle cwd awareness as a part of the fs and it is cheaper to have it in mind right away than to try to retrofit it in the future, esp considering that localfs and our dvcfs/gitfs/etc will be real active users of it not letting the code degrade. The overhead is really only `getcwd`/`chdir` which are easy to to turn into stubs. Though `relpath` is a completely different story and it is not really related to this `chdir` discussion at all, as computing paths relative to some other paths is a common operation (e.g. to know which new directories to create, etc) and we need it no matter what.

> Other APIs can be derived from those 3 APIs, or os.path.* functions can be used on last part of the paths.

I don't think mixing `os.path.` with fs path utils is a good practice, it will turn the code into a mess. To a reasonable extent, all necessary/core path utils should come from the fs itself as it is our `os.path`, and the rest should be built on top of it.

skshetry:
I think `localfs` is a special case. For `dvcfs`/`gitfs`, we could translate the paths at the boundary. 

> Though relpath is a completely different story and it is not really related to this chdir discussion at all, as computing paths relative to some other paths is a common operation

You are right. I was looking from fsspec's perspective, but `relpath` will be useful outside fsspec.


> I don't think mixing `os.path.` with fs path utils is a good practice, it will turn the code into a mess. To a reasonable extent, all necessary/core path utils should come from the fs itself as it is our `os.path`, and the rest should be built on top of it.

`with_name`/`with_suffix`/`suffix`/`name`/`splitext` APIs are not fs-related operations at all. Some of those are not even provided by `os.path`. They are also very simple operations to implement yourself.

martindurant:
Did we not initially discuss having an `fs.path` accessor for all these kind of methods?

efiop:
@skshetry 

> I think localfs is a special case. For dvcfs/gitfs, we could translate the paths at the boundary.

Handling cwd is not really anything special though (and one could argue that it is expected that we do that, since `os.path` and `pathlib` do) and if we can handle it in an elegant way it might be worth it, compared to having at least 1 "special" case + 2 hacks. It would be a different story if it was super complex or had no users, but it really isn't. These are my preliminary thoughts though based on what I've seen in dvc filesystems. I will have a better idea after trying to implement path utils for all fs implementations we have in this repo.

> with_name/with_suffix/suffix/name/splitext APIs are not fs-related operations at all. Some of those are not even provided by os.path. They are also very simple operations to implement yourself.

Most of those are just useful utils based on slim API. Again, I will have a better idea after trying to implement path utils for all fs implementations in this repo.

efiop:
@martindurant We did initially, but after using that in dvc for awhile, the conclusion we've reached is that `fs.<method>` makes more sense from the usability perspective https://github.com/fsspec/filesystem_spec/issues/747#issuecomment-1296356031

martindurant:
>  fs.<method> makes more sense from the usability perspective

We're making a pretty big API! I suppose it will be fairly rare for implementations to have to override many/any of the methods.

ap--:
> I think we should start with a small set of APIs. commonpath, join_path and split_path seems fundamental to me.

If `join()`, `split()`, `splitdrive()`, `normcase()`, and `isabs()` equivalent classmethods are included, file system classes could be used as pathlib-compatible "flavours" in the future, which would greatly simplify universal_pathlib's code base and would allow them to integrate well with stdlib pathlib.

_The link below shows the `PathModuleBase` base class (formerly 'Flavour') that stdlib pathlib will use to implement its string parsing._

https://github.com/barneygale/pathlib-abc/blob/b18f7eca2888f9cedbf3147d8e37aecec0b12713/pathlib_abc/__init__.py#L144-L187

efiop:
Hey folks. I think I have to admit defeat here for now, after me getting laid off from Iterative, I'm no longer actively using this and just don't have the time to spare for this anymore. I'll close this for now, if anyone is interested - please feel free to fork and take over. Thank you.

PR #1525, opened 2024-02-07
Update ruff to 0.2.1

Skylion007:
Updates ruff to v0.2.1 and modified the config accordingly


martindurant:
I know we can auto-update these, but in this case it doesn't even make any difference for the code, and having to make or approve PRs constantly is a bit annoying. Do you have thoughts on how to keep up?

Skylion007:
@martindurant Just enable pre-commit CI and you can have it make the update PRs on a schedule (every 2 weeks, every 2 months, etc).

PR #1531, opened 2024-02-28
POC: transactions on simplecache

martindurant:


fsspec/implementations/cached.py
```diff
@@ -868,15 +888,18 @@ def discard(self):
...
+    def __repr__(self) -> str:
+        return f"LocalTempFile: {self.path}"
+
     def __getattr__(self, item):
-        return getattr(self.fh, item)
+        try:
+            return getattr(self.fh, item)
+        except AttributeError:
+            pass
+        raise AttributeError(item)
```
Skylion007, line 905: why not use a `raise from` here to reraise AttributeError and not through away in the info of the previous attribute error?

Skylion007: It will look like it failed from this object, but it will give additional information about it? It will just additional information. I'm worried about AttributeError information being deleted here.

martindurant: fine...

PR #1535, opened 2024-03-05
Preserve Local cp_file metadata. Fixes #1524

### LINKED ISSUE #1524: `LocalFileSystem` clobbers file permissions after cp_file
Hi,

I am trying to copy a local file with the following permissions:
```
-rwxrwxr-x 1 rhorvath rhorvath 7994 
```
after copying the file it becomes:
```
-rw-rw-r--  1 rhorvath rhorvath 7994
```
I have narrowed the reason for this down to the cp_file function in `LocalFileSystem` where `shutil.copyfile(path1, path2)` doesn't copy the permission mode. This could be solved by using [shutil.copy()](https://docs.python.org/3/library/shutil.html#shutil.copy) and preserving the source file's permission mode.

Skylion007:
Preserves permissions when copying local file. Fixes #1524

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -122,6 +122,7 @@ class LocalFileSystem(AbstractFileSystem):
             self.makedirs(self._parent(path2), exist_ok=True)
         if self.isfile(path1):
             shutil.copyfile(path1, path2)
+            shutil.copystat(path1, path2)
         elif self.isdir(path1):
             self.mkdirs(path2, exist_ok=True)
         else:


Skylion007:
Actually, the more I think about this. The less sure I am about whether it should preserve this copied metadata by default. A lot of os system calls in Python do not... It wouldn't be add as a configuration option though...

martindurant:
Well... the default implementation for cp (like `open(out).write(open(in).read())` ) transfers only data, since it is generic over al backends. I don't think any of the backend cp implementations explicitly transfer metadata, although it may be good to do an audit on this. 

posix/bash's cp *does not* copy all access attributes:
```
$ chmod a+wrx TEMP
$ cp TEMP TEMP2
$ ll TEMP*
-rwxrwxrwx  1 mdurant  staff  1513  6 Mar 09:29 TEMP
-rwxr-xr-x  1 mdurant  staff  1513  6 Mar 09:29 TEMP2
```

PR #1536, opened 2024-03-06
Fix #1476 : do not ignore expand arg in open_files

### LINKED ISSUE #1476: `fsspec.open()` doesn't work with square brackets in a filename
For example, `fsspec.open("filename[test].csv")`.

Is this by design or a bug? (python builtin `open` function works in this case, at least for python 3.11)

Skylion007:
Fixes #1476 . Turns out it was caused by a bug where an arg was not respected when calling `open_files` . This PR fixes that bug and adds some detailed regression test for that behavior

### DIFF
--- a/fsspec/tests/test_core.py
+++ b/fsspec/tests/test_core.py
@@ -247,7 +247,7 @@ def test_pickle_after_open_open():
 # Add tests for more file systems and for more glob magic later
 glob_magic_characters = ("[", "]", "!")
 
-# Unit tests for issue #1476 (expand arg ignored in open file)
+# === Start of unit tests for issue #1476 (expand arg ignored in open file) ===
 
 
 @pytest.mark.parametrize("char", glob_magic_characters)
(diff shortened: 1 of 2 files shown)


martindurant:
Well done, looks like you found it :)

fsspec/tests/test_core.py
```diff
@@ -239,6 +239,104 @@ def test_pickle_after_open_open():
     of2.close()
 
 
+# Define a list of special glob characters.
+# Note that we need to escape some characters and also consider file system limitations.
+# '*' and '?' are excluded because they are not valid for many file systems.
+# Similarly, we're careful with '{', '}', and '@' as their special meaning is
+# context-specific and might not be considered special for filenames.
+# Add tests for more file systems and for more glob magic later
+glob_magic_characters = ("[", "]", "!")
```
martindurant, line 248: Include `*` and `?` ?

Skylion007:
@martindurant '*' and '?' aren't valid in Windows file systems... How should I modify the test suite?

martindurant:
We have a WIN variable somewhere (`sys.platform == "nt"`, I think), so you can define a different set of characters in a simple if.

Skylion007:
> We have a WIN variable somewhere (`sys.platform == "nt"`, I think), so you can define a different set of characters in a simple if.

True, but someone could be running tests locally on a Linux system mounted on NTFS or some other filesystem that doesn't support these chars and the tests would fail.

martindurant:
I see what you mean, but I am not worried about that - it's really unlikely to have ntfs as the temp location.

PR #1544, opened 2024-03-13
Limit coroutines using a pool instead of chunks

sisp:
I've changed the implementation of the function `_run_coros_in_chunks()` by using a coroutine pool instead of running coroutines in chunks.

The goal of this function is to limit the number of simultaneous coroutines, but chunking may lead to inferior throughput when coroutines have different execution times, as the next chunk will run only once all coroutines in the previous chunk have completed. In contrast, a coroutine pool will run the next coroutine in the queue as soon as any of the running coroutines completes.

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -239,21 +239,19 @@ async def _run_coros_in_chunks(
         batch_size = len(coros)
 
     assert batch_size > 0
-    results = []
-    for start in range(0, len(coros), batch_size):
-        chunk = [
-            asyncio.Task(asyncio.wait_for(c, timeout=timeout))
-            for c in coros[start : start + batch_size]
-        ]
+    semaphore = asyncio.BoundedSemaphore(batch_size)
+
+    async def _worker(coro):
+        async with semaphore:
+            return await coro
+
+    tasks = []
+    for coro in coros:
+        task = asyncio.create_task(_worker(asyncio.wait_for(coro, timeout=timeout)))
         if callback is not DEFAULT_CALLBACK:
-            [
-                t.add_done_callback(lambda *_, **__: callback.relative_update(1))
-                for t in chunk
-            ]
-        results.extend(
-            await asyncio.gather(*chunk, return_exceptions=return_exceptions),
-        )
-    return results
+            task.add_done_callback(lambda *_, **__: callback.relative_update(1))
+        tasks.append(task)
+    return await asyncio.gather(*tasks, return_exceptions=return_exceptions)
 
 
 # these methods should be implemented as async by any async-able backend
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -239,19 +239,39 @@ async def _run_coros_in_chunks(
         batch_size = len(coros)
 
     assert batch_size > 0
-    semaphore = asyncio.BoundedSemaphore(batch_size)
 
-    async def _worker(coro):
-        async with semaphore:
-            return await coro
+    async def _run_coro(coro):
+        try:
+            return await asyncio.wait_for(coro, timeout=timeout)
+        except Exception as e:
+            if not return_exceptions:
+                raise
+            return e
+        finally:
+            callback.relative_update(1)
 
-    tasks = []
-    for coro in coros:
-        task = asyncio.create_task(_worker(asyncio.wait_for(coro, timeout=timeout)))
-        if callback is not DEFAULT_CALLBACK:
-            task.add_done_callback(lambda *_, **__: callback.relative_update(1))
-        tasks.append(task)
-    return await asyncio.gather(*tasks, return_exceptions=return_exceptions)
+    coros = iter(coros)
+    stop = False
+    pending = set()
+    results = []
+
+    while pending or not stop:
+        while len(pending) < batch_size and not stop:
+            try:
+                coro = next(coros)
+            except StopIteration:
+                stop = True
+            else:
+                pending.add(asyncio.ensure_future(_run_coro(coro)))
+
+        if not pending:
+            break
+
+        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
+        while done:
+            results.append(await done.pop())
+
+    return results
 
 
 # these methods should be implemented as async by any async-able backend
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -240,36 +240,33 @@ async def _run_coros_in_chunks(
 
     assert batch_size > 0
 
-    async def _run_coro(coro):
+    async def _run_coro(coro, i):
         try:
-            return await asyncio.wait_for(coro, timeout=timeout)
+            return await asyncio.wait_for(coro, timeout=timeout), i
         except Exception as e:
             if not return_exceptions:
                 raise
-            return e
+            return e, i
         finally:
             callback.relative_update(1)
 
-    coros = iter(coros)
-    stop = False
+    i = 0
+    n = len(coros)
+    results = [None] * n
     pending = set()
-    results = []
 
-    while pending or not stop:
-        while len(pending) < batch_size and not stop:
-            try:
-                coro = next(coros)
-            except StopIteration:
-                stop = True
-            else:
-                pending.add(asyncio.ensure_future(_run_coro(coro)))
+    while pending or i < n:
+        while len(pending) < batch_size and i < n:
+            pending.add(asyncio.ensure_future(_run_coro(coros[i], i)))
+            i += 1
 
         if not pending:
             break
 
         done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
         while done:
-            results.append(await done.pop())
+            result, k = await done.pop()
+            results[k] = result
 
     return results
 
--- a/ci/environment-friends.yml
+++ b/ci/environment-friends.yml
@@ -21,6 +21,7 @@ dependencies:
   - flake8
   - black
   - google-cloud-core
+  - google-cloud-storage
   - google-api-core
   - google-api-python-client
   - httpretty


sisp:
The check `CI / gcsfs-pytest` is failing with this error:

```
FAILED gcsfs/gcsfs/tests/test_core.py::test_sign - ImportError: cannot import name 'storage' from 'google.cloud' (unknown location)
```

But I doubt it's related to this PR.

fsspec/asyn.py
```diff
@@ -239,21 +239,19 @@ async def _run_coros_in_chunks(
...
-    results = []
-    for start in range(0, len(coros), batch_size):
-        chunk = [
-            asyncio.Task(asyncio.wait_for(c, timeout=timeout))
-            for c in coros[start : start + batch_size]
-        ]
+    semaphore = asyncio.BoundedSemaphore(batch_size)
+
+    async def _worker(coro):
+        async with semaphore:
```
martindurant, line 245: With this approach, I agree that a limited number of IO tasks are concurrently running/waiting, but is there any potential downside of simply having a very large number of tasks created and waiting on this semaphore? The previous incantation would only have the given number of top-level tasks at a time.

fsspec/asyn.py
```diff
@@ -239,21 +239,19 @@ async def _run_coros_in_chunks(
...
+            return await coro
+
+    tasks = []
+    for coro in coros:
+        task = asyncio.create_task(_worker(asyncio.wait_for(coro, timeout=timeout)))
```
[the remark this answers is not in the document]
martindurant: 👍

fsspec/asyn.py
sisp: Good question. I'm not aware of any downsides but can't prove there aren't any. I was also considering to use [`asyncio.as_completed`](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed) and pass an iterable of `asyncio.Task`s, but it also materializes this iterator as a set internally.

martindurant: Let's leave this question open a little while, to see if any knowledgable person comments.

sisp: FWIW, I've used this implementation to limit the simultaneous execution of 70,000 coroutines and it worked fine.

martindurant: Would you like to try making a queue? It might turn out to be simpler.

Skylion007: https://death.andgravity.com/limit-concurrency This recent blogpost has some really nice solutions to this problem.

sisp: Thanks for the pointers, @Skylion007! 👌 I got a queue-based solution working today except when `return_exceptions=False`. But the blog article you referenced offers a much more comprehensive analysis and different solution.

sisp: I've pushed an implementation based on `asyncio.wait` inspired by https://death.andgravity.com/limit-concurrency#asyncio-wait. I think this looks good and tests are passing.

fsspec/asyn.py
```diff
@@ -239,20 +239,38 @@ async def _run_coros_in_chunks(
...
+                coro = next(coros)
+            except StopIteration:
+                stop = True
+            else:
+                pending.add(asyncio.ensure_future(_run_coro(coro)))
+
+        if not pending:
+            break
+
+        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
```
martindurant, line 270: It feels like this will alter the ordering of the results versus the inputs. Can you please try doing a `cat` with many files, to see if the order is preserved? I don't know that we have tests with genuinely significant numbers and latency.

sisp: Ah, yes, I'm quite sure the order is not preserved. I think neither a queue-based approach nor this one would be able to preserve the order because coroutines are popped when done and appended to the results list. But since the coroutines input is a list and not a generator, we know its length. So, we can allocate a same-length `results` list and insert the results at the same index of the corresponding coroutine.

Let me add this quickly for you to review.

sisp: Mh, but list insertions at some index have complexity O(n). :confused:

sisp: Done.

fsspec/asyn.py
```diff
@@ -239,20 +239,35 @@ async def _run_coros_in_chunks(
...
+    results = [None] * n
+    pending = set()
+
+    while pending or i < n:
+        while len(pending) < batch_size and i < n:
+            pending.add(asyncio.ensure_future(_run_coro(coros[i], i)))
+            i += 1
+
+        if not pending:
+            break
```
sisp, line 264: I wonder whether this condition is actually needed. :thinking:

The outer `while` loop continues only if `pending` is not empty or there are still coroutines left to add to `pending`. If `pending` is empty, then the inner `while` loop will _always_ add at least one coroutine to `pending` because `len(pending) < batch_size` is always true (because `pending` is empty and `batch_size > 0` according to the assertions further above). And if `pending` is not empty, then `i < n` may not hold (i.e. no coroutines are left), but then `if not pending` is `False` anyway.

fsspec/asyn.py
martindurant: So you would need a dict, assuming the inputs or coroutines are hashable

sisp: Sorry, I've updated by comment in the meantime. Setting items in lists is O(1), so all is good.

fsspec/asyn.py
martindurant: I doubt it hurts - we can check in coverage whether the line ever gets hit; but very large scale calls may behave differently and reveal corner cases. It's worth remembering that function may get called simultaneously from multiple threads.

fsspec/asyn.py
Skylion007: I actually think the linked solution may be more efficient. Does this solution run into the issue that the next batch does not start until the current batch finishes? That means with large batch size, future workers will be stalled waiting on the slowest worker?

Or am I misunderstanding the current batch async method.

fsspec/asyn.py
Skylion007: A bit concerned this may still be inefficient in scheduling the workers. IE: it waits until the batch is completed before the next batch is scheduled: https://github.com/fsspec/filesystem_spec/pull/1544#discussion_r1526491315

fsspec/asyn.py
sisp: > Does this solution run into the issue that the next batch does not start until the current batch finishes? That means with large batch size, future workers will be stalled waiting on the slowest worker?

Correct, the slowest coroutine in a batch determines the runtime of the whole batch and the next batch won't start before all coroutines in the current batch have finished.

fsspec/asyn.py
sisp: I don't think this happens here becausw of `return_when=asyncio.FIRST_CIMPLETED`.

sisp: When I print `len(pending)` like this

```diff
     pending = set()
 
     while pending or i < n:
+        print("DEBUG: len(pending) =", len(pending))
         while len(pending) < batch_size and i < n:
             pending.add(asyncio.ensure_future(_run_coro(coros[i], i)))
             i += 1
```

and modify the `test_run_coros_in_chunks` test in `tests/test_async.py` like this

```diff
     async def runner():
         nonlocal total_running
+        import random
 
         total_running += 1
-        await asyncio.sleep(0)
+        await asyncio.sleep(random.uniform(0.0, 1.0))
         if total_running > 4:
             raise ValueError("More than 4 coroutines are running together")
         total_running -= 1
```

then the printed info is this for me:

<details>

<summary>Log</summary>

```
DEBUG: len(pending) = 0
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 2
DEBUG: len(pending) = 1
DEBUG: len(pending) = 0
DEBUG: len(pending) = 0
DEBUG: len(pending) = 0
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 2
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 2
DEBUG: len(pending) = 1
DEBUG: len(pending) = 0
DEBUG: len(pending) = 0
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 2
DEBUG: len(pending) = 1
DEBUG: len(pending) = 0
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 3
DEBUG: len(pending) = 2
DEBUG: len(pending) = 1
```

</details>

I believe it works as we expect. WDYT, @martindurant?

fsspec/asyn.py
Skylion007: I think some of the solutions I linked would have alleviated that issue so it could have eagerly continued the next batch.

sisp: The solution that @martindurant merged is essentially https://death.andgravity.com/limit-concurrency#asyncio-wait, which you linked.

PR #1545, opened 2024-03-15
[Docs] Add a Prevailing Library User

JalinWang:
I unexpectedly discovered that the data tool of the most popular Deep Learning/Transformer library also relies on fsspec, and given its widespread popularity, I believe it would be worthwhile to contribute a pull request (PR) for it :)

https://github.com/huggingface/datasets/issues/6214
https://github.com/search?q=repo%3Ahuggingface%2Fdatasets%20fsspec&type=code

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -53,6 +53,8 @@ The following libraries use ``fsspec`` internally for path and file handling:
    maintainable and modular data science code
 #. `pyxet`_, a Python library for mounting and
    accessing very large datasets from XetHub
+#. `Huggingface🤗 Datasets`_, a popular library to 
+   load&manipulate data for Deep Learning models
 
 ``fsspec`` filesystems are also supported by:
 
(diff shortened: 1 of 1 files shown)


martindurant:
Huggingface doesn't just use fsspec, but has a dedicated implementation, as listed here: https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations .

Do you think that's not prominent enough?

JalinWang:
Apologies for the delay in response. I raised this pull request from the perspective of a new user. 

As a newcomer, I only glanced through the [documentation index page](https://filesystem-spec.readthedocs.io/en/latest/) to understand its overall functionality/popularity and potential usage in my current or future projects. It's unlikely for newcomers to delve deeply into the documentation, navigate to the API page, and scroll through half of it to find the "Other Known Implementations" entry, especially considering they are the ones who need the "Who uses fsspec?" section the most :(

I believe it would be helpful to at least link `Other Known Implementations` in the `Who uses fsspec?` section.
Additionally, considering the popularity of the Hugging Face repository (e.g., currently ranked #30 according to a [Gitstar Ranking](https://gitstar-ranking.com/repositories)), I don't see an issue with mentioning it twice in the documentation, similar to how the DVC repository is presented.

image

martindurant:
Sure, your suggestion is fine - can you please add a "see known implementations" link too, then? Probably the who-uses section hasn't been updated in a long time, and a lot of improvements could be made.

PR #1547, opened 2024-03-15
improve HTTP debug logging

Skylion007:
Use `exc_info` to provide the exception info instead.


martindurant:
SMB still being annoying - rerunning! In the previous PR, I made it retry on initial connect, but it seems like the connection can drop any time (in CI only); it would be nice sometime to add a general retry method to SMB which re-establishes the connection every time it goes away.

PR #1550, opened 2024-03-17
Fix recursive copy in WebHDFS

dsalaj:
Recursive copy with `WebHDFS` implementation does not handle copying directories. This simple patch should match the behavior of recursive copy in `WebHDFS` to that of the `LocalFileSystem` implementation.


dsalaj:
I am willing to write the additional tests for recursive copy of `WebHDFS` if needed.

PR #1551, opened 2024-03-18
Allow kwargs passed to open() with overrides

martindurant:


fsspec/core.py
```diff
@@ -456,6 +456,8 @@ def open(
     - For implementations in separate packages see
       https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations
     """
+    kw = {"expand": False}
+    kw.update(kwargs)
```
Skylion007, line 460: i would use dict.setdefault in the future. :)

fsspec/implementations/local.py
```diff
@@ -127,6 +127,14 @@ def cp_file(self, path1, path2, **kwargs):
         else:
             raise FileNotFoundError(path1)
 
+    def isfile(self, path):
```
Skylion007, line 130: We really never supported these? 0o0

martindurant: Previously this was supported via a call to info() in the superclass, which does stat to get sizes and permissions, and so is slower.

fsspec/core.py
martindurant: Following the conversation, `expand` should probably be promoted to the signature of open() to help those wanting to be explicit.

PR #1556, opened 2024-03-22
fix: always convert blocks from list to set when loading cache metadata

### LINKED ISSUE #1555: `AttributeError: 'list' object has no attribute 'update'` in cache_metadata.py
We're seeing an occasional error after turning on blockcache in our production environments where the cached objects are failing to save with this traceback (some of it has been elided for brevity:

```
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cached.py", line 449, in <lambda>
    return lambda *args, **kw: getattr(type(self), item).__get__(self)(
                   │       │                │      │             └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       │                │      └ 'open'
                   │       │                └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       └ {'mode': 'rb'}
                   └ ('some-file.parquet',)
  File "/app/.venv/lib/python3.11/site-packages/fsspec/spec.py", line 1293, in open
    f = self._open(
        │    └ <function CachingFileSystem._open at 0x7bc9772b9ee0>
        └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cached.py", line 449, in <lambda>
    return lambda *args, **kw: getattr(type(self), item).__get__(self)(
                   │       │                │      │             └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       │                │      └ '_open'
                   │       │                └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       └ {'mode': 'rb', 'block_size': None, 'autocommit': True, 'cache_options': None}
                   └ ('some-file.parquet',)
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cached.py", line 368, in _open
    self.save_cache()
    │    └ <function CachingFileSystem.save_cache at 0x7bc9772b9b20>
    └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cached.py", line 449, in <lambda>
    return lambda *args, **kw: getattr(type(self), item).__get__(self)(
                   │       │                │      │             └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       │                │      └ 'save_cache'
                   │       │                └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
                   │       └ {}
                   └ ()
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cached.py", line 206, in save_cache
    self._metadata.save()
    │    │         └ <function CacheMetadata.save at 0x7bc9772b9760>
    │    └ <fsspec.implementations.cache_metadata.CacheMetadata object at 0x7bc8e036c990>
    └ <fsspec.implementations.cached.CachingFileSystem object at 0x7bc9381dbf50>
  File "/app/.venv/lib/python3.11/site-packages/fsspec/implementations/cache_metadata.py", line 212, in save
    blocks.update(c["blocks"])
    │             └ {'original': 'some-file.parquet',...
    └ []

AttributeError: 'list' object has no attribute 'update'
```

I dug into the code a little bit and I _think_ this is because in `CacheMetadata.save()` the metadata is being loaded from JSON using `CacheMetadata._load()`, which doesn't convert the `blocks` field of each cached file from a `list` to a `set` (this is handled separately in `CacheMetadata.load()`).

It's a bit odd that we only see the error occasionally but perhaps we have a lot of cache misses?

sd2k:
Currently the `load()` method handles the conversion but the `_load()` method doesn't, even though it's used by `save()`.

This commit moves the conversion to the `_load()` method so that the blocks are always of the expected type.

I've struggled to find a way to test this, would appreciate any advice there!

Fixes #1555.

### DIFF
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -58,13 +58,13 @@ class CacheMetadata:
         try:
             with open(fn, "r") as f:
                 loaded = json.load(f)
-                for c in loaded.values():
-                    if isinstance(c.get("blocks"), list):
-                        c["blocks"] = set(c["blocks"])
-                return loaded
         except ValueError:
             with open(fn, "rb") as f:
-                return pickle.load(f)
+                loaded = pickle.load(f)
+        for c in loaded.values():
+            if isinstance(c.get("blocks"), list):
+                c["blocks"] = set(c["blocks"])
+        return loaded
 
     def _save(self, metadata_to_save: Detail, fn: str) -> None:
         """Low-level function to save metadata to specific file"""
--- a/fsspec/implementations/cache_metadata.py
+++ b/fsspec/implementations/cache_metadata.py
@@ -57,7 +57,11 @@ class CacheMetadata:
         """Low-level function to load metadata from specific file"""
         try:
             with open(fn, "r") as f:
-                return json.load(f)
+                loaded = json.load(f)
+                for c in loaded.values():
+                    if isinstance(c.get("blocks"), list):
+                        c["blocks"] = set(c["blocks"])
+                return loaded
         except ValueError:
             with open(fn, "rb") as f:
                 return pickle.load(f)
@@ -152,11 +156,7 @@ class CacheMetadata:
         for fn, _, _ in self._scan_locations():
             if os.path.exists(fn):
                 # TODO: consolidate blocks here
-                loaded_cached_files = self._load(fn)
-                for c in loaded_cached_files.values():
-                    if isinstance(c["blocks"], list):
-                        c["blocks"] = set(c["blocks"])
-                cached_files.append(loaded_cached_files)
+                cached_files.append(self._load(fn))
             else:
                 cached_files.append({})
         self.cached_files = cached_files or [{}]


martindurant:
I think this makes sense, and indeed if you can't think of a way to expose the original error, I can't think of how to test it explicitly (except that nothing new breaks).

martindurant:
It seems the specific error you were wanting to avoid now shows up in two of the runs
```
>                           blocks.update(c["blocks"])
E                           AttributeError: 'list' object has no attribute 'update'
```

sd2k:
Ah, thanks, it looks like those tests were skipped when I ran them locally originally. I've installed all the required dependencies now and pushed a fix; it looks like we need to convert the field to a set even when loading from pickle. Tests should pass now!

Skylion007:
beartest could be used to test these as well as asserts. mypy should be sufficient too with proper type hints.

sd2k:
@martindurant we're running into this a lot in prod - would it be possible to get a new release of fsspec with this fix in? :pray:

martindurant:
I'll get it done in the next couple of daya

PR #1558, opened 2024-03-25
Set expand to True by default but make configurable

martindurant:
Fixes https://github.com/fsspec/gcsfs/issues/616 and perhaps others

cc @Skylion007 - this is what I was proposing, so you can override with argument to open(), for the whole session using the config, or temporarily using the module variable. The only question: should the same default chain exist in open_files?

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -8,7 +8,7 @@ from glob import has_magic
 from pathlib import Path
 
 # for backwards compat, we export cache things from here too
-from .caching import (  # noqa: F401
+from fsspec.caching import (  # noqa: F401
     BaseCache,
     BlockCache,
     BytesCache,
@@ -16,9 +16,10 @@ from .caching import (  # noqa: F401
     ReadAheadCache,
     caches,
 )
-from .compression import compr
-from .registry import filesystem, get_filesystem_class
-from .utils import (
+from fsspec.compression import compr
+from fsspec.config import conf
+from fsspec.registry import filesystem, get_filesystem_class
+from fsspec.utils import (
     _unstrip_protocol,
     build_name_function,
     infer_compression,
@@ -396,6 +397,9 @@ def url_to_fs(url, **kwargs):
     return fs, urlpath
 
 
+DEFAULT_EXPAND = conf.get("open_expand", True)
+
+
 def open(
     urlpath,
     mode="rb",
@@ -404,6 +408,7 @@ def open(
     errors=None,
     protocol=None,
     newline=None,
+    expand=None,
     **kwargs,
 ):
     """Given a path or paths, return one ``OpenFile`` object.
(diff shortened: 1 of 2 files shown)


Skylion007:
I propose this should not be the default. expand should not be the default as it seems like a footgun to use this feature without explicitly enabling it.

fsspec/core.py
```diff
@@ -396,6 +397,9 @@ def url_to_fs(url, **kwargs):
     return fs, urlpath
 
 
+DEFAULT_EXPAND = conf.get("open_expand", True)
```
Skylion007, line 400: I don't think `DEFAULT_EXPAND` should be True. It's a surprising behavior/footgun. It violates the Zen of Python `explicit is better than implicit`.

Skylion007 (changes requested):
We really should not have `open_expand=True`. It was a defect before and we shouldn't complicate future support just for the sake of bw compatibility for an unsupported edgecase in one popular library `(pandas)`.

Skylion007:
`open_files` has always taken an expand=True, expand=False arg, there was just a bug where it wasn't respected before. I don't think we should support this old behavior as it was an error. It's difficult to throw a deprecation warning here either and the lack of semantic versioning in this repo complicates things. We could add some better logging maybe on the filenotfound path is a filename has magic chars maybe? We could provide a suggestion to the user to explicitly set the expand=True etc in that case.

`fsspec.open's` call structure is so similar to builtins.open that I often use it as a drop in replacement. There really only a couple of edge cases (append filemode etc), opening the file outside a with context, that requires any adjustment. As such, I'm strongly against reverting the default behavior to the old, buggy one.

Finally, the old behavior, even when working properly is a footgun. Even if the glob is expanded, it's unexpected that will it open the first file it matches without erroring. What if I want to `open` a glob of CSV files and have special handling for that if `fsspec.open` returns file not found? The old behavior would incorrectly only open the first `csv` file silently and that could lead silent data corruption. The new behavior would require me to explicitly handle globs myself.

martindurant:
> We could add some better logging maybe on the filenotfound path is a filename has magic chars maybe?

This is reasonable: catch FileNotFound in open(), and if there are any special characters, add extra information to the exception saying they might want to try expand=True.

martindurant:
Everyone happy with the wording and test here?

shobsi:
> > We could add some better logging maybe on the filenotfound path is a filename has magic chars maybe?
> 
> This is reasonable: catch FileNotFound in open(), and if there are any special characters, add extra information to the exception saying they might want to try expand=True.

I think this is the best trade-off having learnt that the old behavior was an unintended one.

fsspec/core.py
```diff
@@ -428,6 +444,13 @@ def open(
     newline: bytes or None
         Used for line terminator in text mode. If None, uses system default;
         if blank, uses no translation.
+    expand: bool or Nonw
+        Whether to regard file paths containing special glob characters as needing
+        expansion (finding the first match) or absolute. Setting False allows using
+        paths which do embed such characters. If None (default), this argument
+        takes its value from the DEFAULT_EXPAND module variable, which takes
+        its initial value from the "open_expand" config value at startup, which will
+        be True if not set.
```
martindurant, line 453:
```suggestion
        be False if not set.
```

Issue #1559, opened 2024-03-25
fs.mv swallows copy errors


hugabora:
fsspec.py
https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L1071
definition of copy has on_error param

mv does not take or pass this parameter into copy
https://github.com/fsspec/filesystem_spec/blob/master/fsspec/spec.py#L1173

Result is that errors from copy get swallowed and not raised when recursive=True

PR #1560, opened 2024-03-25
dev/hugabora/issue-1559 mv swallows copy errors for recursive

hugabora:
Created an issue ticket https://github.com/fsspec/filesystem_spec/issues/1559
Added on_error to fsspec mv and it is passed on now

```python
# fsspec-test.py
import s3fs

source_path = './fsspec-test.py'
target_path = 's3://good-bucket/foo'
move_path = 's3://BAD_BUCKET/foo'

s3 = s3fs.S3FileSystem(anon=False)
s3.put(source_path, target_path)
s3.mv(target_path, move_path, recursive=True, on_error='raise')
```

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1170,12 +1170,12 @@ class AbstractFileSystem(metaclass=_Cached):
             raise FileNotFoundError(path)
         return sorted(out)
 
-    def mv(self, path1, path2, recursive=False, maxdepth=None, **kwargs):
+    def mv(self, path1, path2, recursive=False, maxdepth=None, on_error=None, **kwargs):
         """Move file(s) from one location to another"""
         if path1 == path2:
             logger.debug("%s mv: The paths are the same, so no files were moved.", self)
         else:
-            self.copy(path1, path2, recursive=recursive, maxdepth=maxdepth)
+            self.copy(path1, path2, recursive=recursive, maxdepth=maxdepth, on_error)
             self.rm(path1, recursive=recursive)
 
     def rm_file(self, path):
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1175,7 +1175,7 @@ class AbstractFileSystem(metaclass=_Cached):
         if path1 == path2:
             logger.debug("%s mv: The paths are the same, so no files were moved.", self)
         else:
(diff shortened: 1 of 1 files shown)


fsspec/spec.py
```diff
@@ -1170,12 +1170,12 @@ def expand_path(self, path, recursive=False, maxdepth=None, **kwargs):
...
 
-    def mv(self, path1, path2, recursive=False, maxdepth=None, **kwargs):
+    def mv(self, path1, path2, recursive=False, maxdepth=None, on_error=None, **kwargs):
         """Move file(s) from one location to another"""
         if path1 == path2:
             logger.debug("%s mv: The paths are the same, so no files were moved.", self)
         else:
-            self.copy(path1, path2, recursive=recursive, maxdepth=maxdepth)
+            self.copy(path1, path2, recursive=recursive, maxdepth=maxdepth, on_error=on_error)
             self.rm(path1, recursive=recursive)
```
martindurant, line 1179: Also pass to `rm` ?

martindurant (commented):
I would appreciate a test showing that this makes a difference, i.e., pass a non-existent path to be deleted and see it raise in one case and be ignored i the other.

fsspec/spec.py
hugabora: I did not see this parameter in the rm methods' definitions. So not sure how to pass that into these.
Likely those never swallow errors, only copy does w recursive, right?

hugabora:
1. 

> I would appreciate a test showing that this makes a difference, i.e., pass a non-existent path to be deleted and see it raise in one case and be ignored i the other.

I provided a sample code in this PR text, that is how I tested the change.

I did not create a test case as code though.
I looked at `test_spec.py` I do not see a test for `mv` to build on.
Can you advise on how to do it, or just add that the way it is best for the project?

martindurant:
You should choose one of the implementations, any one that depends on the default implementation in AbstractFileSystem. So LocalFileSystem probably does not (could try), but the tests for MemoryFileSystem and FTPFileSystem are good places (see in fsspec/implementations/tests), and they have many of the generic tests covering the superclass.

To be more thorough, in fsspec/tests/abstract are bulk operation tests that will get applied to many backends (including s3fs), and `mv` would be a natural thing to sit alongside cp/put/get - so you could copy the pattern there.

PR #1561, opened 2024-03-26
spec.py, AbstractFileSystem._ls_from_cache: Avoid looking up key in dircache twice

KellyWalker:


KellyWalker:
@martindurant the CI/ 3.8-pytest check failed because of some tests in test_smb.py. Looking over that test file, it does not seem like the change I made should have an affect on the behavior of the tests that failed. Is it possible the failure was unrelated and the check could be re-run?

martindurant:
There is known flakiness in SMB that I haven't had a chance to trace yet. Rerunning.

PR #1566, opened 2024-04-04
Tracking cache requests

betolink:
As discussed in https://github.com/fsspec/filesystem_spec/discussions/1527 this PR adds a few extra variables in the cache implementation to keep track of bytes requested vs cache hits. The numbers can be accessed directly or via the logs in debug mode. 

This is also useful to benchmark the I/O behavior of the different cache implementations.

### DIFF
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -84,7 +84,7 @@ class BaseCache:
                 cache type  :   {self.__class__.__name__} 
                 block size  :   {self.blocksize}
                 block count :   {self.nblocks}
-                cache size  :   {self.size}
+                file size  :   {self.size}
                 cache hits  :   {self.hit_count}
                 cache misses:   {self.miss_count}
                 total requested bytes: {self.total_requested_bytes}
@@ -233,12 +233,20 @@ class FirstChunkCache(BaseCache):
     name = "first"
 
     def __init__(self, blocksize: int, fetcher: Fetcher, size: int) -> None:
+        if blocksize > size:
+            # this will buffer the whole thing
+            blocksize = size
         super().__init__(blocksize, fetcher, size)
         self.cache: bytes | None = None
 
     def _fetch(self, start: int | None, end: int | None) -> bytes:
         start = start or 0
-        end = end or self.size
+        if start > self.size:
+            logger.debug("FirstChunkCache: requested start > file size")
+            return b""
+
+        end = min(end, self.size)
+
         if start < self.blocksize:
             if self.cache is None:
                 self.miss_count += 1
@@ -248,12 +256,15 @@ class FirstChunkCache(BaseCache):
                     self.cache = data[: self.blocksize]
                     return data[start:]
                 self.cache = self.fetcher(0, self.blocksize)
+                self.total_requested_bytes += self.blocksize
             part = self.cache[start:end]
             if end > self.blocksize:
                 self.total_requested_bytes += end - self.blocksize
                 part += self.fetcher(self.blocksize, end)
+            self.hit_count += 1
             return part
         else:
+            self.miss_count += 1
             self.total_requested_bytes += end - start
             return self.fetcher(start, end)
 
@@ -370,8 +381,8 @@ class BlockCache(BaseCache):
         start_pos = start % self.blocksize
         end_pos = end % self.blocksize
 
+        self.hit_count += 1
         if start_block_number == end_block_number:
-            self.hit_count += 1
             block: bytes = self._fetch_block_cached(start_block_number)
             return block[start_pos:end_pos]
 
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -735,7 +735,6 @@ class BackgroundBlockCache(BaseCache):
         self._fetch_future: Future[bytes] | None = None
         self._fetch_future_lock = threading.Lock()
 
-
     def cache_info(self) -> UpdatableLRU.CacheInfo:
         """
         The statistics on the block cache.
--- a/fsspec/caching.py
+++ b/fsspec/caching.py
@@ -80,6 +80,7 @@ class BaseCache:
         self.total_requested_bytes = 0
 
     def __repr__(self) -> str:
+        # TODO: use rich for better formatting
         return f"""
             cache type  :   {self.__class__.__name__} 
             block size  :   {self.blocksize}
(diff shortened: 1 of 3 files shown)


betolink:
Hi @martindurant! I think this PR is ready for a review in case you have time this week.

fsspec/caching.py
```diff
@@ -256,12 +303,6 @@ def __init__(
         self.maxblocks = maxblocks
         self._fetch_block_cached = functools.lru_cache(maxblocks)(self._fetch_block)
 
-    def __repr__(self) -> str:
```
martindurant, line 259: There is an argument that repr should give details of what the object *is* (ideally, how to recreate it), but str should be the formatted version with details.

fsspec/spec.py
```diff
@@ -1841,11 +1841,20 @@ def read(self, length=-1):
             length = self.size - self.loc
         if self.closed:
             raise ValueError("I/O operation on closed file.")
-        logger.debug("%s read: %i - %i", self, self.loc, self.loc + length)
         if length == 0:
             # don't even bother calling fetch
             return b""
         out = self.cache._fetch(self.loc, self.loc + length)
+
+        logger.debug(
```
martindurant, line 1849: This seems like too much information here. Such logging should be controlled by the logger in the cache module, not here, especially since we may have a cacher that does nothing, or no cacher at all for backends that don't support it.

fsspec/tests/test_caches.py
```diff
@@ -15,26 +21,124 @@ def test_cache_getitem(Cache_imp):
 
 
 def test_block_cache_lru():
-    cache = BlockCache(4, letters_fetcher, len(string.ascii_letters), maxblocks=2)
+    """
```
martindurant, line 24: This is a comment, not a docstring

martindurant (commented):
Thank you, this looks good on the whole, particularly the set of tests.

I have some comments for you to consider.

fsspec/caching.py
```diff
@@ -202,24 +234,39 @@ class FirstChunkCache(BaseCache):
     name = "first"
 
     def __init__(self, blocksize: int, fetcher: Fetcher, size: int) -> None:
+        if blocksize > size:
```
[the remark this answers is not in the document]
betolink: The actual bytes returned are important of course but I'd lean towards measuring the cache behavior, if we only measure in one place we lose that granularity.

fsspec/caching.py
```diff
@@ -202,24 +234,39 @@ class FirstChunkCache(BaseCache):
...
                 if end > self.blocksize:
+                    self.total_requested_bytes += end
                     data = self.fetcher(0, end)
                     self.cache = data[: self.blocksize]
                     return data[start:]
                 self.cache = self.fetcher(0, self.blocksize)
+                self.total_requested_bytes += self.blocksize
             part = self.cache[start:end]
             if end > self.blocksize:
+                self.total_requested_bytes += end - self.blocksize
```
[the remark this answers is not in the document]
betolink: Same as above, perhaps we are missing another variable to keep track of the `total_bytes_returned`?

fsspec/caching.py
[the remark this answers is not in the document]
martindurant: OK, actually since `fetcher` is probably just a function, maybe there is nowhere else to keep this information.

fsspec/caching.py
betolink: This was moved to the parent class so all the cache implementations have a clearer picture of what they are and their current state, currently only some implementations have a `__repr__`.

martindurant: Yes, right. The comment came here to show how these reprs used to be, more in line with the typical python `<MyClass: details>` style.

fsspec/spec.py
betolink: Agree, maybe this should be better in `caching.py`

fsspec/caching.py
betolink: Ah OK, should we use this convention? something like this (note that `total_returned_bytes` is not tracked yet)

```python
        return f"""
            <{self.__class__.__name__}:
            block size  :   {self.blocksize}
            block count :   {self.nblocks}
            file size   :   {self.size}
            cache hits  :   {self.hit_count}
            cache misses:   {self.miss_count}
            total requested bytes: {self.total_requested_bytes}
            total returned bytes: {self.total_returned_bytes}>
        """
```

I was thinking of using `rich` to have the rich representation, a bar with colors to represent which parts of the file have been cached or something easy to visualize.

martindurant: Rich or HTML (for jupyter) would be nice, but fsspec shouldn't depend on either. There could be a separate pprint or similar method for that.

> the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

https://docs.python.org/3/reference/datamodel.html#object.__repr__

betolink: Yeah that would be for a future PR, do you think the proposed  `__repr__` above fits the bill of the convention `<MyClass: details>`?

fsspec/spec.py
betolink: Let me backtrack here, I think having this information in one place is more useful for traceability.

```
<File-like object S3FileSystem, file.h5> read: 432000000 - 440000000, total: 276824064, cache hits: 34, cache misses: 33
```

is not that different from 

```
<File-like object S3FileSystem, file.h5> read: 432000000 - 440000000
```

If we move it to the cache implementations we'll have to put those calls and the file objects together (percolate the logs?) in the case we use no cache or the cache does nothing we'll only see zeroes.

martindurant: Since we introspect the state of the cache anyway, maybe we can just have the cache* fields produced by a method on the cache objects, where the BaseCache's implementation returns no numbers at all? Does that sound too complex?

betolink: That's a good approach, something like this?

```python
logger.debug("%s read: %i - %i, %s,
    self,
    self.loc,
    self.loc + length,
    self.cache.state()
)

```

martindurant:
I can't fix your linter failure, since the PR is from master branch. You can run precommit locally, e.g., 
```
>>> pre-commit install
>>> pre-commit run -a
```

betolink:
Ok now it should pass and I also addressed the comments on formatting the logs for cache implementations that do nothing @martindurant

PR #1567, opened 2024-04-07
LocalFileSystem restore _strip_protocol signature

ap--:
Hello,

I noticed that the `LocalFileSystem._strip_protocol` signature changed in #1477, when running the universal_pathlib test-suite against the current fsspec version.

To me it seems that the intention of #1477 was initially to prevent `"C:/"` or `"C:\\"` to be reduced to `"C:"` by `_strip_protocol`, but in addition it introduced function signature changes in `fsspec.implementations.local` and minor changes in `fsspec.mapping`.

I created this PR as a basis for discussion if the signature change could be avoided. In this PR I reverted the changes to `LocalFileSystem` and `make_path_posix` but kept the changes to the tests (first commit) and then provide an alternative implementation that avoids the function signature change.

I would also be happy to try to add more tests around the `LocalFileSystem._parent`, `LocalFileSystem._strip_protocol` and `make_path_posix` behavior for edge-cases if there is interest. But windows is not my main OS for daily work, so I am very likely not aware of most of them.

Cheers,
Andreas

### DIFF
--- a/ci/environment-py38.yml
+++ b/ci/environment-py38.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - pip
-  - git
+  - git <2.45.0
   - py
   - pip:
     - hadoop-test-cluster
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -3,7 +3,6 @@ import io
 import logging
 import os
 import os.path as osp
-import re
 import shutil
 import stat
 import tempfile
@@ -215,7 +214,18 @@ class LocalFileSystem(AbstractFileSystem):
     @classmethod
     def _parent(cls, path):
         path = cls._strip_protocol(path)
-        return osp.dirname(path) or cls.root_marker
+        if os.sep == "/":
+            # posix native
+            return path.rsplit("/", 1)[0] or "/"
+        else:
+            # NT
+            path_ = path.rsplit("/", 1)[0]
+            if len(path_) <= 3:
+                if path_[1:2] == ":":
+                    # nt root (something like c:/)
+                    return path_[0] + ":/"
+            # More cases may be required here
+            return path_
 
     @classmethod
     def _strip_protocol(cls, path):
@@ -228,9 +238,33 @@ class LocalFileSystem(AbstractFileSystem):
             path = path[8:]
         elif path.startswith("local:"):
             path = path[6:]
-        drive, path = osp.splitdrive(make_path_posix(path))
-        path = path.rstrip("/") or cls.root_marker
-        return drive + path
+
+        path = make_path_posix(path)
+        if os.sep != "/":
+            # This code-path is a stripped down version of
+            # > drive, path = ntpath.splitdrive(path)
+            if path[1:2] == ":":
+                # Absolute drive-letter path, e.g. X:\Windows
+                # Relative path with drive, e.g. X:Windows
+                drive, path = path[:2], path[2:]
+            elif path[:2] == "//":
+                # UNC drives, e.g. \\server\share or \\?\UNC\server\share
+                # Device drives, e.g. \\.\device or \\?\device
+                if (index1 := path.find("/", 2)) == -1 or (
+                    index2 := path.find("/", index1 + 1)
+                ) == -1:
+                    drive, path = path, ""
+                else:
+                    drive, path = path[:index2], path[index2:]
+            else:
+                # Relative path, e.g. Windows
+                drive = ""
+
+            path = path.rstrip("/") or cls.root_marker
+            return drive + path
+
+        else:
+            return path.rstrip("/") or cls.root_marker
 
     def _isfilestore(self):
         # Inheriting from DaskFileSystem makes this False (S3, etc. were)
@@ -243,47 +277,53 @@ class LocalFileSystem(AbstractFileSystem):
         return os.chmod(path, mode)
 
 
-def make_path_posix(path, sep=os.sep):
-    """Make path generic"""
-    if isinstance(path, (list, set, tuple)):
-        return type(path)(make_path_posix(p) for p in path)
-    if "~" in path:
-        path = osp.expanduser(path)
-    if sep == "/":
-        # most common fast case for posix
-        if path.startswith("/"):
-            return path
-        if path.startswith("./"):
-            path = path[2:]
-        return f"{os.getcwd()}/{path}"
-    if (
-        (sep not in path and "/" not in path)
-        or (sep == "/" and not path.startswith("/"))
-        or (sep == "\\" and ":" not in path and not path.startswith("\\\\"))
-    ):
-        # relative path like "path" or "rel\\path" (win) or rel/path"
-        if os.sep == "\\":
-            # abspath made some more '\\' separators
-            return make_path_posix(osp.abspath(path))
+def make_path_posix(path):
+    """Make path generic for current OS"""
+    if not isinstance(path, str):
+        if isinstance(path, (list, set, tuple)):
+            return type(path)(make_path_posix(p) for p in path)
         else:
-            return f"{os.getcwd()}/{path}"
-    if path.startswith("file://"):
-        path = path[7:]
-    if re.match("/[A-Za-z]:", path):
-        # for windows file URI like "file:///C:/folder/file"
-        # or "file:///C:\\dir\\file"
-        path = path[1:].replace("\\", "/").replace("//", "/")
-    if path.startswith("\\\\"):
-        # special case for windows UNC/DFS-style paths, do nothing,
-        # just flip the slashes around (case below does not work!)
-        return path.replace("\\", "/")
-    if re.match("[A-Za-z]:", path):
-        # windows full path like "C:\\local\\path"
-        return path.lstrip("\\").replace("\\", "/").replace("//", "/")
-    if path.startswith("\\"):
-        # windows network path like "\\server\\path"
-        return "/" + path.lstrip("\\").replace("\\", "/").replace("//", "/")
-    return path
+            path = str(stringify_path(path))
+    if os.sep == "/":
+        # Native posix
+        if path.startswith("/"):
+            # most common fast case for posix
+            return path
+        elif path.startswith("~"):
+            return osp.expanduser(path)
+        elif path.startswith("./"):
+            path = path[2:]
+        elif path == ".":
+            path = ""
+        return f"{os.getcwd()}/{path}"
+    else:
+        # NT handling
+        if path[0:1] == "/" and path[2:3] == ":":
+            # path is like "/c:/local/path"
+            path = path[1:]
+        if path[1:2] == ":":
+            # windows full path like "C:\\local\\path"
+            if len(path) <= 3:
+                # nt root (something like c:/)
+                return path[0] + ":/"
+            path = path.replace("\\", "/")
+            return path
+        elif path[0:1] == "~":
+            return make_path_posix(osp.expanduser(path))
+        elif path.startswith(("\\\\", "//")):
+            # windows UNC/DFS-style paths
+            return "//" + path[2:].replace("\\", "/")
+        elif path.startswith(("\\", "/")):
+            # windows relative path with root
+            path = path.replace("\\", "/")
+            return f"{osp.splitdrive(os.getcwd())[0]}{path}"
+        else:
+            path = path.replace("\\", "/")
+            if path.startswith("./"):
+                path = path[2:]
+            elif path == ".":
+                path = ""
+            return f"{make_path_posix(os.getcwd())}/{path}"
 
 
 def trailing_sep(path):
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -283,7 +283,9 @@ def make_path_posix(path):
         if isinstance(path, (list, set, tuple)):
             return type(path)(make_path_posix(p) for p in path)
         else:
-            path = str(stringify_path(path))
+            path = stringify_path(path)
+            if not isinstance(path, str):
+                raise TypeError(f"could not convert {path!r} to string")
     if os.sep == "/":
         # Native posix
         if path.startswith("/"):
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -116,8 +116,8 @@ class LocalFileSystem(AbstractFileSystem):
         return osp.lexists(path)
 
     def cp_file(self, path1, path2, **kwargs):
-        path1 = self._strip_protocol(path1).rstrip("/")
-        path2 = self._strip_protocol(path2).rstrip("/")
+        path1 = self._strip_protocol(path1)
+        path2 = self._strip_protocol(path2)
         if self.auto_mkdir:
             self.makedirs(self._parent(path2), exist_ok=True)
         if self.isfile(path1):
@@ -146,8 +146,8 @@ class LocalFileSystem(AbstractFileSystem):
         return self.cp_file(path1, path2, **kwargs)
 
     def mv_file(self, path1, path2, **kwargs):
-        path1 = self._strip_protocol(path1).rstrip("/")
-        path2 = self._strip_protocol(path2).rstrip("/")
+        path1 = self._strip_protocol(path1)
+        path2 = self._strip_protocol(path2)
         shutil.move(path1, path2)
 
     def link(self, src, dst, **kwargs):
@@ -171,7 +171,7 @@ class LocalFileSystem(AbstractFileSystem):
             path = [path]
 
         for p in path:
-            p = self._strip_protocol(p).rstrip("/")
+            p = self._strip_protocol(p)
             if self.isdir(p):
                 if not recursive:
                     raise ValueError("Cannot delete directory, set recursive=True")
@@ -214,11 +214,8 @@ class LocalFileSystem(AbstractFileSystem):
 
     @classmethod
     def _parent(cls, path):
-        path = cls._strip_protocol(path).rstrip("/")
-        if "/" in path:
-            return path.rsplit("/", 1)[0]
-        else:
-            return cls.root_marker
+        path = cls._strip_protocol(path)
+        return osp.dirname(path) or cls.root_marker
 
     @classmethod
     def _strip_protocol(cls, path):
@@ -231,7 +228,9 @@ class LocalFileSystem(AbstractFileSystem):
             path = path[8:]
         elif path.startswith("local:"):
             path = path[6:]
-        return make_path_posix(path).rstrip("/") or cls.root_marker
+        drive, path = osp.splitdrive(make_path_posix(path))
+        path = path.rstrip("/") or cls.root_marker
+        return drive + path
 
     def _isfilestore(self):
         # Inheriting from DaskFileSystem makes this False (S3, etc. were)
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -3,6 +3,7 @@ import io
 import logging
 import os
 import os.path as osp
+import re
 import shutil
 import stat
 import tempfile
@@ -13,12 +14,6 @@ from fsspec.core import get_compression
 from fsspec.utils import isfilelike, stringify_path
 
 logger = logging.getLogger("fsspec.local")
-
-
-def _remove_prefix(text: str, prefix: str):
-    if text.startswith(prefix):
-        return text[len(prefix) :]
-    return text
 
 
 class LocalFileSystem(AbstractFileSystem):
@@ -121,8 +116,8 @@ class LocalFileSystem(AbstractFileSystem):
         return osp.lexists(path)
 
     def cp_file(self, path1, path2, **kwargs):
-        path1 = self._strip_protocol(path1, remove_trailing_slash=True)
-        path2 = self._strip_protocol(path2, remove_trailing_slash=True)
+        path1 = self._strip_protocol(path1).rstrip("/")
+        path2 = self._strip_protocol(path2).rstrip("/")
         if self.auto_mkdir:
             self.makedirs(self._parent(path2), exist_ok=True)
         if self.isfile(path1):
@@ -151,8 +146,8 @@ class LocalFileSystem(AbstractFileSystem):
         return self.cp_file(path1, path2, **kwargs)
 
     def mv_file(self, path1, path2, **kwargs):
-        path1 = self._strip_protocol(path1, remove_trailing_slash=True)
-        path2 = self._strip_protocol(path2, remove_trailing_slash=True)
+        path1 = self._strip_protocol(path1).rstrip("/")
+        path2 = self._strip_protocol(path2).rstrip("/")
         shutil.move(path1, path2)
 
     def link(self, src, dst, **kwargs):
@@ -176,7 +171,7 @@ class LocalFileSystem(AbstractFileSystem):
             path = [path]
 
         for p in path:
-            p = self._strip_protocol(p, remove_trailing_slash=True)
+            p = self._strip_protocol(p).rstrip("/")
             if self.isdir(p):
                 if not recursive:
                     raise ValueError("Cannot delete directory, set recursive=True")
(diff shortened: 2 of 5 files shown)


martindurant:
@fleming79 , this undoes much of your code, but maintains the tests for correctness, while reimplementing consistency for _strip_protocol. Do you have thoughts?

fleming79:
I have no issue with removing the option 'remove_trailing_slash' if it isn't required. It was provided because there were `.rstrip('/')` calls scattered about.

The question I pose is - which code is more maintainable/performant/correct?

ap--:
Thanks for the quick response!

Regarding maintainability, I would prefer if the method signatures would agree with `AbstractFileSystem`. For example, enforcing #1104 (where possible) would make downstream development easier.

Regarding performance, I'm not a big fan of micro-benchmarks anymore, but I ran one (checking 10000 generated URIs) and the implementation in this PR seems to be equally fast/slow (within margin of error) as `2024.3.1` and `2024.3.0` for `_strip_protocol`. It is ~2% slower for `_parent`. This was on an M2max arm macbook, only tested with `py3.11` and only run a few times. Variation between runs is larger than the differences within a run.

<details>
<summary>[click to expand] benchmark for `_strip_protocol` and `_parent`</summary>

```
❯ asv run
· Creating environments
· Discovering benchmarks
·· Uninstalling from virtualenv-py3.11
·· Building 518984d4 <master> for virtualenv-py3.11...
·· Installing 518984d4 <master> into virtualenv-py3.11.
· Running 6 total benchmarks (3 commits * 1 environments * 2 benchmarks)
[ 0.00%] · For filesystem_spec commit 518984d4 <master>:
[ 0.00%] ·· Benchmarking virtualenv-py3.11
[ 8.33%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[25.00%] ··· bench_strip_protocol.Suite.time_parent                        207±0.8ms
[33.33%] ··· bench_strip_protocol.Suite.time_split_protocol                  205±2ms
[33.33%] · For filesystem_spec commit 4bd16f64 <2024.3.0^0>:
[33.33%] ·· Building for virtualenv-py3.11.
[33.33%] ·· Benchmarking virtualenv-py3.11
[41.67%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[58.33%] ··· bench_strip_protocol.Suite.time_parent                        207±0.7ms
[66.67%] ··· bench_strip_protocol.Suite.time_split_protocol                  204±2ms
[66.67%] · For filesystem_spec commit c9eaf46e <local-fs-restore-strip-protocol>:
[66.67%] ·· Building for virtualenv-py3.11....
[66.67%] ·· Benchmarking virtualenv-py3.11
[75.00%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[91.67%] ··· bench_strip_protocol.Suite.time_parent                          211±1ms
[100.00%] ··· bench_strip_protocol.Suite.time_split_protocol                 204±2ms
```


```python
# pip install asv

import itertools
import random
import string
from fsspec import get_filesystem_class

random.seed(0)

def _make_random_path():
    # build a random length path
    pth_len = random.randint(5, 40)
    return "".join(random.sample(string.ascii_letters + "/", k=pth_len))

def _make_uris(n):
    # create uris with and without protocols and with and without netlocs
    it_proto_netloc = itertools.cycle(
        itertools.product(
            ["file", "local", "wrong", None],
            ["netloc", ""]
        )
    )

    for _ in range(n):
        proto, netloc = next(it_proto_netloc)
        pth = _make_random_path()
        if proto and netloc:
            yield f"{proto}://{netloc}/{pth}"
        elif proto:
            yield f"{proto}:/{pth}"
        else:
            yield f"{netloc}/{pth}"

uris = list(_make_uris(10000))


class Suite:

    def setup(self):
        self.fs_cls = get_filesystem_class("file")
        self.uris = uris

    def teardown(self):
        del self.uris

    def time_split_protocol(self):
        for uri in self.uris:
            self.fs_cls._strip_protocol(uri)

    def time_parent(self):
        for uri in self.uris:
            self.fs_cls._parent(uri)
```

</details>


To be able to answer if this implementation is more or less correct would require more tests. I'll try to find some time to come up with more cases for the test-suite.

Cheers,
Andreas

martindurant:
I don't think we can trust timing differences <5%; in these cases it can even matter which order you run them (whether the CPU is warm, etc). The interesting thing would be to check on Windows (but they have lots of path styles).

fleming79:
The original `make_path_posix` contains quite a lot of strip protocol code in it. It also includes 'sep' as an argument implying it can work for an alternate os. 
I'd prefer to see the newer code tweaked to remove the signature.

Here is a starting point.

```




def make_path_posix(path):
    """Make path generic for current OS"""
    if not isinstance(path, str):
        if isinstance(path, (list, set, tuple)):
            return type(path)(make_path_posix(p) for p in path)
        else:
            path = str(stringify_path(path))
    if os.sep == "/":
        # Native posix
        if path.startswith("/"):
            # most common fast case for posix
            return path.rstrip("/") or "/"
        elif path.startswith("~"):
            return make_path_posix(osp.expanduser(path))
        elif path.startswith("./"):
            path = path[2:]
            path = f"{os.getcwd()}/{path}"
            return path.rstrip("/") or "/"
        return f"{os.getcwd()}/{path}"
    else:
        # NT handling
        if len(path) > 1:
            if path[1] == ":":
                # windows full path like "C:\\local\\path"
                if len(path) <= 3:
                    # nt root (something like c:/)
                    return path[0] + ":/"
                path = path.replace("\\", "/").replace("//", "/")
                return path
            elif path[0] == "~":
                return make_path_posix(osp.expanduser(path))
            elif path.startswith(("\\\\", "//")):
                # windows UNC/DFS-style paths
                return "//" + path[2:].replace("\\", "/").replace("//", "/")
        return make_path_posix(osp.abspath(path))


```

```

    @classmethod
    def _strip_protocol(cls, path):
        path = stringify_path(path)
        if path.startswith(("file:",'local')):
            if os.sep == '\\' and path.startswith("file:///"):
                path = path[8:]
            elif path.startswith("file://"):
                path = path[7:]
            elif path.startswith("file:"):
                path = path[5:]
            elif path.startswith("local://"):
                path = path[8:]
            elif path.startswith("local:"):
                path = path[6:]
        drive, path = osp.splitdrive(make_path_posix(path))
        path = path.rstrip("/") or cls.root_marker
        return drive + path

```

ap--:
Thank you @fleming79 for the proposed implementations. I'll try them out shortly, when I find some more time.

I've been collecting more test cases for `_strip_protocol` (still need to do `_parent`). In the test cases I basically try to document current behavior.

So far I noticed that between `v2024.3.0` and `v2024.3.1` there are the following differences in `LocalFileSystem._strip_protocol` behavior (note, that a lot of the test cases are probably just uncommon edge-cases):

**on posix**

| input        | v2024.3.0          | v2024.3.1           |
|--------------|--------------------|---------------------|
| `file://`    | `/current_workdir` | `/current_workdir/` |
| `file://./`  | `/current_workdir` | `/current_workdir/` |
| `./`         | `/current_workdir` | `/current_workdir/` |

**on windows**

| input                        | v2024.3.0             | v2024.3.1                              |
|------------------------------|-----------------------|----------------------------------------|
| `file:/foo/bar`              | `c:/foo/bar`          | `c:/Users/user_home/foo/bar`           |
| `local:/path`                | `c:/path`             | `c:/Users/user_home/path`              |
| `s3://bucket/key`            | `s3://bucket/key`     | `c:/Users/user_home/s3:/bucket/key`    |
| `file:///`                   | `c:/`                 | `c:/Users/user_home/`                  |
| `file:////remote/share/path` | `//remote/share/path` | `c:/Users/user_home/remote/share/path` |


I will write more tests for `_parent`.

(Update: the failing windows test, succeeds locally on a windows py3.11. I'll investigate a bit further...)

ap--:
Performance comparison on windows python==3.11, shows that the `2023.3.1` implementation is the fastest and the implementation in this PR is significantly worse.

I'll update this benchmark once I test the implementation provided in the comment above.

```
(venv) PS C:\User\user\development\filesystem_spec> asv run
· Creating environments.........
· Discovering benchmarks
·· Uninstalling from virtualenv-py3.11
·· Building 4bd16f64 <2024.3.0^0> for virtualenv-py3.11...........
·· Installing 4bd16f64 <2024.3.0^0> into virtualenv-py3.11.
· Running 6 total benchmarks (3 commits * 1 environments * 2 benchmarks)
[ 0.00%] · For filesystem_spec commit 4bd16f64 <2024.3.0^0>:
[ 0.00%] ·· Benchmarking virtualenv-py3.11
[ 8.33%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[25.00%] ··· bench_strip_protocol.Suite.time_parent                                   22.5±0.3ms
[33.33%] ··· bench_strip_protocol.Suite.time_split_protocol                           20.3±0.2ms
[33.33%] · For filesystem_spec commit 516b0626 <local-fs-restore-strip-protocol>:
[33.33%] ·· Building for virtualenv-py3.11........
[33.33%] ·· Benchmarking virtualenv-py3.11
[41.67%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[58.33%] ··· bench_strip_protocol.Suite.time_parent                                   43.3±0.3ms
[66.67%] ··· bench_strip_protocol.Suite.time_split_protocol                           25.7±0.3ms
[66.67%] · For filesystem_spec commit 47b445ae <2024.3.1^0>:
[66.67%] ·· Building for virtualenv-py3.11...........
[66.67%] ·· Benchmarking virtualenv-py3.11
[75.00%] ··· Running (bench_strip_protocol.Suite.time_parent--)..
[91.67%] ··· bench_strip_protocol.Suite.time_parent                                   19.5±0.3ms
[100.00%] ··· bench_strip_protocol.Suite.time_split_protocol                          16.5±0.2ms
```

ap--:
Note: the failing CI runs seem to be caused by a new git release? https://github.com/pypa/setuptools_scm/issues/1038

fleming79:
@ap--  It looks like you've made some good changes.

> PR https://github.com/fsspec/filesystem_spec/pull/1477 added
string coercion to make_path_posix, causing the newly added tests [here](https://github.com/ap--/filesystem_spec/blob/8ef54b032077ad0814f95ce5160926cc8c4c9081/fsspec/implementations/tests/test_local.py#L589-L601) to
> fail. I marked them as xfail for now. It would be great to get your feedback
> on this. My opinion is that these cases should raise an error. On current master
> make_posix_path(None) will return '/cwd/None'.

I agree those tests are useful and an error should probably be raised for invalid types somehow. I guess removing the coercion would suffice, but maybe it would be better to also check the return type from the call to `stringify_path`.

martindurant:
> causing the newly added tests [here](https://github.com/ap--/filesystem_spec/blob/8ef54b032077ad0814f95ce5160926cc8c4c9081/fsspec/implementations/tests/test_local.py#L589-L601) to
fail. I marked them as xfail for now.

Do those fail with a different exception type now, or just return garbage pathstrings?

> the failing CI runs seem to be caused by a new git release? https://github.com/pypa/setuptools_scm/issues/1038

This is for the s3fs build, right? We could pin its requirement on setuptools_scm, wait for a new release, or move that build to hatch (as fsspec has been).

ap--:
> Do those fail with a different exception type now, or just return garbage pathstrings?

Current master returns garbage: 

```python
>>> from fsspec.implementations.local import make_path_posix
>>> make_path_posix(object())
'/Users/poehlmann/development/filesystem_spec/<object object at 0x100664b80>'
```

Possible fixes are:
1. change `path = str(stringify_path(path))` to `path = stringify_path(path)` which will then cause an `AttributeError` to be raised due to the missing attribute `".startswith"`. (This is asserted in the test)
2. check the instance type of `stringify_path(path)` and raise a TypeError if not `str`
3. raise a TypeError in the fallthrough case in `stringify_path`: https://github.com/fsspec/filesystem_spec/blob/df65747ac92e3493fae6e82081026c8f6df8ba94/fsspec/utils.py#L357-L358


> This is for the s3fs build, right? We could pin its requirement on setuptools_scm, wait for a new release, or move that build to hatch (as fsspec has been).

I think this happens when during the s3fs install fsspec gets installed. The traceback shows it failing in hatchling, which uses the hatch-vcs plugin to get the current version via setuptools_scm. setuptools_scm in turn has a compatibility issue with the newest git version which seems to be used on the Github Actions runner already via the conda test-env environment. I'll see if downgrading git will fix the issue.

martindurant:
> I'll see if downgrading git will fix the issue.

A specific version of git could be pinned in the CI conda environment file

ap--:
That made it work ☺️ The py3.11 test failed because the smb tests failed.

martindurant:
rerunning ... I have yet to come up with a way to get SMB to reliably pass every time.

martindurant:
I'm happy with this. @fleming79, any comments?

ap--:
Awesome.

So we just need a decision regarding `make_path_posix` behavior:

1. `make_path_posix(None)` returns `/cwd/None` (`fsspec>=2024.3.1` and **this PR**)
2. `make_path_posix(None)` raises `AttributeError` (`fsspec<=2024.3.0`)
3. `make_path_posix(None)` raises `TypeError` (might be most intuitive for users?)

martindurant:
`2.` seems fine to me, but happy to catch the error and convert to TypeError too (no strong feeling)

fleming79:
I'm happy with `2` .

In `stringify_path` I suggest changing from ` isinstance(filepath, pathlib.Path)` to ` isinstance(filepath, pathlib.PurePath). `Path` is a subclass of `PurePath`. and would mean  `PurePosixPath` and `PureWindowsPath` are also converted to strings (which they should). Perhaps some tests involving PurePaths could be added?

ap--:
I made the changes and added tests for `stringify_path`.

> In `stringify_path` I suggest changing from ` isinstance(filepath, pathlib.Path)` to `isinstance(filepath, pathlib.PurePath).`Path`is a subclass of`PurePath`. and would mean `PurePosixPath`and`PureWindowsPath` are also converted to strings (which they should). Perhaps some tests involving PurePaths could be added?

`PurePath` already implements the `__fspath__` method, which made the isinstance check after `hasattr('fspath')` redundant.

shcheklein:
@martindurant  @ap--  Folks, quick question. Does this implementation change the behavior of methods like `ls` (by removing `strip_trailing_slash`)? E.g. when I do: 

`client.fs("file:///Users/ivan/Projects/dvcx/.github/dependabot.yml/")` (mind the trailing slash)

in case if `dependabot.yml` *file* exists it returns a file (instead of raising an exception since we are asking about the directory).

is it an expected behavior?

ap--:
Hi @shcheklein 

Looking into this, the behavior of `LocalFileSystem.ls` is consistent across most versions. The only difference is `2024.3.1`. In #1477, which was included in `2024.3.1`. A side effect was that the behavior of `LocalFileSystem().ls` changed in regard to the case you described.

Specifically #1477 started raising a `NotADirectory` exception for `LocalFileSystem().ls`, if `<some-path>/` was used as an argument and `<some-path>` was a file.

The previous behavior `fsspec<=2024.3.0` (= returning a list with one item --- the path to the file) was restored by this PR.


<details>

<summary>Click to see behavior of LocalFileSystem.ls for different fsspec versions</summary>

```python
# cat test_1567-issuecomment-2563160414.py
import fsspec
import pytest

def test_local_ls_file_exists(tmp_path):

    file = tmp_path.joinpath("dependabot.yml")
    file.write_text("hello world!")

    urlpath = file.as_uri()
    urlpath_with_trailing_slash = f"{urlpath}/"

    path = fs._strip_protocol(urlpath)

    assert fs.ls(urlpath) == [path]
    assert fs.ls(urlpath_with_trailing_slash) == [path]

```

```python
# cat noxfile.py
import nox

@nox.session(python=["3.10"])
@nox.parametrize("fsspec", ["2024.2.0", "2024.3.0", "2024.3.1", "2024.5.0", "2024.12.0"])
def tests(session: nox.Session, fsspec) -> None:
    session.install(f"fsspec=={fsspec}", "pytest")
    session.run("pytest", "-v")


```



```console
❯ nox
nox > Running session tests-3.10(fsspec='2024.2.0')
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/tests-3-10-fsspec-2024-2-0
nox > python -m pip install fsspec==2024.2.0 pytest
nox > pytest -v
================================================================ test session starts =================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0 -- /home/poehlmann/Development/fsspec#1567/.nox/tests-3-10-fsspec-2024-2-0/bin/python
cachedir: .pytest_cache
rootdir: /home/poehlmann/Development/fsspec#1567
collected 1 item

test_1567-issuecomment-2563160414.py::test_local_ls_file_exists PASSED                                                                         [100%]

================================================================= 1 passed in 0.04s ==================================================================
nox > Session tests-3.10(fsspec='2024.2.0') was successful.
nox > Running session tests-3.10(fsspec='2024.3.0')
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/tests-3-10-fsspec-2024-3-0
nox > python -m pip install fsspec==2024.3.0 pytest
nox > pytest -v
================================================================ test session starts =================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0 -- /home/poehlmann/Development/fsspec#1567/.nox/tests-3-10-fsspec-2024-3-0/bin/python
cachedir: .pytest_cache
rootdir: /home/poehlmann/Development/fsspec#1567
collected 1 item

test_1567-issuecomment-2563160414.py::test_local_ls_file_exists PASSED                                                                         [100%]

================================================================= 1 passed in 0.04s ==================================================================
nox > Session tests-3.10(fsspec='2024.3.0') was successful.
nox > Running session tests-3.10(fsspec='2024.3.1')
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/tests-3-10-fsspec-2024-3-1
nox > python -m pip install fsspec==2024.3.1 pytest
nox > pytest -v
================================================================ test session starts =================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0 -- /home/poehlmann/Development/fsspec#1567/.nox/tests-3-10-fsspec-2024-3-1/bin/python
cachedir: .pytest_cache
rootdir: /home/poehlmann/Development/fsspec#1567
collected 1 item

test_1567-issuecomment-2563160414.py::test_local_ls_file_exists FAILED                                                                         [100%]

====================================================================== FAILURES ======================================================================
_____________________________________________________________ test_local_ls_file_exists ______________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-poehlmann/pytest-27/test_local_ls_file_exists0')

    def test_local_ls_file_exists(tmp_path):

        file = tmp_path.joinpath("dependabot.yml")
        file.write_text("hello world!")

        urlpath = file.as_uri()
        urlpath_with_trailing_slash = f"{urlpath}/"

        path = fs._strip_protocol(urlpath)

        assert fs.ls(urlpath) == [path]
>       assert fs.ls(urlpath_with_trailing_slash) == [path]

test_1567-issuecomment-2563160414.py:19:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.nox/tests-3-10-fsspec-2024-3-1/lib/python3.10/site-packages/fsspec/implementations/local.py:66: in ls
    info = self.info(path)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fsspec.implementations.local.LocalFileSystem object at 0x7fdb63812e60>
path = '/tmp/pytest-of-poehlmann/pytest-27/test_local_ls_file_exists0/dependabot.yml/', kwargs = {}

        if isinstance(path, os.DirEntry):
            # scandir DirEntry
            out = path.stat(follow_symlinks=False)
            link = path.is_symlink()
            if path.is_dir(follow_symlinks=False):
                t = "directory"
            elif path.is_file(follow_symlinks=False):
                t = "file"
            else:
                t = "other"
            path = self._strip_protocol(path.path)
        else:
            # str or path-like
>           out = os.stat(path, follow_symlinks=False)
E           NotADirectoryError: [Errno 20] Not a directory: '/tmp/pytest-of-poehlmann/pytest-27/test_local_ls_file_exists0/dependabot.yml/'

.nox/tests-3-10-fsspec-2024-3-1/lib/python3.10/site-packages/fsspec/implementations/local.py:92: NotADirectoryError
============================================================== short test summary info ===============================================================
FAILED test_1567-issuecomment-2563160414.py::test_local_ls_file_exists - NotADirectoryError: [Errno 20] Not a directory: '/tmp/pytest-of-poehlmann/pytest-27/test_local_ls_file_exists0/dependabot.yml/'
================================================================= 1 failed in 0.06s ==================================================================
nox > Command pytest -v failed with exit code 1
nox > Session tests-3.10(fsspec='2024.3.1') failed.
nox > Running session tests-3.10(fsspec='2024.5.0')
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/tests-3-10-fsspec-2024-5-0
nox > python -m pip install fsspec==2024.5.0 pytest
nox > pytest -v
================================================================ test session starts =================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0 -- /home/poehlmann/Development/fsspec#1567/.nox/tests-3-10-fsspec-2024-5-0/bin/python
cachedir: .pytest_cache
rootdir: /home/poehlmann/Development/fsspec#1567
collected 1 item

test_1567-issuecomment-2563160414.py::test_local_ls_file_exists PASSED                                                                         [100%]

================================================================= 1 passed in 0.04s ==================================================================
nox > Session tests-3.10(fsspec='2024.5.0') was successful.
nox > Running session tests-3.10(fsspec='2024.12.0')
nox > Creating virtual environment (virtualenv) using python3.10 in .nox/tests-3-10-fsspec-2024-12-0
nox > python -m pip install fsspec==2024.12.0 pytest
nox > pytest -v
================================================================ test session starts =================================================================
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0 -- /home/poehlmann/Development/fsspec#1567/.nox/tests-3-10-fsspec-2024-12-0/bin/python
cachedir: .pytest_cache
rootdir: /home/poehlmann/Development/fsspec#1567
collected 1 item

test_1567-issuecomment-2563160414.py::test_local_ls_file_exists PASSED                                                                         [100%]

================================================================= 1 passed in 0.04s ==================================================================
nox > Session tests-3.10(fsspec='2024.12.0') was successful.
nox > Ran multiple sessions:
nox > * tests-3.10(fsspec='2024.2.0'): success
nox > * tests-3.10(fsspec='2024.3.0'): success
nox > * tests-3.10(fsspec='2024.3.1'): failed
nox > * tests-3.10(fsspec='2024.5.0'): success
nox > * tests-3.10(fsspec='2024.12.0'): success

```

</details>


It would probably be good to add a test case and clearly define what should happen. `ls` on unix raises an error in the case you described.

```console
❯ touch abc
❯ ls abc
abc

❯ ls abc/
ls: cannot access 'abc/': Not a directory
```

To be on the safe side this should be checked on windows too, to see if the behavior is potentially different there.

Happy holidays,
Andreas

shcheklein:
@ap-- amazing research, thanks! 

> t would probably be good to add a test case and clearly define what should happen. ls on unix raises an error in the case you described.

I guess, yes, my biggest question was if we can consider this behavior as a bug or is it intentional for some reason. 

It makes sense to me to be aligned with `ls` and to be aligned with other implementations.

@martindurant what is your take and who is making these decisions?

martindurant:
> who is making these decisions?

In practice, myself. I would rather not do that alone, and if we can get people together to define the expected behaviours across filesystem implementations, it would help greatly. 
[Tests](https://github.com/fsspec/filesystem_spec/tree/master/fsspec/tests/abstract) and [associated documentation](https://filesystem-spec.readthedocs.io/en/latest/copying.html#expected-behavior) was an attempt to set up such standards for bulk copy/rm operations.

martindurant:
> It makes sense to me to be aligned with ls and to be aligned with other implementations.

Yes, we would certainly like consistency across implementations. There will always be some unavoidable inconsistencies, however, caused by inherent differences (e.g., on blob storage, a directory/prefix and file can have the same name, but not locally).

ap--:
> if we can get people together to define the expected behaviours across filesystem implementations, it would help greatly.

I should be able to free up some time in the next weeks to start working on abstract tests and docs for the urlpath parsing interface of `AbstractFileSystem`.


```python
class AbstractFileSystem:

    protocol: str | tuple[str, ...]
    root_marker: Literal["/", ""]
    sep: Literal["/"]

    @classmethod
    def _strip_protocol(cls, path: str) -> str:
        ...

    def unstrip_protocol(self, name: str) -> str:
        ...
        
    @staticmethod
    def _get_kwargs_from_urls(path: str) -> dict[str, Any]:
        ...

    @classmethod
    def _parent(cls, path: str) -> str:
        ...

```

PR #1569, opened 2024-04-11
Use "post-release" versioning

martindurant:
For version 2024.3.1, further dev commits or working tree will have versions like 2024.3.1.XXX rather than 2024.3.2.XXX to allow s3fs and gcsfs's requirements in CI to be met.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,6 +135,7 @@ Homepage = "https://github.com/fsspec/filesystem_spec"
 
 [tool.hatch.version]
 source = "vcs"
+raw-options = {'version_scheme'='post-release'}
 
 [tool.hatch.build.hooks.vcs]
 version-file = "fsspec/_version.py"


fleming79:
I think you may also need to modify the dependency requirement in s3fs. Currently it is set as `fsspec==2024.3.1` for s3fs. 

Perhaps `fsspec==2024.3.1.*` might be suitable.

martindurant:
> Perhaps fsspec==2024.3.1.* might be suitable.

No, this would not match "2024.3.1" , I believe

martindurant:
```
$ pip install fsspec==2024.3.1.* --force-reinstall
zsh: no matches found: fsspec==2024.3.1.*
```

martindurant:
I am wrong
```
$ pip install "fsspec==2024.3.1.*" --force-reinstall
Collecting fsspec==2024.3.1.*
  Using cached fsspec-2024.3.1-py3-none-any.whl.metadata (6.8 kB)
Using cached fsspec-2024.3.1-py3-none-any.whl (171 kB)
Installing collected packages: fsspec
Successfully installed fsspec-2024.3.1
```

martindurant:
Indeed. Thanks for the ping, here's to remembering to do that.

PR #1576, opened 2024-04-16
Do not swallow errors on mv

### LINKED ISSUE #1559: fs.mv swallows copy errors

Skylion007:
fixes #1559 . This prevents data corruption when using fsspec.mv and adds a unit test.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -53,7 +53,7 @@ The following libraries use ``fsspec`` internally for path and file handling:
    maintainable and modular data science code
 #. `pyxet`_, a Python library for mounting and
    accessing very large datasets from XetHub
-#. `Huggingface🤗 Datasets`_, a popular library to 
+#. `Huggingface🤗 Datasets`_, a popular library to
    load&manipulate data for Deep Learning models
 
 ``fsspec`` filesystems are also supported by:
--- a/fsspec/tests/abstract/mv.py
+++ b/fsspec/tests/abstract/mv.py
@@ -1,0 +1,29 @@
+from unittest.mock import patch
+
+import pytest
+
+import fsspec
+
+
+def test_move_raises_error_with_tmpdir(tmpdir):
+    # Create a file in the temporary directory
+    source = tmpdir.join("source_file.txt")
+    source.write("content")
+
+    # Define a destination that simulates a protected or invalid path
+    destination = tmpdir.join("non_existent_directory/destination_file.txt")
+
+    # Instantiate the filesystem (assuming the local file system interface)
+    fs = fsspec.filesystem("file")
+
+    # Patch the entire `mv` method to control the behavior fully
+    with patch.object(fs, "mv", autospec=True) as mock_mv:
+        # Configure the mock to raise an OSError when called
+        mock_mv.side_effect = OSError("Unable to move the file")
+
+        # Use the actual file paths as string
+        with pytest.raises(OSError) as excinfo:
+            fs.mv(str(source), str(destination))
+
+        # Assert the message of the raised OSError matches the expected message
+        assert "Unable to move the file" in str(excinfo.value)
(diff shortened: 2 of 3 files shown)


docs/source/index.rst
```diff
@@ -53,7 +53,7 @@ The following libraries use ``fsspec`` internally for path and file handling:
    maintainable and modular data science code
 #. `pyxet`_, a Python library for mounting and
    accessing very large datasets from XetHub
-#. `Huggingface🤗 Datasets`_, a popular library to 
+#. `Huggingface🤗 Datasets`_, a popular library to
```
Skylion007, line 56: This is due a lint failure on master

Skylion007:
Current failres appear to be flakes

martindurant:
Do you think `mv` should have an `on_error` of its own? I can see an argument that we should not delete any file if the copy phase failed, or maybe that we should delete, but only for those files that successfully moved.

Skylion007:
@martindurant Regardless, we should never delete the src files if the copy failed, wouldn't you agree? Otherwise, the src files could be lost and data corruption could occur.

martindurant:
Agreed, but should we delete the ones that did copy ok? If we raise here, we lose the information on which ones copied and which did not.

Skylion007:
> Agreed, but should we delete the ones that did copy ok? If we raise here, we lose the information on which ones copied and which did not.

Hmm, we could try, but we don't guarantee that would could even delete them I suppose. That could be a fix in a subsequent PR for sure, but the current behavior of deleting the files whether the copy succeeds or not is extremely dangerous.

Whether this is a big or not, `on_error` only ignores FileNotFoundErrors, and does not ignore PermissionDeniedError or other issues which is better than I feared.

To be honest, the error handling, path expansion, unit tests and just generally consistency of all all the spec utility functions are in dire need of a clean up.

fsspec/tests/abstract/mv.py
```diff
@@ -0,0 +1,57 @@
...
+    # Create a file in the temporary directory
+    source = tmpdir.join("source_file.txt")
+    source.write("content")
+
+    # Create a protected directory (non-writable)
+    protected_dir = tmpdir.mkdir("protected_directory")
+    protected_path = str(protected_dir)
+
+    # Set the directory to read-only
+    if os.name == "nt":
```
martindurant, line 35: Does this clean up after the test?

martindurant: Oh sorry, I see you did this below; it would leave orphan directories if the test fails before getting there, so a fixture would be better

martindurant:
OK, I'm going to merge this as being better behaviour than before; however, we should 
- fix copy to allow on_error="return", returning the list of files that were successfully copied or exception instances for each failure. The output list will be the same length as the number of file copies attempted
- allow mv to have on_error="return"|"omit" (or "ignore"), which will delete only those files which copied. "return" would provide the exceptions back.

fsspec/tests/abstract/mv.py
Skylion007: Would this leave orphaned directories though @martindurant ? Wouldn't the tmpdir fixture clean them up? Or am I missing? Or would the permission issue prevent that...

Skylion007:
@martindurant Why not just allow on_error to be a callback object? That seems like it could be useful instead of adding more and more strings that backends potentially need to support?

martindurant:
> allow on_error to be a callback object?

If you can figure out a way to phrase that... but of course, we have quite some code depending on the current structure. Maybe an enum is a good idea, so at least the possibilities are clearly defined.

fsspec/tests/abstract/mv.py
martindurant: That's what I am thining. I don't know how smart the tmpdir fixture is, I am surprised if it considers permissions.

PR #1591, opened 2024-04-23
Stringify url in `url_to_fs`

mariosasko:
Call `stringify_path` on the `url` value in `url_to_fs` to be consistent with `get_fs_token_paths`.


mariosasko:
This minor inconsistency has caused us problems in `datasets`: https://github.com/huggingface/datasets/issues/6829.

martindurant:
I am sad to see the proliferation of Path in python (jokes)

PR #1592, opened 2024-04-23
Add bulk_mkdirs

martindurant:


fsspec/spec.py
```diff
@@ -306,6 +306,9 @@ def makedirs(self, path, exist_ok=False):
         """
         pass  # not necessary to implement, may not have directories
 
+    def bulk_makedirs(self, path, **kw):
```
Skylion007, line 309: Why do we need a special function here instead of modifying the makedirs function?

martindurant: I had imagined it like cp/cp_file

Skylion007: Why can't we just override makedirs to no check if the dirrectory exists for object stores since we directories aren't really a concept there?

martindurant: _makedirs is overidden by implementations already, so every implementation would have to make the list-of-paths check or do some sort of super()

Skylion007: Ah, I see now what you mean... BTW, @martindurant this might be a good time to start introducing the typing.override hints I mentioned in a Github issue.

Skylion007: I am still confused is this suppose to overridden by the implementation? Are there are implementations that override it currently?

martindurant: No there are none, but s3fs would be the first, since we don't want to check the existence of every directory, only the root bucket, or try to make the bucket multiple times.

martindurant: If you can think of a simpler way to do it, happy to hear!

PR #1602, opened 2024-05-15
Start enforcing bugbear rules (B)

DimitriPapadopoulos:


fsspec/implementations/cached.py
```diff
@@ -438,9 +437,6 @@ def __getattribute__(self, item):
             "cache_size",
             "pipe_file",
             "pipe",
-            "isdir",
```
martindurant, line 441: These ones, plus `_mkcache` above should very likely be kept in this list - the caching filesystem can care about intercepting the target filesystem for these operations. On what basis are they being removed?

DimitriPapadopoulos: They are duplicate items of a set.

fsspec/caching.py
```diff
@@ -597,7 +597,7 @@ def __init__(
 
             self.data = dict(zip(offsets, blocks))
         else:
-            self.data = data
+            self.data = {}
```
Skylion007, line 600:
```suggestion
            self.data = dict()
```
`{}` is an empty set, not an empty dict. I'm surprised mypy didn't throw a fit. 

Technically behavior change since the empty dict is copy instead of pass by reference, but should be fine I suppose.

Skylion007 (changes requested):
1 major bug in fsspec caching. Otherwise, looks good to me. Can we also enable the 'B' checks in ruff in pyproject.toml?

fsspec/caching.py
DimitriPapadopoulos: I could change to:
```python
        else if data is None:
            self.data = dict()
        else:
            self.data = data
```
However, I believe the most common idiom is to handle `None` and empty containers the same: create a new empty container.

martindurant: > {} is an empty set, not an empty dict

This is not so

```python
>>> type({})
dict
```

This is because `{}` was around in the language before sets existed at all (dicts were there from the start)

martindurant: ruff says:
```
fsspec/caching.py:600:25: C408 Unnecessary `dict` call (rewrite as a literal)
```

Skylion007: Whoops, got them mixed up. 👍

fsspec/caching.py
```diff
@@ -597,7 +597,7 @@ def __init__(
 
             self.data = dict(zip(offsets, blocks))
         else:
-            self.data = data
+            self.data = dict()
```
Skylion007, line 600:
```suggestion
            self.data = {}
```
Apologies for my incorrect feedback before.

fsspec/caching.py
DimitriPapadopoulos: Perhaps rule C408 should be disabled.

Skylion007 (approved):

Skylion007:
Please uncomment this line to enable the bugbear rules:
https://github.com/fsspec/filesystem_spec/blob/0bb3f26c412d7ad9b2d52a5c32265014709d1c1f/pyproject.toml#L152

DimitriPapadopoulos:
We're not there yet, the following rules need to be have not yet been applied:
* [B007](https://docs.astral.sh/ruff/rules/unused-loop-control-variable/)
* [B011](https://docs.astral.sh/ruff/rules/assert-false/)
* [B019](https://docs.astral.sh/ruff/rules/cached-instance-method/)
* [B020](https://docs.astral.sh/ruff/rules/loop-variable-overrides-iterator/)
* [B026](https://docs.astral.sh/ruff/rules/star-arg-unpacking-after-keyword-arg/)
* [B028](https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/)
* [B034](https://docs.astral.sh/ruff/rules/re-sub-positional-args/)
* [B904](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/)

My fear was that the maintainers wouldn't accept all rules, so I started with the most useful. Do you think I should enforce them all?

martindurant:
I believe going piecemeal is the best approach, as you have been doing

DimitriPapadopoulos:
What I can do is enable `B` rules in general, and ignore the above specific rules.

fsspec/caching.py
```diff
@@ -574,7 +574,7 @@ def __init__(
         blocksize: int,
         fetcher: Fetcher,
         size: int,
-        data: dict[tuple[int, int], bytes] = {},
+        data: dict[tuple[int, int], bytes] = None,
```
martindurant, line 577: The type hint is now wrong. It could be made into a class attribute hint instead...

fsspec/compression.py
```diff
@@ -139,7 +139,7 @@ def _fetch_range(self, start, end):
 try:
     import snappy
 
-    snappy.compress
+    snappy.compress  # noqa: B018
```
martindurant, line 142: We could call this on something
```suggestion
    snappy.compress(b"0")
```

fsspec/tests/test_core.py
```diff
@@ -21,7 +21,9 @@
 
 
 @contextmanager
-def tempzip(data={}):
+def tempzip(data=None):
+    if data is None:
+        data = {}
```
martindurant, line 26: I usually write 
```suggestion
    data = data or {}
```

fsspec/tests/test_utils.py
```diff
@@ -381,7 +381,7 @@ def client(self):
             return mock
 
         def func_2(self):
-            assert False, "have to overwrite this"
+            assert False, "have to overwrite this"  # noqa: B011
```
martindurant, line 384: `raise AssertError` ?

fsspec/compression.py
DimitriPapadopoulos: Perhaps even `b""`?

fsspec/tests/test_core.py
DimitriPapadopoulos: I prefer it too, but other maintainers find it less readable. Will change it.

fsspec/tests/test_utils.py
DimitriPapadopoulos: Or perhaps disable this rule in all tests?

fsspec/caching.py
DimitriPapadopoulos: Indeed `dict` would be consistent with `dict(zip(offsets, blocks))`, two lines above.

fsspec/caching.py
DimitriPapadopoulos: I've added `Optional[]` for consistency with the other arguments, which all have a type hint.

fsspec/compression.py
martindurant: Apparently b"" is fine - I had assumed it needed some content.

fsspec/tests/test_utils.py
martindurant: This is certainly a special case, it's not normal to `assert False`! On second thoughts, I'm OK with the original `noqa` too.

martindurant:
I see you added additional rules one by one - is the process now finished?

fsspec/tests/test_utils.py
DimitriPapadopoulos: I changed to `raise AssertionError` in the mean time. Is that OK?

Skylion007 (approved):

fsspec/compression.py
```diff
@@ -139,7 +139,7 @@ def _fetch_range(self, start, end):
 try:
     import snappy
 
-    snappy.compress
+    snappy.compress(b"")
```
Skylion007, line 142: Wait, this one isn't quite right, it's an attribute look up that could be an assert statement I suppose

DimitriPapadopoulos: Indeed, I guess the idea was that the `compress(b"")` call cannot fail; any error has to be an attribute look up. I can revert to the pure attribute look up with `# noqa`, it had the merit to be more explicit.

martindurant: An assert in regular code is not recommended, since they can be turned off for the whole interpreter session. I think the function call is fine, or if we wanted a different function call to appease linting, we could do a `getattr`.

Background: `snappy-python` is not the only installable to provide a package called `snappy` (not my fault!), so this is a way to differentiate. Actually, snappy-compressing whole files (as opposed to buffers within files) is super-rare, so we could perhaps even remove this, rather than increase import time by checking this. We should probably not be importing anything at import time!

```
In [1]: import fsspec
In [2]: import sys
In [3]: "snappy" in sys.modules
Out[3]: True
```

I intend to merge this PR as is; @DimitriPapadopoulos or @Skylion007 , you are most welcome to make a PR to defer imports in the fsspec.compression module.

PR #1607, opened 2024-05-22
Update README.md - Add monthly downloads badge

ayushjain01:
Added a monthly downloads badge from pip Trends to the README.
View more at - https://piptrends.com/widgets/fsspec

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 [![Anaconda-Server Badge](https://anaconda.org/conda-forge/fsspec/badges/version.svg)](https://anaconda.org/conda-forge/fsspec)
 ![Build](https://github.com/fsspec/filesystem_spec/workflows/CI/badge.svg)
 [![Docs](https://readthedocs.org/projects/filesystem-spec/badge/?version=latest)](https://filesystem-spec.readthedocs.io/en/latest/?badge=latest)
-[![PyPi downloads](https://img.shields.io/pypi/dm/fsspec?label=pypi%20downloads&style=flat)](https://pepy.tech/project/fsspec)
 [![fsspec Downloads Last Month](https://assets.piptrends.com/get-last-month-downloads-badge/fsspec.svg 'fsspec Downloads Last Month by pip Trends')](https://piptrends.com/package/fsspec)
 
 A specification for pythonic filesystems.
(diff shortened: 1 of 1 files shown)


martindurant:
How is this different from the "pypi downloads" immediately above?

ayushjain01:
The stats are slightly more accurate and refreshed daily, you can see the significant difference in the download counts between the two badges. Also there are more options in terms of badges and widgets to display stats about the package.
You can check them out here - https://piptrends.com/widgets/fsspec and https://piptrends.com/package/fsspec

martindurant:
So you would say this should _replace_ the other badge? Having both seems silly.

ayushjain01:
yes it should

ixmatus:
@martindurant a colleague of this person tried to open this exact PR against a project I'm the maintainer for as well (https://github.com/ixmatus/inflector/pull/16). This is a promotional campaign and they (or colleagues) appear to be doing it on reddit too, I think this is suspicious activity. They are creating an external dependency on a web property that (as far as I can tell) is not officially blessed by pypi. I strongly recommend reverting this change.

Users can get package statistics from `libraries.io` which is linked to directly by pypi itself.

martindurant:
Thanks for the heads up, @ixmatus : #1636 . @ayushjain01 , you have until tomorrow to respond.

ixmatus:
@martindurant I appreciate you're giving them the benefit of the doubt and a chance to object but I thought you should know I went through quite a few python github projects and left this warning where they've shot gunned this exact change out.

The innocent interpretation is that it's promotional and SEO link juicing their site. The nefarious interpretation is that they've now acquired first-time contributor status to your repository so GitHub now gives them the "Contributor" label on any subsequent PRs they open (a little far fetched social engineering attack vector but I work in security so I tend towards paranoid).

ayushjain01:
Thank you for your feedback and concerns. I'd like to clarify that this is not a promotional campaign. As stated in the PR description, our intention is simply to share a tool we have developed, which we believe could be beneficial to the community.

Our badge system offers highly accurate statistics, which is why we wanted to make it available. If the link to the pip trends page is a concern, we are more than willing to remove it.

We respect your decision if you prefer not to add more badges to your README. Our primary goal is to address any misunderstandings and assure you that this is a genuine effort to contribute to the open-source community by sharing our work on GitHub.

Thank you for considering our contribution.

martindurant:
Thanks for replying @ayushjain01 . In fact, I ended up simply removing the download badge - fsspec doesn't need to advertise it's stats, and users can find the information for themselves if they wish.

PR #1608, opened 2024-05-22
Fix ruff 0.4.4 issue

DimitriPapadopoulos:


DimitriPapadopoulos:
I don't see that error locally. I wonder which version of _ruff_ is being used by CI. That's the downside of using pytest-ruff, it seems unable to handle the version of _ruff_. Why use it in the first place?

DimitriPapadopoulos:
It's actually _pre-comit_:
```
▶ Run pre-commit run --show-diff-on-failure --color=always --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check docstring is first.................................................Passed
check json...........................................(no files to check)Skipped
check yaml...............................................................Passed
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
```
I can fix the version of _ruff_ used by _pre-commit_.

PR #1610, opened 2024-05-22
Disable more ruff rules

DimitriPapadopoulos:
They are not compatible with `ruff format` or are actual anti-patterns.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,6 +189,23 @@ ignore = [
     # Fix these codes later
     "G004",
     "PERF203",
+    # https://github.com/astral-sh/ruff/issues/7871
+    "UP038",
+    # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
+    "W191",
+    "E111",
+    "E114",
+    "E117",
+    "D206",
+    "D300",
+    "Q000",
+    "Q001",
+    "Q002",
+    "Q003",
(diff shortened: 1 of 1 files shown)


martindurant:
@Skylion007 , thoughts ?

pyproject.toml
```diff
@@ -189,6 +189,23 @@ ignore = [
     # Fix these codes later
     "G004",
     "PERF203",
+    # https://github.com/astral-sh/ruff/issues/7871
+    "UP038",
```
Skylion007, line 193: This should be disabled, but it should also be tabled by setting the target python version low enough.

DimitriPapadopoulos: Could you provide more precise guidance? Right now, the minimal supported version is defined in  `pyproject.toml`:
```toml
requires-python = ">=3.8"
````
What would you like to change? Also, why should the minimal supported version of Python depend on a ruff rule?

martindurant: @Skylion007 , are you OK with having "UP038" in this list for now?

Skylion007: Leave it disabled, it doesn't affect us anyway. 

If it's really triggering despite the Python version, this might be a ruff a bug that we should file upstream though. It's suppose to have version checks to prevent it suggesting upgrades which break the min version.

Issue #1572, opened 2024-04-14
2024.3.1: pytest fails in 5 `fsspec/implementations/tests/test_dbfs.py::test_dbfs*` units


kloczek:
I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix> using `installer` module
- run pytest with $PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>
- build is performed in env which is *`cut off from access to the public network`* (pytest is executed with `-m "not network"`)

<details>
<summary>Here is pytest output:</summary>

```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-fsspec-2024.3.1-2.fc37.x86_64/usr/lib64/python3.10/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-fsspec-2024.3.1-2.fc37.x86_64/usr/lib/python3.10/site-packages
+ /usr/bin/pytest -ra -m 'not network'
============================= test session starts ==============================
platform linux -- Python 3.10.14, pytest-8.1.1, pluggy-1.4.0
rootdir: /home/tkloczko/rpmbuild/BUILD/filesystem_spec-2024.3.1
configfile: pyproject.toml
plugins: vcr-1.0.2, mock-3.14.0, asyncio-0.23.6
asyncio: mode=strict
collected 1618 items / 8 skipped

fsspec/implementations/tests/local/local_test.py ....................... [  1%]
........................................................................ [  5%]
........................................                                 [  8%]
fsspec/implementations/tests/memory/memory_test.py ..................... [  9%]
........................................................................ [ 14%]
..........................................                               [ 16%]
fsspec/implementations/tests/test_archive.py ........................... [ 18%]
..........................................................ssssssssssssss [ 22%]
sss                                                                      [ 22%]
fsspec/implementations/tests/test_cached.py ............................ [ 24%]
..............ssssss...............................                      [ 27%]
fsspec/implementations/tests/test_common.py ssss                         [ 28%]
fsspec/implementations/tests/test_data.py ..                             [ 28%]
fsspec/implementations/tests/test_dbfs.py FFFFFss                        [ 28%]
fsspec/implementations/tests/test_dirfs.py ............................. [ 30%]
[...]

PR #1611, opened 2024-05-22
Remove tests from package

cachitas:
Once installed, the package got ~11x bigger with the release of `2024.5.0`.

I see that you are now shipping files under `tests` folders that were not included before. Going through the latest commits, I couldn't tell if this was on purpose or just overlooked with the transition to the **hatch** build system.

In this PR I set an `exclude` rule to mimic the state of the distributed packages as they were in version `2024.3.1`.

I am available to discuss this further or make any adjustments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,8 +140,7 @@ raw-options = {'version_scheme'='post-release'}
 [tool.hatch.build.hooks.vcs]
 version-file = "fsspec/_version.py"
 
-[tool.hatch.build.targets.wheel]
-package = ["fsspec"]
+[tool.hatch.build]
 exclude = ["**/tests/*", "!**/tests/abstract/"]
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,6 +140,10 @@ raw-options = {'version_scheme'='post-release'}
 [tool.hatch.build.hooks.vcs]
 version-file = "fsspec/_version.py"
 
+[tool.hatch.build.targets.wheel]
+package = ["fsspec"]
+exclude = ["**/tests/*", "!**/tests/abstract/"]
+
 [tool.ruff]
 target-version = "py38"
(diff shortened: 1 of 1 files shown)


martindurant:
I am regretting trying out hatch. Too many things to take care of. Once you have the template, it's all the same, but still...

martindurant:
(I see the wheel package size of 172kiB for 2024.3.1 and 316kiB for 2024.5.0, where does "11x" come from?)

martindurant:
I'm not sure we care about installed size, but I'm happy to hear otherwise. However, it does seem reasonable to keep the same behaviour as pre-hatch. I'll leave this open a couple of days to see if there are more opinions.

cachitas:
I agree, the size shouldn't be an issue nowadays. I just mentioned it because it was what triggered me into looking more closely to this new release, wondering what could be going on to justify such a difference.

mgorny:
While I don't mind removing them from wheels/installed files, Linux distributions would really appreciate tests back in sdist, since we actually run them.

martindurant:
Some seem to test them anyway: https://github.com/fsspec/filesystem_spec/issues/1572

Actually, the trend seems to have been to move tests out of the source directory altogether.

How would one include tests in one release type and not the other?

mgorny:
I suppose the simplest way would be to move tests out of package directory, e.g. into the top-level `test` directory inside the git repo. Then you wouldn't have to exclude parts of the package to avoid installing them.

cachitas:
Note that in this PR I ended up not removing all test files, I just applied the exclusions that were in place before migrating to hatch.

> How would one include tests in one release type and not the other?

Here, I applied the same rule to both `sdist` and `wheel`, but you can specify different rules for different targets (see [example](https://hatch.pypa.io/1.9/config/build/#patterns) from hatch documentation).

```toml
[tool.hatch.build.targets.wheel]
...

[tool.hatch.build.targets.sdist]
...
```

This way is possible to exclude all test related files from the `wheel` and keep some/all in `sdist`.

PR #247, opened 2020-03-17
JSON serialise

martindurant:
Fixes #199

PR #1612, opened 2024-05-28
Extend JSON serialization capabilities

DarkLight1337:
#247 first introduced JSON serialization of filesystem objects. However, it fails to handle `storage_args` and `storage_options` that contain `Path` or other filesystems. This can occur in a few cases, such as:

- When passing a `Path` instance to construct a `DirFileSystem` (not officially documented but the `Path` object is automatically converted to a string in the initializer, so it is functional)
- When passing a filesystem to wrapper filesystems such as `CachingFileSystem`, `DirFileSystem` and `ReferenceFileSystem`.

This PR makes it possible to serialize such filesystems.

### DIFF
--- a/fsspec/json.py
+++ b/fsspec/json.py
@@ -56,11 +56,15 @@ class FilesystemJSONDecoder(json.JSONDecoder):
     def try_resolve_fs_cls(cls, dct: Dict[str, Any]):
         with suppress(Exception):
             if "cls" in dct:
-                fs_cls = _import_class(dct["cls"])  # if this fails, we get None
-                if issubclass(fs_cls, AbstractFileSystem):
-                    return fs_cls
-            if "protocol" in dct:
-                return get_filesystem_class(dct["protocol"])
+                try:
+                    fs_cls = _import_class(dct["cls"])  # if this fails, we get None
+                    if issubclass(fs_cls, AbstractFileSystem):
+                        return fs_cls
+                except Exception:
+                    if "protocol" in dct:
+                        return get_filesystem_class(dct["protocol"])
+
+                    raise
 
         return None
 
--- a/fsspec/json.py
+++ b/fsspec/json.py
@@ -57,11 +57,11 @@ class FilesystemJSONDecoder(json.JSONDecoder):
         with suppress(Exception):
             if "cls" in dct:
                 try:
-                    fs_cls = _import_class(dct["cls"])  # if this fails, we get None
+                    fs_cls = _import_class(dct["cls"])
                     if issubclass(fs_cls, AbstractFileSystem):
                         return fs_cls
                 except Exception:
-                    if "protocol" in dct:
+                    if "protocol" in dct:   # Fallback if cls cannot be imported
                         return get_filesystem_class(dct["protocol"])
 
                     raise
--- a/fsspec/json.py
+++ b/fsspec/json.py
@@ -61,7 +61,7 @@ class FilesystemJSONDecoder(json.JSONDecoder):
                     if issubclass(fs_cls, AbstractFileSystem):
                         return fs_cls
                 except Exception:
-                    if "protocol" in dct:   # Fallback if cls cannot be imported
+                    if "protocol" in dct:  # Fallback if cls cannot be imported
                         return get_filesystem_class(dct["protocol"])
 
                     raise
--- a/fsspec/json.py
+++ b/fsspec/json.py
@@ -45,10 +45,7 @@ class FilesystemJSONDecoder(json.JSONDecoder):
         with suppress(Exception):
             fqp = dct["cls"]
 
-            try:
-                path_cls = _import_class(fqp)
-            except Exception:
-                raise
+            path_cls = _import_class(fqp)
 
             if issubclass(path_cls, PurePath):
                 return path_cls
@@ -58,18 +55,12 @@ class FilesystemJSONDecoder(json.JSONDecoder):
     @classmethod
     def try_resolve_fs_cls(cls, dct: Dict[str, Any]):
         with suppress(Exception):
-            fqp = dct["cls"]
-
-            try:
-                fs_cls = _import_class(fqp)
-            except Exception:
-                if "protocol" in dct:
-                    return get_filesystem_class(dct["protocol"])
-
-                raise
-
-            if issubclass(fs_cls, AbstractFileSystem):
-                return fs_cls
+            if "cls" in dct:
+                fs_cls = _import_class(dct["cls"])  # if this fails, we get None
+                if issubclass(fs_cls, AbstractFileSystem):
+                    return fs_cls
+            if "protocol" in dct:
+                return get_filesystem_class(dct["protocol"])
 
         return None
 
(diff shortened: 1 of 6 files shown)


fsspec/spec.py
```diff
@@ -115,6 +118,59 @@ class AbstractFileSystem(metaclass=_Cached):
...
+    storage_args: Tuple[Any, ...]
+    storage_options: Dict[str, Any]
+
+    class _JSONEncoder(json.JSONEncoder):
+        def default(self, o: Any) -> Any:
+            if isinstance(o, AbstractFileSystem):
+                return o.to_dict()
+            if isinstance(o, PurePath):
+                cls = type(o)
+                return {"cls": f"{cls.__module__}.{cls.__name__}", "str": str(o)}
```
martindurant, line 131: This is effectively pickling and a potential security hole (deser requires only an import, not an execute)

fsspec/spec.py
```diff
@@ -115,6 +118,59 @@ class AbstractFileSystem(metaclass=_Cached):
...
+                parse_constant=parse_constant,
+                strict=strict,
+                object_pairs_hook=object_pairs_hook,
+            )
+
+        def custom_object_hook(self, dct: Dict[str, Any]):
+            if "cls" in dct:
+                try:
+                    cls = _import_class(dct["cls"])
+                except (ImportError, ValueError, RuntimeError, KeyError):
```
martindurant, line 161: What are the various possible exceptions here?

fsspec/spec.py
```diff
@@ -115,6 +118,59 @@ class AbstractFileSystem(metaclass=_Cached):
...
+                    return AbstractFileSystem.from_dict(dct)
+
+                if issubclass(cls, PurePath):
+                    return cls(dct["str"])
+                else:
+                    return AbstractFileSystem.from_dict(dct)
```
martindurant, line 167: If anything else besides Path or filesystem slips into the JSON here, this line will fail

fsspec/spec.py
```diff
@@ -1425,17 +1465,51 @@ def from_json(blob):
...
+        Dictionary with keys cls (the python location of this class),
+        protocol (text name of this class's protocol, first one in case of
+        multiple), args (positional args, usually empty), and all other
+        kwargs as their own keys.
+        """
+        cls = type(self)
+        proto = self.protocol
+
+        return dict(
+            cls=f"{cls.__module__}.{cls.__name__}",
```
martindurant, line 1485: Perhaps prefer with ":", which is slightly more readable?

martindurant (approved):
Very minor comments

fsspec/spec.py
DarkLight1337: Could you elaborate? Unlike pickling, this does not allow arbitrary code execution since the object needs to be importable in the first place.

fsspec/spec.py
DarkLight1337: I copied this from the original code.

fsspec/spec.py
DarkLight1337: I copied this from the original code, but it also supports `module:class` format so I'll change it.

fsspec/spec.py
DarkLight1337: I'll make it more robust.

fsspec/spec.py
DarkLight1337: I have made the code more robust regardless of the exception type.

fsspec/spec.py
martindurant: Importing a module executes code. Now, most python packages don't have side-effects on import, and of course only installed packages are importable. Furthermore, the code very reasonably checks for if `issubclass(path_cls, PurePath)` when loading. I would suggest a single sentence in the load-json docstring saying that it will do imports.

We do this kind of thing all the time (e.g., when creating filesystem instances by looking in the registry), but it's nice to be more explicit about this kind of thing.

fsspec/spec.py
```diff
@@ -115,6 +119,87 @@ class AbstractFileSystem(metaclass=_Cached):
...
+                    return fs_cls
+
+            return None
+
+        def custom_object_hook(self, dct: Dict[str, Any]):
+            if "cls" in dct:
+                if (obj_cls := self.try_resolve_fs_cls(dct)) is not None:
+                    return AbstractFileSystem.from_dict(dct)
+                if (obj_cls := self.try_resolve_path_cls(dct)) is not None:
+                    return obj_cls(dct["str"])
```
martindurant, line 196: If these both fail, you end up with just a dictionary?

fsspec/spec.py
```diff
@@ -115,6 +119,87 @@ class AbstractFileSystem(metaclass=_Cached):
...
+
+        @classmethod
+        def try_resolve_path_cls(cls, dct: Dict[str, Any]):
+            with suppress(Exception):
+                fqp = dct["cls"]
+
+                try:
+                    path_cls = _import_class(fqp)
+                except Exception:
+                    raise
```
martindurant, line 166: why catch Exception only to raise it again, all within the outer suppress block?

martindurant (commented):
The whole encoder/decoder classes are quite long now, there's no reason they need to be in the main `spec` module - can they be pulled into a different file?

fsspec/spec.py
DarkLight1337: Yes, that is the default behaviour when decoding a dictionary from JSON.

fsspec/spec.py
DarkLight1337: It's just to have a consistent control flow with the code for decoding the filesystem class. If you think it's too verbose then I can simplify it.

fsspec/spec.py
DarkLight1337: That sounds good, thanks for pointing that out!

DarkLight1337:
> The whole encoder/decoder classes are quite long now, there's no reason they need to be in the main `spec` module - can they be pulled into a different file?

Sure - I'll move it to `fsspec.json`.

DarkLight1337:
I have finished addressing your comments.

fsspec/json.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def try_resolve_path_cls(cls, dct: Dict[str, Any]):
+        with suppress(Exception):
+            fqp = dct["cls"]
+
+            try:
+                path_cls = _import_class(fqp)
+            except Exception:
+                raise
```
martindurant, line 51:
```suggestion
            path_cls = _import_class(fqp)
```

fsspec/json.py
```diff
@@ -0,0 +1,86 @@
...
+            try:
+                fs_cls = _import_class(fqp)
+            except Exception:
+                if "protocol" in dct:
+                    return get_filesystem_class(dct["protocol"])
+
+                raise
+
+            if issubclass(fs_cls, AbstractFileSystem):
+                return fs_cls
```
martindurant, line 72:
```suggestion
            if "cls" in dct:
                fs_cls = _import_class(dct["cls"])  # if this fails, we get None
                if issubclass(fs_cls, AbstractFileSystem):
                    return fs_cls
            if "protocol" in dct:
                return get_filesystem_class(dct["protocol"])
```

PR #1613, opened 2024-05-29
Add dependencies for building docs

DarkLight1337:
I found that I couldn't build the documentation just from installing `dev` and `test` dependencies. It turns out that `pyproject.toml` currently does not specify the dependencies for building the documentation.

This PR adds the `doc` extra which includes such dependencies, making it easier to set up the environment for building documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,13 @@ CI runtime. For local use, pick a version suitable for you.
 mamba create -n fsspec -c conda-forge  python=3.9 -y
 conda activate fsspec
 
-# Standard dev test install.
-pip install -e ".[dev,test]"
+# Standard dev install with docs and tests.
+pip install -e ".[dev,doc,test]"
 
 # Full tests except for downstream
 pip install s3fs
 pip uninstall s3fs
-pip install -e .[dev,test_full]
+pip install -e .[dev,doc,test_full]
 pip install s3fs --no-deps
 pytest -v
 
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -3,9 +3,3 @@ channels:
   - defaults
 dependencies:
   - python=3.9
-  - docutils<0.17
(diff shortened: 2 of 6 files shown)


martindurant:
This essentially replicates the environment in https://github.com/fsspec/filesystem_spec/blob/master/docs/environment.yml . It seems like a bad idea to have the environment in two places.

DarkLight1337:
> This essentially replicates the environment in https://github.com/fsspec/filesystem_spec/blob/master/docs/environment.yml . It seems like a bad idea to have the environment in two places.

I have addressed this in the latest commit.

martindurant:
```
  - (b'{"key": "value"}')
  ?           -
  + (b'{"key":"value"}')
```
Apparently the JSON serialization changed format slightly

DarkLight1337:
> ```
>   - (b'{"key": "value"}')
>   ?           -
>   + (b'{"key":"value"}')
> ```
> 
> Apparently the JSON serialization changed format slightly

Seems that this was introduced by #1562. Imo we should fix that in a separate PR.

martindurant:
The tests passed at the time, so some version got updated. Since it's a one-character fix, I'd be happy to see it here.

DarkLight1337:
> The tests passed at the time, so some version got updated. Since it's a one-character fix, I'd be happy to see it here.

The problem is not that trivial. Seems that the separator in `json.dumps` depends on whether `json` or `ujson` is being used to serialize the data.

martindurant:
It's find to importorskip("ujson") for the test, and we know that the package will be there in CI.

DarkLight1337:
> It's find to importorskip("ujson") for the test, and we know that the package will be there in CI.

I found a better solution, which is to use the same json module to encode the dictionary for the reference value.

PR #1627, opened 2024-06-15
Fix `AbstractFileSystem.from_dict` not being idempotent

DarkLight1337:
The existing implementation mutates the input dictionary which causes the function to fail if called multiple times in succession. This PR fixes this issue.


martindurant:
Good spot.

PR #1628, opened 2024-06-17
DirFS: Fix `pipe(dict[str, bytes])`

lgeiger:
`.pipe()` can be called with a dictionary as input. This is currently not supported by `DirFS` as `._join()` silently maps the dictionary input to a list which causes a cryptic error message error when trying to use `.pipe()` with dictionary input.


lgeiger:
Thanks for the fast review, really appreciated 🚀 

Would it be possible to release a new `fsspec` version that includes this fix?

Issue #1626, opened 2024-06-14
HTTPFileSystem breaks when range requests are not supported


dholth:
```
from fsspec.core import url_to_fs

ufs, url = url_to_fs("https://example.org/")
f = ufs.open("https://example.org/no-range-requests/data.txt")
f.seek(8192)
f.read(1)
```

Will raise `# ValueError: The HTTP server doesn't appear to support range requests.`

Different caching options seem to be ineffective.

martindurant:
Specifically: 
- HTTPFile, which supports seeking, sends range requests and is refused by the server
- HTTPStreamFile does not use range, but does not support seek().
- HTTPStreamFile ought to support some caching modes, particularly "all" and "first", but is hard-coded to "none".

dholth:
IIUC, for my use case I could

- use `simplecache::` or `filecache::`
- enhance those so that the cache only keeps a file or two if I'm worried about disk space, like python TemporaryFile's delete-file-on-close
- Edit HTTPStreamFile so that it can fetch a range, if that range is right after whatever we've fetched so far, and use the file-level "all" cache instead of the filesystem wrappers `simplecache::`

martindurant:
cf https://github.com/fsspec/filesystem_spec/discussions/1629

PR #1631, opened 2024-06-19
http file system directed to stream by an "Accept-Ranges": "none" response

itcarroll:
## Background

`HTTPFileSystem._open` chooses between `HTTPFile` or `HTTPStreamFile`, and only the former allows random access, by using the `"Range"` header in subsequent HTTP GET requests. Servers do not always respond predictably to partial content requests, but they have the option of explicitly discouraging the use of `"Range"` by including `"Accept-Ranges": "none"` in response headers. It's rarely done, but the lack of a `"Accept-Ranges"` response does not (in practice) indicate the lack of support for the `"Range"` header.

## Changes

Previously, `HTTPFileSystem._open` only used `HTTPStreamFile` when NOT block caching (which is not the default) or when the content size could not be determined in advance (i.e. rarely). This PR makes a response header with `"Accept-Range": "none"` into another rare cause for the switch to streaming.

- `HTTPFileSystem._open` ~~now always makes a `self.info` call (even given `size`) 🤨~~ may see a "partial" key in `self.info` results
- `http._file_info` checks for the `"Accept-Ranges"` header in the response
- The return value of `HTTPFileSystem._info` may include `"partial": False`
- Tested with `test_no_range_support`
  - in its first parameterization of `header`, the `ignore_range` value was irrelevant and removed
  - a new parameterization was added with `accept_range` with `conftest` updated to implement

## Alternative

It seems like an alternative approach is to stick with `HTTPFile` but force `_fetch_all` for read. Please comment if that is preferred ... I haven't figured out what does and doesn't use caching.

~~## Draft~~

~~I'm in the process of working with a NASA DAAC that does not provide partial content to test the approach, and will remove the "Draft" status when successful. Reviews in the meantime are very welcome.~~ ~~Ugh, I give up waiting for them. Please review when able!~~


fsspec/implementations/http.py
```diff
@@ -358,9 +358,10 @@ def _open(
         kw = self.kwargs.copy()
         kw["asynchronous"] = self.asynchronous
         kw.update(kwargs)
-        size = size or self.info(path, **kwargs)["size"]
+        info = self.info(path, **kwargs)
```
martindurant, line 361: We really don't want to do this unless necessary. I think that, in the case that `size` is provided, we can ssume that ranges should work and we can do random access. This moves the responsibility to the calling code.

itcarroll: Will change to:

```suggestion
        partial = True
        if size is None:
            info = self.info(path, **kwargs)
            size = info["size"]
            partial = info.get("partial", partial)
        session = sync(self.loop, self.set_session)
        if block_size and size and partial:
```

itcarroll:
Testing with the DAAC providing `"Accept-Range": "none"` and `earthaccess` jumped through one hoop (getting `HTTPStreamFile` rather than `HTTPFile`, but revealed another. When I get an `HTTPStreamFile` through `earthaccess.open`, there's an error in `fsspec.caching.BaseCache._fetch` because `self.size` is not set. First step to resolving is making a test to cover it, but I'm currently not sure how to trigger this use of caching directloy (i.e. without `earthaccess.open`). Investigating.

martindurant:
You seem to have the same situation as https://github.com/fsspec/filesystem_spec/issues/1626 - there are some suggestions in that thread of what might be done.

itcarroll:
I think I'm actually running into nsidc/earthaccess#610. If I understand correctly what you say about the cache type being hard coded to "none" (i.e. BaseCache), then there should not have been a call to the cache (and no problem with the lack of `self.size`) if not for the `earthaccess` bug.

If they agree, I will mark this ready for review. Caching on `HTTPStreamFile` would be great, but a separate (harder!) contribution.

itcarroll:
Waiting for a review of nsidc/earthaccess#620, and I won't forget to pick this up after.

martindurant:
OK thanks :)

itcarroll:
@martindurant The downstream issue has resolved as I'd hoped. I've sync'd this branch and ran tests locally. Ready for workflow approval and review. Thanks for your patience!

martindurant:
Perfect, thank you.

PR #1632, opened 2024-06-22
Fix dict representation not being JSON serializable

DarkLight1337:
This PR fixes `AbstractFileSystem.to_dict()` failing to serialize recursive filesystem/paths which caused the output to not be JSON serializable.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1486,10 +1486,15 @@ class AbstractFileSystem(metaclass=_Cached):
 
         json_decoder = FilesystemJSONDecoder()
 
-        def deserialize(obj: Any):
+        def deserialize(obj: Any) -> Any:
             if isinstance(obj, dict):
-                return json_decoder.custom_object_hook(obj)
-            
+                obj = json_decoder.custom_object_hook(obj)
+                if isinstance(obj, dict):
+                    return {k: deserialize(v) for k, v in obj.items()}
+
+            if isinstance(obj, (list, tuple)):
+                return [deserialize(v) for v in obj]
+
             return obj
 
         dct = dict(dct)  # Defensive copy
@@ -1501,10 +1506,10 @@ class AbstractFileSystem(metaclass=_Cached):
         dct.pop("cls", None)
         dct.pop("protocol", None)
 
-        args = tuple(deserialize(arg) for arg in dct.pop("args", ()))
-        kwargs = {k: deserialize(v) for k, v in dct.items()}
-
-        return cls(*args, **kwargs)
+        return cls(
+            *deserialize(dct.pop("args", ())),
+            **deserialize(dct),
+        )
 
     def _get_pyarrow_filesystem(self):
         """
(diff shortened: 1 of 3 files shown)


fsspec/tests/test_spec.py
```diff
@@ -912,6 +984,20 @@ def test_dict_idempotent():
...
+    b = DummyTestFS(2, bar=Path("baz"))
+    c = DummyTestFS(3, baz=b)
+
+    outa = a.to_dict()
+    outb = b.to_dict()
+    outc = c.to_dict()
+
+    json.dumps(outa)
+    json.dumps(outb)
+    json.dumps(outc)
```
martindurant, line 998: These show there is no error, but should we be checking their output too?

fsspec/spec.py
```diff
@@ -1437,14 +1437,28 @@ def to_dict(self) -> Dict[str, Any]:
         multiple), ``args`` (positional args, usually empty), and all other
         keyword arguments as their own keys.
         """
+        from .json import FilesystemJSONEncoder
+
+        json_encoder = FilesystemJSONEncoder()
+
+        def serialize(obj: Any) -> Any:
```
martindurant, line 1444: Can we put these functions (serialize/deserialize) into the `json` module? They don't appear to depend on `self`. I'm not totally sure why this was necessary at all, I thought the whole purpose of the Encoder is to use it to walk the object tree,

fsspec/spec.py
```diff
@@ -1437,14 +1437,28 @@ def to_dict(self) -> Dict[str, Any]:
...
+            if isinstance(obj, (str, int, float, bool)):
+                return obj
+            if isinstance(obj, dict):
```
martindurant, line 1447: consider `collections.abc.Mapping` and `collections.abc.Iterable`? Note that the latter also catches strings, which would then need to be tested for first.

fsspec/tests/test_spec.py
DarkLight1337: The output is already checked by another test (`test_dict_fs_attr`). Here we specifically check whether the dict is serializable.

fsspec/spec.py
DarkLight1337: The dict methods do not use the JSON encoder/decoder since those are only capable of converting to/from JSON strings directly

fsspec/spec.py
DarkLight1337: Sure!

DarkLight1337:
Looks like we might need to lock the NumPy version to avoid binary incompatibility with SciPy.

martindurant:
You mean this cryptic message on windows only?
> OSError: Error loading libscipy_openblas64_-fb1711452d4d8cee9f276fd1449ee5c7.dll; The operation completed successfully.

I don't know where the scipy dependency is coming from, but we could remove it in the windows run.

DarkLight1337:
> You mean this cryptic message on windows only?
> 
> > OSError: Error loading libscipy_openblas64_-fb1711452d4d8cee9f276fd1449ee5c7.dll; The operation completed successfully.
> 
> I don't know where the scipy dependency is coming from, but we could remove it in the windows run.

Yeah.

Issue #1638, opened 2024-06-27
DirFileSystem.ls(path, detail=True) hits AssertionError with S3FileSystem


metadaddy:
This is a similar issue to #924.

**Code**

```python
fs = DirFileSystem(f'/{bucket_name}', S3FileSystem())

files = fs.ls('/')
```

**Error**

```text
Traceback (most recent call last):
  File "/Users/ppatterson/src/b2_zip_files/app.py", line 40, in <module>
    files = fs.ls('/')
            ^^^^^^^^^^
  File "/Users/ppatterson/live_read_demo/b2_zip_files/lib/python3.11/site-packages/fsspec/implementations/dirfs.py", line 253, in ls
    entry["name"] = self._relpath(entry["name"])
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ppatterson/live_read_demo/b2_zip_files/lib/python3.11/site-packages/fsspec/implementations/dirfs.py", line 70, in _relpath
    assert path.startswith(prefix)
AssertionError
```

The problem is that `DirFileSystem.__init__` sets `self.path` via `fs._strip_protocol(path)`, which returns `/bucket_name`. However, `S3FileSystem._ls` returns filenames of the form `bucket_name/path/to/file.ext` (no leading `/`). Hence, in the following code from `DirFileSystem._relpath`, the assertion fails:

```python
prefix = self.path + self.fs.sep # '/bucket_name/'
assert path.startswith(prefix) # 'bucket_name/path/to/file.ext' does not start with '/bucket-name/'!
```

It doesn't look like it's possible to set a flag to have S3FileSystem prepend a `/` to paths, so I think the solution is to make the leading `/` optional in `DirFileSystem._relpath`. Something like:

```python
    def _relpath(self, path):
        if isinstance(path, str):
            if not self.path:
                return path
            if path == self.path:
                return ""
            prefix = self.path + self.fs.sep if path.startswith(self.fs_sep) else self.path[1:] + self.fs.sep
            assert path.startswith(prefix)
            return path[len(prefix) :]
        return [self._relpath(_path) for _path in path]
```

I'll file a PR with this change.

PR #1639, opened 2024-06-27
DirFS: Handle paths with no leading `/` (#1638)

metadaddy:
If you wrap an `S3FileSystem` with `DirFileSystem`, `ls()` breaks, since `DirFileSystem` assumes that the underlying FS returns paths with a leading `/`, but `S3FileSystem` does not (see #1638).

This change to _relpath() removes the leading `/` from the `prefix` used in constructing the relative path when the incoming `path` has no leading `/`, but the `DirFileSystem` path does.

### DIFF
--- a/fsspec/implementations/dirfs.py
+++ b/fsspec/implementations/dirfs.py
@@ -66,7 +66,9 @@ class DirFileSystem(AsyncFileSystem):
                 return path
             # We need to account for S3FileSystem returning paths that do not
             # start with a '/'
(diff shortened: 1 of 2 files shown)


martindurant:
Thanks for sorting this out.
It sort of feels untidy, but I can't think of a better way to do this :)

metadaddy:
Yeah - I spent a while looking at DirFS and S3FS before concluding that this was the way. The root cause, I think, is that `fsspec` itself doesn't specify whether or not a leading slash is required/expected, while DirFS assumes it will be present.

Thanks for fixing the `lint` nit!

PR #1640, opened 2024-06-29
Update ruff to 0.5.0 & add rules

Skylion007:
Updates ruff to latest version and enables a few new useful rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,18 +154,20 @@ select = [
     # fix noqas in fsspec/implementations/http.py
     "ASYNC",
     "B",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -154,18 +154,20 @@ select = [
     # fix noqas in fsspec/implementations/http.py
     "ASYNC",
     "B",
-    "I",    # isort
```
Skylion007, line 157: The ISORT rule shouldn't be needed since we are using ruff fmt now anyway.

martindurant:
+1

PR #1641, opened 2024-06-29
Improve exception trace with RUF B904

Skylion007:
Improve the exception trace of all exceptions in filesystem spec by using the "from" keyword to specify a cause. I also enable a ruff rule B904 to require this cause in the future. I also enable RUF015 which ensure that we always access the first instance of iterable in constant time instead of using a linear time list or tuple constructor. I also had to change one of the exception catches to fix this in memory mapping.

Overall, this should prevent fsspec from swallowing the cause of exceptions nearly as much and return more info to the user when an exception occurs.


fsspec/implementations/dbfs.py
```diff
@@ -264,9 +264,9 @@ def _send_to_api(self, method, endpoint, json):
             try:
                 exception_json = e.response.json()
             except Exception:
-                raise e
+                raise e from None
```
martindurant, line 267: This is weird, no?
Is this to suppress the "while an exception was being handled" message

fsspec/implementations/dbfs.py
```diff
@@ -212,9 +212,9 @@ def mv(
             )
         except DatabricksException as e:
             if e.error_code == "RESOURCE_DOES_NOT_EXIST":
-                raise FileNotFoundError(e.message)
+                raise FileNotFoundError(e.message) from e
             elif e.error_code == "RESOURCE_ALREADY_EXISTS":
-                raise FileExistsError(e.message)
+                raise FileExistsError(e.message) from e
```
martindurant, line 217: I wonder if there is a level of verbosity in the upstream exception when we shouldn't do this. We already extracted the salient information from `e`.

martindurant (approved):
This was all auto-generated, correct?

I wonder if there any cases in which `raise from ...` causes bloat in the printout. That aside, I am in favour, es it at least provides the theoretical option for calling code to be more discriminating when handing our exceptions.

fsspec/implementations/dbfs.py
Skylion007: We don't get the traceback from e

Skylion007:
I manually added these to clarify.

Issue #1571, opened 2024-04-12
SMB: errors ignored when connecting


frafra:
The system tries to connect 5 times in a row, ignoring all errors. Since the for loop has no `else` branch, the function does not raise any error when all tentative failed.

https://github.com/fsspec/filesystem_spec/blob/05e7d80ab2affaa01505ff2602b0097c38ad7688/fsspec/implementations/smb.py#L123-L136

martindurant:
You are quite right! Would you like to implement a fix?

frafra:
I would suggest dropping the multiple tentative: if the connection is unstable, it should be up to the user to rely on some automatic retry mechanism such as tenacity or backoff. What is your opinion on that?

martindurant:
I don't know what the user's workflow might be or how they would apply those tools, but in our own CI pipelines, SMB *is* unstable for some reason, and this loop makes tests fail less often.

frafra:
Since it is a problem with how the CI is configured, I would suggest applying such workaround to the CI only,

martindurant:
Plenty of other implementations have retry loops for connection errors, why do you think it's a problem here?

frafra:
Well, I am not familiar with SMB, but generally a failing connection hints at some underlying issues. If SMB is intrinsically unstable, it would be nice to have a reference to in the code (why wait 0.1 seconds? why trying just 5 times?).
Automatic retry can be a feature whenever it is needed, but the developer should be in control of that; otherwise one could end up with a connection which usually takes 5 tentative, and sometimes 6, so it would apparently fail randomly without even having a warning logged by the library.

Is SMB supposed to randomly fail? Is it a documented behavior? Or is it a common misconfiguration or glitch that the developer/user should be made aware of?

GalLadislav:
Hi, i would add to this discussion with my problem. In my environment there are set login restrictions that lock my account after 3 or 5 invalid logins. One day everything worked good, but another for some reason i started seeing:
`smbprotocol.exceptions.SMBAuthenticationError: Failed to authenticate with server: SpnegoError (1): SpnegoError (7): Major (458752): No credentials were supplied, or the credentials were unavailable or inaccessible, Minor (2529639053): No Kerberos credentials available (default cache: FILE:/tmp/krb5cc_1000), Context: Processing security token, Context: Unable to negotiate common mechanism`
[...]

PR #1650, opened 2024-07-20
Add exception handling for SMBFileSystem init connection errors

GalLadislav:
Related to #1571 

This PR implements exception handling for `SMBFileSystem._connect()` which brings changes to the current implementation:
- reraise exceptions related to authentication issues 
- reraise ValueError exceptions related to configuration
- retry on connection issues raised by socket
- retry all other exceptions
- distinguished attempts and retries, where always one attempt to register a session is made

Retrying all other exceptions could be dropped in the future once all exceptions suited for retry are identified. Until then, we should probably not affect current behavior as much as we should.

Other feature proposed in this PR is raising latest exception after all retries are exhausted. This will at least inform a user about excepted issue. (related question is whether to implement ExceptionGroup to inform about all exceptions)

Additionally, an implementation for exponential wait time growth from 0 to requested time (included) is added. 
The wait time is calculated using exponential function. For factor=1 all wait times will be equal to `register_session_retry_wait`. For any number of retries, the last wait time will be equal to `register_session_retry_wait` and for retries>1 the first wait time will be equal to `register_session_retry_wait / factor`.  
By default a factor is made 10 as most optimal (quick first attempts)

### DIFF
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -150,7 +150,7 @@ class SMBFileSystem(AbstractFileSystem):
                 # that the credentials are invalid and not a network issue.
                 raise
             except ValueError as exc:
-                if re.match(r"\[Errno -\d+]", str(exc)):
+                if re.findall(r"\[Errno -\d+]", str(exc)):
                     # This exception is raised by the smbprotocol.transport:Tcp.connect
                     # and originates from socket.gaierror (OSError). These exceptions might
                     # be raised due to network instability. We will retry to connect.
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -160,7 +160,7 @@ class SMBFileSystem(AbstractFileSystem):
         # will be equal to `wait`. For any number of retries the last wait time will be
         # equal to `wait` and for retries>2 the first wait time will be equal to `wait / factor`.
         wait_times = iter(
-            factor ** (n / n_waits - 1) * wait_time for n in range(1, n_waits + 1)
+            factor ** (n / n_waits - 1) * wait_time for n in range(0, n_waits + 1)
         )
 
         for attempt in range(self.register_session_retries + 1):
--- a/fsspec/implementations/smb.py
+++ b/fsspec/implementations/smb.py
@@ -110,16 +110,16 @@ class SMBFileSystem(AbstractFileSystem):
         register_session_retries: int
             Number of retries to register a session with the server. Retries are not performed
             for authentication errors, as they are considered as invalid credentials and not network
-            issues.
+            issues. If set to negative value, no register attempts will be performed.
         register_session_retry_wait: int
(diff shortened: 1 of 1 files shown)


fsspec/implementations/smb.py
```diff
@@ -138,9 +141,35 @@ def _connect(self):
...
+                    # All another ValueError exceptions should be raised, as they are not
+                    # related to network issues.
+                    raise exc
+            except Exception as exc:
+                # Save the exception and retry to connect. This except might be dropped
+                # in the future, once all exceptions suited for retry are identified.
+                retried_errors.append(exc)
+
+            if retry < self.register_session_retries - 1:
+                time.sleep(1)
```
martindurant, line 168: May as well make the retry time configurable and ideally exponential, `base * factor**retry`

GalLadislav: That is great idea. Should the exponential option be also configurable? To avoid confusing a user?

martindurant: That's an interesting idea, but not the way around it is is other implementations ([example](https://github.com/fsspec/gcsfs/blob/main/gcsfs/retry.py#L129)). At least, then I would call it "max_wait_time", but probably better to have "base_wait_time" and backoff factor.

fsspec/implementations/smb.py
```diff
@@ -128,7 +156,26 @@ def _port(self):
     def _connect(self):
         import time
 
-        for _ in range(self.register_session_retries):
+        if self.register_session_retries <= -1:
```
martindurant, line 159: Is there no way to connect, then? Perhaps should have `_connect(self, retries=None)` and allow overriding the original value if the method is to be called later.

GalLadislav: I assumed that _connect is private method only used for connecting on initialization. Also i tried to not change the actual behavior as much as possible, only target the issue at hand (compatibility issues). 
Since it is possible to connect later using fs methods and kwargs, i just places the option to not connect ( which was there since register_session_retries implementation ).

GalLadislav: Here, i am just making assumptions about fsspec and set consistencies for its features. Just by looking at SMBFileSystem implementation, i can se that each of its methods call smbclient which tries to find registered session OR create new session using passed kwargs. Following example is a working usage:
```python
from fsspec.implementations.smb import SMBFileSystem
smb = SMBFileSystem(host="mysmbhost", register_session_retries=-1)
...
smb.listdir("/sharefolder", username="user", password="password")
```
I do not know why would anyone wanted to use it like this. Maybe it could be useful where credentials are supplied at later time (i am here really only speculating). Either way i just followed already present option to not connect by setting retries to 0.

martindurant: I didn't realise that smbclient had an auto-connect if it hadn't already been done. Now I'm worried what might happen if a user wanted to connect to multiple SMB services, but let's leave that for now.

PR #1659, opened 2024-08-12
localfs: reduce stat calls during info

skshetry:
For symlinks, `LocalFileSystem.info()` was making 3 stat calls, first for the symlink file itself, second for the link target, and third to get `size` information of the link target.

This is wasteful, and `size` can be gathered from `stat` of the link target.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -79,6 +79,14 @@ class LocalFileSystem(AbstractFileSystem):
                 t = "file"
             else:
                 t = "other"
+
+            size = out.st_size
+            if link:
+                try:
+                    out2 = path.stat(follow_symlinks=True)
+                    size = out2.st_size
+                except OSError:
+                    size = 0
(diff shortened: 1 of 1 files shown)


fsspec/implementations/local.py
```diff
@@ -95,20 +104,15 @@ def info(self, path, **kwargs):
...
+            "size": size,
             "type": t,
             "created": out.st_ctime,
             "islink": link,
         }
         for field in ["mode", "uid", "gid", "mtime", "ino", "nlink"]:
             result[field] = getattr(out, f"st_{field}")
-        if result["islink"]:
+        if link:
             result["destination"] = os.readlink(path)
```
martindurant, line 115: Since we are doing micro-optimisation, might the destination path also be known?

skshetry: I don't think there is a way to get `destination` path without doing a `readlink`. `stat` struct does not have a path or a target path.

PR #1660, opened 2024-08-12
Apply more ruff rules

DimitriPapadopoulos:


fsspec/core.py
```diff
@@ -706,7 +706,7 @@ def _expand_paths(path, name_function, num):
         assert len(path) == num
         paths = list(path)
     else:
-        raise ValueError(
+        raise TypeError(
```
DimitriPapadopoulos, line 709: This is a breaking change. I have committed it for documentation purposes, but perhaps I should discard the relevant commit.

martindurant: I don't see a big reason to change it. Both say "you passed the wrong thing" in this context.

DimitriPapadopoulos: `TypeError` is for arguments of the wrong type, while `ValueError` is for argument with a wrong value.

martindurant: > TypeError is for arguments of the wrong type, while ValueError is for argument with a wrong value.

I know that's technically the case, but here that can be several types; let's not change it.

DimitriPapadopoulos: How is the number of possible types relevant?

Any way, I have removed that commit.

fsspec/implementations/reference.py
```diff
@@ -159,7 +159,7 @@ def open_refs(field, record):
             path = self.url.format(field=field, record=record)
             data = io.BytesIO(self.fs.cat_file(path))
             df = self.pd.read_parquet(data, engine="fastparquet")
-            refs = {c: df[c].values for c in df.columns}
+            refs = {c: df[c].to_numpy() for c in df.columns}
```
martindurant, line 162: I'm not sure that these are equivalent

fsspec/tests/test_utils.py
```diff
@@ -261,7 +261,6 @@ def test_common_prefix(paths, out):
     (
         (["/path1"], "/path2", False, ["/path2"]),
         (["/path1"], "/path2", True, ["/path2/path1"]),
-        (["/path1"], "/path2", False, ["/path2"]),
```
martindurant, line 264: Why remove cases?

fsspec/tests/test_utils.py
```diff
@@ -279,18 +278,6 @@ def test_common_prefix(paths, out):
             True,
             ["/path2/more/path1", "/path2/more/path2"],
         ),
-        (
```
martindurant, line 282: Why remove cases?

fsspec/tests/test_utils.py
DimitriPapadopoulos: The explanation is hidden in the commit message: the list contains duplicate cases.

fsspec/implementations/reference.py
DimitriPapadopoulos: The documentation of [DataFrame.values](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.values.html#pandas.DataFrame.values) reads:
> Return a Numpy representation of the DataFrame.
> 
> #### ⚠️ Warning
> 
> We recommend using [DataFrame.to_numpy()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_numpy.html#pandas.DataFrame.to_numpy) instead.

martindurant: This is Series.values ( https://pandas.pydata.org/docs/reference/api/pandas.Series.values.html ), which suggests `.array` is the more direct successor. By default, `.to_numpy` has `copy=False`, so maybe it amounts to the same thing. At least we should put in that kwarg explicitly to prevent surprises.

Issue #1662, opened 2024-08-14
Slow find on ZipFileSystem


johandahlberg:
I noticed that calling `find` on `ZipFileSystem` can be very slow for large archives, taking upwards to two minutes on zip-files containing ~5000 files. I have created an alternative `find` implementation (see below), that relies on the `infolist` in the zip file, which is considerable faster (~0.2 seconds).

I'm wondering if you would be interested in bringing these changes in to `fsspec`? For our particular use-case (reading hive-style partitioned parquet files from zip-archives) this is absolutely necessary, but I am not sure for how many others this is an issue, so I thought I'd ask.

```
class _CustomZipFileSystem(ZipFileSystem):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def find(self, path, maxdepth=None, withdirs=False, detail=False, **kwargs):
        if maxdepth is not None and maxdepth < 1:
            raise ValueError("maxdepth must be at least 1")

        result = {}

        def _below_max_recursion_depth(path):
            if not maxdepth:
                return True

            depth = len(path.split("/"))
            return depth <= maxdepth

        for zip_info in self.zip.infolist():
            file_name = zip_info.filename
            if not file_name.startswith(path.lstrip("/")):
                continue

            # zip files can contain explicit or implicit directories
            # hence the need to either add them directly or infer them
            # from the file paths
            if zip_info.is_dir():
                if withdirs:
                    if not result.get(file_name) and _below_max_recursion_depth(
                        file_name
                    ):
                        result[file_name.strip("/")] = (
                            self.info(file_name) if detail else None
                        )
                    continue
                else:
                    continue  # Skip along to the next entry if we don't want to add the dirs

            if not result.get(file_name):
                if _below_max_recursion_depth(file_name):
                    result[file_name] = self.info(file_name) if detail else None

                # Here we handle the case of implicitly adding the
                # directories if they have been requested
                if withdirs:
                    directories = file_name.split("/")
                    for i in range(1, len(directories)):
                        dir_path = "/".join(directories[:i]).strip(
                            "/"
                        )  # remove the trailing slash, as this is not expected
                        if not result.get(dir_path) and _below_max_recursion_depth(
                            dir_path
                        ):
                            result[dir_path] = {
                                "name": dir_path,
                                "size": 0,
                                "type": "directory",
                            }

        return result if detail else sorted(list(result.keys()))
```

And some tests for this:

```
class TestCustomZipFileSystem:
[...]

PR #1664, opened 2024-08-16
Improve performance find zip archive

johandahlberg:
This is an attempt to address #1662. Working on a code base that makes extensive us of zip archives containing parquet files that need to be read on the fly to load data, I noticed that `find` was a major bottleneck in that process.

I have made a zip specific implementation of `find` here that relies on the file list of the zip file, rather than explicitly walking.

Here is how I measured the performance:

```
from fsspec.implementations.zip import ZipFileSystem

# Example achieve with roughly 900 files, in some deep directory structures
file_system = ZipFileSystem("example.zip")
%timeit file_system.find("/")
```

Performance for find on current `master`
```
2.14 s ± 146 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Performance for find with this fix applied:
```
318 µs ± 11.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Let me know what you think.

### DIFF
--- a/fsspec/implementations/zip.py
+++ b/fsspec/implementations/zip.py
@@ -137,52 +137,41 @@ class ZipFileSystem(AbstractArchiveFileSystem):
         if maxdepth is not None and maxdepth < 1:
             raise ValueError("maxdepth must be at least 1")
 
-        result = {}
-
         def _below_max_recursion_depth(path):
             if not maxdepth:
-                return True
+                return False
 
             depth = len(path.split("/"))
-            return depth <= maxdepth
+            return depth > maxdepth
 
-        for zip_info in self.zip.infolist():
-            file_name = zip_info.filename
-            if not file_name.startswith(path.lstrip("/")):
+        result = {}
+
+        # Remove the leading slash, as the zip file paths are always
+        # given without a leading slash
+        path = path.lstrip("/")
+
+        self._get_dirs()
+
+        # To match posix find, if an exact file name is given, we should
+        # return only that file
+        if path in self.dir_cache and self.dir_cache[path]["type"] == "file":
+            result[path] = self.dir_cache[path]
+            return result if detail else [path]
+
+        for file_path, file_info in self.dir_cache.items():
+            if not (path == "" or file_path.startswith(path + "/")):
                 continue
 
-            # zip files can contain explicit or implicit directories
-            # hence the need to either add them directly or infer them
-            # from the file paths
-            if zip_info.is_dir():
-                if withdirs:
-                    if not file_name in result and _below_max_recursion_depth(
-                        file_name
-                    ):
-                        result[file_name.strip("/")] = (
-                            self.info(file_name) if detail else None
-                        )
+            if _below_max_recursion_depth(file_path):
                 continue
 
-            if file_name not in result:
-                if _below_max_recursion_depth(file_name):
-                    result[file_name] = self.info(file_name) if detail else None
-
-                # Here we handle the case of implicitly adding the
-                # directories if they have been requested
+            if file_info["type"] == "directory":
                 if withdirs:
-                    directories = file_name.split("/")
-                    for i in range(1, len(directories)):
-                        dir_path = "/".join(directories[:i]).strip(
-                            "/"
-                        )  # remove the trailing slash, as this is not expected
-                        if not result.get(dir_path) and _below_max_recursion_depth(
-                            dir_path
-                        ):
-                            result[dir_path] = {
-                                "name": dir_path,
-                                "size": 0,
-                                "type": "directory",
-                            }
+                    if file_path not in result:
+                        result[file_path.strip("/")] = file_info
+                continue
+
+            if file_path not in result:
+                result[file_path] = file_info if detail else None
 
         return result if detail else sorted(result)
--- a/fsspec/implementations/zip.py
+++ b/fsspec/implementations/zip.py
@@ -144,8 +144,6 @@ class ZipFileSystem(AbstractArchiveFileSystem):
             depth = len(path.split("/"))
             return depth > maxdepth
 
-        result = {}
-
         # Remove the leading slash, as the zip file paths are always
         # given without a leading slash
         path = path.lstrip("/")
@@ -160,6 +158,7 @@ class ZipFileSystem(AbstractArchiveFileSystem):
 
         self._get_dirs()
 
+        result = {}
         # To match posix find, if an exact file name is given, we should
         # return only that file
         if path in self.dir_cache and self.dir_cache[path]["type"] == "file":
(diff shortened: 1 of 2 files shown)


fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
             out.name = info["name"]
         return out
+
+    def find(self, path, maxdepth=None, withdirs=False, detail=False, **kwargs):
+        if maxdepth is not None and maxdepth < 1:
+            raise ValueError("maxdepth must be at least 1")
+
+        result = {}
+
+        def _below_max_recursion_depth(path):
```
martindurant, line 142: maxdepth really exists for filesystems doing an explicit ls(), to prevent having to walk very deep trees. In this case, we consider all files anyway, so the code can be much simpler by ignoring maxdepth.

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+            if not maxdepth:
+                return True
+
+            depth = len(path.split("/"))
+            return depth <= maxdepth
+
+        for zip_info in self.zip.infolist():
+            file_name = zip_info.filename
+            if not file_name.startswith(path.lstrip("/")):
```
martindurant, line 151:
```suggestion
            if not file_name.startswith(path.lstrip("/")) + "/":
```
? 

So that find('apath') doesn't match 'apathlonger', only 'apath/nested'.

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+                continue
+
+            # zip files can contain explicit or implicit directories
+            # hence the need to either add them directly or infer them
+            # from the file paths
+            if zip_info.is_dir():
+                if withdirs:
+                    if not result.get(file_name) and _below_max_recursion_depth(
```
martindurant, line 159:
```suggestion
                    if not file_name in result and _below_max_recursion_depth(
```
seems more obvious

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+                        file_name
+                    ):
+                        result[file_name.strip("/")] = (
+                            self.info(file_name) if detail else None
+                        )
+                    continue
+                else:
+                    continue  # Skip along to the next entry if we don't want to add the dirs
+
+            if not result.get(file_name):
```
martindurant, line 169:
```suggestion
            if file_name not in result:
```

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+                if withdirs:
+                    if not result.get(file_name) and _below_max_recursion_depth(
+                        file_name
+                    ):
+                        result[file_name.strip("/")] = (
+                            self.info(file_name) if detail else None
+                        )
+                    continue
+                else:
+                    continue  # Skip along to the next entry if we don't want to add the dirs
```
martindurant, line 167:
```suggestion
                continue
```
?

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+
+            if not result.get(file_name):
+                if _below_max_recursion_depth(file_name):
+                    result[file_name] = self.info(file_name) if detail else None
```
martindurant, line 171: How is `info` got here? Don't we know everything about this path at this point?

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+                            "/"
+                        )  # remove the trailing slash, as this is not expected
+                        if not result.get(dir_path) and _below_max_recursion_depth(
+                            dir_path
+                        ):
+                            result[dir_path] = {
+                                "name": dir_path,
+                                "size": 0,
+                                "type": "directory",
+                            }
```
martindurant, line 188: I feel like this code is duplicated from elsewhere. That might be unavoidable, but perhaps worth thinking about refactoring something into utils.

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,59 @@ def _open(
...
+
+        return result if detail else sorted(result.keys())
```
martindurant, line 190:
```suggestion
        return result if detail else sorted(result)
```

fsspec/implementations/zip.py
johandahlberg: I agree that removing it would make this quite a lot simpler. I wondering, since `maxdepth` is part of the interface if you think that there is any risk that people using it are relying on it only returning files from a specific `maxdepth`? I think that might be a case for keeping it in?

fsspec/implementations/zip.py
johandahlberg: Yeah. That makes sense.

johandahlberg:
Thank you for the review @martindurant! I refactored the code quite a bit, to make use of the `dir_cache` instead of directly accessing the zip file. I think it made the code easier to follow.

Regarding removing the `maxdepth` that would be a change in the behavior compared to the `find` method on the `AbstractFileSystem` - since my initial strategy here was to try to implement something that had parity in the output I kept it in. If you still want me to remove it let me know and I'll do it (or feel free to push any changes you want to this branch if you prefer that).

I implemented it as suggested by you now, but just filtering for it at the end if it is set.

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,53 @@ def _open(
...
+            return depth > maxdepth
+
+        # Remove the leading slash, as the zip file paths are always
+        # given without a leading slash
+        path = path.lstrip("/")
+        path_parts = list(filter(lambda s: bool(s), path.split("/")))
+
+        def _matching_starts(file_path):
+            file_parts = filter(lambda s: bool(s), file_path.split("/"))
+            for a, b in zip(path_parts, file_parts):
```
martindurant, line 154: This could be expressed with `any` or `all` in one line, I think

martindurant:
I wonder, should this code go up into AbstractArchiveFileSystem, since it may have a similar performance benefit for libarchive or tar filesystems? I'm not sure if find is slow there or if there's anything else to worry about.

fsspec/implementations/zip.py
```diff
@@ -132,3 +132,53 @@ def _open(
...
+                if a != b:
+                    return False
+            return True
+
+        self._get_dirs()
```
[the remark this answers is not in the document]
johandahlberg: Yeah. It's in the same ball park.

johandahlberg:
@martindurant Let me know if you are happy with this or if there is anything else you'd like to see. When it comes to moving to evaluating this libarchive/tar file systems I unfortunately don't think that I will have the bandwidth to take that on.

martindurant:
> I unfortunately don't think that I will have the bandwidth to take that on.

Understood - maybe someone else gets the itch if indeed speed is a problem for the other ones.

johandahlberg:
@martindurant thank you for a very enjoyable review process, and thanks from brining this in. It will help our use-case immensely.

PR #1676, opened 2024-09-18
WIP: Add type annotations to base spec

edgarrmondragon:
See https://github.com/fsspec/filesystem_spec/issues/625#issuecomment-2332915497

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -10,7 +10,17 @@ import weakref
 from errno import ESPIPE
 from glob import has_magic
 from hashlib import sha256
-from typing import Any, ClassVar, Dict, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Literal,
+    List,
+    Tuple,
+    overload,
+)
 
 from .callbacks import DEFAULT_CALLBACK
 from .config import apply_config, conf
@@ -25,6 +35,12 @@ from .utils import (
     stringify_path,
     tokenize,
 )
+
+if TYPE_CHECKING:
+    import datetime
+
+    from .caching import BaseCache
+    from .mapping import FSMap
 
 logger = logging.getLogger("fsspec")
 
@@ -184,8 +200,16 @@ class AbstractFileSystem(metaclass=_Cached):
     def __reduce__(self):
         return make_instance, (type(self), self.storage_args, self.storage_options)
 
+    @overload
     @classmethod
-    def _strip_protocol(cls, path):
+    def _strip_protocol(cls, path: str) -> str: ...
+
+    @overload
+    @classmethod
+    def _strip_protocol(cls, path: List[str]) -> List[str]: ...
+
+    @classmethod
+    def _strip_protocol(cls, path) -> str | List[str]:
         """Turn path from fully-qualified to file-system-specific
 
         May require FS-specific handling, e.g., for relative paths or links.
@@ -277,7 +301,7 @@ class AbstractFileSystem(metaclass=_Cached):
         if self._intrans:
             self._invalidated_caches_in_transaction.append(path)
 
-    def mkdir(self, path, create_parents=True, **kwargs):
+    def mkdir(self, path: str, create_parents: bool = True, **kwargs):
         """
         Create directory entry at path
 
@@ -295,7 +319,7 @@ class AbstractFileSystem(metaclass=_Cached):
         """
         pass  # not necessary to implement, may not have directories
 
-    def makedirs(self, path, exist_ok=False):
+    def makedirs(self, path: str, exist_ok: bool = False):
         """Recursively make directories
 
         Creates directory at path and any intervening required directories.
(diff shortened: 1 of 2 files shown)


martindurant:
Thanks for giving this a start. Perhaps you already appreciate that this process will _not_ be easy for the package, and spec might have been the hardest place to start (since all these methods are overridden in various duck ways in the backend implementations).

mon:
Preface: Just a random user.

This is a really good start (I love me some types), though I feel like it could do with some minor improvements
- a few `object` annotations that could probably be narrowed to a `Protocol`
- Inconsistent typing for inferred kwargs (explicitly specified `recursive: bool = False` vs implied `on_error="omit"`)
- `open` could do with an overload to differentiate binary and text streams

...however, I feel like it could just devolve into a bunch of minor nitpicks, whereas this already provides lots of value as-is.

What's the strategy towards a merge here? Grab a bunch of popular downstream consumers and ensure these don't cause conflicts with any overloaded duck typing happening there?

martindurant:
@mon : as with my initial comments, I am happy to see typing leak slowly into fsspec, but the duck-type design of the (old!) code means that it will be hard to have it everywhere and certainly very hard to get mypy happy throughout the codebase, including downstream implementations.

It gets complicated fast, because of issues like people demanding support for Path instances where originally only strings were allowed. The end-user API is based on convenience, but library developers using fsspec want strict definitions - requirements that are hard to simultaneously meet.

> What's the strategy towards a merge here? Grab a bunch of popular downstream consumers and ensure these don't cause conflicts with any overloaded duck typing happening there?

It shouldn't be possible to break code with types alone, right?
However, I need people more experiences with type-ifying to review the code. It is not for *my* convenience, but for others, so it will not be high on my priority list.

mon:
> It shouldn't be possible to break code with types alone, right?

At runtime, there will be no effect, unless you type annotate forward/circular references. But these are found as soon as the module is imported, so shouldn't ever be a problem for you, only the person adding the types (and easily fixed by gating the import behind an `if TYPE_CHECKING` branch).

My worry was there might just be some downstream projects that, for example, extend the valid argument types for some functions, and making them specifically typed (instead of the default `Unknown`) will cause type errors from tools like Pylance. Personally I don't think that's an issue, and imperfect types is better than none.

fsspec/spec.py
```diff
@@ -381,7 +405,7 @@ def _ls_from_cache(self, path):
         except KeyError:
             pass
 
-    def walk(self, path, maxdepth=None, topdown=True, on_error="omit", **kwargs):
+    def walk(self, path: str, maxdepth=None, topdown=True, on_error="omit", **kwargs):
```
Skylion007, line 408: nit: on_error has only a few valid values, right? We should make it a Union of Type Literals

fsspec/spec.py
```diff
@@ -2040,29 +2096,29 @@ def close(self):
...
             self.close()
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"<File-like object {type(self.fs).__name__}, {self.path}>"
 
     __repr__ = __str__
 
-    def __enter__(self):
+    def __enter__(self) -> "AbstractBufferedFile":
```
Skylion007, line 2120: May want to consider typing_extensions.Self, if it's not already a project dependency, we should add as it's one of the top 10 most popular pypi downloads anyway.

Skylion007:
Nit: we may want to enable the relevent type checking rules in RUFF at some point.

fsspec/spec.py
```diff
@@ -606,7 +637,7 @@ def glob(self, path, maxdepth=None, **kwargs):
                 depth_double_stars = path[idx_double_stars:].count("/") + 1
                 depth = depth - depth_double_stars + maxdepth
             else:
-                depth = None
+                depth = None  # type: ignore[assignment]
```
Skylion007, line 640: Type depth above as an Optional and you won't need this type ignore.

Skylion007:
> @mon : as with my initial comments, I am happy to see typing leak slowly into fsspec, but the duck-type design of the (old!) code means that it will be hard to have it everywhere and certainly very hard to get mypy happy throughout the codebase, including downstream implementations.
> 
> It gets complicated fast, because of issues like people demanding support for Path instances where originally only strings were allowed. The end-user API is based on convenience, but library developers using fsspec want strict definitions - requirements that are hard to simultaneously meet.
> 
> > What's the strategy towards a merge here? Grab a bunch of popular downstream consumers and ensure these don't cause conflicts with any overloaded duck typing happening there?
> 
> It shouldn't be possible to break code with types alone, right? However, I need people more experiences with type-ifying to review the code. It is not for _my_ convenience, but for others, so it will not be high on my priority list.

In Python 3.9 or newer, it should not.

Skylion007:
@martindurant I have a lot of typing experience and am happy to look at this PR if you would like.

martindurant:
@Skylion007 yes please, much appreciated!

slmnsh:
Is this PR abandoned? I would love type hints in fsspec and can contribute as well.

martindurant:
I'll let the contributors in this PR reply, but it seems to have gone quiet indeed. @slmnsh : if you want to take this code and push forward or start again, you would be welcome to.

edgarrmondragon:
> Is this PR abandoned? I would love type hints in fsspec and can contribute as well.

Yeah, I don't think I'll be able to work on this for the foreseeable future.

Issue #1683, opened 2024-09-19
Add option to disable AutoAddPolicy for SSH backend and use known_hosts and ssh config file


mxmlnkn:
`AutoAddPolicy` makes the user vulnerable to man-in-the-middle attacks. I think, it should not be added by default and/or there should be an option to disable/enable it. If the user-wide `~/.ssh/known_hosts` file is correctly loaded, then the `AutoAddPolicy` should likely not be necessary in the first place. For SSH, this policy change needs to be enabled with `ssh -o StrictHostKeychecking=no ...` or via the SSH config file.

Loading the SSH config file is yet another feature request of mine. For example, currently, things like `ssh cluster` works because I have configured an alias for that in my `~/.ssh/config`, but `fsspec.open("ssh://cluster")` does not work because paramiko does not automatically load the usual OpenSSH files even though it has helper APIs to load them manually.

martindurant:
I am no expert at SSH, so would be happy to see reasonable changes like this in a PR. Also, be aware that [sshfs](https://github.com/fsspec/sshfs) might be more fully featured.

mxmlnkn:
I have opened a PR. Thanks for the pointer to Python sshfs based on asyncssh. I didn't see this mentioned in the API reference's list of [known implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#other-known-implementations). It is confusing having so many alternatives. Should the built-in one be deprecated then (if and only if sshfs turns out to be much more feature-complete)?

martindurant:
> I didn't see this mentioned

This is definitely an omission!

> Should the built-in one be deprecated then

I don't directly maintain sshfs; I would be happy to have it supersede the paramiko implementation if they think there would be no lost functionality.

mxmlnkn:
> > Should the built-in one be deprecated then
> 
> I don't directly maintain sshfs; I would be happy to have it supersede the paramiko implementation if they think there would be no lost functionality.

Any opinions about this @shcheklein @efiop ? Is it feature complete? What was the motivation for that SSHFS implementation? Is it better than the one shipped with fsspec? Is it still being worked on?

I was also wondering about the maintainers with access to the repository, but that seems to have been sorted out in https://github.com/fsspec/sshfs/pull/45.

As for the issues I have:

[...]

PR #1684, opened 2024-09-19
Parse ssh_config and known_hosts files

mxmlnkn:
Implements #1683


martindurant:
I wonder, is there a way to test this?

PR #1687, opened 2024-09-23
Skip test_deep_parq without kerchunk installed

QuLogic:
Similar to `test_append_parquet` (#1523).

### DIFF
--- a/fsspec/implementations/tests/test_reference.py
+++ b/fsspec/implementations/tests/test_reference.py
@@ -762,6 +762,7 @@ def test_append_parquet(lazy_refs, m):
 
 
 def test_deep_parq(m):
+    pytest.importorskip("kerchunk")
     zarr = pytest.importorskip("zarr")
     lz = fsspec.implementations.reference.LazyReferenceMapper.create(
         "memory://out.parq", fs=m


martindurant:
OK, but I don't immediately see why the test needs kerchunk. You are right that this is kerchunk-derived functionality, though.

QuLogic:
> OK, but I don't immediately see why the test needs kerchunk.

Here is the original backtrace I should have posted:
```
________________________________ test_deep_parq ________________________________
m = <fsspec.implementations.memory.MemoryFileSystem object at 0x7f43626c1be0>
    def test_deep_parq(m):
        zarr = pytest.importorskip("zarr")
        lz = fsspec.implementations.reference.LazyReferenceMapper.create(
            "memory://out.parq", fs=m
        )
        g = zarr.open_group(lz, mode="w")
        g2 = g.create_group("instant")
        g2.create_dataset(name="one", data=[1, 2, 3])
>       lz.flush()
fsspec/implementations/tests/test_reference.py:772: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
fsspec/implementations/reference.py:493: in flush
    self.write(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <fsspec.implementations.reference.LazyReferenceMapper object at 0x7f433a757ed0>
field = 'instant/one', record = 0, base_url = None, storage_options = None
    def write(self, field, record, base_url=None, storage_options=None):
        # extra requirements if writing
>       import kerchunk.df
E       ModuleNotFoundError: No module named 'kerchunk'
fsspec/implementations/reference.py:407: ModuleNotFoundError
```

martindurant:
OK, I had overlooked that. It could be worked around.

PR #1690, opened 2024-09-23
TST: Don't hard code server port in tests

QuLogic:
This allows running tests in parallel without running into "Port already in use" errors.


fsspec/tests/conftest.py
```diff
@@ -68,10 +78,12 @@ def do_GET(self):
         if file_path.endswith("/") and file_path.rstrip("/") in self.files:
             file_path = file_path.rstrip("/")
         file_data = self.files.get(file_path)
+        if callable(file_data):
+            file_data = file_data(self.server.server_port)
         if "give_path" in self.headers:
             return self._respond(200, data=json.dumps({"path": self.path}).encode())
         if "redirect" in self.headers and file_path != "/index/realfile":
-            new_url = f"http://127.0.0.1:{port}/index/realfile"
+            new_url = f"http://127.0.0.1:{self.server.server_port}/index/realfile"
```
martindurant, line 86: self.server.realfile?

fsspec/tests/conftest.py
```diff
@@ -169,13 +181,13 @@ def do_HEAD(self):
...
 def serve():
-    server_address = ("", port)
+    server_address = ("", 0)
     httpd = HTTPServer(server_address, HTTPTestHandler)
     th = threading.Thread(target=httpd.serve_forever)
     th.daemon = True
     th.start()
     try:
-        yield f"http://127.0.0.1:{port}"
+        yield f"http://127.0.0.1:{httpd.server_port}"
```
martindurant, line 190: Since we are choosing to become port agnostic, this may be a good time to also become server address agnostic. `httpd.server_name` should be populated here, and might be other than the standard loopback, e.g., for systems that are IPv6 only.

fsspec/tests/conftest.py
```diff
@@ -185,4 +197,5 @@ def serve():
 @pytest.fixture(scope="module")
 def server():
     with serve() as s:
-        yield s
+        server = SimpleNamespace(address=s, realfile=_make_realfile(s))
```
martindurant, line 200: named_tuple or dict? I haven't really seen SimpleNamespace used much. In effect, it's a dictionary here. May as well include server name and other strings that are templated throughout the tests.

martindurant (commented):
At first glance all looks well here, but I have not yet reviewed in detail. 

I have one main comment for you to consider before moving on.

fsspec/tests/conftest.py
QuLogic: `self.server` is the `HTTPServer`, not the `server` fixture, and doesn't have a `realfile` attribute. But this could probably use `_make_realfile` instead.

fsspec/tests/conftest.py
QuLogic: I thought that might be a bit overkill for tests.

Are there really any other strings that could be put in here? I replaced all the `realfile`s which seemed like the only common ones.

martindurant: Yeah, OK, maybe you're right :)

martindurant:
Test failures with a bunch of "Cannot connect to host 0.0.0.0:50610" look real - but only on windows.

QuLogic:
Unfortunately, it looks like `HTTPServer.server_address` doesn't expand to a concrete address, and I guess unlike Linux, Windows doesn't allow connecting to 0.0.0.0. But maybe I should have used `server_name`?

martindurant:
Well done :)

PR #1692, opened 2024-09-26
Adds engine option to `LazyReferenceMapper`

norlandrhagen:
Defaults to 'fastparquet', but allows for 'pyarrow'

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -158,7 +158,7 @@ class LazyReferenceMapper(collections.abc.MutableMapping):
             """cached parquet file loader"""
             path = self.url.format(field=field, record=record)
             data = io.BytesIO(self.fs.cat_file(path))
-            df = self.pd.read_parquet(data, engine="fastparquet")
+            df = self.pd.read_parquet(data, engine="pyarrow")
             refs = {c: df[c].to_numpy() for c in df.columns}
             return refs
 
@@ -465,7 +465,7 @@ class LazyReferenceMapper(collections.abc.MutableMapping):
         self.fs.mkdirs(f"{base_url or self.out_root}/{field}", exist_ok=True)
         df.to_parquet(
             fn,
-            engine="fastparquet",
+            engine="pyarrow",
             storage_options=storage_options
             or getattr(self.fs, "storage_options", None),
             compression="zstd",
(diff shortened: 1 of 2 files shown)


martindurant:
Are you aware of arro3, a new minimalist arrow implementation that does include a parquet reader? Maybe would be perfect for this, but perhaps still too immature to use.

norlandrhagen:
> Are you aware of arro3, a new minimalist arrow implementation that does include a parquet reader? Maybe would be perfect for this, but perhaps still too immature to use.

I just stumbled on [that](https://github.com/kylebarron/arro3) yesterday! Seems like a great idea, but yeah, it might be a bit too early to use 🤷

fsspec/implementations/reference.py
```diff
@@ -126,11 +132,20 @@ def __init__(
             Encode urls as pandas.Categorical to reduce memory footprint if the ratio
             of the number of unique urls to total number of refs for each variable
             is greater than or equal to this number. (default 10)
+        engine: Literal["fastparquet","pyarrow"]
+            Engine choice for reading parquet files. (default is "fastparquet")
         """
+
+        from importlib.util import find_spec
+
+        if find_spec("pyarrow") is None:
```
martindurant, line 141: This gets called even if engine is "fastparquet", no? That would make this unusable if pyarrow isn't installed.

fsspec/implementations/reference.py
```diff
@@ -463,18 +478,28 @@ def write(self, field, record, base_url=None, storage_options=None):
 
         fn = f"{base_url or self.out_root}/{field}/refs.{record}.parq"
         self.fs.mkdirs(f"{base_url or self.out_root}/{field}", exist_ok=True)
+
+        if self.engine == "pyarrow":
+            df_backend_kwargs = {}
```
martindurant, line 483: There are probably some good defaults here?

fsspec/implementations/reference.py
martindurant: (could test by monkeypatch on find_spec)

norlandrhagen: Great catch!

fsspec/implementations/reference.py
norlandrhagen: Good call! I added a few, I'll see how the tests go :) 

Absolutely not a parquet expert, so we'll see.

PR #1696, opened 2024-09-29
Make SFTPFile.seek return the new position and .write return the number of written bytes

### LINKED ISSUE #1695: File object opened via SSH is inconsistent with the file object API
This is kind of a duplicate of the [paramiko issue](https://github.com/paramiko/paramiko/issues/2452). But, I think it could also be fixed in fsspec. I'm not sure whether fsspec does define a standard for the file object. If not, maybe it should? I find it highly unexpected that for the file object:

 - seek does not return the new position
 - write does not return the number of bytes written
 - probably other issues?

This could be fixed here instead of upstream by adding a small wrapper around the paramiko file object. Of course, the same [discussion](https://github.com/fsspec/filesystem_spec/issues/1683#issuecomment-2371738666) about fsspec/sshfs applies here.

mxmlnkn:
Fixes #1695

### DIFF
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -13,6 +13,35 @@ from ..utils import infer_storage_options
 logger = logging.getLogger("fsspec.sftp")
 
 
+def _patch_SFTPFile(file):
+    """
+    This patcher tries to rectify the file API of paramiko.sftp_file.SFTPFile
+    to be more consistent with io.IOBase.
+    https://github.com/paramiko/paramiko/issues/2452
+    """
+
+    if getattr(file, "_patched_by_fsspec", False):
+        return file
+    file._patched_by_fsspec = True
+
+    self = file
+
+    real_seek = self.seek
+    def seek(offset: int, whence: int = 0) -> int:
+        result = real_seek(offset, whence)
+        return self.tell() if result is None else result
+    self.seek = seek
+
+    real_write = self.write
+    def write(data) -> int:
+        old_offset = self.tell()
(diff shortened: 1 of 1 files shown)


fsspec/implementations/sftp.py
```diff
@@ -13,6 +13,35 @@
 logger = logging.getLogger("fsspec.sftp")
 
 
+def _patch_SFTPFile(file):
```
martindurant, line 16: Hah, ugly :( (not a criticism of your work, just a shame)
Does this create reference cycles on `self` and the closure?

Other options might be:
- patch the upstream ftp.open to return a subclass `FsspecSShFile(SFTPFile, IOBase)`
- make our own class with the upstream file object as an attribute and pass all the methods through except for these two

Do these sound more palatable?

mxmlnkn: Yeah, I kinda agree that it is not the most beautiful. I also considered a separate file object class, but I didn't want to copy 10+ methods with the danger of missing some. Subclassing also does not work because the object is created inside SFTPClient.open.

I have also opened a PR to fix this [upstream](https://github.com/paramiko/paramiko/pull/2456). I guess it is fine to wait for that to be merged and simply close this PR and issue here. I might have been too impatient. Fixing this here has the advantage that it will work with older versions of paramiko just fine, though.

martindurant: Ah, I'm glad you made that PR. In that case let's wait - it doesn't appear that this part of the API gets touched much.

Issue #1697, opened 2024-09-29
Further test_zip failures with Python 3.13: `_compresslevel` vs. `compress_level`


mgorny:
```pytb
$ pytest fsspec/implementations/tests/test_zip.py -vv
========================================================= test session starts =========================================================
platform linux -- Python 3.13.0rc2, pytest-8.3.3, pluggy-1.5.0 -- /tmp/filesystem_spec/.venv/bin/python3.13
cachedir: .pytest_cache
rootdir: /tmp/filesystem_spec
configfile: pyproject.toml
collected 17 items                                                                                                                    

fsspec/implementations/tests/test_zip.py::test_info PASSED                                                                      [  5%]
fsspec/implementations/tests/test_zip.py::test_fsspec_get_mapper PASSED                                                         [ 11%]
fsspec/implementations/tests/test_zip.py::test_not_cached PASSED                                                                [ 17%]
fsspec/implementations/tests/test_zip.py::test_root_info PASSED                                                                 [ 23%]
fsspec/implementations/tests/test_zip.py::test_write_seek PASSED                                                                [ 29%]
fsspec/implementations/tests/test_zip.py::test_rw PASSED                                                                        [ 35%]
fsspec/implementations/tests/test_zip.py::test_mapper PASSED                                                                    [ 41%]
fsspec/implementations/tests/test_zip.py::test_zip_glob_star PASSED                                                             [ 47%]
fsspec/implementations/tests/test_zip.py::test_append PASSED                                                                    [ 52%]
fsspec/implementations/tests/test_zip.py::test_find_returns_expected_result_detail_true FAILED                                  [ 58%]
fsspec/implementations/tests/test_zip.py::test_find_returns_expected_result_detail_false PASSED                                 [ 64%]
fsspec/implementations/tests/test_zip.py::test_find_returns_expected_result_detail_true_include_dirs FAILED                     [ 70%]
fsspec/implementations/tests/test_zip.py::test_find_returns_expected_result_detail_false_include_dirs PASSED                    [ 76%]
fsspec/implementations/tests/test_zip.py::test_find_returns_expected_result_path_set PASSED                                     [ 82%]
fsspec/implementations/tests/test_zip.py::test_find_with_and_without_slash_should_return_same_result PASSED                     [ 88%]
fsspec/implementations/tests/test_zip.py::test_find_should_return_file_if_exact_match PASSED                                    [ 94%]
[...]

PR #1700, opened 2024-10-02
Limit fields checked in ZIP tests

martindurant:
Fixes https://github.com/fsspec/filesystem_spec/issues/1697


martindurant:
@mgorny , can you test, please?

mgorny:
I can confirm that the tests pass on 3.13.0rc3 with this change.

PR #1703, opened 2024-10-03
Avoid use of _pygit2.Blob.hex, which has been removed in pygit2 >= 1.15

### LINKED ISSUE #1702: AttributeError: '_pygit2.Blob' object has no attribute 'hex'
```python3
import fsspec
f = fsspec.open("git://tests/")
print(f.fs.ls("/"))
```

Results in:

```
  File "/home/user/.local/lib/python3.10/site-packages/fsspec/implementations/git.py", line 93, in ls
    "hex": obj.hex,
AttributeError: '_pygit2.Blob' object has no attribute 'hex'
```

After downgrading with: `python3 -m pip install --user --force-reinstall 'pygit2==1.14`, it works again. It seems that the API break occurred in 1.15. Seems like they don't follow semantic versioning.

This dependency version requirement should be added to the pyproject.toml, or worked around.

mxmlnkn:
Fixes #1702

### DIFF
--- a/fsspec/implementations/git.py
+++ b/fsspec/implementations/git.py
@@ -80,7 +80,7 @@ class GitFileSystem(AbstractFileSystem):
                         {
                             "type": "directory",
                             "name": "/".join([path, obj.name]).lstrip("/"),
-                            "hex": obj.hex,
+                            "hex": str(obj.id),
                             "mode": f"{obj.filemode:o}",
                             "size": 0,
                         }
@@ -90,7 +90,7 @@ class GitFileSystem(AbstractFileSystem):
                         {
                             "type": "file",
                             "name": "/".join([path, obj.name]).lstrip("/"),
-                            "hex": obj.hex,
+                            "hex": str(obj.id),
                             "mode": f"{obj.filemode:o}",
                             "size": obj.size,
                         }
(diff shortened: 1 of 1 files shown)


fsspec/implementations/git.py
```diff
@@ -70,6 +70,9 @@ def _get_kwargs_from_urls(path):
         return out
 
     def ls(self, path, detail=True, ref=None, **kwargs):
+        def get_hex(obj):
```
martindurant, line 73: I suppose rather than redefine this on every call, it could be a top-level function in the module.

mxmlnkn: I force-pushed to simply use `str(obj.id)` and tested it up until pygit2 1.4, which is the oldest version I was able to install. I suspect it works with all versions though.

mxmlnkn: Based on the [changelog](https://github.com/libgit2/pygit2/blob/master/CHANGELOG.md#0202-2014-02-04) it seems to have been added in pygit2 0.20.2 in 2014.

> New Object.id deprecates Object.oid https://github.com/libgit2/pygit2/pull/322

martindurant: Quite recent, then :)

PR #1706, opened 2024-10-04
Apply ruff/flake8-bugbear preview rule B901

DimitriPapadopoulos:
See [return-in-generator (B901)](https://docs.astral.sh/ruff/rules/return-in-generator/):

> Using `return {value}` in a generator function was syntactically invalid in Python 2. In Python 3 `return {value}` can be used in a generator; however, the combination of `yield` and `return` can lead to confusing behavior, as the `return` statement will cause the generator to raise `StopIteration` with the value provided, rather than returning the value to the caller.


fsspec/spec.py
```diff
@@ -428,11 +428,10 @@ def walk(self, path, maxdepth=None, topdown=True, on_error="omit", **kwargs):
         except (FileNotFoundError, OSError) as e:
             if on_error == "raise":
                 raise
-            elif callable(on_error):
+            if callable(on_error):
                 on_error(e)
-            if detail:
-                return path, {}, {}
-            return path, [], []
+            yield path, {}, {} if detail else path, [], []
```
martindurant, line 433:
```suggestion
            yield (path, {}, {}) if detail else (path, [], [])
```

fsspec/spec.py
```diff
@@ -428,11 +428,11 @@ def walk(self, path, maxdepth=None, topdown=True, on_error="omit", **kwargs):
...
             if on_error == "raise":
                 raise
-            elif callable(on_error):
+            if callable(on_error):
                 on_error(e)
-            if detail:
-                return path, {}, {}
-            return path, [], []
+            result = (path, {}, {}) if detail else (path, [], [])
+            yield result
```
DimitriPapadopoulos, line 434: Am I correct that we want both?
1. Yield an empty result (`yield`).
2. Raise `StopIteration` (`return`).

martindurant: Correct, the generator should stop after yielding the single tuple. I don't know why `return` should not be the simple way to achieve that...

fsspec/spec.py
```diff
@@ -428,11 +428,11 @@ def walk(self, path, maxdepth=None, topdown=True, on_error="omit", **kwargs):
...
+            return result
```
martindurant, line 435: You can't both yield and return the result, else there will be two tuples produced, not just once. I think you are right that the non-return variant of this is
```
yield result
raise StopIteration
```
which to my mind is far less attractive than `return result`.

fsspec/spec.py
martindurant: Testing this, it seems you are right, and the original code is equivalent to `raise StopIteration` by itself with no value at all. 

```python
In [4]: def genme():
   ...:     if True:
   ...:         return 1
   ...:     yield 2
   ...:

In [5]: list(genme())
Out[5]: []
```

DimitriPapadopoulos: Here's the error:
```
  Full diff:
  - []
  + [
  +     (
  +         '/do_not_exist',
  +         [],
  +         [],
  +     ),
  + ]
```
Actually, I think we do not want to yield anything. So `return` was correct. However, I don't understand why `return path, [], []` instead of a mere `return`.

martindurant: Before we were returning nothing (even though the return statement had a value), now we yield the tuple first.

PR #1710, opened 2024-10-08
Test with Python 3.13

edgarrmondragon:
Starting as draft since this seems to be blocked by:

- https://github.com/libgit2/pygit2/issues/1317
- https://github.com/apache/arrow/issues/43519

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,7 +13,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


martindurant:
There is no py3.13 on conda yet at all, but it should be with us soon. We may need a separate trimmed down environment specifically for it until all the other packages catch up.

martindurant:
Seems all good. @edgarrmondragon - anything to add here?

edgarrmondragon:
> Seems all good. @edgarrmondragon - anything to add here?

good to go from my end 😄

PR #1732, opened 2024-10-22
Provide default _fetch_range implementation for File

martindurant:
Since this was not marked as NotImplemented


martindurant:
@kylebarron

PR #1734, opened 2024-10-22
Fix broken async reference file system `_cat_file` method

mpiannucci:
Adds missing return in `_cat_file` method of `ReferenceFileSystem`

Im not sure how to best test this because i dont think any of the tests from test_reference use an async filesystem and s3 maybe shouldnt be tested here? 

Closes https://github.com/fsspec/filesystem_spec/issues/1735

### DIFF
--- a/fsspec/implementations/tests/test_reference.py
+++ b/fsspec/implementations/tests/test_reference.py
@@ -513,6 +513,28 @@ def test_cat_file_ranges(m):
     assert fs.cat_file("d", 1, -3) == other[4:10][1:-3]
 
 
+@pytest.mark.asyncio
+async def test_async_cat_file_ranges():
+    fs = fsspec.filesystem(
+        "reference",
+        fo={
+            "version": 1,
+            "refs": {
+                "reference_time/0": [
+                    "http://noaa-nwm-retro-v2-0-pds.s3.amazonaws.com/full_physics/2017/201704010000.CHRTOUT_DOMAIN1.comp",
+                    39783,
+                    12,
+                ],
+            },
+        },
+        remote_protocol="http",
+        remote_options={"asynchronous": True},
+        asynchronous=True,
+    )
+
+    assert await fs._cat_file("reference_time/0") == b'x^K0\xa9d\x04\x00\x03\x13\x01\x0f'
+
+
 @pytest.mark.parametrize(
     "fo",
     [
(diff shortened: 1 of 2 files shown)


martindurant:
Can we make a test? It will need to run in an async func, with an async backend (http?) - can we guarantee something that won't disappear?

mpiannucci:
The async function part is easy, are you ok with pointing to s3 data via http similar to the kerchunk tests? Thats the easiest thing i can think of doing for now

martindurant:
I suppose we can reconsider if it starts to fail. The download should be small though (part of a file only), as the GHA CI runners are pretty weak.

mpiannucci:
Okay I have a test added and working well!

mpiannucci:
@martindurant do you have any idea why this test is failing? It works fine locally...

martindurant:
I also ran it successfully locally. Is there a chance the server explicitly won't take connections from CI?

You might try 
```
    fs={"https": fss}
```
(because the default protocol is "http", so maybe another filesystem instance is being silently created)

martindurant:
Or, perhaps passing `remote_protocol` . Actually, it is https? I might have been confused.

martindurant:
@mpiannucci : so now it passes. I don't know why, but an http FS must have leaked from another (async) test case. Probably worth tracking that down at some point, or to make a FS that lasts all session *or* make sure that when an FS cleans itself, it also gets removed from the instance cache.

mpiannucci:
Thanks for the help!

fsspec/implementations/tests/test_reference.py
```diff
@@ -513,6 +513,35 @@ def test_cat_file_ranges(m):
...
+    fss = fsspec.filesystem("https", asynchronous=True)
+    session = await fss.set_session()
+
+    fs = fsspec.filesystem(
+        "reference",
+        fo={
+            "version": 1,
+            "refs": {
+                "reference_time/0": [
+                    "https://noaa-nwm-retro-v2-0-pds.s3.amazonaws.com/full_physics/2017/201704010000.CHRTOUT_DOMAIN1.comp",
```
QuLogic, line 528: This seems to require the network; should this test get a `vcr`?

martindurant: I suppose we are assuming public archival S3 data should always be available.

QuLogic: S3 might be, but the network won't (e.g., no network on Fedora builders, and probably many other distributions.)

martindurant: True, we should have a pytest mark for tests that require network (at all). There will always be some. There are also other possible edges that have come up before, such as the default/only loopback interface being IPv6 rather than IPv4.

Issue #1726, opened 2024-10-17
Ask transfer tosfs to the fsspec official org


yanghua:
Hi community:

I am the contributor of [tosfs](https://github.com/volcengine/tosfs), which is an implementation of fsspec. I filed this issue to ask if we can transfer it to the fsspec official org.

Now I would like to introduce tosfs briefly.

Tosfs is an SDK that implements fsspec and aims to provide a convenient Python filesystem interface for [TOS（Tinder Object Storage）](https://www.volcengine.com/docs/6349/74820). By the way, TOS is the object store service of the volcano engine(Cloud as a Service Provider owned by [ByteDance](https://www.bytedance.com/en/)).

If any question or guidance need to be followed please let me know.

Thanks,
Vino

martindurant:
Please remind me on Monday, when I'll have more time to look into it. Is tosfs mentioned in the fsspec "known implementations" docs?

yanghua:
>> Please remind me on Monday, when I'll have more time to look into it.

Ok.

>> Is tosfs mentioned in the fsspec "known implementations" docs?

No, it hasn't been added to "known implementations", I will add it.

yanghua:
The doc pr is here #1728

yanghua:
@martindurant Following your suggestion, ping you @martindurant on Monday.

yanghua:
Hi @martindurant I have sent the admin permission invitation. thanks~

yanghua:
@martindurant , thanks, close this issue.

PR #1739, opened 2024-10-24
Register tosfs to fsspec registry

yanghua:
Since we have transferred the [tosfs](https://github.com/fsspec/tosfs) into fsspec org. The related issue is here: #1726 .
IMO, we can add it to the default registry.


yanghua:
@martindurant please take a look, when you have time, thanks.

yanghua:
Hi @martindurant Would you please have a look? Very appreciate.

PR #1743, opened 2024-10-31
Fix parsing s3 accesspoint url

wang2bo2:
Parsing S3 access point URI, like s3://arn:aws:s3:us-west-2:1234:accesspoint/abc/123.jpg, causes problems in `infer_storage_options` for the `:accesspoint` was parsed as port number. This PR addresses that.


martindurant:
Whilst this appears to do the right thing, I wonder if you see any way that this can be isolated in s3fs rather than here?

martindurant:
ping @wang2bo2

martindurant:
I think it would make sense to look for "accesspoint" in s3fs, but still use to infer_storage_options in the common case. Perhaps the accesspoint branch could simply modify the URL into something that makes sense. An ARN description isn't very much like a URL anyway...

PR #1745, opened 2024-11-05
Add async wrapper for sync FS

moradology:
This PR adds support for arbitrarily wrapping synchronous filesystems to make them compatible with interfaces which might expect `AsynchronousFileSystem` instances. Both instance and class-level wrapping behavior is provided and tested. This is a feature that will be of use downstream to enable kerchunk/zarr-python3 compatibility and is related to the efforts [here](https://github.com/fsspec/kerchunk/issues/514)

### DIFF
--- a/fsspec/implementations/asyn_wrapper.py
+++ b/fsspec/implementations/asyn_wrapper.py
@@ -1,0 +1,80 @@
+import asyncio
+import functools
+from fsspec.asyn import AsyncFileSystem
+
+
+def async_wrapper(func, obj=None):
+    """
+    Wraps a synchronous function to make it awaitable.
+
+    Parameters
+    ----------
+    func : callable
+        The synchronous function to wrap.
+    obj : object, optional
+        The instance to bind the function to, if applicable.
+
+    Returns
+    -------
+    coroutine
+        An awaitable version of the function.
+    """
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        self = obj or args[0]
+        return await asyncio.to_thread(func, *args, **kwargs)
+    return wrapper
+
+
+class AsyncFileSystemWrapper(AsyncFileSystem):
+    """
+    A wrapper class to convert a synchronous filesystem into an asynchronous one.
+
+    This class takes an existing synchronous filesystem implementation and wraps all
+    its methods to provide an asynchronous interface.
+
+    Parameters
+    ----------
+    sync_fs : AbstractFileSystem
+        The synchronous filesystem instance to wrap.
+    """
+    def __init__(self, sync_fs, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


martindurant:
cf https://github.com/fsspec/kerchunk/pull/524

moradology:
Hey @martindurant - would you prefer this (or some version thereof) implemented over in kerchunk rather than here?

martindurant:
The only reason to have it in kerchunk, is to indicate a level of experimentation and allow for releasing whenever is convenient. I suppose ice/zarr has no problem depending on it. I don't think it really matters.

moradology:
OK, digging into things a little bit, I think there's a decent argument for pushing this behavior upstream to `fsspec` itself for the sake of not requiring a `kerchunk` dependency in `zarr-python`. In particular, this method will need to have access to the wrapper: https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/storage/remote.py#L159-L168

The alternative might be to provide it as a util in `zarr-python` itself, as `kerchunk` also depends on `zarr-python`.
image

martindurant:
It can stay here, perhaps it'll come in useful for some other uses besides zarr.

fsspec/implementations/asyn_wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+        An awaitable version of the function.
+    """
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        self = obj or args[0]
+        return await asyncio.to_thread(func, *args, **kwargs)
+    return wrapper
+
+
+class AsyncFileSystemWrapper(AsyncFileSystem):
```
martindurant, line 30: This class should appear somewhere in the documentation, even though it's not one we expect users to directly access.

moradology:
I still need to get the docs updated here. Other than that, I think I think this should be good. I've been chasing other test failures for zarr3 integration, but will plan to get docs updated + CI fixed ready for potential merger tomorrow

martindurant:
OK, thank you. Let me know.

moradology:
KK, I think this is ready. Wasn't *entirely* sure how best to handle `fsid`, so that might deserve some critical attention but otherwise things look OK

martindurant:
> how best to handle fsid

I think what you've done is fine.

martindurant:
I did the lint, but please check [this](https://github.com/fsspec/filesystem_spec/pull/1745/commits/34074c17c677d66f49157c097b1c14a7786ecd13#diff-234fdfcf7b8bd75c3ec9902d38b77327ededfb9b0dc01aed46ffc12af6df62ffL25) - is there a chance that `self` gets lost?

moradology:
Just looked things over again. It appears to me that line was kept in error and should be ditched

PR #1755, opened 2024-11-21
Ensure reference FS wraps any sync filesystems

moradology:
This PR guards against the possibility (and likelihood, as `LocalFileSystem` is the default value here) that non-async filesystems are used in `ReferenceFileSystem`. This is important, as `ReferenceFileSystem` is, itself, an `AsyncFileSystem` and it is currently failing to provide functional methods required by that interface (cf. https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/reference.py#L809)

### DIFF
--- a/fsspec/implementations/asyn_wrapper.py
+++ b/fsspec/implementations/asyn_wrapper.py
@@ -46,6 +46,7 @@ class AsyncFileSystemWrapper(AsyncFileSystem):
         super().__init__(*args, **kwargs)
         self.asynchronous = True
         self.fs = sync_fs
+        self.protocol = self.fs.protocol
         self._wrap_all_sync_methods()
 
     @property
--- a/fsspec/implementations/asyn_wrapper.py
+++ b/fsspec/implementations/asyn_wrapper.py
@@ -45,27 +45,27 @@ class AsyncFileSystemWrapper(AsyncFileSystem):
     def __init__(self, sync_fs, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.asynchronous = True
-        self.fs = sync_fs
-        self.protocol = self.fs.protocol
+        self.sync_fs = sync_fs
+        self.protocol = self.sync_fs.protocol
         self._wrap_all_sync_methods()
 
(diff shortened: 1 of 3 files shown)


martindurant:
You are thinking of a referenceFS called from an async context?

moradology:
Test failures here are pretty strange. Don't think these changes should have any effect on `test_async.py` - certainly nothing obvious

martindurant:
Those tests seem to be counting the number of coroutines, but now you are wrapping the coroutines inside coroutines? Maybe there's a more rigorous way to count.

martindurant:
Actually this is the error. The stuff about async is in the cleanup stage for those coroutines (which ought to be awaited in the test even when they fail, but this doesn't cause a failure)
```
m = <fsspec.implementations.memory.MemoryFileSystem object at 0x7f1b6cc6dc10>

    def test_fss_has_defaults(m):
        fs = fsspec.filesystem("reference", fo={})
        assert None in fs.fss
    
        fs = fsspec.filesystem("reference", fo={}, remote_protocol="memory")
>       assert fs.fss[None].protocol == "memory"
E       AssertionError: assert 'abstract' == 'memory'
E         
E         - memory
E         + abstract
```

martindurant:
To fix the async thing (if you wish to try), probably in this block from `_run_coros_in_chunks`, all remaining coroutines should be cancelled and awaited
```
        try:
            return await asyncio.wait_for(coro, timeout=timeout), i
        except Exception as e:
            if not return_exceptions:
                # <- cancel pending coroutines
                raise
            return e, i
```
(or maybe try/except at `await asyncio.wait(..)` a few lines down)

moradology:
The remaining s3fs test failure is pretty strange! Looks to me like conda is installing the wrong version? Came across this: https://github.com/eWaterCycle/ewatercycle/pull/468

Not sure of the best way to test it, unfortunately, as this is pretty specific to CI. Looks like we may be able to simply pin the version over in s3fs (or maybe here?): `pyopenssl>=24.0.0`

PR #1769, opened 2025-01-09
asyn_wrapper: avoid creating async version of open method

skshetry:
The `_open` method is part of an `AbstractFileSystem` interface, and is supposed to be synchronous.

### DIFF
--- a/fsspec/implementations/asyn_wrapper.py
+++ b/fsspec/implementations/asyn_wrapper.py
@@ -57,8 +57,9 @@ class AsyncFileSystemWrapper(AsyncFileSystem):
         """
         Wrap all synchronous methods of the underlying filesystem with asynchronous versions.
         """
+        excluded_methods = {"open"}
         for method_name in dir(self.sync_fs):
-            if method_name.startswith("_"):
+            if method_name.startswith("_") or method_name in excluded_methods:
                 continue
 
             attr = inspect.getattr_static(self.sync_fs, method_name)


fsspec/implementations/asyn_wrapper.py
```diff
@@ -57,8 +57,9 @@ def _wrap_all_sync_methods(self):
         """
         Wrap all synchronous methods of the underlying filesystem with asynchronous versions.
         """
+        excluded_methods = {"open"}
         for method_name in dir(self.sync_fs):
-            if method_name.startswith("_"):
+            if method_name.startswith("_") or method_name in excluded_methods:
                 continue
```
skshetry, line 63: Alternatively, this can be changed to following,

```python
for method in dir(AsyncFileSystem):
  if not method.startswith("_"):
      continue
  if not asyncio.iscoroutinefunction(getattr(AsyncFileSystem, method, None)):
      continue
  smethod = method[1:]
  f = getattr(self.sync_fs, smethod, None)
  if callable(f) and not asyncio.iscoroutinefunction(f):
      mth = async_wrapper(f, obj=self)
      setattr(self, method, mth)
      if not mth.__doc__:
          mth.__doc__ = f.__doc__
```

and maybe moved to `fsspec.asyn`.

martindurant: There is similar code in https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L915 to differentiate methods defined in Async versus AsbtractFilesystem.

martindurant:
To be clear: I am happy with the changes as given, I can merge it. I wonder if there will be other methods that need adding to the list in the future.

skshetry:
> To be clear: I am happy with the changes as given, I can merge it. I wonder if there will be other methods that need adding to the list in the future.

I see that the current implementation async-ifies APIs that are not in `AsyncFileSystem` (eg: `lexists`, `read_bytes`, etc). What's the use-case of `AsyncFileSystemWrapper` - is it to imitate `AsyncFileSystem` or is it to provide async methods for all sync methods of a filesystem (which is the current implementation)?

If the latter, then this might be a good-enough workaround. If the former, I can push the aforementioned patch.

 In any case, I am only interested in the `_open` API at the moment. So, I'll leave that up to you to decide. :)

skshetry:
@martindurant, what do you suggest? Are you okay with the PR or do you want me to try the alternative way as discussed in https://github.com/fsspec/filesystem_spec/pull/1769#discussion_r1908353030?

martindurant:
I think the PR is fine as is - if you say it's ready. The rest was only suggestions of how it could be made more complete maybe, which can be left for the future or not done at all.

skshetry:
Then I think this should be ready. :)

PR #1770, opened 2025-01-10
Add newline inclusive disclaimer

diederikvdv:
Hi, I got confused due to some ambiguity with the readline and readlines function.
I propose adding a disclaimer that the newline character is also included.

Some extra references about similar functions within Python

`splitlines` has the argument `keeplinebreaks` to make this choice for including or excluding newlines explicit. It defaults to removing the newline.
https://www.w3schools.com/python/ref_string_splitlines.asp

The documentation of `readlines` also doesn't state anything about including or excluding the newlines. It defaults to keeping the newline: similar to the fsspec implementation.
https://www.w3schools.com/python/ref_file_readlines.asp

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -2131,7 +2131,7 @@ class AbstractBufferedFile(io.IOBase):
         return b"".join(out)
 
     def readline(self):
-        """Read until first occurrence of newline character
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks!

PR #1778, opened 2025-01-21
referencefs: add streaming _open support

### LINKED ISSUE #1771: ReferenceFileSystem: streaming support for `_open`
Hi, I am using `ReferenceFileSystem` as a sort of virtual filesystem, similar to how you have described it in [stackoverflow - Does fsspec support virtual filesystems such as pyfileysystem](https://stackoverflow.com/a/74395247).

It works great for my use-case, but I have encountered an issue - the `_open` API reads the entire file instead of streaming it.

https://github.com/fsspec/filesystem_spec/blob/30af5e1d1f201d8681faf0ca163c4c9509de69a3/fsspec/implementations/reference.py#L1102-L1104

This behaviour is expected and is documented as such:

https://github.com/fsspec/filesystem_spec/blob/30af5e1d1f201d8681faf0ca163c4c9509de69a3/fsspec/implementations/reference.py#L597-L599

I’m curious if there’s a specific reason `_open` was implemented to load the entire file instead of allowing for streaming access. Could it be that I’m misusing `ReferenceFileSystem`? If not, I’d be happy to work on a PR to implement streaming support. Let me know if this would be useful!

EDIT: I'm basically using it as follows, for pyarrow to preserve partitioning format that it infers from filepath.

```python
path = "s3://bucket/parquets/first_name=Alice/5c6de-0.parquet"
fs = ReferenceFileSystem(fo={path: ["/path/to/a/local/cache"]}
ds = dataset(path, filesystem=fs, **kwargs)
```

skshetry:
Closes #1771.


fsspec/implementations/reference.py
```diff
@@ -1100,8 +1100,28 @@ def _dircache_from_items(self):
...
     def _open(self, path, mode="rb", block_size=None, cache_options=None, **kwargs):
-        data = self.cat_file(path)  # load whole chunk into memory
-        return io.BytesIO(data)
+        part_or_url, start0, end0 = self._cat_common(path)
+        if isinstance(part_or_url, bytes):
+            return io.BytesIO(part_or_url[start0:end0])
+
+        protocol, _ = split_protocol(part_or_url)
+        if start0 is None and end0 is None:
+            return self.fss[protocol]._open(
```
skshetry, line 1109: Ideally, we could put these in `ReferenceFile`, but I did not want to go through an indirection that's not required.

martindurant: This is fine. Maybe add a comment to say what it's doing.

martindurant (approved):
This looks totally good to me.

There is apparently something being picked up by codespell.

The dask installation failure in the `downstream` run would be fixed by removing dask-expr from the environment. Please do fix this.

The error in the s3fs might already be fixed, but is in any case due to a recent release of botocore and nothing to do with this PR.

fsspec/implementations/reference.py
skshetry: Added a small. PTAL.

PR #1779, opened 2025-01-27
spec: fix spelling error in docstring for fs.walk

skshetry:


skshetry:
@martindurant, would you be open to add a pre-commit hook for codespell instead of using github actions?

martindurant:
> would you be open to add a pre-commit hook for codespell instead of using github actions?

Sure. This was probably meant to be there already (and my IDE does it automatically for me anyway, which is why I wouldn't have noticed).

martindurant:
Merging this - remaining CI issues can be fixed elsewhere.

PR #1781, opened 2025-01-27
drop `dask-expr` dependency from downstream tests

skshetry:


martindurant:
This follows the release of dask 2025.1.0, in which dask-expr was merged into dask itself (and dask-expr explicitly depends on only earlier versions of dask).

skshetry:
Got it. This PR should be ready. :-)

Issue #1788, opened 2025-02-10
`LocalFileSystem.ls()` with `details=False` slow


FrankEssenberger:
Hi fsspec developers,

I realized when using a later version of fsspec than the ones from 2023 that the `.ls()` was super slow. The reason came in in this issue PR: #1479 where always the `info` is taken for all files

![Image](https://github.com/user-attachments/assets/dd721727-e71d-468a-aa6d-de4efb886849)

and then thrown away - for a large number of files this can become costly and the `detail` flag can not be used to reduce work. Before the mentioned PR the `os.listdir` was used which just list the number of files.

I just wanted to ask if I can open a PR in which the`ls` used the `os.listdir` again for detail=False and if the path is a file it will just return the path to the file - no much use to call it for a file with not details but should of cours work. And if the `details=True` the code stays as is?

Let me know what you think.

Best
Frank

martindurant:
I would be happy to see such a PR.

PR #1789, opened 2025-02-11
Improve Performance on `LocalFileSystem.ls()` if `detail=False`

FrankEssenberger:
Relates to #1788 

Only get the infos on a file if they are requested by the user - otherwise return just the path as usual.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,25 @@
 Changelog
 =========
 
+2025.2.0
+--------
+
+Enhancements
+
+- add open() to referenceFS (#1778)
+
+Fixes
+
+- don't make async open() in async-wrapper (#1769)
+- fix CI following dask-expr upstream change (#1781)
+- cope with zarr3 "Buffer" objects in referenceFS (#1784)
+
+Other
+
+- use itemgetter in archiveFS (#1764)
+- document that newline is included in readline(s) (#1770)
+- format/spelling (#1774, 1779, 1780)
+
 2024.12.0
 ---------
 
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -452,7 +452,7 @@ def open(
     newline: bytes or None
         Used for line terminator in text mode. If None, uses system default;
         if blank, uses no translation.
-    expand: bool or Nonw
+    expand: bool or None
         Whether to regard file paths containing special glob characters as needing
         expansion (finding the first match) or absolute. Setting False allows using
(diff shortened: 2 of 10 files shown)


martindurant:
The failures are not your fault (also seen in #1790 ), I'll fix early next week.

martindurant:
The windows failure I think is genuine - we are getting windows path separators inside an otherwise posix path in the ls() code path.

martindurant:
I think using pathlib is the wrong choice for two reasons:
- it is considerably slower than string manipulation
- you will get windows paths (in the test), but actually we want the real code to return posix paths always, that's the contract fsspec promises. So instead of fixing the test, we need to fix the code.

Hah, yeah, instead of removing the one errant "\\", it made *all* the separators like that:
```
  -     'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_ls_on_files0\\file_2.json',
  ?        ^^     ^^           ^^       ^^     ^^    ^^                     ^^        ^^                 ^^
  +     'C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_ls_on_files0/file_2.json',
```

FrankEssenberger:
ah I was not aware of the `we want the real code to return posix paths always` paradime - who uses windows anywaty :-). No jokes aside there is the `make_path_posix` this should do the trick if I use this in the test properly.

martindurant:
Can you do a speed test, please, and then I'll merge this. I ask because make_path_posix isn't free either, but should still be faster than info().

FrankEssenberger:
> Can you do a speed test, please, and then I'll merge this. I ask because make_path_posix isn't free either, but should still be faster than info().

sure.  Since the `info` calls the `make_path_posix` as well it must be faster. I would assume the `path.stat()` is the costly thing to do in the `info` and if I add a very deep structure with a lot of data the aggregation in the test it will take time. What do you think would be a fair setup for such a test? Is there some test creating a representative data layout?

martindurant:
I meant a quick and dirty test, perhaps on your home directory - no need to go crazy or to add code.

martindurant:
Running the following locally:
```
In [1]: import fsspec
In [2]: fs = fsspec.filesystem("file")
In [3]: %timeit fs.ls("", detail=False)
```
I get 85us on this branch and 171us on master. So 👍 .

FrankEssenberger:
Thanks for taking care of the review.

PR #1791, opened 2025-02-16
feat(gist): fsspec file system for GitHub gists (resolves #888)

### LINKED ISSUE #888: Add gist FS
We have a github implementation, which I find surprisingly convenient. We should also do one for gists, which also have a reasonable API https://docs.github.com/en/rest/reference/gists . We could mirror the github implementation and require user/gist_id, or only require user and auto list their gists (which have hex-like IDs, not sensible names). I tend towards the latter.
Within a gist you only have a single flat list of files.

`"https://api.github.com/users/martindurant/gists"` lists my gists, AND the files within each of them, giving size and URL to the raw download. The call can be optionally paginated.

lmmx:
This PR introduces a new filesystem backend, `GistFileSystem`, which allows read-only access to files within a single GitHub Gist (as suggested in #888). I'd find this really useful in combination with _Universal Pathlib_ (also an fsspec project)!

- Gists are essentially flat collections of files, so there is no subdirectory concept. (Technically they are git repos that can store dirs too but we only need to support them as flat file lists, that's all the website UI will show them as)
- The implementation is closely based on `GithubFileSystem` but simplified for a single gist.
- Supports both public and private gists, latter needed user/token (PAT).

Users can do:

```python
import fsspec

# For a public gist
fs = fsspec.filesystem("gist", gist_id="729837f14264089288178a5f632221ab")
print(fs.ls(""))  # lists files
with fs.open("test1.txt", "rb") as f:
    print(f.read().decode())
```

For a private gist, the same but also passing `username` and `token` args.

- [x] Implemented FS (methods: `ls`, `_open`, `cat`, `invalidate_cache`), read-only impl
- [x] Added to registry (alphabetically)
- [x] Added a test - I changed the gist ID to [one by martindurant](https://gist.github.com/martindurant/16bee4256595d3b6814be139ab1bd54e) so you wouldn't have to worry about relying on someone else preserving their artifact for your tests to pass, also it's fairly small so shouldn't be slow to load
- [x] Added **documentation** in `docs/source/api.rst`.
- [x] Verified that read-only operations (ls, cat, open) are working with public gists.

## Example usage

Below is a short snippet showing how to retrieve files from a public gist:

```python
import fsspec

gist_id = "16bee4256595d3b6814be139ab1bd54e"
print("Gist ID:", gist_id)
fs = fsspec.filesystem("gist", gist_id=gist_id)
file_list = fs.ls("")
print("Files in the Gist (via fsspec):", file_list)
contents = fs.cat(["gistfile1.txt"])
print(contents["gistfile1.txt"].decode()[:120] + "\n...")
```

⇣

```
Gist ID: 16bee4256595d3b6814be139ab1bd54e
Files in the Gist (via fsspec): ['gistfile1.txt']
import astropy.io.fits._tiled_compression as tiled
from astropy.io import fits
import numcodecs
import fsspec
import zar
...
```

### DIFF
--- a/fsspec/implementations/gist.py
+++ b/fsspec/implementations/gist.py
@@ -1,0 +1,232 @@
+import requests
+
+from ..spec import AbstractFileSystem
+from ..utils import infer_storage_options
+from .memory import MemoryFile
+
+
+class GistFileSystem(AbstractFileSystem):
+    """
+    Interface to files in a single GitHub Gist.
+
+    Provides read-only access to a gist's files. Gists do not contain
+    subdirectories, so file listing is straightforward.
+
+    Parameters
+    ----------
+    gist_id : str
+        The ID of the gist you want to access (the long hex value from the URL).
+    filenames : list[str] (optional)
+        If provided, only make a file system representing these files, and do not fetch
+        the list of all files for this gist.
+    sha : str (optional)
+        If provided, fetch a particular revision of the gist. If omitted,
+        the latest revision is used.
+    username : str (optional)
+        GitHub username for authentication (required if token is given).
+    token : str (optional)
+        GitHub personal access token (required if username is given).
+    timeout : (float, float) or float, optional
+        Connect and read timeouts for requests (default 60s each).
+    kwargs : dict
+        Stored on `self.request_kw` and passed to `requests.get` when fetching Gist
+        metadata or reading ("opening") a file.
+    """
+
+    protocol = "gist"
+    gist_url = "https://api.github.com/gists/{gist_id}"
+    gist_rev_url = "https://api.github.com/gists/{gist_id}/{sha}"
+
+    def __init__(
+        self,
+        gist_id,
+        filenames=None,
+        sha=None,
+        username=None,
+        token=None,
+        timeout=None,
+        **kwargs,
+    ):
+        super().__init__()
+        self.gist_id = gist_id
+        self.filenames = filenames
+        self.sha = sha  # revision of the gist (optional)
+        if (username is None) ^ (token is None):
+            # Both or neither must be set
+            if username or token:
+                raise ValueError("Auth requires both username and token, or neither.")
+        self.username = username
+        self.token = token
+        self.request_kw = kwargs
+        # Default timeouts to 60s connect/read if none provided
+        self.timeout = timeout if timeout is not None else (60, 60)
+
+        # We use a single-level "directory" cache, because a gist is essentially flat
+        self.dircache[""] = self._fetch_file_list()
+
+    @property
+    def kw(self):
+        """Auth parameters passed to 'requests' if we have username/token."""
+        if self.username is not None and self.token is not None:
+            return {"auth": (self.username, self.token), **self.request_kw}
+        return self.request_kw
+
+    def _fetch_gist_metadata(self):
+        """
+        Fetch the JSON metadata for this gist (possibly for a specific revision).
+        """
+        if self.sha:
+            url = self.gist_rev_url.format(gist_id=self.gist_id, sha=self.sha)
+        else:
+            url = self.gist_url.format(gist_id=self.gist_id)
+
+        r = requests.get(url, timeout=self.timeout, **self.kw)
+        if r.status_code == 404:
+            raise FileNotFoundError(
+                f"Gist not found: {self.gist_id}@{self.sha or 'latest'}"
+            )
+        r.raise_for_status()
+        return r.json()
+
+    def _fetch_file_list(self):
+        """
+        Returns a list of dicts describing each file in the gist. These get stored
+        in self.dircache[""].
+        """
+        meta = self._fetch_gist_metadata()
+        if self.filenames:
+            available_files = meta.get("files", {})
+            files = {}
+            for fn in self.filenames:
+                if fn not in available_files:
+                    raise FileNotFoundError(fn)
+                files[fn] = available_files[fn]
+        else:
+            files = meta.get("files", {})
+
+        out = []
+        for fname, finfo in files.items():
+            if finfo is None:
+                # Occasionally GitHub returns a file entry with null if it was deleted
+                continue
+            # Build a directory entry
+            out.append(
+                {
+                    "name": fname,  # file's name
+                    "type": "file",  # gists have no subdirectories
+                    "size": finfo.get("size", 0),  # file size in bytes
+                    "raw_url": finfo.get("raw_url"),
+                }
+            )
+        return out
+
+    @classmethod
+    def _strip_protocol(cls, path):
+        """
+        Remove 'gist://' from the path, if present.
+        """
+        # The default infer_storage_options can handle gist://username:token@id/file
+        # or gist://id/file, but let's ensure we handle a normal usage too.
+        # We'll just strip the protocol prefix if it exists.
+        path = infer_storage_options(path).get("path", path)
+        return path.lstrip("/")
+
+    @staticmethod
+    def _get_kwargs_from_urls(path):
+        """
+        Parse 'gist://' style URLs into GistFileSystem constructor kwargs.
+        For example:
+          gist://:TOKEN@<gist_id>/file.txt
+          gist://username:TOKEN@<gist_id>/file.txt
+        """
+        so = infer_storage_options(path)
+        out = {}
+        if "username" in so and so["username"]:
+            out["username"] = so["username"]
+        if "password" in so and so["password"]:
+            out["token"] = so["password"]
+        if "host" in so and so["host"]:
+            # We interpret 'host' as the gist ID
+            out["gist_id"] = so["host"]
+
+        # Extract SHA and filename from path
+        if "path" in so and so["path"]:
+            path_parts = so["path"].rsplit("/", 2)[-2:]
+            if len(path_parts) == 2:
+                if path_parts[0]:  # SHA present
+                    out["sha"] = path_parts[0]
+                if path_parts[1]:  # filename also present
+                    out["filenames"] = [path_parts[1]]
+
+        return out
+
+    def ls(self, path="", detail=False, **kwargs):
+        """
+        List files in the gist. Gists are single-level, so any 'path' is basically
+        the filename, or empty for all files.
+
+        Parameters
+        ----------
+        path : str, optional
+            The filename to list. If empty, returns all files in the gist.
+        detail : bool, default False
+            If True, return a list of dicts; if False, return a list of filenames.
+        """
+        path = self._strip_protocol(path or "")
+        # If path is empty, return all
+        if path == "":
+            results = self.dircache[""]
+        else:
+            # We want just the single file with this name
+            all_files = self.dircache[""]
+            results = [f for f in all_files if f["name"] == path]
+            if not results:
+                raise FileNotFoundError(path)
+        if detail:
+            return results
+        else:
+            return sorted(f["name"] for f in results)
+
+    def _open(self, path, mode="rb", block_size=None, **kwargs):
+        """
+        Read a single file from the gist.
+        """
+        if mode != "rb":
+            raise NotImplementedError("GitHub Gist FS is read-only (no write).")
+
+        path = self._strip_protocol(path)
+        # Find the file entry in our dircache
+        matches = [f for f in self.dircache[""] if f["name"] == path]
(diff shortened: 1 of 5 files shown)


martindurant:
Thanks for providing! I haven't had a chance to look yet, but I will soon :)

lmmx:
Most welcome, no worries! 😃

martindurant:
Quick suggestion: it would be good to enable bundling the gist ID with the URL:
```
with fsspec.open("gist://16bee4256595d3b6814be139ab1bd54e@/test1.txt", "rb")
```
like github: allows. It would require enabling extracting kwargs from the URL.

fsspec/implementations/gist.py
```diff
@@ -0,0 +1,211 @@
...
+        if not raw_url:
+            raise FileNotFoundError(f"No raw_url for file: {path}")
+
+        r = requests.get(raw_url, timeout=self.timeout, **self.kw)
+        if r.status_code == 404:
+            raise FileNotFoundError(path)
+        r.raise_for_status()
+        return MemoryFile(path, None, r.content)
+
+    def cat(self, path, recursive=False, on_error="raise", **kwargs):
```
martindurant, line 193: I'm surprised this doesn't work already with the parent version in AbstractFileSystem

fsspec/implementations/gist.py
```diff
@@ -0,0 +1,211 @@
...
+    gist_id : str
+        The ID of the gist you want to access (the long hex value from the URL).
+    sha : str (optional)
+        If provided, fetch a particular revision of the gist. If omitted,
+        the latest revision is used.
+    username : str (optional)
+        GitHub username for authentication (required if token is given).
+    token : str (optional)
+        GitHub personal access token (required if username is given).
+    timeout : (float, float) or float, optional
```
martindurant, line 28: Perhaps more general set of arguments to pass to requests?

fsspec/implementations/gist.py
```diff
@@ -0,0 +1,211 @@
...
+            # Both or neither must be set
+            if username or token:
+                raise ValueError("Auth requires both username and token, or neither.")
+        self.username = username
+        self.token = token
+        # Default timeouts to 60s connect/read if none provided
+        self.timeout = timeout if timeout is not None else (60, 60)
+
+        # We use a single-level "directory" cache, because a gist is essentially flat
+        self.dircache[""] = self._fetch_file_list()
```
martindurant, line 54: It is unusual to fetch file listings on init, but I think it makes sense in this case.

fsspec/implementations/gist.py
```diff
@@ -0,0 +1,211 @@
...
+            return results
+        else:
+            return sorted(f["name"] for f in results)
+
+    def invalidate_cache(self, path=None):
+        """
+        Clear the dircache. If path is given, we could refetch—but for gist,
+        we typically refetch everything in one shot anyway.
+        """
+        self.dircache.clear()
```
martindurant, line 165: Interesting way to "refresh". There could be a refresh= kwarg on ls(); but actually I'd be happy to treat the dircache as immutable for the purposes of this implementation.

martindurant (commented):
As I read through, I found that you answered almost all of my questions elsewhere in the code :)

I don't think there's a good way to test this thoroughly, but at least we can reasonably expect gist to be available whenever GHA is running.

martindurant:
Please ping me when I should have another look

lmmx:
Thanks for reviewing Martin, gotten sidetracked in a CI fixing rabbit hole this week I've thankfully emerged and can return to revisit this!

> Please ping me when I should have another look

Will do :saluting_face:

martindurant:
I just noticed this is still stalled. Please ask for help if you need it.

lmmx:
Oh snap I’m sorry, yeah let me take a look…

fsspec/implementations/tests/test_gist.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+import fsspec
+
+
+@pytest.mark.parametrize("gist_id", ["16bee4256595d3b6814be139ab1bd54e"])
```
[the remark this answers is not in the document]
lmmx: Would require the FS impl to support a hash param (will check)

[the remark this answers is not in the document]
lmmx: Done!

fsspec/implementations/gist.py
lmmx: Note:

https://github.com/fsspec/filesystem_spec/blob/a513182cefee141f25f50154fc20291017381037/fsspec/implementations/github.py#L9

martindurant: Yes, you are welcome to remove that line!

fsspec/implementations/gist.py
```diff
@@ -0,0 +1,211 @@
...
+        path = infer_storage_options(path).get("path", path)
+        return path.lstrip("/")
+
+    @staticmethod
+    def _get_kwargs_from_urls(path):
+        """
+        Parse 'gist://' style URLs into GistFileSystem constructor kwargs.
+        For example:
+          gist://:TOKEN@<gist_id>/file.txt
+          gist://username:TOKEN@<gist_id>/file.txt
```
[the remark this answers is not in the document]
lmmx: Great question!

For reference:

```py
>>> infer_storage_options("gist://username:TOKEN@<gist_id>/file.txt")
{'protocol': 'gist', 'path': '/file.txt', 'host': '<gist_id>', 'username': 'username', 'password': 'TOKEN'}
>>> infer_storage_options("gist://:TOKEN@<gist_id>/file.txt")
{'protocol': 'gist', 'path': '/file.txt', 'host': '<gist_id>', 'password': 'TOKEN'}
```

The regular website HTTP gist URL supports SHA by an optional extra path, which makes sense (like how `example.com/bar` ⇒ `example.com/bar/index.html`).

So we could just copy the same scheme. In the test I picked this one

- https://gist.github.com/martindurant/16bee4256595d3b6814be139ab1bd54e

If I click `Raw` I get the file URL with `/raw/$SHA/$filename` appended

- https://gist.githubusercontent.com/martindurant/16bee4256595d3b6814be139ab1bd54e/raw/760905f9f222ad41b9c3fd8308cbbd016943c65a/gistfile1.txt

If we search [grep.app](https://grep.app/search?regexp=true&q=gist%3A%2F%2F.%2B%5C%2F.%2B%5C%2F) for anyone else using a `gist://` URL scheme / protocol identifier we can see at least 1 such example ([here](https://github.com/luvit/lit/blob/42333e558e1e8a73a174e83d05d52044423034df/libs/core.lua#L519))

```lua
    "^gist://([^/]+)/(.+)/?$", "https://gist.github.com/%1/%2.git",
```

I read that as meaning they are writing it as `gist://$gist_id/$sha` (with an optional terminating `/`).

You **can't** put it with another `@`

```py
>>> infer_storage_options("gist://TOKEN@<gist_id>file.txt@mysha")
{'protocol': 'gist', 'path': '', 'host': 'mysha', 'username': 'TOKEN@<gist_id>file.txt'}
```

You could put it in a URL query:

```py
>>> infer_storage_options("gist://TOKEN@<gist_id>file.txt?mysha")
{'protocol': 'gist', 'path': '', 'host': '<gist_id>file.txt', 'username': 'TOKEN', 'url_query': 'mysha'}
```

or with a URL parameter that you'd need to parse

```py
>>> infer_storage_options("gist://TOKEN@<gist_id>file.txt?rev=mysha")
{'protocol': 'gist', 'path': '', 'host': '<gist_id>file.txt', 'username': 'TOKEN', 'url_query': 'rev=mysha'}
```

fsspec/implementations/gist.py
lmmx: Yeah I think it's required when you want all the files, but having thought about it a bit there'd also be cases where you want one specific file and would thus be able to forego the file listing, I suspect?

For context I was trying to pull in all files from a gist, but there's no reason why you couldn't populate a single file "file system" of a single file (for a gist it isn't a bad idea, they often are a single file).

fsspec/implementations/gist.py
lmmx: Hmmm anything in mind? You mean like headers and things? :thinking:

martindurant: OK, I suppose general kwargs is fine, if so documented.

fsspec/implementations/gist.py
lmmx: Ah sorry, I think I wrote this to fit in with the other impls - I wasn't using it myself. I'll cut this method.

fsspec/implementations/gist.py
lmmx: Oh, on review the `**kwargs` go to the base `AbstractFileSystem` class which calls them `**storage_options` in its init signature, and pops a few but then passes them to `DirCache` which does nothing with them. So this is not in fact handled by the gist file system ! Will fix.

lmmx:
```diff
M: fsspec/implementations/gist.py
@@ -26,7 +26,8 @@ class GistFileSystem(AbstractFileSystem):
     timeout : (float, float) or float, optional
         Connect and read timeouts for requests (default 60s each).
     kwargs : dict
-        Passed to AbstractFileSystem base class.
+        Stored on `self.request_kw` and passed to `requests.get` when fetching Gist
+        metadata or reading ("opening") a file.
     """
 
     protocol = "gist"
@@ -36,7 +37,7 @@ class GistFileSystem(AbstractFileSystem):
     def __init__(
         self, gist_id, sha=None, username=None, token=None, timeout=None, **kwargs
     ):
-        super().__init__(**kwargs)
+        super().__init__()
         self.gist_id = gist_id
         self.sha = sha  # revision of the gist (optional)
         if (username is None) ^ (token is None):
@@ -45,6 +46,7 @@ class GistFileSystem(AbstractFileSystem):
                 raise ValueError("Auth requires both username and token, or neither.")
         self.username = username
         self.token = token
+        self.request_kw = kwargs
         # Default timeouts to 60s connect/read if none provided
         self.timeout = timeout if timeout is not None else (60, 60)
 
@@ -55,8 +57,8 @@ class GistFileSystem(AbstractFileSystem):
     def kw(self):
         """Auth parameters passed to 'requests' if we have username/token."""
         if self.username is not None and self.token is not None:
-            return {"auth": (self.username, self.token)}
-        return {}
+            return {"auth": (self.username, self.token), **self.request_kw}
+        return self.request_kw
 ```

File: fsspec/implementations/gist.py (hunk 3 above)
Comment (lmmx) [reply fsspec/implementations/gist.py:52]: Hm no turns out there is no single file API endpoint so it would have to do the same metadata request for all files it currently does, or else it wouldn't be able to have the size of the file in question (and since the dircache is immutable now, we wouldn't be able to get it).

File: fsspec/implementations/gist.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (lmmx) [reply fsspec/implementations/gist.py:140]: Implemented and tested !

```py
    filenames : list[str] (optional)
        If provided, only make a file system representing these files, and do not fetch
        the list of all files for this gist.
```

It's just a filter on the file system but that's still useful when you expect the file to exist.

The file list fetch now verifies an expected filenames list if one was passed

```py
    def _fetch_file_list(self):
        """
        Returns a list of dicts describing each file in the gist. These get stored
        in self.dircache[""].
        """
        meta = self._fetch_gist_metadata()
        if self.filenames:
            available_files = meta.get("files", {})
            files = {}
            for fn in self.filenames:
                if fn not in available_files:
                    raise FileNotFoundError(fn)
                files[fn] = available_files[fn]
        else:
            files = meta.get("files", {})
```

fsspec/implementations/gist.py
lmmx: I implemented it regardless, it's still useful to know you'll get a `FileNotFoundError` if the file isn't in the gist

lmmx:
Cool, all done. A "round trip" might be nice too

lmmx:
Checks passed, 3.10 failed with an intermitten HTTP error from conda repodata, LGTM

martindurant:
I like it! Let's put it in, and see if the public has feedback once using it.

QuLogic:
These tests need a vcr or a network marker.

martindurant:
A network marker is fair. I don't think it's worth setting up vcr for this.

PR #1798, opened 2025-02-27
dirfs: Ensures consistency of details returned by the “ls” and “info” methods.

lmignon:
Before this changes, the name field into the information returned by the 'info' method was an absolute path while the name for the same entry returned by the 'ls' method was a relative path. As a result, the name in the info of the 'info' method was unusable for api calls.


martindurant:
Thanks for fixing this oversight

lmignon:
@martindurant Thank you for the quick merge.  Waiting now for a new release ;-)

martindurant:
> Waiting now for a new release

I agree we are due.

PR #1799, opened 2025-03-05
Add _pipe_file and test to http

kephale:
This adds a `_pipe_file` implementation to `HTTPFileSystem`. This was necessary for write support of http filesystems for writing zarr image data in https://github.com/copick/copick-server.

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -543,7 +543,7 @@ class HTTPFileSystem(AsyncFileSystem):
 
         session = await self.get_session()
 
-        async with self.session.put(url, data=value, headers=headers, **kwargs) as r:
+        async with session.put(url, data=value, headers=headers, **kwargs) as r:
             r.raise_for_status()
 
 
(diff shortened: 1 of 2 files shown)


fsspec/implementations/http.py
```diff
@@ -522,6 +522,34 @@ async def _isdir(self, path):
...
+            Additional parameters to pass to the HTTP request
+        """
+        url = self._strip_protocol(path)
+        headers = kwargs.pop("headers", {})
+        headers["Content-Length"] = str(len(value))
+
+        if not hasattr(self, "session"):
+            import aiohttp
+
+            self.session = aiohttp.ClientSession()
```
martindurant, line 547:
```suggestion
        session = await self.get_session()
```

fsspec/implementations/http.py
```diff
@@ -522,6 +522,34 @@ async def _isdir(self, path):
...
+
+        async with self.session.put(url, data=value, headers=headers, **kwargs) as r:
+            r.raise_for_status()
+        return True
```
martindurant, line 551: Is `True` copied from somewhere?

martindurant (commented):
Thanks, looks like a good addition.

fsspec/implementations/http.py
kephale: Upps, fixed! I was debugging a buggy server at the time, which is where that came from 😅

PR #1801, opened 2025-03-07
Fix typo

kephale:
@martindurant eek, we had a typo in https://github.com/fsspec/filesystem_spec/pull/1799!


martindurant:
Also uncovered a bug in the test server implementation :)

PR #1802, opened 2025-03-07
fix(url_to_fs): mangle arg to avoid collision with kwargs

shcheklein:
Fixes https://github.com/iterative/dvc/issues/10700

[dvc_fs](https://dvc.org/doc/api-reference/dvcfilesystem) is using url as an fs config option. So, it now collides in the url_to_fs with the main argument.

We can do urlpath for consistency, like we do in get_fs_token_paths, but _url feels a bit safer for the future? Let me know if you have any preferences folks.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -366,13 +366,13 @@ def _un_chain(path, kwargs):
     return out
 
 
-def url_to_fs(url, **kwargs):
+def url_to_fs(_url, **kwargs):
     """
     Turn fully-qualified and potentially chained URL into filesystem instance
 
     Parameters
     ----------
-    url : str
+    _url : str
         The fsspec-compatible URL
     **kwargs: dict
         Extra options that make sense to a particular storage connection, e.g.


martindurant:
Having the first and required parameters start with `_` seems wrong to me. If we must change it, it should be `urlpath`. Of course, the function name implies it _should_ be just `url` - is it certain this can't be fixed in dvc instead?

shcheklein:
> Having the first and required parameters start with _ seems wrong to me

yes, agreed, also have mixed feeling to be honest about this ... and url_to_fs is user facing

>  is it certain this can't be fixed in dvc instead?

now that I think about this - yes, I think we can make it happen w/o breaking the compatibility (support both params for a while) ... @skshetry anything I'm missing specifically on this?

Let me close this, and research on the DVC side a bit more.

skshetry:
One alternative here could be to make `url` a positional argument. That, however, could be considered a breaking change.

```python
def url_to_fs(url, /, **kwargs):
   print(url, kwargs)

url_to_fs("url", url="myurl")
# url {"url": "myurl"}
```

`url` is a common argument name. I see it is used in one of the filesystem within fsspec.

https://github.com/fsspec/filesystem_spec/blob/66340a15ef6a11aa89b4f23a4370c53316d85362/fsspec/implementations/jupyter.py#L15

PR #1804, opened 2025-03-11
Do not force err

etobella:
Otherwise, a KeyError is raised if err has not been defined

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -248,7 +248,7 @@ def get_filesystem_class(protocol):
         try:
             register_implementation(protocol, _import_class(bit["class"]))
         except ImportError as e:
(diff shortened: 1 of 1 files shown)


etobella:
There is an example of misconfiguration that is rising this problem:

https://github.com/fsspec/filesystem_spec/blob/master/fsspec/registry.py#L75-L77

luisDIXMIT (approved):
LGTM!

martindurant:
This is fine; but the case of no error message means that the class in question should not be unimportable - because they are defined in this repo and have no further requirements. The specific import for async-wrapper is fixed in https://github.com/fsspec/filesystem_spec/pull/1806

PR #1807, opened 2025-03-14
tar index as json file

pgierz:
Local PR to check if tests run sensibly.

### DIFF
--- a/fsspec/implementations/tar.py
+++ b/fsspec/implementations/tar.py
@@ -1,4 +1,6 @@
+import json
 import logging
+import pathlib
 import tarfile
 
 import fsspec
@@ -89,16 +91,23 @@ class TarFileSystem(AbstractArchiveFileSystem):
         self._index()
 
     def _index(self):
-        # TODO: load and set saved index, if exists
-        out = {}
-        for ti in self.tar:
-            info = ti.get_info()
-            info["type"] = typemap.get(info["type"], "file")
-            name = ti.get_info()["name"].rstrip("/")
-            out[name] = (info, ti.offset_data)
+        if self.index_store is not None and pathlib(self.index_store).exists():
+            # NOTE(PG): Not sure if JSON is the best way to go here, but it's
+            #           simple and human-readable.
+            with self.index_store.open("r") as f:
+                self.index = json.load(f)
+        else:
+            out = {}
+            for ti in self.tar:
+                info = ti.get_info()
+                info["type"] = typemap.get(info["type"], "file")
+                name = ti.get_info()["name"].rstrip("/")
+                out[name] = (info, ti.offset_data)
 
-        self.index = out
-        # TODO: save index to self.index_store here, if set
+            self.index = out
+            if self.index_store is not None:
+                with self.index_store.open("w") as f:
+                    json.dump(out, f)
 
     def _get_dirs(self):
         if self.dir_cache is not None:


pgierz:
Still a draft, I need to tweak a few things, I'll mark it as ready soon.

pgierz:
It might be sensible to also check against a "Truthy" value here and default to a filename that is similar to the name of the actual tar file. At the moment, I assumed that the user was giving a file-like string.

pgierz:
Alright, this seems to work the way I would imagine it to. I'd be very happy to get some feedback if this is useful or not 🙂

martindurant:
Thanks for having a go at this! Please ping me next week for a thorough review.

Before proceeding, I wonder if there are any existing TAR index formats and/or index file naming conventions that we should attempt to follow. I agree that JSON is a sensible alternative if not.

You might also be interested in https://github.com/pauldmccarthy/indexed_gzip , which allows for starting to read at checkpoints throughout the GZIP stream rather than always reading from the beginning as normal. If the two ideas were combined, you could random-access to the beginning of any file in the archive.

pgierz:
> Before proceeding, I wonder if there are any existing TAR index formats and/or index file naming conventions that we should attempt to follow. I agree that JSON is a sensible alternative if not.

I wasn't able to find anything, and perhaps the most interesting answer was [here](https://superuser.com/questions/655739/extract-single-file-from-huge-tgz-file/655802#655802), which "long story short" seems to indicate that indexed tar with standard GNU tools is not supported, and therefore, there is also no standard for naming such an index.

> You might also be interested in https://github.com/pauldmccarthy/indexed_gzip , which allows for starting to read at checkpoints throughout the GZIP stream rather than always reading from the beginning as normal. If the two ideas were combined, you could random-access to the beginning of any file in the archive.

Perhaps also this? https://github.com/colon3ltocard/pyindexedtar

I would have a try at adding some of the ideas there, and let you know later in the week.

fsspec/implementations/tar.py
```diff
@@ -84,21 +86,39 @@ def __init__(
         self.tar = tarfile.TarFile(fileobj=self.fo)
         self.dir_cache = None
 
-        self.index_store = index_store
+        if isinstance(index_store, (str, pathlib.Path)):
```
martindurant, line 89: I think it's reasonable to only allow str here. That would allow for the index to be stored _anywhere_ (in another storage backend), whereas Path is local specific. If you use fsspec.open(), then you could include anything that it can handle.

fsspec/implementations/tar.py
```diff
@@ -84,21 +86,39 @@ def __init__(
...
-            name = ti.get_info()["name"].rstrip("/")
-            out[name] = (info, ti.offset_data)
-
-        self.index = out
-        # TODO: save index to self.index_store here, if set
+        if self.index_store is not None and self.index_store.exists():
+            # NOTE(PG): Not sure if JSON is the best way to go here, but it's
+            #           simple and human-readable.
+            logger.debug(f"Reloading from {self.index_store}")
+            with self.index_store.open("r") as f:
```
martindurant, line 104: `fsspec.open(self.index_store)`

fsspec/implementations/tar.py
```diff
@@ -84,21 +86,39 @@ def __init__(
...
+            out = {}
+            for ti in self.tar:
+                info = ti.get_info()
+                info["type"] = typemap.get(info["type"], "file")
+                info["name"] = name = info["name"].rstrip("/")
+                out[name] = (info, ti.offset_data)
+
+            self.index = out
+            if self.index_store is not None:
+                with self.index_store.open("w") as f:
```
martindurant, line 117: `fsspec.open(self.index_store, "wt")`

martindurant (commented):
I think the implementation is fine aside a couple of comments.

Would appreciate a test (that the index file gets created and looks correct) and text to describe what happens.

Extension to tar.gz or other types can follow later, if there is interest.

fsspec/implementations/tar.py
pgierz: I like that idea. Does `fsspec` have some way to check if a file exists, akin to `pathlib.Path(...).exists()`? I could do try/except and catch for a `FileNotFoundError`, unless there is already a way to check that.

martindurant: It would be two-step, create a filesystem instance (fsspec.url_to_fs), and use .exists on that - so probably fine to use try/except here. We have discussed elsewhere whether we should have top-level operations like exists(url) that dispatch to the implementations like fsspec.open, but nothing has been done yet (cf fsspec.generic)

vadimkantorov:
Regarding JSON as a tar index, might be good to also support .json.gz for smaller index size (for tars containing a lot of long file paths and dozens of thousands of small files), or maybe also some other existing binary formats like parquet... Also a question is that it might be good for this index to be sorted by file paths (to enable fast search by file path), but it's also important sometimes to preserve the index of the record in the initial tar (maybe the offset can be used for this purpose as well)...

Issue #1438, opened 2023-11-24
Reading Files from Git LFS Repo


johko:
Hey,

I'm trying to read files from a GitHub LFS Repo (https://github.com/openai/dalle3-eval-samples/tree/main) but only get the pointers to the actual large files (the images in the repo), instead of the binaries.

Is there any way of reading these files from an LFS repo with fsspec?

My current testing code is:

```python
import fsspec

github_repo = fsspec.get_mapper("github://openai:dalle3-eval-samples@main")
for file_name in github_repo:
    file = github_repo[file_name]

```

martindurant:
We don't have such an integration. I don't know how LFS works in detail, but I imagine it's not too complex, if you would like to implement it. I know that some git-based data services which integrate already with fsspec (dvc, lakefs, xet, maybe others).

johko:
Thanks for the really quick response @martindurant .

I have to admit I don't know too much of the inner workings of LFS myself. But sounds like a fun project to investigate, so if I find the time I'll implement it :slightly_smiling_face:

martindurant:
It doesn't look terrible:
```python
>>>  print(github_repo["t2i_compbench/sdxl/complex_val/The%20black%20camera%20was%20next%20to%20the%20white%20tripod._000160.png"].decode())
version https://git-lfs.github.com/spec/v1
oid sha256:ac618aaf4f05d1a938323f4d37d78877d03b5afb2d4f04af183f298d60e33b55
size 1133715
```

thomasgilgenast:
I have recently encountered a situation where this feature would be useful to me.

I have opened #1810 as one possible proposal for how to add Git LFS support to the github implementation and would welcome feedback on it @johko @martindurant

PR #1810, opened 2025-03-14
rework github _open() implementation to support LFS

thomasgilgenast:
#1438 raised the question of whether it should be possible to read Git LFS files from GitHub via the github implementation.

This PR is an unsolicited proposal of one possible path forward to supporting Git LFS files hosted on GitHub via the github implementation. It reworks the way the github implementation accesses content from github repos, enabling Git LFS file support in the process.

Before this PR, the content of github files was always fetched using the `rurl` (the https://raw.githubusercontent.com/ URL). This URL works well for fetching normal files tracked by Git, but does not work for Git LFS files.

In this PR, the `_open()` function in the github implementation is reworked to instead use the GitHub API to fetch the content. This API endpoint is documented in the GitHub API docs [here](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content).

The way this API works is unfortunately more complicated than the `rurl`. For files <1MB in size, the file's contents will be base64 encoded and included in the JSON response under the `content` key. For files >1MB in size, the `content` key will have an empty string value, and the only way to get its content is to make an additional GET request to a specific URL provided under the `download_url` key of the JSON response.

The downsides of the new reworked implementation of `_open()` are
1. More logic is required - sometimes the content can be found in `content` and other times we need to use `download_url`
2. Two requests are needed for files >1MB in size (one to the content API and another to hit `download_url`)

The upside is that the `download_url` points to the actual file contents for files tracked by Git LFS. This means that the same github implementation will work for normal files (both smaller and larger than 1MB) and also for Git LFS files.

I would be happy to take feedback on whether or not these downsides are worth the improved support for LFS, or if this should be moved to a new, separate "github_lfs" implementation to avoid complicating the current github implementation.

In addition to reworking the `_open()` function, I also removed the implementation of `cat()`, which was also referencing `rurl`. I am not super familiar with why `cat()` was overridden in the github implementation. I believe that if we still want to use the http implementation of `cat()` here, then that should still be possible. Let me know if this would be required to move forward with this PR.

Finally, I've added a few basic tests to cover the added/modified functionality. These tests interact with the real GitHub API, so it may not be desirable to leave them in the test suite as-is - let me know what would be preferred here and I would be happy to adjust.

### DIFF
--- a/fsspec/implementations/github.py
+++ b/fsspec/implementations/github.py
@@ -1,6 +1,6 @@
-import requests
+import base64
 
-import fsspec
+import requests
 
 from ..spec import AbstractFileSystem
 from ..utils import infer_storage_options
@@ -36,7 +36,7 @@ class GithubFileSystem(AbstractFileSystem):
     """
 
     url = "https://api.github.com/repos/{org}/{repo}/git/trees/{sha}"
-    rurl = "https://raw.githubusercontent.com/{org}/{repo}/{sha}/{path}"
+    content_url = "https://api.github.com/repos/{org}/{repo}/contents/{path}?ref={sha}"
     protocol = "github"
     timeout = (60, 60)  # connect, read timeouts
 
@@ -219,21 +219,35 @@ class GithubFileSystem(AbstractFileSystem):
     ):
         if mode != "rb":
             raise NotImplementedError
-        url = self.rurl.format(
+
+        # construct a url to hit the GitHub API's repo contents API
+        url = self.content_url.format(
             org=self.org, repo=self.repo, path=path, sha=sha or self.root
         )
+
+        # make a request to this API, and parse the response as JSON
         r = requests.get(url, timeout=self.timeout, **self.kw)
         if r.status_code == 404:
             raise FileNotFoundError(path)
         r.raise_for_status()
-        return MemoryFile(None, None, r.content)
+        content_json = r.json()
 
-    def cat(self, path, recursive=False, on_error="raise", **kwargs):
-        paths = self.expand_path(path, recursive=recursive)
-        urls = [
-            self.rurl.format(org=self.org, repo=self.repo, path=u, sha=self.root)
-            for u, sh in paths
-        ]
-        fs = fsspec.filesystem("http")
-        data = fs.cat(urls, on_error="return")
-        return {u: v for ((k, v), u) in zip(data.items(), urls)}
+        # if the response's content key is not empty, try to parse it as base64
+        if content_json["content"]:
+            content = base64.b64decode(content_json["content"])
+
+            # as long as the content does not start with the string
+            # "version https://git-lfs.github.com/"
+            # then it is probably not a git-lfs pointer and we can just return
+            # the content directly
+            if not content.startswith(b"version https://git-lfs.github.com/"):
+                return MemoryFile(None, None, content)
+
+        # we land here if the content was not present in the first response
+        # (regular file over 1MB or git-lfs tracked file)
+        # in this case, we get the content from the download_url
+        r = requests.get(content_json["download_url"], timeout=self.timeout, **self.kw)
+        if r.status_code == 404:
+            raise FileNotFoundError(path)
+        r.raise_for_status()
+        return MemoryFile(None, None, r.content)
--- a/fsspec/implementations/github.py
+++ b/fsspec/implementations/github.py
@@ -2,8 +2,10 @@ import base64
 
 import requests
 
+from ..asyn import get_loop, sync
 from ..spec import AbstractFileSystem
 from ..utils import infer_storage_options
+from .http import HTTPFile, HTTPFileSystem
 from .memory import MemoryFile
 
 # TODO: add GIST backend, would be very similar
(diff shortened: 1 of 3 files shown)


martindurant:
To clarify: this does not use LFS directly, so there's no way to enable git-lfs for normal (non github) git repos, correct?

martindurant:
To clarify: this does not use LFS directly, so there's no way to enable git-lfs for normal (non github) git repos, correct?

thomasgilgenast:
Yes, this PR not use LFS directly, and therefore it only works for getting Git LFS file content from GitHub and cannot be used with Git LFS repos that are not hosted on GitHub. I see it as a narrow "upgrade" to the github implementation (which was already limited to repos hosted on GitHub, but would not work as-expected for LFS files).

On the other hand, this PR does not require managing a local clone of the Git LFS repo and does not require the user to have git or git-lfs installed.

I guess there's a separate question about if the best path forward is to instead add LFS support to the git implementation, or to add a new git_lfs implementation that would similarly interact with a clone on local disk.

If I understand correctly, the git implementation is for dealing with local git repos that have already been cloned to local disk, while the github implementation feels a lot more like http in that it treats github as "just another source of remote files" (albeit one with versioning powers via `sha`). For my specific use case, I would personally strongly prefer to avoid locally cloning the entire Git LFS repo to local disk (e.g., if I just need one or two files from the repo and if the repo contains multiple large non-LFS files that will make the initial cloning step slow). I also strongly prefer not to have to manually manage the local clone of the repo. But I could also understand if this my personal use case does not match the intent behind the original issue.

fsspec/implementations/tests/test_github.py
```diff
@@ -0,0 +1,41 @@
...
+def test_github_open_small_file():
+    # test opening a small file <1 MB
+    with fsspec.open("github://mwaskom:seaborn-data@4e06bf0/penguins.csv") as f:
+        assert f.readline().startswith(b"species,island")
+
+
+def test_github_open_large_file():
+    # test opening a large file >1 MB
+    with fsspec.open("github://mwaskom:seaborn-data@83bfba7/brain_networks.csv") as f:
+        assert f.readline().startswith(b"network,1,1,2,2")
```
martindurant, line 13: Question: this will download the whole file in one shot, even though it's "large". Do you think we can generate a buffered file (like a HTTPFile), given that we expect the remote to accept HTTP range requests?

thomasgilgenast: This sounds like a great idea - thank you for mentioning this. I will rework this and then rewrite this test to use a range request to avoid fetching the whole file over the network.

martindurant:
You are quite right, LFS for local git would be different, then. I think your "upgrade" is a fine scope to keep this PR to.

fsspec/implementations/tests/test_github.py
thomasgilgenast: Do you have thoughts on if it would also be important to introduce other concepts from the http implementation such as support for async, reuse of an aiohttp ClientSession for the multiple calls to the GitHub API, etc.? Or would it be fine to just use HTTPFile and ensure that range requests are supported, but otherwise not reach for full feature parity with the http implementation in this first PR?

thomasgilgenast: I guess maybe my question may not make sense to ask if the relevant async support is already "enabled" just by virtue of the `asynchronous` kwarg on the HTTPFile constructor. I should probably take a closer look before asking this question 😅 .

martindurant: I did not think about the async-ness. We probably *don't* want this for github, since there is a fairly onerous API rate limit, even if you come with a developer token.

thomasgilgenast: In a751042 I tweaked `_open()` to return an HTTPFile wrapping the `download_url` in cases where the initial response from the GitHub API does not include the file content already. I believe this should enable streaming and range queries.

fsspec/implementations/github.py
```diff
@@ -219,21 +226,42 @@ def _open(
...
+            self.http_fs,
+            content_json["download_url"],
+            session=self.session,
+            block_size=block_size,
+            autocommit=autocommit,
+            cache_options=cache_options,
+            size=content_json["size"],
+            loop=self.loop,
+            **kwargs,
+        )
```
thomasgilgenast, line 267: I think there might be an alternative here to instead just do something like

```suggestion
        return self.http_fs.open(
            content_json["download_url"],
            mode=mode,
            block_size=block_size,
            cache_options=cache_options,
            autocommit=autocommit,
            **kwargs,
        )
```

which avoids some of the complexity of creating the HTTPFile directly.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ full = [
 fuse = ["fusepy"]
 gcs = ["gcsfs"]
 git = ["pygit2"]
-github = ["requests"]
+github = ["fsspec[http]", "requests"]
```
thomasgilgenast, line 59: I've added the http extra here since now the github backend depends on HTTPFileSystem which requires aiohttp.

fsspec/implementations/github.py
thomasgilgenast: I've gone ahead and made this change in 75a6f4e but would be happy to revert if it's going in the wrong direction.

pyproject.toml
martindurant: Can a package depend on itself? Maybe list aiohttp here instead. 

But actually it's optional - for normal small files, you don't need it. Do you think there's an easy way to defer the import and use of HTTPFileSystem (+ appropriate error message if it fails)?

martindurant:
That will do nicely, thank you

thomasgilgenast:
Thank you @martindurant for the helpful comments, patience with my meandering, and quick merge!

PR #1820, opened 2025-04-01
add positional argument

ruoyeruolan:
# Fix compatibility issue with PyTorch Geometric

The current implementation of LocalFileSystem.mv() was causing "TypeError: takes 3 positional arguments but 4 were given" when called from PyTorch Geometric's fs.py. 

This PR explicitly adds the `recursive` parameter to the method signature, allowing it to be properly called with this parameter without causing positional argument errors.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -155,6 +155,11 @@ class LocalFileSystem(AbstractFileSystem):
         return self.cp_file(path1, path2, **kwargs)
 
     def mv(self, path1, path2, recursive: bool = True, **kwargs):
+        """Move files/directories
+        
+        For the specific case of local, all ops on directories are recursive and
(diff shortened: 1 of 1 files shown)


martindurant:
Please add documentation to the method, that this is always actually recursive, and the value of the argument makes no difference.

fsspec/implementations/local.py
```diff
@@ -154,7 +154,12 @@ def get_file(self, path1, path2, callback=None, **kwargs):
     def put_file(self, path1, path2, callback=None, **kwargs):
         return self.cp_file(path1, path2, **kwargs)
 
-    def mv(self, path1, path2, **kwargs):
+    def mv(self, path1, path2, recursive: bool = True, **kwargs):
+        """Move files/directories
+
```
martindurant, line 159:
```suggestion

```

PR #1826, opened 2025-04-24
fix: keep same permissions when using transaction

### LINKED ISSUE #1797: transaction changes default group/guest permissions
Hi,

Saving a file via `fsspec.open()` results in default (macOS) permissions 644.
Saving a file via `fs.transaction, fsopen()` strips group and guest permissions to 600.

Is this intended?

Here's a code snippet:

```
import fsspec

# correct permissions 644
with fsspec.open('correct_permissions.bin', 'wb') as f:
    f.write(b'hello')

# incorrect permissions 600
fs, urlpath = fsspec.core.url_to_fs('incorrect_permissions.bin')
with fs.transaction, fs.open(urlpath, 'wb') as f:
    f.write(b'hello')
```

Thanks!

siemdejong:
Uses https://stackoverflow.com/a/44130549 to check the current umask, temporarily (one line) change it to 0o666, and apply the permission modification.

Fixes #1797 

I'm not sure if this is the right approach, but it passes the test on my machine.

### DIFF
--- a/fsspec/implementations/local.py
+++ b/fsspec/implementations/local.py
@@ -417,6 +417,13 @@ class LocalFileOpener(io.IOBase):
         if self.autocommit:
             raise RuntimeError("Can only commit if not already set to autocommit")
         shutil.move(self.temp, self.path)
+        try:
+            # Get the current umask and set it temporarily to 0o666.
+            umask = os.umask(0o666)
+            os.umask(umask)
+            os.chmod(self.path, 0o666 & ~umask)
+        except RuntimeError:
+            pass
 
     def discard(self):
         if self.autocommit:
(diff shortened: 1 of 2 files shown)


fsspec/implementations/local.py
```diff
@@ -417,6 +417,13 @@ def commit(self):
         if self.autocommit:
             raise RuntimeError("Can only commit if not already set to autocommit")
         shutil.move(self.temp, self.path)
+        try:
+            # Get the current umask and set it temporarily to 0o666.
+            umask = os.umask(0o666)
```
martindurant, line 422: Can we do this just once, and cache the `umask` - so that we don't have to do this on every file?

siemdejong: Good point. Addressed it with an `lru_cache`.

PR #1834, opened 2025-04-30
DBFS - list fixes.

tomerm-iguazio:
1) Support the "details" field in error responses (reproduced when listing a non-existent file).
2) Allow returning an empty list when a directory has no contents.


martindurant:
Is there any way to generate this exception so that we can test it?

tomerm-iguazio:
Tests added:) @martindurant

PR #1840, opened 2025-05-07
docs: correct typo in api.rst

guoci:


martindurant:
Actually, both "gs" and "gcs" work equivalently

guoci:
Is that documented somewhere?
Google Cloud CLI does not recognise `gcs://`
```
$ gcloud storage cp somefile gcs://<redacted>
ERROR: (gcloud.storage.cp) Unrecognized scheme "gcs"
```

martindurant:
This is not the gcloud CLI :)

The definition is [here](https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L282) and many examples of using gcsfs prefer "gcs", since that came first.

PR #1852, opened 2025-06-01
fix: missing support for `update_cond='never'` condition in fsspec.generic.rsync

obendidi:
The `update_cond='never'` was missing in the `if-else` block, this PR add support of that


martindurant:
Thank you - this looks like a well fixed oversight. We ought to have a test for it, if you might provide one.

PR #1853, opened 2025-06-03
docs: add link to new implementation for Microsoft drive API

lmignon:


martindurant:
Is "microsoft drive" the same as "onedrive", the sync-local-files thing? My understanding of the MS ecosystem isn't very good...

lmignon:
> Is "microsoft drive" the same as "onedrive", the sync-local-files thing? My understanding of the MS ecosystem isn't very good...

The current implementation uses the `drive` services from the msgraph api. These services works with OneDrive and Sharepoint drives. I ran the testsuite with a Sharepoint drive with success.

PR #1859, opened 2025-06-11
Local file system following symbolic links

annesophie-cls:
Dear all,

This PR adds an option to LocalFileSystem called `follow_symlinks `allowing to follow symbolic links when walking in a file system.

Please check it out and let me know for required changes

Cheers

Anne-Sophie


martindurant:
info() will cover a lot, but I feel like there are other places where links might or not be followed. Can we have some tests, starting with your original problem? You might need to make tests behave differently on windows versus posix.

PR #1863, opened 2025-06-13
Apply Sourcery suggestiosn and fix typos

DimitriPapadopoulos:


martindurant:
I'm not convinced that every change is an improvement, but it all seems to be OK. Are you planning to do more work here?

DimitriPapadopoulos:
Do you have in mind the changes from `str + str` to `f"{str}{str}"`? Not sure about them either. While f-strings are much faster and considered more readable in a majority of situations, they might not be an improvement here. I can revert the this class of changes if you agree, absolutely not a problem.

I don't intend adding more changes (although Sourcery suggests more). I do think it might be worthwhile to add Sourcery to the repository, it provides AI assistance for PR review. I find it useful, although not bulletproof or astounding.

Issue #1865, opened 2025-06-14
`DatabricksFileSystem.ls()` returns empty list for directories in some scenarios.


mkoistinen:
I'm using `fsspec` to access a number of filesystems including `file`, `fs`, `hdfs`, `webhdfs` and `dbfs`. Of these, only `dbfs` is returning an empty list in some conditions where I request the contents of a Parquet dataset directory. Here's a repro case:

``` python
from __future__ import annotations

import os
from typing import Any
from types import MethodType

import fsspec

if __name__ == "__main__":

    fs = fsspec.filesystem(
        "dbfs",
        instance=os.environ["DATABRICKS_HOST"],
        token=os.environ["DATABRICKS_TOKEN"],
    )

    for path in [
        "/mkoistinen@host.com/parquet_files",          # PATH A
        "/mkoistinen@host.com/parquet_files/bank/",    # PATH B
    ]:
        listing = fs.ls(path=path, detail=False)  # NOTE: the value for `detail` makes no difference
        if listing and isinstance(listing[0], dict):
            print(f"ls({path})\n{'\n'.join(f.get('name') for f in listing)}", end="\n\n")  # noqa
        else:
            print(f"ls({path})\n{'\n'.join(listing)}", end="\n\n")
```
When run with 2025.5.1 (and earlier versions, I tried all released 2025 versions), the output of this run is:

``` bash
ls(/mkoistinen@howso.com/parquet_files)
/mkoistinen@host.com/parquet_files/adult
/mkoistinen@host.com/parquet_files/adult.parquet
/mkoistinen@host.com/parquet_files/bank
/mkoistinen@host.com/parquet_files/bank.parquet
/mkoistinen@host.com/parquet_files/iris
/mkoistinen@host.com/parquet_files/iris.parquet
/mkoistinen@host.com/parquet_files/tmp

ls(/mkoistinen@host.com/parquet_files/adult)
/mkoistinen@host.com/parquet_files/adult
```

If I disable `PATH A` in the list, I get:

``` bash
ls(/mkoistinen@host.com/parquet_files/adult)
/mkoistinen@host.com/parquet_files/adult/part_18aacbf9-6844-4038-8929-c45028ef870c.parquet
/mkoistinen@host.com/parquet_files/adult/part_4950eed3-56d3-4f28-962e-46c9f728f5c4.parquet
[...]

PR #1873, opened 2025-06-20
Prevent returning cached entry if the entry is degenerate

mkoistinen:
When `DatabricksFileSystem.ls` is called with a directory, it will query the API and build a directory listing, then cache it.

If a subsequent `ls` is called on a directory that is child to the first directory, the results that are returned are for the child directory itself, as though it were just a file. The correct behavior is to get the contents of this child directory.

This PR ignores such responses so that `ls` will call the API again with the child directory to get its contents. These results are then cached as usual and available for future repeated `ls` calls with the child directory.

This addresses issue: #1865 by overriding `AbstractFileSystem._ls_from_cache` with a new implementation that only uses the parent's cache entry if it indicates that `path` is not a "directory". If it is a "directory", nothing is returned to allow for `DatabricksFileSystem.ls` to make a new API call to DBFS to get the contents of that path.

Also, if `path` is not found in the cached listing of its parent (if it exists), then it strongly suggests that `path` was created since the parent was cached. Using this information, `DatabricksFileSystem.ls` will invalidate the parent's cache entry (by deleting it) before continuing.


fsspec/implementations/dbfs.py
```diff
@@ -70,7 +90,16 @@ def ls(self, path, detail=True, **kwargs):
...
         """
-        out = self._ls_from_cache(path)
+        try:
+            out = self._ls_from_cache(path)
+        except FileNotFoundError:
+            # This happens if the `path`'s parent was cached, but `path` is not
+            # there. This suggests that `path` is new since the parent was
+            # cached. Attempt to invalidate parent's cache before continuing.
+            with suppress(Exception):
+                del self.dircache[self._parent(path)]
```
martindurant, line 100:
```suggestion
            self.dircache.pop(self._parent(path))
```

mkoistinen:
I think I've fixed the 3.9 test issues (annotation issues). Not sure if there's anything I can do about the 3.12 ones (rate-limiting issues)

fsspec/implementations/dbfs.py
martindurant: I think `pop` is perfectly descriptive of the operation (better than a context and indent), and you need not assign it to the local variable `_` :)

martindurant:
OK, let's give this a try and see how it goes for others

mkoistinen:
Sorry for the delay, my annual leave happened =) The VCR cassettes are re-recorded and the tests should pass now (they do locally).

martindurant:
Sorry it took me a while to notice the update here.
This will do for now, and I hope we don't hear complaints of too many listings lookups ;)
Thanks!

PR #1874, opened 2025-06-23
Use standard library compression.zstd on python >= 3.14

lukeshingles:
[PEP 748](https://peps.python.org/pep-0784/) adds zstandard compression to the standard library from version 3.14 (currently in beta). This PR changes the import behaviour for std to try compression.zstd first before trying to import zstandard. The test dependency on zstandard is restricted to python_version < '3.14'.

### DIFF
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -155,19 +155,26 @@ except ImportError:
     pass
 
 try:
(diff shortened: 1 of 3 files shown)


martindurant:
Should we wait until 3.14 is released?

PR #1878, opened 2025-06-25
fix: add tos scheme to registry

yanghua:
The `tosfs` is a wrong scheme. The right scheme is `tos`


martindurant:
Do you think anyone might already be using "tosfs" as a protocol, should we keep both? I do see that "tos" is indeed the protocol at https://github.com/fsspec/tosfs/blob/main/tosfs/core.py#L104

yanghua:
Sounds reasonable.

yanghua:
@martindurant, please take a look. Thank you.

PR #1893, opened 2025-07-22
fix: download directories with http fs

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

This PR includes two fixes:

- Skip directories when getting file from remote http to local
- Unquote file names when creating local files

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -1,10 +1,11 @@
 import asyncio
 import io
 import logging
+import os
 import re
 import weakref
 from copy import copy
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 import aiohttp
 import yarl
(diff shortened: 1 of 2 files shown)


fsspec/implementations/http.py
```diff
@@ -253,6 +254,9 @@ async def _get_file(
         kw.update(kwargs)
         logger.debug(rpath)
         session = await self.set_session()
+        if await self._isdir(rpath):
```
martindurant, line 257: This will result in essentially every file getting downloaded twice, once to check the directory. This information should already be available in _get(), so we can skip there rather than here. It doesn't really make sense to call _get_file() on a path that is not considered a file.

martindurant (commented):
Thanks for the fix! I have a comment about where best to insert it.

On the issue of unquote, do you have an example of the kind of thing you were seeing? This is essentially an HTTP href-> path translation? If yes, maybe it should be done at listing time?

PR #1905, opened 2025-08-07
localmemory: init

### LINKED ISSUE #1904: create multiple separate memory filesystems
[fsspec/implementations/memory.py](https://github.com/fsspec/filesystem_spec/blob/c46db870f828f7f1318ed7a3dc26ecc8b48b3880/fsspec/implementations/memory.py#L17)

```py
class MemoryFileSystem(AbstractFileSystem):
    """A filesystem based on a dict of BytesIO objects

    This is a global filesystem so instances of this class all point to the same
    in memory filesystem.
    """

    store: ClassVar[dict[str, Any]] = {}  # global, do not overwrite!
    pseudo_dirs = [""]  # global, do not overwrite!
```

... ok, but i need multiple different memory filesystems in parallel

```py
assert (
    fsspec.filesystem("memory://").store !=
    fsspec.filesystem("memory://").store
)
```

this should be as simple as moving `store` and `pseudo_dirs` to `def __init__`

maybe add a new protocol `localmemory`

milahu:
fix #1904 

also fix the `memory` filesystem:

> AttributeError: 'MemoryFileSystem' object has no attribute '_intrans'

all tests are passing:

> `nix-shell -p python3.pkgs.{aiohttp,pytest,requests,numpy,pytest-asyncio} --run 'pytest -k localmemory -k memory --maxfail=1'`


fsspec/implementations/localmemory.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class LocalMemoryFileSystem(MemoryFileSystem):
+    """A filesystem based on a dict of BytesIO objects
+
+    This is a local filesystem so different instances of this class
+    point to different memory filesystems.
+    """
+
+    pseudo_dirs = None
```
martindurant, line 17: Why set the class variable even if none of the instances see it?

fsspec/implementations/localmemory.py
```diff
@@ -0,0 +1,27 @@
...
+    protocol = "localmemory"
+    root_marker = "/"
+    _intrans = False
+    cachable = False  # same as: skip_instance_cache = True
+
+    def __init__(self, *args, **kwargs):
+        self.logger = logger  # global
```
martindurant, line 24: Why bother having this on this instance? We only even have one logger anyway.

fsspec/implementations/memory.py
```diff
@@ -25,6 +25,11 @@ class MemoryFileSystem(AbstractFileSystem):
     pseudo_dirs = [""]  # global, do not overwrite!
     protocol = "memory"
     root_marker = "/"
+    _intrans = False
+
+    def __init__(self, *args, **kwargs):
+        self.logger = logger
+        super().__init__(*args, **kwargs)
```
martindurant, line 32:
```suggestion
```

martindurant (commented):
I wasn't expecting a new implementation - I thought you were thinking to change the existing one. This, of course, never has a chance to affect the existing tests and fixtures.

I wouldn't even bother putting this in a new module, since it's so similar to rmemoryFS.

milahu:
success. merged with `MemoryFileSystem`
i only had to change `fsspec/spec.py`

```diff
-        skip = kwargs.pop("skip_instance_cache", False)
+        skip = kwargs.get("skip_instance_cache", False)
```

to pass `skip_instance_cache` to init

i wanted to avoid adding a new parameter like `local_memory=True`
because this would have required setting two parameters:
`local_memory=True, skip_instance_cache=True`

milahu:
sorry, abandoned

(i would prefer "a flag in init" to not pollute the classes namespace)

PR #1909, opened 2025-08-24
Feature, Add a lookup of Last-Modified in http file headers

TheLazzziest:


martindurant:
I am happy to include this. Please post a link to the RFC for posterity (I was not aware of it).

TheLazzziest:
Hi @martindurant! Thank you for your quick feedback. I can, but it seems that the original document is a quite confusing in this regard. 

Don't you mind if I mention the document this way: [RFC: 1123](https://datatracker.ietf.org/doc/html/rfc1123#section-5.2.14) -> [RFC: 822#Section-5](https://datatracker.ietf.org/doc/html/rfc822#section-5)

It should become more obvious where the actual knowledge comes from.

In addition to that, I used this [resource](https://wiki.freepascal.org/RFC_1123_Time_Format) for a more distilled explanation. Let me know if that works.

martindurant:
Thanks for the links, that will do nicely. I'll try to see why the CI test is reporting filenotfound for a URL that clearly does exist. 

This is also my first time finding out about dlt, which I only briefly glanced at so far. It seems to have some ideas in common with [intake](https://intake.readthedocs.io/en/latest/).

martindurant:
It turns out the failure was temporary.

PR #1911, opened 2025-08-30
fix: Ensure kwargs passed via `expand_path`

ReubenFrankel:
Pass through kwargs to `expand_path` calls where possible to ensure they are propagated to `ls` implementation by default.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -892,7 +892,7 @@ class AbstractFileSystem(metaclass=_Cached):
         dict of {path: contents} if there are multiple paths
         or the path has been otherwise expanded
         """
-        paths = self.expand_path(path, recursive=recursive)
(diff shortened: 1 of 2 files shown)


martindurant:
I agree, but is there any particular situation in which this was a problem for you, perhaps something that could be made into a test?

ReubenFrankel:
I think I was trying to use `download`/`get` and wanted to pass some kwargs through to my `ls` implementation, which required this change. Happy to write some tests -- would those fit best in `fss spec/tests/test_spec.py`?

martindurant:
>  Happy to write some tests -- would those fit best in fss spec/tests/test_spec.py

Yes, that would be fine. You will see that many of the general-purpose tests use one of the simpler filesystem implementations, though in this case you would need one that actually makes use of some kwarg.

PR #1912, opened 2025-09-02
changelog

martindurant:


martindurant:
Ah, I must have missed the changelog bullet for it. Would you mind adding?

PR #1922, opened 2025-09-28
Add missing open-modes for memory filesystem (fixes #1921)

### LINKED ISSUE #1921: MemoryFilesSstem missing valid modes for open()
```python
import fsspec

fs = fsspec.filesystem('memory')

with fs.open('hello', 'w+b') as f:
    f.write(b'world')
```

results in:

```shell
$ python3 /tmp/filemode.py
  File "/tmp/try2.py", line 5, in <module>
    with fs.open('hello', 'w+b') as f:
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "virtualenvs/app-tNTrAeIV-py3.12/lib/python3.12/site-packages/fsspec/spec.py", line 1338, in open
    f = self._open(
        ^^^^^^^^^^^
  File "virtualenvs/app-tNTrAeIV-py3.12/lib/python3.12/site-packages/fsspec/implementations/memory.py", line 211, in _open
    raise ValueError(f"unsupported file mode for {name}: {mode!r}")
ValueError: unsupported file mode for MemoryFileSystem: 'w+b'
```

The `open()` method should respect all valid modes, including "w+b", "a+b", "x+b". 
Will follow up with a PR to fix this behavior.

Note also that the memory filesystem does not repsect "read-only"/"write-only" properties; However fixing this is more involved change, and may impact backward compatibilty for users that rely on such (broken) behavior, and nobody complained so far.

orenl:
Resolves #1921

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -187,10 +187,10 @@ class MemoryFileSystem(AbstractFileSystem):
             parent = self._parent(parent)
             if self.isfile(parent):
                 raise FileExistsError(parent)
-        if mode in ["rb", "ab", "r+b"]:
+        if mode in ["rb", "ab", "r+b", "a+b"]:
             if path in self.store:
                 f = self.store[path]
-                if mode == "ab":
+                if "a" in mode:
                     # position at the end of file
                     f.seek(0, 2)
                 else:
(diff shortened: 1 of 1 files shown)


martindurant:
This if fine, I will merge. However, I'm not sure how useful those extra modes are - essentially all remote FSs only allow "w" and "r" and even "a" needs emulating.

PR #1923, opened 2025-09-28
Fix lazy reference handling for .parquet/.parq files

### LINKED ISSUE #1919: HTTP .parquet reference files mis-interpreted as JSON files due to isfile check in ReferenceFileSystem
I want to report a bug in fsspec / kerchunk related to how .parquet reference files served over HTTP are handled. The issue is that HTTP .parquet refs get misinterpreted as JSON refs, because of the isfile(...) logic in ReferenceFileSystem.

Reproduction / Example

There is a .parquet reference file available via HTTP, e.g.
https://data-taos.ifremer.fr/kerchunk/copernicus_NWSHELF_2022.parquet/

When doing xr.open_dataset(uri, engine="kerchunk"), internally fsspec calls in fsspec/implementations/reference.py:

```
ref_fs, fo2 = fsspec.core.url_to_fs(fo, **dic)
if ref_fs.isfile(fo2):
    # text JSON
    with fsspec.open(fo, "rb", **dic) as f:
        logger.info("Read reference from URL %s", fo)
        text = json.load(f)
    self._process_references(text, template_overrides)
else:
    # Lazy parquet refs
    logger.info("Open lazy reference dict from URL %s", fo)
    self.references = LazyReferenceMapper(
        fo2,
        fs=ref_fs,
        cache_size=cache_size,
    )
```

Because ref_fs.isfile(fo2) returns True for that HTTP URI (since the server responds with 200 OK), it enters the JSON branch. But the content is not a JSON reference dict — it's either a directory listing, an HTML index, or an HTTP page, not the actual parquet file. This leads to a JSON parsing error (e.g. JSONDecodeError).

```
File /srv/conda/envs/notebook/lib/python3.12/site-packages/fsspec/implementations/reference.py:705, in ReferenceFileSystem.__init__(self, fo, target, ref_storage_args, target_protocol, target_options, remote_protocol, remote_options, fs, template_overrides, simple_templates, max_gap, max_block, cache_size, **kwargs)
    703     with fsspec.open(fo, "rb", **dic) as f:
    704         logger.info("Read reference from URL %s", fo)
--> 705         text = json.load(f)
    706     self._process_references(text, template_overrides)
    707 else:
    708     # Lazy parquet refs

JSONDecodeError: Expected object or value
```

I precise that in local (obviously) the .parquet can be open properly and is not corrupted.

capetienne:
fix: only treat .parquet/.parq files as lazy reference dicts

- Add utility get_file_extension() to extract extension from URL/path  
- Update condition to `ref_fs.isfile(fo2) and ext in {"parq","parquet"}`  
- Prevent misinterpretation of parquet ref files as JSON  
- Add unit tests 

- closes #1919

### DIFF
--- a/fsspec/implementations/reference.py
+++ b/fsspec/implementations/reference.py
@@ -22,7 +22,11 @@ from fsspec.asyn import AsyncFileSystem
 from fsspec.callbacks import DEFAULT_CALLBACK
 from fsspec.core import filesystem, open, split_protocol
 from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
-from fsspec.utils import get_file_extension, isfilelike, merge_offset_ranges, other_paths
+from fsspec.utils import (
+    isfilelike,
+    merge_offset_ranges,
+    other_paths,
+)
 
 logger = logging.getLogger("fsspec.reference")
 
@@ -698,7 +702,9 @@ class ReferenceFileSystem(AsyncFileSystem):
                 **(ref_storage_args or target_options or {}), protocol=target_protocol
             )
             ref_fs, fo2 = fsspec.core.url_to_fs(fo, **dic)
-            if get_file_extension(fo2) in {"parq", "parquet"}:
+            if ".json" not in fo2 and (
+                fo.endswith(("parq", "parquet", "/")) or ref_fs.isdir(fo2)
+            ):
                 # Lazy parquet refs
                 logger.info("Open lazy reference dict from URL %s", fo)
                 self.references = LazyReferenceMapper(
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -338,6 +338,7 @@ def test_get_protocol(par):
     url, outcome = par
     assert get_protocol(url) == outcome
 
+
 @pytest.mark.parametrize(
     ["url", "expected"],
     (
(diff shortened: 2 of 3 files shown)


fsspec/implementations/reference.py
```diff
@@ -698,20 +698,20 @@ def __init__(
...
             )
             ref_fs, fo2 = fsspec.core.url_to_fs(fo, **dic)
-            if ref_fs.isfile(fo2):
-                # text JSON
-                with fsspec.open(fo, "rb", **dic) as f:
-                    logger.info("Read reference from URL %s", fo)
-                    text = json.load(f)
-                self._process_references(text, template_overrides)
-            else:
+            if get_file_extension(fo2) in {"parq", "parquet"}:
```
martindurant, line 701: The path for parqut should be a directory - can we include anything finishing with "/"? Or does that get lost before making it here?

It might be better to swap the conditions around, and require .json for non-parquet? I have a feeling that should be more standard.

keewis: > and require .json for non-parquet?

the main issue with that is that as a directory, the kerchunk "parquet" format will most likely not be compressed, while `.json` may be, and thus have a suffix. This, in turn, will make file extensions harder to extract properly, e.g. the full file extension may be  `.json.zstd`. Not sure if that's actually a problem in practice, though.

PR #1929, opened 2025-10-07
Add marker for chained filesystems for path passthrough

### LINKED ISSUE #1767: CachedFileSystem is special in URL chaining
I'm trying to implement a file system that always replaces the `'a'` open mode with `'w'`. The purpose is to support TensorBoard on filesystems (local mounts of remote network FS, which is `LocalFileSystem` in `fsspec`) that don't support the append writing mode. The new `TruncateFileSystem` works like `CachedFileSystem`, but it modifies the `open` parameter while forwarding all other calls to the underlying file system.

However, when I tried to use `url_to_fs` with my file system, I noticed the following behavior:

image

I believe this behavior occurs due to the special handling in the following code:  
https://github.com/fsspec/filesystem_spec/blob/90c7cd9e6c939fc37341fd793831a399753ebfd9/fsspec/core.py#L361-L362

This makes it difficult for third-party implementations to achieve a similar URL chaining behavior as the `cached` file system.  

Here’s my idea: There are two types of URL chaining. One type uses a **file** from the underlying file system (e.g., `zip`), while the other type uses the entire file system from another implementation (e.g., `cached` and m `truncate`). These two styles and their implementations should be clearly documented (currently, the URL chaining documentation lacks a description of this behavior for the `cached` file system). Furthermore, third-party implementations should be able to achieve the same behavior without needing to use `CachedFileSystem` as a base class.  

One possible solution is to create a new marker class, `ChainedFileSystem`, that inherits from `AbstractFileSystem` but introduces no new implementation. Then, make `CachedFileSystem` inherit from `ChainedFileSystem`.

timkpaine:
fixes  #1767

### DIFF
--- a/fsspec/implementations/chained.py
+++ b/fsspec/implementations/chained.py
@@ -19,4 +19,5 @@ class ChainedFileSystem(AbstractFileSystem):
 
     Additional functionality may be added in the future.
     """
+
     protocol: ClassVar[str] = "chained"


fsspec/tests/test_chained.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self, target_protocol="", target_options=None, **kwargs):
+        super().__init__(**kwargs)
+        self.fs = filesystem(target_protocol, **target_options)
+
+class MyNonChainedFS(AbstractFileSystem):
+    protocol = "mynonchain"
+
+@pytest.fixture(scope="module")
+def register_fs():
+    register_implementation(MyChainedFS.protocol, MyChainedFS)
```
martindurant, line 21: Note that these will remained registered for the rest of the test session. I don't expect a problem from this.

fsspec/implementations/chained.py
```diff
@@ -0,0 +1,9 @@
+from typing import ClassVar
+
+from fsspec import AbstractFileSystem
+
+__all__ = ("ChainedFileSystem",)
+
+
+class ChainedFileSystem(AbstractFileSystem):
```
martindurant, line 8: A comprehensive docstring here would be nice, and this should be added to the API docs

fsspec/implementations/chained.py
```diff
@@ -0,0 +1,9 @@
...
+    chained: ClassVar[str] = "chained"
```
martindurant, line 9:
```suggestion
    protocol: ClassVar[str] = "chained"
```
?

martindurant:
Thanks for working on this!

martindurant:
Is there any need to explicitly consider async filesystems here?

timkpaine:
> Is there any need to explicitly consider async filesystems here?

I would think this needs to be considered more generically, e.g. what happens if you chain an async fs inside a sync fs? I would think basically no implementations handle this right now. We could consider having

```python
AbstractChainedFilesystem
ChainedFileSystem
AsyncChainedFileSystem
```

To provide the correct indications via inheritance.

PR #1932, opened 2025-10-16
Mark dir, async wrapper fs as chained

timkpaine:
- `async_wrapper` was missing from known implementations and wrongly detected by regex logic in core
- dirfs and async_wrapper marked as chained, chaining detection logic updated to pass in `bit` if its being used (in which case you need to pass args in via open)

I am validating largely by testing https://github.com/1kbgz/fsspec-union/blob/544723c93bc6d72071494c040d0c28edda5ee088/fsspec_union/tests/conftest.py#L10
which uses (abuses?) chaining to construct a read-through cache, and https://github.com/1kbgz/fsspec-python/blob/216dee3f3b95cd944910acca8b98626d4795ca9a/fsspec_python/tests/conftest.py#L63 which uses chaining to mount the fs as an import path.

### DIFF
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -18,7 +18,7 @@ from fsspec.caching import (  # noqa: F401
 )
 from fsspec.compression import compr
 from fsspec.config import conf
-from fsspec.registry import filesystem, get_filesystem_class
+from fsspec.registry import available_protocols, filesystem, get_filesystem_class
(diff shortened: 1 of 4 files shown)


martindurant:
This all looks fine and I am merging it now so it makes the release.
Please come back and write some tests, probably around dirFS, which exercises the new code path.

PR #1946, opened 2025-11-12
Add support for Python 3.14 and drop EOL 3.9

hugovk:
Follow on from https://github.com/fsspec/s3fs/pull/993.


martindurant (commented):
Can you comment on why strict=True needs to be added to zip(), but mode="r" (or "rt") needs to be removed from open() when these are both the default values? I would rather not have the big diff if possible.

PR #1951, opened 2025-11-20
fix(asyn): support sync style implementation in async _rm_file

Sbargaoui:
# Fix AsyncFileSystem._rm_file() to support sync-style implementations

## Summary

Fixes architectural inconsistency between sync and async file deletion APIs that prevents `rm_file()` from working on async filesystems that implement `_rm()` instead of `_rm_file()`.

## Problem

The async and sync APIs have inverted delegation patterns:

**Sync (AbstractFileSystem):**
```python
rm_file(path) → _rm(path)  # subclasses implement _rm()
```

**Async (AsyncFileSystem):**
```python
_rm(path) → _rm_file(path)  # but _rm_file() raises NotImplementedError
```

This breaks external async filesystems (like `adlfs.AzureBlobFileSystem` where this issue was encoutered) that follow the sync pattern and implement `_rm()` but not `_rm_file()`:
- `fs.rm()` works (calls the overridden `_rm()`)
- `fs.rm_file()` fails with `NotImplementedError` (calls `_rm_file()` which isn't implemented)

## Root Cause

```python
# fsspec/asyn.py
async def _rm_file(self, path, **kwargs):
    raise NotImplementedError  # Wrong extension point

async def _rm(self, path, recursive=False, batch_size=None, **kwargs):
    path = await self._expand_path(path, recursive=recursive)
    return await _run_coros_in_chunks(
        [self._rm_file(p, **kwargs) for p in reversed(path)],  # Calls unimplemented _rm_file
        batch_size=batch_size,
        nofiles=True,
    )
```

## Proposal

Make `_rm_file()` detect if subclass implements `_rm()` and delegate to it:

```python
# fsspec/asyn.py
async def _rm_file(self, path, **kwargs):
    if type(self)._rm is not AsyncFileSystem._rm:
        return await self._rm(path, recursive=False, batch_size=1, **kwargs)
    raise NotImplementedError
```

### Modified Files

1. **fsspec/asyn.py**
   - Added delegation logic to `_rm_file()`

2. **fsspec/tests/test_async.py**
   - Added `test_rm_file_with_rm_implementation()`: Tests sync-style pattern (implements `_rm()`)
   - Added `test_rm_file_with_rm_file_implementation()`: Tests async-style pattern (implements `_rm_file()`)
   - Added `test_rm_file_without_implementation()`: Tests error handling

Open to all feedback and happy to iterate !

### DIFF
--- a/fsspec/implementations/tests/test_local.py
+++ b/fsspec/implementations/tests/test_local.py
@@ -929,37 +929,35 @@ def test_strip_protocol_no_authority(uri, expected, cwd, current_drive):
 
 @pytest.mark.parametrize(
     "uri, expected",
-    (
-        [
-            ("file:/path", "/path"),
-            ("file:///path", "/path"),
-            ("file:////path", "//path"),
-            ("local:/path", "/path"),
-            ("s3://bucket/key", "{cwd}/s3://bucket/key"),
-            ("/path", "/path"),
-            ("file:///", "/"),
-        ]
-        if not WIN
-        else [
-            ("file:c:/path", "c:/path"),
-            ("file:/c:/path", "c:/path"),
-            ("file:/C:/path", "C:/path"),
-            ("file://c:/path", "c:/path"),
-            ("file:///c:/path", "c:/path"),
-            ("local:/path", "{current_drive}/path"),
-            ("s3://bucket/key", "{cwd}/s3://bucket/key"),
-            ("c:/path", "c:/path"),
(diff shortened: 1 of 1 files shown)


Sbargaoui:
Good point! I can update the check to use `inspect.iscoroutinefunction(self._rm)` combined with the identity check.

That way it'll cover subclasses with custom `async def _rm()` are detected (even through inheritance) and make sure that base implementation doesn't recurse into itself.

PR #1958, opened 2025-12-06
feat: Allow specifying SSL/TLS protocol version for FTP-TLS connections.

HviorForgeFlow:
This change introduces support for implicit FTPS connections in the FTPFileSystem. Previously, only explicit FTPS was supported.

A new parameter security tls has been added to the FTPFileSystem constructor. This parameter allows users to specify the SSL/TLS protocol to be used for implicit FTPS connections,
such as TLSv1.2.

Key changes:
- Introduced ImplicitFTPTLS class to handle automatic SSL wrapping for implicit FTPS.
- Added security tls parameter to FTPFileSystem to allow selection of SSL/TLS protocols (e.g., "tlsv1_2").
- Updated the connection logic to differentiate between explicit and implicit FTPS based on the security tls parameter.

This enhancement provides greater flexibility and compatibility for connecting to various FTP servers requiring implicit FTPS with specific security protocols.

### DIFF
--- a/fsspec/implementations/tests/test_ftp.py
+++ b/fsspec/implementations/tests/test_ftp.py
@@ -8,7 +8,7 @@ import pytest
 
 import fsspec
 from fsspec import open_files
-from fsspec.implementations.ftp import FTPFileSystem
+from fsspec.implementations.ftp import FTPFileSystem, ImplicitFTPTLS
 
 ftplib = pytest.importorskip("ftplib")
 here = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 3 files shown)


martindurant:
Thank you for contributing. I expect to be able to look through this shortly.

fsspec/implementations/tests/test_ftp.py
```diff
@@ -43,6 +59,16 @@ def test_tls(ftp, tls, exp_cls):
...
 
 
+@pytest.mark.parametrize(
+    "tls,exp_cls",
+    (("tlsv1_2", ImplicitFTPTLS),),
+)
+def test_tls_v12(ftp_tlsv12, tls, exp_cls):
+    host, port = ftp_tlsv12
+    fs = FTPFileSystem(host, port, tls=tls)
+    assert isinstance(fs.ftp, exp_cls)
```
martindurant: Maybe also execute some operation here, anything would do.

HviorForgeFlow:
Looks like the workflow approval got rejected when I squash + force push the commits.
Added a read check on the same test method.

martindurant:
I believe approval is required every time if you are not already a contributor. For future reference, we squash-merge in this repo anyway, so no need to consolidate your commits in the PR branch.

PR #1984, opened 2026-02-02
Remove prefetching loop

TLCFEM:


fsspec/caching.py
```diff
@@ -391,19 +391,8 @@ def _fetch(self, start: int | None, end: int | None) -> bytes:
         if start >= self.size or start >= end:
             return b""
 
-        # byte position -> block numbers
-        start_block_number = start // self.blocksize
-        end_block_number = end // self.blocksize
```
TLCFEM: The calculation of `end_block_number` is not correct.

PR #2012, opened 2026-04-13
Fix closing contexts of tar and zip files and add test 3.14t in CI

ariostas:
I ran into an issue while testing Uproot with Python 3.14t in this PR https://github.com/scikit-hep/uproot5/pull/1619. Since garbage collection on freethreaded mode is a bit looser, it doesn't end up immediately closing contexts when an object is dropped, and you end up with warnings like `ResourceWarning: unclosed file ...`. To solve this, we can explicitly call `__exit__` in `__del__`. I also added `close()` to the tar implementation, since it was missing.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   linux:
-    name: ${{ matrix.PY }}-pytest
+    name: ${{ matrix.PY }}${{ matrix.FREETHREADING && 't' || '' }}-pytest
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


ikrommyd:
is there a particular reason fsspec is not being tested in 3.14t atm @martindurant ?

martindurant:
> is there a particular reason fsspec is not being tested in 3.14t

Only that no one has added it - feel free to do so.

ikrommyd:
Should it be added here and turn this into a 3.14t testing/compatibility PR?

martindurant:
Yes, let's try

ariostas:
I made some final tweaks and tested the CI in my fork and everything works: https://github.com/ariostas/filesystem_spec/actions/runs/24408370609/job/71297942303

PR #2019, opened 2026-04-23
Updating `adl` error message

anjaliratnam-msft:
We are removing support for ADLS Gen1 in this [PR](https://github.com/fsspec/adlfs/pull/545), so users will no longer be able to use the `AzureDatalakeFileSystem` class. This PR is for updating the `adl://` protocol error message to instruct users to use the `az://` protocol for Azure Blob Storage and Azure Data Lake Storage Gen2.

### DIFF
--- a/fsspec/registry.py
+++ b/fsspec/registry.py
@@ -67,8 +67,9 @@ known_implementations = {
     "adl": {
         "class": "adlfs.AzureDatalakeFileSystem",
         "err": (
-            "ADLS Gen1 is no longer supported. Please install adlfs and use the "
-            "`az` protocol for Azure Datalake Gen2 and Azure Blob Storage."
+            "Azure Data Lake Storage Gen1 is retired and no longer supported. Please "
+            "install adlfs and use the `az://` protocol to access Azure Blob Storage "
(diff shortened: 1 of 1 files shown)


fsspec/registry.py
```diff
@@ -66,7 +66,10 @@ def register_implementation(name, cls, clobber=False, errtxt=None):
     },
     "adl": {
         "class": "adlfs.AzureDatalakeFileSystem",
-        "err": "Install adlfs to access Azure Datalake Gen1",
+        "err": (
+            "ADLS Gen1 is no longer supported. Please install adlfs and use the "
+            "`az` protocol for Azure Datalake Gen2 and Azure Blob Storage."
```
kyleknap: For the error message, let's make it:
```
Azure Data Lake Storage Gen1 is retired and no longer supported. Please install adlfs and use the az:// protocol to access Azure Blob Storage and Azure Data Lake Storage Gen2 instead.
```
Mainly just cleaned up the wording and make sure the full official product name for the services.

Looks good otherwise.

kyleknap (approved):
Looks good to me! :ship:

PR #2028, opened 2026-04-30
Fix: Omit `threading.get_ident()` from cache token for async implementations in sync mode

### LINKED ISSUE #2020: Remove threading.get_ident() from instance cache token for asynchronous=False instances
**The Problem:**
Currently, the `AbstractFileSystem` instance cache incorporates `threading.get_ident()` into its cache token (in `fsspec/spec.py`). 

This means that if multiple Python threads request the same filesystem (e.g., `fsspec.filesystem("gs")`), `fsspec` completely bypasses the cache and creates a brand-new filesystem instance and a separate `aiohttp.ClientSession` for every single thread.

In libraries like Hugging Face `datasets`, which heavily utilize multi-threading (via `thread_map`) to fetch metadata. Every background thread creates a new FS instance, forcing redundant authentication, TLS handshaking, and connection pooling overhead. This defeats the purpose of the cache and wastes significant time and bandwidth.

**Historical Context:**
By tracing the `fsspec` git history, it appears this thread-specific caching is a vestige of an older event-loop architecture:

1. In **PR #572** (Commit `b252369`), `fsspec` implemented thread-local event loops (`loops[ident] = loop`). Because `aiohttp.ClientSession` is bound to a specific event loop, `threading.get_ident()` was correctly added to the cache token so each calling thread got its own FS instance and its own event loop.
2. In **PR #590** (Commit `57d6ce92`), the architecture was completely revamped. The thread-local event loops were removed in favor of a single, global background IO loop (`fsspecIO`). 
3. However, when the event loop became a global singleton, `threading.get_ident()` was never removed from the cache token.

**Why it is safe to remove now (for `asynchronous=False`):**
Because `fsspec` currently uses a single global `fsspecIO` event loop for all synchronous calls (`asynchronous=False`), all `sync()` coroutine executions are inherently funneled into that single background thread. 

Therefore, multiple calling Python threads can perfectly and safely share the exact same `AsyncFileSystem` instance (and its underlying `aiohttp` session). The session is only ever mutated by the cooperative `fsspecIO` background thread.

**Proposed Solution:**
We propose modifying `fsspec/spec.py` to remove `threading.get_ident()` from the caching token if `asynchronous=False` (the default behavior). 

For `asynchronous=True`, retaining a loop-specific or thread-specific token is still necessary, as advanced users may be running multiple event loops across different threads, and `aiohttp` sessions cannot cross event loop boundaries.

This change would massively improve performance and reduce connection overhead for synchronous, multi-threaded workloads interacting with remote filesystems.

yuxin00j:
**The Problem:**
Currently, the `AbstractFileSystem` instance cache incorporates `threading.get_ident()` into its cache token. For workloads using multi-threading (like Hugging Face `datasets` fetching metadata), this completely bypasses the cache, forcing `fsspec` to create a brand-new filesystem instance and `aiohttp.ClientSession` for every single thread. This results in redundant authentication and connection overhead, defeating the purpose of the cache.

**The Solution:**
As discussed in #2020, because `fsspec` currently uses a single global `fsspecIO` event loop for all synchronous calls (`asynchronous=False`), all `sync()` coroutine executions are inherently funneled into that single background thread. Therefore, multiple calling Python threads can perfectly and safely share the exact same `AsyncFileSystem` instance.

This PR modifies `fsspec/spec.py` to conditionally remove `threading.get_ident()` from the caching token if:
1. The class uses an async implementation (`async_impl = True`).
2. The user requests a synchronous interface (`asynchronous=False`).

For purely synchronous filesystems (e.g., FTP) or when `asynchronous=True`, the thread ID remains in the cache token to ensure thread safety and avoid cross-loop boundary violations.

**Changes:**
- Conditionally generated the cache `token` in `_Cached.__call__`.
- Added tests in `fsspec/tests/test_cache_threads.py` to explicitly verify the caching behavior across multiple threads for both `AsyncFileSystem` and `AbstractFileSystem` under different modes.

Closes #2020

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -59,12 +59,12 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
       - name: Setup conda
-        uses: conda-incubator/setup-miniconda@v3
+        uses: conda-incubator/setup-miniconda@v4
         with:
           environment-file: ci/environment-win.yml
 
--- a/ci/environment-friends.yml
+++ b/ci/environment-friends.yml
@@ -24,7 +24,7 @@ dependencies:
   - google-api-python-client
   - httpretty
   - aiobotocore
-  - moto
+  - moto !=5.2.0
   - flask
   - pip:
     - git+https://github.com/fsspec/s3fs
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -5,4 +5,3 @@ dependencies:
   - python=3.11
   - s3fs
   - pytest
-  - pytest-asyncio
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -2,7 +2,10 @@ name: test_env
 channels:
   - conda-forge
 dependencies:
-  - python=3.11
+  - python=3.13
   - s3fs
   - pytest
   - pytest-asyncio
+  - pytest-mock
+  - requests
+  - numpy
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -5,3 +5,4 @@ dependencies:
   - python=3.11
   - s3fs
   - pytest
+  - pytest-asyncio
--- a/ci/environment-win.yml
+++ b/ci/environment-win.yml
@@ -4,4 +4,3 @@ channels:
 dependencies:
   - python=3.11
   - s3fs
-  - pytest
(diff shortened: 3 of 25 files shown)


yuxin00j:
Hi @martindurant, can we retry the CI for this PR? Thanks.

martindurant:
It's your specific new tests that are failing, but only for older python versions.

yuxin00j:
Hi @martindurant, I have updated the test. Could you help run the CI again?

martindurant:
We seem to be missing some test dependencies in the windows CI env. Maybe they should all be in the [test] env define in pyproject. I added pytest-asyncio directly to the env, but there are more.

martindurant:
@yuxin00j , let's leave investigating what's going on with the install process for windows another time.

PR #2030, opened 2026-05-03
fix: pass on_error when mv calls copy

Yonghui-Lee:
#1576 added an explicit `on_error="raise"` to the internal `copy()` call inside `mv()` so that a failing recursive copy would not be followed by deleting the source. The keyword was misspelled as `onerror`, so it was silently absorbed by `**kwargs` and `copy()` fell back to its default — which for recursive calls is `on_error="ignore"`. 

The tests added in #1576 didn't catch this because they only ran against `LocalFileSystem`, whose `mv()` override calls `shutil.move()` directly and never reaches the patched line.

### DIFF
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -112,6 +112,18 @@ def test_mv_same_paths(m):
     assert m.exists("src/file.txt")
 
 
+def test_mv_recursive_propagates_cp_file_errors(m, monkeypatch):
+    m.pipe("/src/a.txt", b"hello")
+
+    def failing_cp_file(path1, path2, **kwargs):
+        raise FileNotFoundError(path1)
+
+    monkeypatch.setattr(m, "cp_file", failing_cp_file)
+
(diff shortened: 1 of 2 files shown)


fsspec/implementations/tests/test_memory.py
```diff
@@ -112,6 +112,18 @@ def test_mv_same_paths(m):
...
+    m.pipe("/src/a.txt", b"hello")
+
+    def failing_cp_file(path1, path2, **kwargs):
+        raise FileNotFoundError(path1)
+
+    monkeypatch.setattr(m, "cp_file", failing_cp_file)
+
+    with pytest.raises(FileNotFoundError):
+        m.mv("/src", "/dst", recursive=True)
+
```
martindurant, line 127: So we should check that the original file didn't get deleted?

Yonghui-Lee: Done

martindurant:
(s3fs error due to https://github.com/getmoto/moto/issues/10010 )

PR #2031, opened 2026-05-03
replay DEP: de-duplicate and sort optional dependencies (#2021)

neutrinoceros:
Reverts the revert PR fsspec/filesystem_spec#2026

I'm opening as a draft to try and experiment with a remediation strategy


neutrinoceros:
One unexpected (to me) detail that seems extremely important here: the `2026.4.0` tag for this repo doesn't appear to be a parent of the current `master` branch, which makes the dynamic version I get `2026.3.0+...` (which is incorrect), and this violates `gcsfs`'s requirement on `fsspec>=2026.4.0`. I rebased this branch onto the `2026.4.0` tag, but now I'm seeing another resolution error with `s3fs`.

neutrinoceros:
https://github.com/fsspec/s3fs/pull/1023 contains the fix to the remaining issue. I checked the patch locally with uv a path dependency (`uv add --optional s3 ../s3fs`) that, combined with having the `2026.4.0` tag as a parent, and this patch, actually fixes all resolution errors I'm aware of. I'll push a temporary patch to the same here for demonstration purposes.

neutrinoceros:
>  I'll push a temporary patch to the same here for demonstration purposes.

scratch that. pip supports dependencies specified as github repos via `requirements.txt`, but apparently not from `pyproject.toml` (which makes sense since this isn't a standard)

martindurant:
> the 2026.4.0 tag for this repo doesn't appear to be a parent of the current master branch

This at least I have fixed. Something happened to the merge commit, I don't know where github put it - should be right now. Perhaps related to the PR metadata issues github was having last week.

neutrinoceros:
Great ! I was worried this was a deliberate practice that was also completely foreign to me. Just curious, how did you fix it ?

It would probably require a new release of `s3fs` including my patch, but otherwise I think this should be stable now (no rush).

I should let you know btw, my end goal would be to add missing lower bounds on fsspec's requirements, so that packages requiring it (including but not limited to astropy) could reliably test against oldest supported version of all their dependencies (including transitive ones). In order to achieve this in a reliable way, I'd also need to add a job to exercise envs resolved to oldest versions here, which is only convenient to do with uv instead of pip. Is this something you'd be interested in ?

neutrinoceros:
@martindurant is there a timeline for the next release of `s3fs` ?

martindurant:
> timeline for the next release of s3fs

Within the next two week, unless there's a good reason to go sooner.

neutrinoceros:
no rush on my side. Thanks for the quick reply !

neutrinoceros:
with `s3fs>=2026.6.0`, this is now stable

PR #2037, opened 2026-05-07
Do not strip trailing slashes from data URIs

joouha:
Trailing slashes are incorrectly stripped from data URIs, resulting in incorrect decoding.

The fix is to override the `_strip_protocol` class method for the `DataFileSystem` to not strip trailing slashes.

Without the fix:
```python
>>> from fsspec.core import url_to_fs
>>> from base64 import b64encode
>>> value = b"ab?"
>>> uri = "data:text/plain;base64," + b64encode(value)
>>> uri
'data:text/plain;base64,YWI/'
>>> fs, fs_url = url_to_fs(uri)
>>> fs_url
... # Trailing slash is stripped
'data:text/plain;base64,YWI'
>>> fs.cat(fs_url)
... # Value is wrong
b'ab' 
>>> fs.cat("data:text/plain,/")
''
```

With the fix:
```python
>>> from fsspec.core import url_to_fs
>>> fs, fs_uri = url_to_fs("data:text/plain;base64,YWI/")
>>> fs.cat(fs_uri)
b'ab?'
>>> fs.cat("data:text/plain,/")
'/'
```

### DIFF
--- a/fsspec/implementations/data.py
+++ b/fsspec/implementations/data.py
@@ -3,6 +3,7 @@ import io
 from urllib.parse import unquote
 
 from fsspec import AbstractFileSystem
+from fsspec.utils import stringify_path
 
 
 class DataFileSystem(AbstractFileSystem):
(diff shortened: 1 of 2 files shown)


martindurant:
Thanks for this. Please include your test case too.

joouha:
Done!

martindurant:
+1

PR #2041, opened 2026-05-26
Type annotations for base spec

### LINKED ISSUE #625: Type annotations?

slmnsh:
Fixes #625

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -50,6 +50,10 @@ jobs:
           pip install -e .[test_full]
           pytest -v
 
+      - name: Run Stub Tests
+        shell: bash -l {0}
+        run: stubtest fsspec.spec
+
   win:
     name: pytest-win
     runs-on: windows-2022
--- a/README.md
+++ b/README.md
@@ -81,6 +81,11 @@ run and corresponding environment file run a set of tests from the dask
 test suite, and very minimal tests against pandas and zarr from the
 test_downstream.py module in this repo.
 
+### Type Hints
+
+fsspec uses type stub files for type hints instead of inline hints. Use ``stubtest <module>``
+to verify if type stubs are in sync with implementation.
+
 ### Code Formatting
 
 fsspec uses [Black](https://black.readthedocs.io/en/stable) to ensure
--- a/fsspec/spec.pyi
+++ b/fsspec/spec.pyi
@@ -1,0 +1,482 @@
+# pyright: reportExplicitAny=none, reportAny=none
+
+import io
+import os
+from collections.abc import Callable, Generator
+from pathlib import Path
+from typing import IO, Any, ClassVar, Literal, Self, TypeAlias, TypedDict, overload
+
+from _typeshed import Incomplete, OpenBinaryMode, OpenTextMode
+
+from fsspec.mapping import FSMap
+
+from .callbacks import DEFAULT_CALLBACK as DEFAULT_CALLBACK
+from .callbacks import Callback
+from .config import conf as conf
+from .dircache import DirCache as DirCache
+from .transaction import Transaction as Transaction
+from .utils import isfilelike as isfilelike
+from .utils import other_paths as other_paths
+from .utils import read_block as read_block
+from .utils import stringify_path as stringify_path
+from .utils import tokenize as tokenize
+
+logger: Incomplete
+
+TPath: TypeAlias = str | os.PathLike[str] | Path
+
+class FileInfo(TypedDict):
+    name: str
+    size: int | None
+    type: Literal["file", "directory"] | str  # noqa: PYI051
+
+def make_instance(cls, args, kwargs: dict[str, Any]): ...
+
+class _Cached(type):
+    def __init__(cls, *args, **kwargs: Any) -> None: ...
+    def __call__(cls, *args, **kwargs: Any): ...
+
+class AbstractFileSystem(metaclass=_Cached):
+    cachable: bool
+    blocksize: Incomplete
+    sep: str
+    protocol: ClassVar[str | tuple[str, ...]]
+    async_impl: bool
+    mirror_sync_methods: bool
+    root_marker: str
+    transaction_type: type[Transaction]
+    storage_args: tuple[Any, ...]
+    storage_options: dict[str, Any]
+    dircache: Incomplete
+    def __init__(self, *args, **storage_options) -> None: ...
+    @property
+    def fsid(self) -> None: ...
+    def __dask_tokenize__(self): ...
+    def __hash__(self): ...
+    def __eq__(self, other): ...
+    def __reduce__(self): ...
+    def unstrip_protocol(self, name: str) -> str: ...
+    @classmethod
+    def current(cls) -> Self: ...
+    @property
+    def transaction(self) -> Transaction: ...
+    def start_transaction(self) -> Transaction: ...
+    def end_transaction(self) -> None: ...
+    def invalidate_cache(self, path: str | None = None) -> None: ...
+    def mkdir(
+        self, path: TPath, create_parents: bool = True, **kwargs: Any
+    ) -> None: ...
+    def makedirs(self, path: TPath, exist_ok: bool = False) -> None: ...
+    def rmdir(self, path: TPath) -> None: ...
+    @overload
+    def ls(
+        self, path: TPath, detail: Literal[True], **kwargs: Any
+    ) -> list[FileInfo]: ...
+    @overload
+    def ls(self, path: TPath, detail: Literal[False], **kwargs: Any) -> list[str]: ...
+    def ls(
+        self, path: TPath, detail: bool = True, **kwargs: Any
+    ) -> list[str] | list[FileInfo]: ...
+    def walk(
+        self,
+        path: TPath,
+        maxdepth: int | None = None,
+        topdown: bool = True,
+        on_error: Literal["omit", "raise"] | Callable[[OSError], Any] = "omit",
+        **kwargs: Any,
+    ) -> Generator[Incomplete, Incomplete]: ...
+    @overload
+    def find(
+        self,
+        path: TPath,
+        maxdepth: int | None,
+        withdirs: bool,
+        detail: Literal[True],
+        **kwargs: Any,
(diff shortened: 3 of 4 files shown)


fsspec/spec.py
```diff
@@ -389,7 +427,40 @@ def _ls_from_cache(self, path):
...
 
-    def walk(self, path, maxdepth=None, topdown=True, on_error="omit", **kwargs):
+    @overload
+    def walk(
+        self,
+        path: TPath,
+        maxdepth: int | None,
+        topdown: bool,
+        on_error: Literal["omit", "raise"] | Callable[[OSError], Any],
+        detail: Literal[True],
```
slmnsh: does it makes sense to make detail kwarg explicit (with default false) for all the functions that takes it?

fsspec/spec.py
```diff
@@ -1354,7 +1504,7 @@ def open(
                 **text_kwargs,
             )
         else:
-            ac = kwargs.pop("autocommit", not self._intrans)
+            ac: bool = kwargs.pop("autocommit", not self._intrans)
```
slmnsh: I would like to make these kwargs explicit with default `None`. What do you think?

fsspec/spec.py
```diff
@@ -33,6 +37,15 @@ def make_instance(cls, args, kwargs):
     return cls(*args, **kwargs)
 
 
+TPath = str | os.PathLike[str] | Path
```
slmnsh: I saw that fsspec/utils.py has just `str` for paths. Should I switch to just `str`?

fsspec/spec.py
slmnsh, line 129: Should I initialize these to empty tuple/dict? I am getting error on extended classes that they need initialize these

fsspec/spec.py
martindurant: The pattern is to accept all the types in methods that users see, but only handle str internally - convert everything as soon as possible.

fsspec/spec.py
```diff
@@ -100,24 +114,27 @@ def __call__(cls, *args, **kwargs):
...
-    async_impl = False
-    mirror_sync_methods = False
-    root_marker = ""  # For some FSs, may require leading '/' or other character
-    transaction_type = Transaction
+    _latest: str | None = None
+    _intrans: bool = False
+    _transaction: Transaction | None = None
+    _invalidated_caches_in_transaction: list[str]
+    async_impl: bool = False
+    mirror_sync_methods: bool = False
```
martindurant: These are class attributes

fsspec/spec.py
martindurant: No, the metaclass does that

fsspec/spec.py
```diff
@@ -285,7 +312,7 @@ def invalidate_cache(self, path=None):
         if self._intrans:
             self._invalidated_caches_in_transaction.append(path)
 
-    def mkdir(self, path, create_parents=True, **kwargs):
+    def mkdir(self, path: TPath, create_parents: bool = True, **kwargs: Any):
```
martindurant: Annotating `Any` doesn't really help...

fsspec/spec.py
```diff
@@ -319,11 +346,21 @@ def makedirs(self, path, exist_ok=False):
...
+    def ls(
+        self, path: TPath, detail: Literal[True], **kwargs: Any
+    ) -> list[FileInfo]: ...
+
+    @overload
+    def ls(self, path: TPath, detail: Literal[False], **kwargs: Any) -> list[str]: ...
+
+    def ls(
+        self, path: TPath, detail: bool = True, **kwargs: Any
+    ) -> list[str] | list[FileInfo]:
```
martindurant: I realise that you are trying to specify: if detail is True, returns list[dict]; if False returns list[str]. Does it really take all this code to do so?

fsspec/spec.py
```diff
@@ -29,10 +34,19 @@
...
     return cls(*args, **kwargs)
 
 
+TPath = str | os.PathLike[str] | Path
+
+
+class FileInfo(TypedDict):
+    name: str
+    size: int | None
+    type: Literal["file", "directory"]
```
martindurant: In general, file info may have many other fields.
The "type" is not limited to those two, there are other possibilities.

fsspec/spec.py
```diff
@@ -389,7 +426,40 @@ def _ls_from_cache(self, path):
...
+        self,
+        path: TPath,
+        maxdepth: int | None,
+        topdown: bool,
+        on_error: Literal["omit", "raise"] | Callable[[OSError], Any],
+        detail: Literal[False],
+        **kwargs: Any,
+    ) -> Generator[tuple[str, list[str], list[str]], None]: ...
+
+    def walk(
```
martindurant: Again, I just find the `@overload`s really hard to read.

fsspec/spec.py
```diff
@@ -389,7 +426,40 @@ def _ls_from_cache(self, path):
...
+        **kwargs: Any,
+    ) -> Generator[tuple[str, dict[str, FileInfo], dict[str, FileInfo]], None]: ...
+
+    @overload
+    def walk(
+        self,
+        path: TPath,
+        maxdepth: int | None,
+        topdown: bool,
+        on_error: Literal["omit", "raise"] | Callable[[OSError], Any],
```
martindurant: on_error probably should be standardised to {"ignore", "return", "raise"}

fsspec/spec.py
martindurant: The point was, not all subclasses support autocommit - so they would have to have their methods changed to also have the explicit kwarg and then ignore it.

fsspec/spec.py
```diff
@@ -1855,20 +2023,28 @@ class AbstractBufferedFile(io.IOBase):
...
-        size=None,
-        **kwargs,
+        fs: AbstractFileSystem,
+        path: str,
+        mode: str = "rb",
+        block_size: int | Literal["default"] | None = "default",
+        autocommit: bool = True,
+        cache_type: Literal["readahead", "none", "mmap", "bytes"] = "readahead",
+        cache_options: dict[str, Any] | None = None,
+        size: int | None =None,
```
martindurant: General question: why is it useful to list the attributes above and also type them here, conveying the information twice?

fsspec/spec.py
```diff
@@ -1855,20 +2023,28 @@ class AbstractBufferedFile(io.IOBase):
...
-        cache_type="readahead",
-        cache_options=None,
-        size=None,
-        **kwargs,
+        fs: AbstractFileSystem,
+        path: str,
+        mode: str = "rb",
+        block_size: int | Literal["default"] | None = "default",
+        autocommit: bool = True,
+        cache_type: Literal["readahead", "none", "mmap", "bytes"] = "readahead",
```
martindurant: This list is not exhaustive, more cache types may be registered at runtime (see the "prefetcher" cache in gcsfs and "known_parts" cache in fsspec.parquet; also "all" is missing).

fsspec/spec.py
```diff
@@ -1902,7 +2078,7 @@ def __init__(
         self.fs = fs
         self.mode = mode
         self.blocksize = (
-            self.DEFAULT_BLOCK_SIZE if block_size in ["default", None] else block_size
+            self.DEFAULT_BLOCK_SIZE if block_size is None or block_size == "default" else block_size
```
martindurant: Why this change?

fsspec/spec.py
```diff
@@ -1954,6 +2130,10 @@ def details(self, value):
     def full_name(self):
         return _unstrip_protocol(self.path, self.fs)
 
+    @property
+    def name(self):
+        return _unstrip_protocol(self.path, self.fs)
```
martindurant: Where does this new property come from?

fsspec/spec.py
```diff
@@ -1994,11 +2174,12 @@ def info(self):
         else:
             raise ValueError("Info not available while writing")
 
+    @override
```
martindurant: What does this override?

martindurant (commented):
Some thoughts.

My biggest gripe is the extended overrides, mostly because of detail=True|False.

fsspec/spec.py
slmnsh: `tell` method from `io.IOBase`

fsspec/spec.py
slmnsh: because type checker can't narrow types with `in` operator.

slmnsh:
@martindurant Thanks for review! I think I should've started with type stubs then later maybe add inline types. type stubs will be cleaner to add than inline types.

I will push type stubs this weekend.

fsspec/spec.py
slmnsh: Yeah but without it the type is `Unknown` and mypy raises warning for that

fsspec/spec.py
slmnsh: sadly yes. python doesn't have proper type manipulation support like typescript [yet](https://peps.python.org/pep-0827/).

fsspec/spec.py
slmnsh: then we need to change the implementation as well

slmnsh:
@martindurant I have added completely separate type stub file. I am planning to add more for known implementations once you review this

fsspec/spec.py
slmnsh: these are two separate typehints one for class members and one for constructor function. python type system can't infer it from constructor function types

martindurant:
*Much* happier with the annotations separate :). Now the main code is not affected, nor even import times. Do you have any thoughts on how best to keep the code and types in sync in the future?

slmnsh:
mypy has `stubtest` for testing stub maybe we can incorporate that

https://mypy.readthedocs.io/en/stable/stubtest.html#automatic-stub-testing-stubtest

martindurant:
Can you please draft a non-required CI run with that, and text instructions for how to go about updating types in the future as the API changes?

slmnsh:
@martindurant Sorry for delays. I added "Run Stub Tests" step for running stub test it will raise error if failed and instructions to readme.

PR #2042, opened 2026-05-31
Prevent attribute error by checking for 'forced' attribute before flushing cache

TLCFEM:
This is not a critical bug; however, if a `NotImplementedError` is raised during initialization due to an invalid mode, attributes such as forced are not created.
Consequently, subsequent cleanup operations (e.g., close or delete) may raise an `AttributeError`.
While this exception is typically suppressed, it may still be exposed in some scenarios.
This PR adds safeguards to prevent such cases and avoid unnecessary visual clutter.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -2222,11 +2222,7 @@ class AbstractBufferedFile(io.IOBase):
             if self.mode == "rb":
                 self.cache = None
             else:
-                if not hasattr(self, "forced"):
-                    # the field is not available if NotImplementedError is thrown in constructor
-                    return
-
-                if not self.forced:
+                if not getattr(self, "forced", True):
                     self.flush(force=True)
 
                 if self.fs is not None:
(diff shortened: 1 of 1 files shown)


martindurant:
Perhaps have `forced` be set early in the init function, or even have a backup class attribute?

TLCFEM:
Fixed, shall be Okay now.

TLCFEM:
Looks like early initialization would cause some logical error elsewhere.

fsspec/spec.py
```diff
@@ -2222,6 +2222,10 @@ def close(self):
             if self.mode == "rb":
                 self.cache = None
             else:
+                if not hasattr(self, "forced"):
+                    # the field is not available if NotImplementedError is thrown in constructor
+                    return
+
                 if not self.forced:
```
martindurant:
```suggestion
                if not getattr(self, "forced", True):
```

would be a simpler way to express the original code

PR #2044, opened 2026-06-05
Forward kwargs from cat_ranges to cat_file (#2016)

### LINKED ISSUE #2016: cat_ranges doesn't pass kwargs to cat_file which  generates block size mismatches which result in no cache usage
I'm trying to use fsspec with non-default block-sizes for a local disk cache and access to remote s3/https resources. It seems that when `cat_ranges` is not used, all works as expected, but when using`cat_ranges`, the cache is bypassed, as cat_ranges always goes out to the network. 

If I've understood the problem correctly,  ``AbstractFileSystem.cat_ranges`` accepts ``**kwargs`` but does not forward them when calling ``cat_file``:

[fsspec/spec.py](https://github.com/fsspec/filesystem_spec/blob/2b5ed0faa8b496b988f7e085aea725c0dbd52a05/fsspec/spec.py#L874): 
```python
 out.append(self.cat_file(p, s, e))   # <-- kwargs silently dropped
```

The consequence is that when ``CachingFileSystem._open`` uses ``block_size`` to look up blocks in its
on-disk cache and ``block_size`` is missing the call falls back to the default value, which (in my case) does not match the block size used to populate the cache, so every ``cat_ranges`` call fetches from the underlying (remote) filesystem instead of reading locally.

## Solution

Locally patching `AbstractFileSystem.cat_ranges` to call `self.cat_file(p, s, e, **kwargs)` restores the  expected cache reuse in our real integration case. I'm not brave enough to deliver a pull request with all your testing environment, but I hope that'd be trivial for you to check.

## Environment

- Python: 3.12.13 (conda-forge, macOS arm64)
- fsspec: 2026.3.0
- s3fs: 2026.3.0

## MWE

I've struggled to demonstrate this in a useful MWE, but in the spirit of trying to provide something, I got my LLM to generate the following which at least demonstrates what is obvious from the line of code above, that the kwargs are dropped ... 

```python
from __future__ import annotations
from fsspec.spec import AbstractFileSystem

class _SpyFS(AbstractFileSystem):
    """Minimal AbstractFileSystem that records kwargs received by cat_file."""

    protocol = "spy"

    def __init__(self):
        # Skip AbstractFileSystem.__init__ to avoid registration side-effects.
        self._kwargs_received: dict | None = None

    def cat_file(self, path, start=None, end=None, **kwargs):
        self._kwargs_received = kwargs
        return b"data"

def main() -> None:
    fs = _SpyFS()

    # Call cat_ranges with an explicit block_size kwarg.
    block_size = 2 * 1024 * 1024
    fs.cat_ranges(["dummy"], [0], [10], block_size=block_size)

    received = fs._kwargs_received
    print(f"kwargs passed to cat_file: {received}")

if __name__ == "__main__":
    main()
```

assinscreedFC:
## Summary

Fixes #2016.

`AbstractFileSystem.cat_ranges` accepts `**kwargs` but did not forward them when calling `cat_file`:

```python
out.append(self.cat_file(p, s, e))   # kwargs silently dropped
```

As reported, this means parameters such as a cache's `block_size` are lost, so `cat_ranges` bypasses an on-disk cache and goes out to the underlying (remote) filesystem on every call. The async `AsyncFileSystem._cat_ranges` already forwards `**kwargs` to `_cat_file`, so this just brings the sync path in line.

@martindurant confirmed in the issue: *"I'd be happy to see kwargs passed through as you suggest for part 1 of this issue."* This PR is scoped to that part 1.

## Change

```python
out.append(self.cat_file(p, s, e, **kwargs))
```

## Test plan

- Added `test_cat_ranges_forwards_kwargs` in `fsspec/tests/test_spec.py` (a recording filesystem asserts the kwarg reaches `cat_file`).
- `pytest fsspec/tests/test_spec.py` — 120 passed.
- `ruff check` / `ruff format --check` — clean.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -874,7 +874,7 @@ class AbstractFileSystem(metaclass=_Cached):
         out = []
         for p, s, e in zip(paths, starts, ends):
             try:
-                out.append(self.cat_file(p, s, e))
+                out.append(self.cat_file(p, s, e, **kwargs))
             except Exception as e:
                 if on_error == "return":
                     out.append(e)
--- a/fsspec/tests/test_spec.py
+++ b/fsspec/tests/test_spec.py
@@ -1456,3 +1456,21 @@ def test_expand_path_with_magic_input():
         "bucket/file?.txt",
     ]
     assert sorted(paths) == sorted(expected)
+
+
+def test_cat_ranges_forwards_kwargs():
+    # See GH#2016: cat_ranges must forward **kwargs to cat_file, otherwise
+    # parameters such as a cache's block_size are silently dropped.
+    received = []
+
+    class RecordingFS(AbstractFileSystem):
+        cachable = False
+
+        def cat_file(self, path, start=None, end=None, **kwargs):
+            received.append(kwargs)
+            return b""
+
+    fs = RecordingFS()
+    fs.cat_ranges(["a", "b"], [0, 0], [1, 1], block_size=42)
+
+    assert received == [{"block_size": 42}, {"block_size": 42}]


martindurant:
Thank you!

bnlawrence:
Thanks from me too!!

PR #2056, opened 2026-06-22
ftp: Make deprecated TLS versions optional

### LINKED ISSUE #2053: PROTOCOL_TLSv* are deprecated (and possibly removed)
As seen [here in the Python docs](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1), these are all deprecated:
https://github.com/fsspec/filesystem_spec/blob/2a2416a75ebc09debe708b5a5435d39871b55e3b/fsspec/implementations/ftp.py#L10-L16

OpenSSL 4 removed these modes, and so Python compiled against it doesn't have these constants at all.

This means that any use of `FTPFileSystem` will fail with a hard error on import, even if these options aren't used at runtime.

QuLogic:
When built against OpenSSL 4, Python 3.15 will not have any of these deprecated symbols.

Fixes #2053

### DIFF
--- a/fsspec/implementations/ftp.py
+++ b/fsspec/implementations/ftp.py
@@ -9,11 +9,11 @@ from ..utils import infer_storage_options, isfilelike
 
 SECURITY_PROTOCOL_MAP = {
     "tls": ssl.PROTOCOL_TLS,
-    "tlsv1": ssl.PROTOCOL_TLSv1,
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):
Thanks for adding support for more recent python versions. When I add this feature I've just ported what it was on a python 3.6 lib, so it makes sense.

👍

martindurant:
Simple, effective

PR #2058, opened 2026-06-24
fix(glob): discard partial dircache entry after prefix-filtered _find

bruno-hays:
## Problem
`_glob` passes a `prefix=` hint to `_find` so backends (s3fs, gcsfs, adlfs) can filter server-side up to the first wildcard. Those backends store the filtered result in `dircache` under the parent directory key — but the entry contains only prefix-matching files, making it a partial (misleading) directory listing.
Consequence: after `glob("dir/train-*")`, a call to `glob("dir/test-*")` or `fs.exists("dir/test-file")` hits the stale partial entry and silently returns nothing / `False`, even though the files exist.
This regression was introduced in 2026.4.0 by #1996.
## Fix
After `_find` returns with a prefixed query, discard the dircache entry for `root`. The result is already captured in `allpaths` so nothing is lost; the next lookup for the same directory fetches a fresh full listing.
```python
if prefix and root:
    self.dircache.pop(root.rstrip("/"), None)

### DIFF
--- a/fsspec/asyn.py
+++ b/fsspec/asyn.py
@@ -871,9 +871,18 @@ class AsyncFileSystem(AbstractFileSystem):
         # gcsfs, s3fs and adlfs can filter server-side up to the first wildcard.
         if prefix:
             kwargs["prefix"] = prefix
+        root_key = root.rstrip("/")
+        pre_cached = prefix and root and root_key in self.dircache
         allpaths = await self._find(
             root, maxdepth=depth, withdirs=withdirs, detail=True, **kwargs
         )
+        # Backends store only prefix-matching files in dircache when prefix= is
+        # used, so a newly created entry is a partial listing that would mislead
+        # subsequent lookups.  Only discard it when _find created it; if the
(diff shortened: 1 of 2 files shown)


martindurant:
I had not anticipated such a solution! Does this miss the case that the dircache was already populated before calling find(), so we end up discarding legitimate listings?

martindurant:
Seems to be causing explicit failures

bruno-hays:
I fixed the test, sorry for oversight.
We indeed did discard existing cache, I added a guard.
After checking the code, it seems like only s3fs is affected by this bug as gcsfs does not cache prefixed listings and adlfs does not write to dircache at all.
I still think it's worth fixing in fsspec as it is where the issue originated and prevents the same error from arising in another backend, and I'm not 100% sure either for other backends anyway.
And it's worth fixing in s3fs as there could be other functions that suffer from the same bug

bruno-hays:
Here is my proposed fix for s3fs: https://github.com/fsspec/s3fs/pull/1034

PR #2060, opened 2026-06-25
ci: use trusted publishing in release pipeline and trigger on tags

### LINKED ISSUE #1866: Switch to trusted publishing for package upload to PyPI in CI
Trusted publishing (with attestations) means I can know for certain that what I download from PyPI is the same artefact which was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing (rather than having to manually review all of the installed files on each release).

See [the Python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing), [the PyPI documentation](https://docs.pypi.org/trusted-publishers/), and [the official pypi-publish GitHub action documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) on trusted publishing - you'll need to configure an environment in PyPI and GitHub. You will be able to remove the PYPI_PASSWORD project secret.

Should be as simple as switching to the pypa/gh-action-pypi-publish action (instead of `twine upload dist/*`) in the deploy job of the [pypipublish workflow](https://github.com/fsspec/filesystem_spec/blob/master/.github/workflows/pypipublish.yaml), and adding environment and permissions to that job.

Yonghui-Lee:
This PR makes release pipeline automatically run tests, build packages, create GitHub releases, and publish to PyPI upon a new tag push.

Close #1866

### Key Changes

1. **CI Workflow** (main.yaml):
   - Added the `workflow_call` trigger to allow the CI workflow to be called by the release workflow.
2. **Release Pipeline** (pypipublish.yaml):
   - **Trigger**: Configured to trigger automatically on CalVer tag pushes (e.g., `2026.6.0`).
   - **CI Dependency**: Added a `ci` job dependency to ensure all tests pass before any build or publish steps run.
   - **Build**: Uses the standard `build` module to generate source distributions and wheels.
   - **GitHub Release**: Automatically creates a GitHub release, generates release notes, and attaches the build artifacts.
   - **PyPI Publish**: Switched from legacy username/password secrets to **Trusted Publishing (OIDC)** using the official `pypa/gh-action-pypi-publish` action.

### Setup Required

Trusted Publishing must be configured on PyPI.

### DIFF
--- a/.github/workflows/pypipublish.yaml
+++ b/.github/workflows/pypipublish.yaml
@@ -1,14 +1,25 @@
-name: Upload Python Package
+name: Release Python Package
 
+# Triggered by a tag push, this workflow creates a GitHub release
+# and publishes the package to PyPI.
 on:
-  release:
-    types: [created]
+  push:
+    tags:
+      - '2[0-9][0-9][0-9].[0-9].[0-9]*'
+      - '2[0-9][0-9][0-9].[0-9][0-9].[0-9]*'
 
 jobs:
-  deploy:
+  ci:
+    uses: ./.github/workflows/main.yaml
+
+  build:
+    name: Build artifacts
+    needs: ci
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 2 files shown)


Yonghui-Lee:
Hi @martindurant. Could you help take a look at this PR? The release process has been validated in gcsfs. We also want to automate the release process of fsspec. Based on this, we can also Implement a mechanism to gate fsspec releases to ensure compatibility with E2E tests.

martindurant:
OK, I suppose it's fine. Will need to do s3fs too, but we can let that wait one more release, since this one is different enough from gcsfs's.

martindurant:
Publisher added on pypi

.github/workflows/pypipublish.yaml
```diff
@@ -1,26 +1,72 @@
...
+          generate_release_notes: true
+          files: dist/*
+
+  publish:
+    name: Upload release to PyPI
+    needs: release
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    environment: pypi
```
EpicWink, line 64:
```suggestion
    environment:
      name: pypi
      url: https://pypi.org/p/fsspec
```

Including URL adds a link to the [deployments page](https://github.com/fsspec/filesystem_spec/deployments/pypi)

PR #2062, opened 2026-07-01
MemoryFileSystem: create the file when appending to a missing path

Sreekant13:
Opening a path that doesn't exist yet in append mode behaves differently on MemoryFileSystem than on LocalFileSystem and the builtin `open`:

```python
import fsspec
fsspec.filesystem("file").open("/tmp/new.txt", "ab")   # creates the file
fsspec.filesystem("memory").open("/new.txt", "ab")     # FileNotFoundError
```

open(path, "ab") creates the file if it's missing, and LocalFileSystem inherits that, so having the memory backend raise FileNotFoundError here is surprising. This makes _open create an empty file when the path isn't in the store yet for append modes (a/ab/a+b), then position at the end as before. Read modes (rb/r+b) are unchanged and still raise.

Added a regression test covering the three append modes.

### DIFF
--- a/fsspec/implementations/memory.py
+++ b/fsspec/implementations/memory.py
@@ -197,6 +197,13 @@ class MemoryFileSystem(AbstractFileSystem):
                     # position at the beginning of file
                     f.seek(0)
                 return f
+            elif "a" in mode:
+                # append modes create the file if it does not exist, matching
+                # builtin open() and LocalFileSystem
+                m = MemoryFile(self, path, kwargs.get("data"))
+                if not self._intrans:
+                    m.commit()
+                return m
             else:
                 raise FileNotFoundError(path)
         elif mode in {"wb", "w+b", "xb", "x+b"}:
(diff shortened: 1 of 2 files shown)


fsspec/implementations/memory.py
```diff
@@ -197,6 +197,13 @@ def _open(
                     # position at the beginning of file
                     f.seek(0)
                 return f
+            elif "a" in mode:
+                # append modes create the file if it does not exist, matching
+                # builtin open() and LocalFileSystem
+                m = MemoryFile(self, path, kwargs.get("data"))
+                if not self._intrans:
+                    m.commit()
```
martindurant, line 205: Should make sure to seek to the end of the file. In the test, it is already at the end, but this is not guaranteed.

Sreekant13:
Good point @martindurant, thanks! Pushed a commit that seeks to the end explicitly after creating the file, so it matches the existing-file append path and no longer relies on the new file being empty. Also updated the regression test to assert the position is at the end.

PR #2064, opened 2026-07-09
fix: Make Cached metaclass instance instantiation thread-safe and fork-safe

### LINKED ISSUE #2065: Concurrent Instantiation bypasses Cache across Threaded Forks
## Description
When running `gcsfs` in highly concurrent environments—specifically utilizing `multiprocessing` processes alongside internal multithreading—the underlying `fsspec` cache logic cannot handle when multiple threads try to instantiate concurrently.

If a Python process forks (e.g., via `multiprocessing.Pool`), the `fsspec` child process clears its inherited instance cache (`os.getpid() != cls._pid`). However, because this cache-clearing check and instantiation process was not thread-safe:
1. All threads evaluate the newly empty cache post-fork.
2. They bypass the cache concurrently.
3. Every single overlapping thread instantiates a new, fully independent `GCSFileSystem` instance and overwrites the local cache.

This leads to significant redundant filesystem instantiations and associated overhead due to multiple parallel clients standing up unnecessarily. 

## Proposed Solution
We have developed a complete patch and will be landing a Pull Request in `fsspec` that introduces:
1. Double-checked `threading.Lock` locking directly inside `_Cached.__call__`.
2. Safe boundaries for `os.getpid() != cls._pid` post-fork cache clearing.
3. Fixing `tokenize()` dependencies on PIDs prior to evaluating the cache.

yuxin00j:
Fix #2065
### Scope
This PR tackles a race condition in the `_Cached` metaclass.

### The Problem
There were three distinct overlapping lock gaps around system forks:
1. **Unprotected Instantiation**: The `super().__call__` instantiation process inside `_Cached` was entirely unprotected, allowing concurrent cache misses to spin up disjoint raw instances independently.
2. **Post-fork Cache Clearing Race**: When a process forks, `os.getpid() != cls._pid` attempts to clear the inherited cache. When 10 new threads simultaneously wake up post-fork and evaluate this, they concurrently bypass the cache logic.
3. **Ghost Tokenization**: `token = tokenize(cls, cls._pid, ...)` historically evaluated before any locking boundary. The very first thread running post-fork evaluated `tokenize()` utilizing the lingering *parent's* PID before updating `cls._pid` inside the lock. Subsequence threads would then evaluate `tokenize()` natively under the child's PID, resulting in permanently ghosted, duplicated Cache hashes for identically parameterized filesystems.

### The Solution
*   **Double-Checked Locking**: Re-instated a standard `_instantiation_lock` in `_Cached.__call__` that strictly gates the underlying filesystem instantiation and cache updates. We also updated `clear_instance_cache()` to respect this exact lock.
*   **Encapsulated Fork Boundaries**: Moved the `os.getpid() != cls._pid` cache clearing evaluation strictly inside the boundaries of the lock. Furthermore, we explicitly re-initialize `cls._instantiation_lock = threading.RLock()` across forks so child processes do not natively inherit and permanent-deadlock on parent-locked states.
*   **Deterministic Tokenizing**: Altered `tokenize()` targeting to natively utilize `os.getpid()` rather than evaluating across `cls._pid`, ensuring deterministic key hashes pre-and-post fork regardless of strict thread scheduling.
*   **Testing**: Added `test_instance_cache_concurrency`, `test_clear_instance_cache_concurrency`, and `test_fork_deadlock` heavily simulating asynchronous thread congestion across cache misses, explicit cache-clear injections, and native OS forks to guarantee locking determinism.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -76,6 +76,12 @@ class _Cached(type):
 
             os.register_at_fork(after_in_child=_reset_lock)
 
+    def _check_cache(cls, token):
+        inst = cls._cache.get(token)
+        if inst is not None:
+            cls._latest = token
+        return inst
+
     def __call__(cls, *args, **kwargs):
         kwargs = apply_config(cls, kwargs)
         extra_tokens = tuple(
@@ -96,9 +102,8 @@ class _Cached(type):
 
         if not skip and cls.cachable:
             if pid == cls._pid:
-                inst = cls._cache.get(token)
+                inst = cls._check_cache(token)
                 if inst is not None:
-                    cls._latest = token
                     return inst
 
             with cls._instantiation_lock:
@@ -106,9 +111,8 @@ class _Cached(type):
                     cls._cache.clear()
                     cls._pid = pid
 
-                inst = cls._cache.get(token)
+                inst = cls._check_cache(token)
                 if inst is not None:
-                    cls._latest = token
                     return inst
 
         obj = super().__call__(*args, **kwargs, **strip_tokenize_options)
@@ -123,13 +127,8 @@ class _Cached(type):
 
         if cls.cachable and not skip:
             with cls._instantiation_lock:
-                if pid != cls._pid:
-                    cls._cache.clear()
-                    cls._pid = pid
-
-                inst = cls._cache.get(token)
+                inst = cls._check_cache(token)
                 if inst is not None:
-                    cls._latest = token
                     return inst
 
                 cls._latest = token
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -95,37 +95,47 @@ class _Cached(type):
             )
         skip = kwargs.pop("skip_instance_cache", False)
 
-        if not skip and cls.cachable and pid == cls._pid:
-            inst = cls._cache.get(token)
-            if inst is not None:
-                cls._latest = token
-                return inst
-
-        with cls._instantiation_lock:
-            if pid != cls._pid:
-                cls._cache.clear()
-                cls._pid = pid
-
-            if not skip and cls.cachable:
+        if not skip and cls.cachable:
+            if pid == cls._pid:
                 inst = cls._cache.get(token)
                 if inst is not None:
                     cls._latest = token
                     return inst
 
-            obj = super().__call__(*args, **kwargs, **strip_tokenize_options)
-            # Setting _fs_token here causes some static linters to complain.
-            obj._fs_token_ = token
-            obj.storage_args = args
-            obj.storage_options = kwargs
-            if obj.async_impl and obj.mirror_sync_methods:
-                from .asyn import mirror_sync_methods
+            with cls._instantiation_lock:
+                if pid != cls._pid:
+                    cls._cache.clear()
+                    cls._pid = pid
 
-                mirror_sync_methods(obj)
+                inst = cls._cache.get(token)
+                if inst is not None:
+                    cls._latest = token
+                    return inst
 
-            if cls.cachable and not skip:
+        obj = super().__call__(*args, **kwargs, **strip_tokenize_options)
+        # Setting _fs_token here causes some static linters to complain.
+        obj._fs_token_ = token
+        obj.storage_args = args
+        obj.storage_options = kwargs
+        if obj.async_impl and obj.mirror_sync_methods:
+            from .asyn import mirror_sync_methods
+
+            mirror_sync_methods(obj)
+
+        if cls.cachable and not skip:
+            with cls._instantiation_lock:
+                if os.getpid() != cls._pid:
+                    cls._cache.clear()
+                    cls._pid = os.getpid()
+
+                inst = cls._cache.get(token)
+                if inst is not None:
+                    cls._latest = token
+                    return inst
+
                 cls._latest = token
                 cls._cache[token] = obj
-            return obj
+        return obj
 
 
 class AbstractFileSystem(metaclass=_Cached):
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -123,9 +123,9 @@ class _Cached(type):
 
         if cls.cachable and not skip:
             with cls._instantiation_lock:
-                if os.getpid() != cls._pid:
+                if pid != cls._pid:
                     cls._cache.clear()
-                    cls._pid = os.getpid()
+                    cls._pid = pid
 
                 inst = cls._cache.get(token)
                 if inst is not None:
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -98,8 +98,8 @@ class _Cached(type):
 
         with cls._instantiation_lock:
             if pid != cls._pid:
-                cls._pid = pid
                 cls._cache.clear()
+                cls._pid = pid
 
             if not skip and cls.cachable:
                 inst = cls._cache.get(token)
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -112,10 +112,9 @@ class _Cached(type):
                     cls._pid = pid
 
         if not skip and cls.cachable:
-            if pid == cls._pid:
-                inst = cls._check_instance_cache(token)
-                if inst is not None:
-                    return inst
+            inst = cls._check_instance_cache(token)
+            if inst is not None:
+                return inst
 
             with cls._instantiation_lock:
                 inst = cls._check_instance_cache(token)
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -81,12 +81,7 @@ class _Cached(type):
             )
         skip = kwargs.pop("skip_instance_cache", False)
 
-        if (
-            not skip
-            and cls.cachable
-            and pid == cls._pid
-            and token in cls._cache
-        ):
+        if not skip and cls.cachable and pid == cls._pid and token in cls._cache:
             cls._latest = token
             return cls._cache[token]
 
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -76,7 +76,7 @@ class _Cached(type):
 
             os.register_at_fork(after_in_child=_reset_lock)
 
-    def _check_cache(cls, token):
+    def _check_instance_cache(cls, token):
         inst = cls._cache.get(token)
         if inst is not None:
             cls._latest = token
@@ -102,7 +102,7 @@ class _Cached(type):
 
         if not skip and cls.cachable:
             if pid == cls._pid:
-                inst = cls._check_cache(token)
+                inst = cls._check_instance_cache(token)
                 if inst is not None:
                     return inst
 
@@ -111,7 +111,7 @@ class _Cached(type):
                     cls._cache.clear()
                     cls._pid = pid
 
-                inst = cls._check_cache(token)
+                inst = cls._check_instance_cache(token)
                 if inst is not None:
                     return inst
 
@@ -127,7 +127,7 @@ class _Cached(type):
 
         if cls.cachable and not skip:
             with cls._instantiation_lock:
-                inst = cls._check_cache(token)
+                inst = cls._check_instance_cache(token)
                 if inst is not None:
                     return inst
 
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -6,6 +6,7 @@ import logging
 import os
 import threading
 import warnings
+import weakref
 from errno import ESPIPE
 from glob import has_magic
 from hashlib import sha256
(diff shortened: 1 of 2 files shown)


fsspec/spec.py
```diff
@@ -71,19 +72,31 @@ def __call__(cls, *args, **kwargs):
             k: kwargs.pop(k) for k in cls._strip_tokenize_options if k in kwargs
         }
         if getattr(cls, "async_impl", False) and not kwargs.get("asynchronous", False):
-            token = tokenize(cls, cls._pid, *args, *extra_tokens, **kwargs)
+            token = tokenize(cls, os.getpid(), *args, *extra_tokens, **kwargs)
```
zhixiangli: Let's calculate os.getpid() only once.

fsspec/spec.py
```diff
@@ -61,6 +61,7 @@ def __init__(cls, *args, **kwargs):
         else:
             cls._cache = {}
         cls._pid = os.getpid()
+        cls._instantiation_lock = threading.RLock()
```
zhixiangli, line 80: It may be a deadlock after forking, as a thread might fork while a lock is held by another thread.

You may follow this pattern to create a new lock per cached class after forking.

https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L157-L166

fsspec/spec.py
zhixiangli: Lock is required during the clearing cache.

yuxin00j:
Hi @martindurant, could you take a look at this PR please? Thanks.

fsspec/spec.py
yuxin00j: Done.

fsspec/spec.py
yuxin00j: Done.

fsspec/spec.py
yuxin00j: Done.

fsspec/spec.py
```diff
@@ -70,20 +80,30 @@ def __call__(cls, *args, **kwargs):
...
             )
         skip = kwargs.pop("skip_instance_cache", False)
-        if os.getpid() != cls._pid:
-            cls._cache.clear()
-            cls._pid = os.getpid()
-        if not skip and cls.cachable and token in cls._cache:
+
+        if not skip and cls.cachable and pid == cls._pid and token in cls._cache:
             cls._latest = token
             return cls._cache[token]
```
zhixiangli: If we check token in cls._cache and then access cls._cache[token] without a lock, is it possible for the token to be cleared from the cache in between those steps, causing a KeyError?

fsspec/spec.py
```diff
@@ -70,20 +80,30 @@ def __call__(cls, *args, **kwargs):
...
-        else:
+
+        with cls._instantiation_lock:
+            if pid != cls._pid:
+                cls._pid = pid
+                cls._cache.clear()
+                cls._instantiation_lock = threading.RLock()
```
zhixiangli: Drop this line?

fsspec/spec.py
```diff
@@ -70,20 +80,30 @@ def __call__(cls, *args, **kwargs):
...
+
+            if not skip and cls.cachable and token in cls._cache:
```
zhixiangli: Use _cache.get(token) and a None check to avoid a race condition.

fsspec/spec.py
```diff
@@ -61,6 +62,19 @@ def __init__(cls, *args, **kwargs):
         else:
             cls._cache = {}
         cls._pid = os.getpid()
+        cls._instantiation_lock = threading.RLock()
+
+        if hasattr(os, "register_at_fork"):
+            cls_ref = weakref.ref(cls)
```
zhixiangli: There can be multiple filesystem classes active at the same time (e.g., LocalFileSystem, MemoryFileSystem, S3FileSystem, GCSFileSystem). If we only stored a single weak reference, we would overwrite it each time a new filesystem class is defined, only keeping track of the last one initialized.

So, weakref.WeakSet() is necessary.

fsspec/spec.py
```diff
@@ -70,34 +84,57 @@ def __call__(cls, *args, **kwargs):
...
+        obj.storage_args = args
+        obj.storage_options = kwargs
+        if obj.async_impl and obj.mirror_sync_methods:
+            from .asyn import mirror_sync_methods
+
+            mirror_sync_methods(obj)
+
+        if cls.cachable and not skip:
+            with cls._instantiation_lock:
+                if pid != cls._pid:
```
zhixiangli: This `if` block can be dropped?

fsspec/spec.py
```diff
@@ -70,34 +84,57 @@ def __call__(cls, *args, **kwargs):
...
+                if inst is not None:
+                    cls._latest = token
+                    return inst
+
+            with cls._instantiation_lock:
+                if pid != cls._pid:
+                    cls._cache.clear()
+                    cls._pid = pid
+
+                inst = cls._cache.get(token)
```
zhixiangli: Could you refactor this block so that it's just implemented once?
```
                inst = cls._cache.get(token)
                if inst is not None:
                    cls._latest = token
                    return inst
```

zhixiangli (changes requested):

fsspec/spec.py
```diff
@@ -33,6 +33,20 @@ def make_instance(cls, args, kwargs):
     return cls(*args, **kwargs)
 
 
+_register_instances = weakref.WeakSet()
+
+
+def _reset_instances_lock():
+    for c in _register_instances:
+        c._instantiation_lock = threading.RLock()
+        c._cache.clear()
```
zhixiangli: Could you confirm if we can delete
```
        c._cache.clear()
        c._pid = os.getpid()

File: fsspec/spec.py (hunk 3 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Done. Refactored to use `_cache.get(token)` in `_check_instance_cache`.

File: fsspec/spec.py (hunk 4 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Done. Discarded the re-initialization of the lock here.

File: fsspec/spec.py (hunk 5 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Done.

File: fsspec/spec.py (hunk 6 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Yes, I've updated it to use a module-level `weakref.WeakSet()` to track all classes safely.

File: fsspec/spec.py (hunk 7 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Done. Dropped.

File: fsspec/spec.py (hunk 8 above)
Comment (yuxin00j) [reply fsspec/spec.py]: Done. Added `_check_instance_cache(token)`.

File: fsspec/spec.py (hunk 9 above)
Comment (yuxin00j) [reply fsspec/spec.py]: It is better to keep it because in a multi-process setup, clearing the cache eagerly in the child process via the fork hook ensures memory and file descriptors from the parent are released immediately, even if this specific class isn't instantiated again.

File: fsspec/spec.py
Comment (zhixiangli) [inline fsspec/spec.py]: Please check if those cases have been cleared.
```
with cls._instantiation_lock:
    inst = cls._cache.get(cls._latest)
    ...
````

File: fsspec/spec.py [hunk 10]
```diff
@@ -70,34 +95,50 @@ def __call__(cls, *args, **kwargs):
...
         if getattr(cls, "async_impl", False) and not kwargs.get("asynchronous", False):
-            token = tokenize(cls, cls._pid, *args, *extra_tokens, **kwargs)
+            token = tokenize(cls, pid, *args, *extra_tokens, **kwargs)
         else:
             token = tokenize(
-                cls, cls._pid, threading.get_ident(), *args, *extra_tokens, **kwargs
+                cls, pid, threading.get_ident(), *args, *extra_tokens, **kwargs
             )
         skip = kwargs.pop("skip_instance_cache", False)
-        if os.getpid() != cls._pid:
```
zhixiangli, line 80: It was unconditional clear in the original code. Let's keep it.
```
if pid != cls._pid:                                                                                   
    with cls._instantiation_lock:
        if pid != cls._pid:
            cls._cache.clear()
            cls._pid = pid

**Review by zhixiangli — APPROVED** [review]

File: fsspec/spec.py [hunk 11]
```diff
@@ -33,6 +33,20 @@ def make_instance(cls, args, kwargs):
     return cls(*args, **kwargs)
 
 
+_register_instances = weakref.WeakSet()
+
+
+def _reset_instances_lock():
+    for c in _register_instances:
+        c._instantiation_lock = threading.RLock()
```
martindurant: Fair enough that we can't use the original lock, but is this lock even useful at all once the instance has been made? I assume there's no way to have an uninstantiated instance at the time of fork.

fsspec/spec.py
```diff
@@ -33,6 +33,20 @@ def make_instance(cls, args, kwargs):
...
+        c._cache.clear()
+        c._pid = os.getpid()
+
+
+if hasattr(os, "register_at_fork"):
```
martindurant: Perhaps separate this out
```
FORK_AVAILABLE=hasattr(os, "register_at_fork")
```

I guess, if it's False, we don't need the `_register_instances` set at all.

fsspec/spec.py
```diff
@@ -70,34 +95,51 @@ def __call__(cls, *args, **kwargs):
...
+                if pid != cls._pid:
+                    cls._cache.clear()
+                    cls._pid = pid
+
+        if not skip and cls.cachable:
+            inst = cls._check_instance_cache(token)
+            if inst is not None:
+                return inst
+
+            with cls._instantiation_lock:
```
martindurant, line 120: Add comment: this is to protect against the race condition that a new instance was created and inserted into the cache since the initial check just above.

fsspec/spec.py
```diff
@@ -70,34 +95,51 @@ def __call__(cls, *args, **kwargs):
...
+        # Setting _fs_token here causes some static linters to complain.
+        obj._fs_token_ = token
+        obj.storage_args = args
+        obj.storage_options = kwargs
+        if obj.async_impl and obj.mirror_sync_methods:
+            from .asyn import mirror_sync_methods
+
+            mirror_sync_methods(obj)
+
+        if cls.cachable and not skip:
```
martindurant, line 137: How is this different from L114?

fsspec/spec.py
```diff
@@ -238,8 +280,10 @@ def current(cls):
 
         If no instance has been created, then create one with defaults
         """
-        if cls._latest in cls._cache:
-            return cls._cache[cls._latest]
+        with cls._instantiation_lock:
```
martindurant: This doesn't seem necessary? We might not have an up-to-date `_latest` occasionally (maybe?), but that's OK.

fsspec/spec.py
```diff
@@ -1614,7 +1658,12 @@ def clear_instance_cache(cls):
         since the instances refcount will not drop to zero until
         ``clear_instance_cache`` is called.
         """
-        cls._cache.clear()
+        lock = getattr(cls, "_instantiation_lock", None)
```
martindurant: Isn't `.clear()` atomic anyway?

fsspec/spec.py
martindurant: Since we are looping over instances, and there may be many instances per class, `.clear()` will be called multiple times on the same cache. Is that OK?

martindurant (commented):
A thorny problem you are solving! 
Can we really not just say "don't use fork"?

martindurant:
To check: there is no way to un-register a function passed via register-at-fork, nor even a way to list them. Is there any way the function can get registered multiple times?

fsspec/spec.py
yuxin00j: Removed the lock here.

fsspec/spec.py
yuxin00j: Done.

fsspec/spec.py
yuxin00j: Done.

fsspec/spec.py
yuxin00j: It verifies that another thread hasn't already completed instantiation and inserted the filesystem into the cache while building the current one.

fsspec/spec.py
yuxin00j: This is actually looping over the filesystem classes, not instances. Renamed _register_instances to _registered_classes to make this clear.

fsspec/spec.py
yuxin00j: Same as above. This is actually filesystem classes, not instances. Renamed it to _registered_classes.

fsspec/spec.py
yuxin00j: Removed the lock.

yuxin00j:
> A thorny problem you are solving! Can we really not just say "don't use fork"?

This pattern exists in major upstream libraries. Fixing the race condition ensures we enforce only one instance per process, allowing for safe, efficient resource sharing after a fork.

yuxin00j:
> To check: there is no way to un-register a function passed via register-at-fork, nor even a way to list them. Is there any way the function can get registered multiple times?

Re-instantiating the lock and calling .clear() should be idempotent.

PR #2071, opened 2026-07-14
Fix get_file_extension when a parent directory name contains a dot

Sreekant13:
`get_file_extension` splits on the last `.` in the entire path, so a `.` in a parent directory name is mistaken for the file's extension:

```python
>>> from fsspec.utils import get_file_extension
>>> get_file_extension("/path/to.dir/file")
'dir/file'   # expected ''  (the file has no extension)
>>> get_file_extension("s3://bucket.name/data/file")
'name/data/file'   # expected ''
```

This splits within the final path component instead, so a file with no extension inside a dotted directory correctly returns `""`. Files that do have an extension are not affected. For example, `"/path/to.dir/file.parquet"` still returns `"parquet"`. Added regression tests.

### DIFF
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -348,6 +348,10 @@ def test_get_protocol(par):
         ("file:///home/user/no_extension", ""),
         ("/local/path/to/file.json", "json"),
         ("relative/path/file.yaml", "yaml"),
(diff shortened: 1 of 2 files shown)


Sreekant13:
Thank you for the quick merge, @martindurant! Much appreciated.

Issue #639, opened 2021-05-14
race conditions in fsspec cache?


nbren12:
I recently had some strange data corruption errors happen when using a cached fsspec filesystem from parallel processes. Seemed like a race condition. Is the cached file system thread/process safe? 

According this this [stackoverflow answer](https://stackoverflow.com/questions/2333872/how-to-make-file-creation-an-atomic-operation) `os.rename` can ensure that parallel writes do not corrupt a single file.

martindurant:
The cached filesystem is largely untested from multiple processes. I would probably expect multi-thread to be OK, though. The current cache implementations might not work well, in general, for large loads. There is an idea elsewhere to try to decouple the logical layer (what gets cached, when) from the storage end (dealing with the disk or other backend). I could use help planning these.

Note that the cache metadata is saved using `shutil.move`, which indeed is the same as `os.rename` where the paths are on the same device. This should probably be used for the files themselves - which would perhaps best be accomplished by using `LocalFileSystem.open(, autocommit=False)` or transactions.

nbren12:
>The cached filesystem is largely untested from multiple processes. 

Yah, I just started playing around with it. It seems like a potentially powerful feature, but as they say: "there are two hard problems in CS..."

>This should probably be used for the files themselves - which would perhaps best be accomplished by using LocalFileSystem.open(, autocommit=False) or transactions.

Refactors aside, I think this would resolve this issue.

martindurant:
Essentially, the cachers call `fs.get(remotepaths, localpaths)` in just a couple of places, so it would be simple to use a temporary location until downloading has completed. I don't think there is any foolproof way to ensure that the temporary file is on the same device as the eventual target, except to put them in the same directory (e.g., "{localpath}.temp.UUID").

nbren12:
> except to put them in the same directory (e.g., "{localpath}.temp.UUID")

this sounds like a good idea. It would also probably not work reliably on NFS.

Danferno:
Was this ever implemented? I'm looking for a process-safe read cache using fsspec, but it's proving harder than anticipated. In my case, multiple Dask workers might simultaneously try to fetch the same file. It is a bit hard to debug (celery + dask + containers), but it seems like sometimes this causes an error when the 2nd worker tries to open a file that is half downloaded by the 1st.

martindurant:
It seems no one has picked up this work. @Danferno , are you in a position to contribute?

Danferno:
[...]

PR #2075, opened 2026-07-15
Make whole-file cache writes atomic

lhoupert:
Addresses the whole-file-cache part of #639 

Whole-file caches (`simplecache`, `filecache`) download cache misses directly to the final cache filename, and readers treat "file exists" as "file is ready",  so a concurrent reader of a key whose download is still in flight can return a partially-written file. 

This implements the fix @martindurant proposed in #639 ("replace `get(remote, local)` with `get(remote, local.temp); mv(local.temp, local)`"): every whole-file download now goes to a `.part` temp file in the cache directory followed by an atomic `os.replace()` .

#2154 and #2163 have since landed the sync `cat_ranges` fixes this PR originally carried, so `SimpleCacheFileSystem.cat_ranges` here is master's with one line changed. What remains is the async side, in methods this PR touches anyway: `filecache`'s `_cat_file`/`_cat_ranges` recorded no metadata, so every call re downloaded, and `check_files` went through the sync `ukey`, which raises inside a running loop.

**For reviewers:**    Batch downloads are now all-or-nothing: if one file in a `cat`/`cat_ranges` batch fails, none of that batch's temp files are renamed into place. 

## Author attestation

- [x] I am a human, these are my changes, and I have reviewed and understood every change and can explain why each is correct.

Implementation and tests were AI-assisted, adapted from our production mitigation in EOPF-Explorer/data-model#220 where Claude helped me root-caused the race; I reviewed and verified every line of this diff.

### DIFF
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -30,6 +30,86 @@ if TYPE_CHECKING:
 logger = logging.getLogger("fsspec.cached")
 
 
+# Cache misses are downloaded to a temporary file in the cache directory and
+# then renamed into place, so that a partially-written download is never
+# visible under its final cache filename: readers treat the existence of that
+# filename as "download complete" (see issue #639). The rename is atomic, and
+# a concurrent duplicate download of the same path is harmless: both
+# temporary files hold identical bytes and the last rename wins. These
+# helpers are module-level functions rather than methods because
+# CachingFileSystem.__getattribute__ dispatches only known method names to
+# the caching class, delegating anything else to the wrapped filesystem.
+
+
+def _temppath(lpath):
+    fd, tmp = tempfile.mkstemp(
+        dir=os.path.dirname(lpath),
+        prefix=os.path.basename(lpath) + ".",
+        suffix=".part",
+    )
+    os.close(fd)
+    return tmp
+
+
+def _remove_tempfile(tmp):
+    try:
+        os.remove(tmp)
+    except OSError:
+        pass
+
+
+def _atomic_get_file(fs, rpath, lpath):
+    tmp = _temppath(lpath)
+    try:
+        fs.get_file(rpath, tmp)
+        os.replace(tmp, lpath)
+    except BaseException:
+        _remove_tempfile(tmp)
+        raise
+
+
+async def _atomic_get_file_async(fs, rpath, lpath, **kwargs):
+    tmp = _temppath(lpath)
+    try:
+        await fs._get_file(rpath, tmp, **kwargs)
+        os.replace(tmp, lpath)
+    except BaseException:
+        _remove_tempfile(tmp)
+        raise
+
+
+def _atomic_get(fs, rpaths, lpaths):
+    tmps = [_temppath(lpath) for lpath in lpaths]
+    try:
+        fs.get(rpaths, tmps)
+        for tmp, lpath in zip(tmps, lpaths):
+            os.replace(tmp, lpath)
+    except BaseException:
+        for tmp in tmps:
+            _remove_tempfile(tmp)
+        raise
+
+
+def _atomic_get_file_decompressed(fs, rpath, lpath, compression, **kwargs):
+    tmp = _temppath(lpath)
+    try:
+        with fs._open(rpath, mode="rb", **kwargs) as f, open(tmp, "wb") as f2:
+            if isinstance(f, AbstractBufferedFile):
+                # want no type of caching if just downloading whole thing
+                f.cache = BaseCache(0, f.cache.fetcher, f.size)
+            comp = infer_compression(rpath) if compression == "infer" else compression
+            f = compr[comp](f, mode="rb")
+            data = True
+            while data:
+                block = getattr(f, "blocksize", 5 * 2**20)
+                data = f.read(block)
+                f2.write(data)
+        os.replace(tmp, lpath)
+    except BaseException:
+        _remove_tempfile(tmp)
+        raise
+
+
 class WriteCachedTransaction(Transaction):
     def complete(self, commit=True):
         rpaths = [f.path for f in self.files]
@@ -603,7 +683,7 @@ class WholeFileCacheFileSystem(CachingFileSystem):
         downfn = [fn for fn, d in zip(downfn0, details) if not d]
         if downpath:
             # skip if all files are already cached and up to date
-            self.fs.get(downpath, downfn)
+            _atomic_get(self.fs, downpath, downfn)
 
             # update metadata - only happens when downloads are successful
             newdetail = [
@@ -687,7 +767,7 @@ class WholeFileCacheFileSystem(CachingFileSystem):
                 paths.remove(p)
 
         if getpaths:
-            self.fs.get(getpaths, storepaths)
+            _atomic_get(self.fs, getpaths, storepaths)
             self.save_cache()
 
         callback.set_size(len(paths))
@@ -704,23 +784,9 @@ class WholeFileCacheFileSystem(CachingFileSystem):
         # call target filesystems open
         self._mkcache()
         if self.compression:
-            with self.fs._open(path, mode="rb", **kwargs) as f, open(fn, "wb") as f2:
-                if isinstance(f, AbstractBufferedFile):
-                    # want no type of caching if just downloading whole thing
-                    f.cache = BaseCache(0, f.cache.fetcher, f.size)
-                comp = (
-                    infer_compression(path)
-                    if self.compression == "infer"
-                    else self.compression
-                )
-                f = compr[comp](f, mode="rb")
-                data = True
-                while data:
-                    block = getattr(f, "blocksize", 5 * 2**20)
-                    data = f.read(block)
-                    f2.write(data)
+            _atomic_get_file_decompressed(self.fs, path, fn, self.compression, **kwargs)
         else:
-            self.fs.get_file(path, fn)
+            _atomic_get_file(self.fs, path, fn)
         self.save_cache()
 
     def _open(self, path, mode="rb", **kwargs):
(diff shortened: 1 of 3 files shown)


fsspec/implementations/cached.py
```diff
@@ -29,6 +30,86 @@
...
+# CachingFileSystem.__getattribute__ dispatches only known method names to
+# the caching class, delegating anything else to the wrapped filesystem.
+
+
+def _temppath(lpath):
+    fd, tmp = tempfile.mkstemp(
+        dir=os.path.dirname(lpath),
+        prefix=os.path.basename(lpath) + ".",
+        suffix=".part",
+    )
```
martindurant: Is there not also a chance of a clash here - should there be random part?

lhoupert: Hi @martindurant !  `mkstemp` already inserts a random token between the prefix and suffix (the name is <key>.<random>.part) so two concurrent downloads of the same key are guaranteed distinct temp files. I've added a comment in making that explicit (029bdb5).  I try to show this in [`test_concurrent_downloads_leave_no_tempfiles` ](https://github.com/fsspec/filesystem_spec/pull/2075/changes#diff-b65b803563b629100ce65766fa29064ec1271f3a139fa23929deb33fd5a8c98eR1529): 4 concurrent downloads of one key must end with exactly one file in the cache dir and no .part leftovers.

fsspec/implementations/cached.py
```diff
@@ -771,7 +840,7 @@ async def _cat_file(self, path, start=None, end=None, **kwargs):
 
         if not fn:
             fn = os.path.join(self.storage[-1], sha)
-            await self.fs._get_file(path, fn, **kwargs)
+            await _atomic_get_file_async(self.fs, path, fn, **kwargs)
```
Sanjays2402: where does `filecache` record this download in `CacheMetadata`? `_cat_file` never calls `_make_local_details` or `save_cache`, so its second read is another cache miss and downloads the file again. The new test only checks returned bytes, so it doesn’t catch that.

fsspec/implementations/cached.py
```diff
@@ -29,6 +30,88 @@
...
+# temporary files hold identical bytes and the last rename wins. These
+# helpers are module-level functions rather than methods because
+# CachingFileSystem.__getattribute__ dispatches only known method names to
+# the caching class, delegating anything else to the wrapped filesystem.
+
+
+def _temppath(lpath):
+    # mkstemp inserts a random token between prefix and suffix and creates the
+    # file with O_EXCL, so concurrent downloads of the same key never clash.
+    fd, tmp = tempfile.mkstemp(
```
Sanjays2402: mkstemp creates 0600 and os.replace keeps that mode, so cache files that used to land at 0644 (umask) are now owner-only. Anything sharing a cache_storage dir between users/processes running as different uids breaks after this. Worth a chmod to 0o666 & ~umask on the tmp file before the replace.

martindurant: Only on POSIX, right?

fsspec/implementations/cached.py
```diff
@@ -29,6 +30,88 @@
 logger = logging.getLogger("fsspec.cached")
 
 
+# Cache misses are downloaded to a temporary file in the cache directory and
+# then renamed into place, so that a partially-written download is never
+# visible under its final cache filename: readers treat the existence of that
+# filename as "download complete" (see issue #639). The rename is atomic, and
+# a concurrent duplicate download of the same path is harmless: both
+# temporary files hold identical bytes and the last rename wins. These
```
Sanjays2402: the "last rename wins" claim holds on posix but not on windows: os.replace onto an existing destination raises PermissionError if any other process has that file open, and that is exactly the state this design creates — reader A opens the cache file right after the first rename, downloader B for the same key then tries to replace it out from under A. pytest-win runs the new concurrency tests too, so this is reachable in CI. would it be enough to swallow OSError around the replace and just drop the temp file when the destination already exists (bytes are identical anyway)?

fsspec/implementations/cached.py
martindurant: Yeah, I think we should just use `open` and avoid the issue altogether (which means this function just generates a filename including a random component). The documentation should indicate that these files are *not* cleared up in case of early exit (except for an exception during replace).

lhoupert: Hi @Sanjays2402 and @martindurant , thank you for spotting this issue. It should be fixed in d9ffb53/9028ef9 where `_temppath` now only generates the name (<key>.<random>.part), then the file is created by the plain open() of whatever downloads it, so mkstemp's fd and 0600 mode are gone. 

I also added in the docs (features.rst and the module comment) that .part files are removed when the download or the final rename raises, but not on early exit, leftover .part files are ignored by the cache and safe to delete. I also added tests covering failed-download cleanup and that a stale .part file is never mistaken for a cache entry.

fsspec/implementations/cached.py
lhoupert: Good catch @Sanjays2402 , thank you!  You are right _cat_file never recorded the download, and my test only checked the returned bytes so it couldn't have caught it... I've tried to address it properly in the follow-up commits 5b94d78 and aeb6cc3:
- misses in `_cat_file`/`_cat_ranges` now record the metadata (with an async-safe ukey lookup, since the sync ukey can't be called inside a running loop for asynchronous=True targets), followed by save_cache(), mirroring the sync cat() path (5b94d78)

The new tests count actual `_get_file` calls now: a second read, including from a fresh instance over the same cache_storage, should perform zero downloads. Happy to adjust if you'd prefer a different approach here.

fsspec/implementations/cached.py
lhoupert: Thanks!  I hadn't considered the Windows rename semantics at all... I went with your suggestion in 5b94d78, then refined it in aeb6cc3 with one change I'd welcome your view on: `_replace_tempfile` swallows the failed rename and drops the temp copy only when the destination holds identical bytes (filecmp), which I think covers the concurrent-duplicate case you describe.

fsspec/implementations/cached.py
```diff
@@ -29,6 +34,110 @@
...
+    try:
+        os.replace(tmp, lpath)
+    except OSError:
+        # On Windows, replacing a file that another process holds open
+        # raises PermissionError. If a concurrent download of the same
+        # path already renamed identical bytes into place, this copy is
+        # redundant and safe to discard; a differing destination (e.g. a
+        # cache refresh racing a reader of the stale copy) must surface
+        # the error rather than let stale bytes pass for fresh ones.
+        if not (os.path.exists(lpath) and filecmp.cmp(tmp, lpath, shallow=False)):
```
martindurant: If I understand, this will always re-read the whole contents of both files (os.stat will never be identical), which could be a bad performance hit. Might be enough to look at the creation time of the two files? Is there any way to know if the target file is done (would that mean it gets closed and there is no error?) ?

lhoupert: Good point! What do you think about the last commit I pushed to address this 2b9ec83?  There is no content reads any more. The temp file is pre-created empty so the download start is stamped on the cache filesystem's own clock, and on a failed rename stat checks decide: the temp copy is dropped only if the destination is a regular file of the same size with mtime at or after that start (written while this download ran, so a complete recent download of the same path).  Claude suggested to use mtime rather than creation time because st_birthtime is not available on all platforms.

lhoupert: > Is there any way to know if the target file is done (would that mean it gets closed and there is no error?) ?

Something else that the fact that a temporary file has been renamed with its final name? (the rename runs only after the download completes: https://github.com/fsspec/filesystem_spec/blob/2b9ec831475d118914fbc28370354fbde6e934c6/fsspec/implementations/cached.py#L103-L116

lhoupert:
Hi @martindurant , is there anything else I could do to move this forward? Thx!

martindurant:
It seems like it behaves the right way now, but the diff has become big for what it does. Can you clean it up or instruct your bot not to make all the little helper functions and all those very extensive comments, please?

lhoupert:
Thank you Martin! I did some cleanup the helpers and the long comments are gone.
I merged main too and two things came up with it:
- I think `filecache._cat_ranges` is broken on master since #2154: on current  master, `await fs._cat_ranges(["memory://f"], [0], [4])` raises `AttributeError: ... has no attribute '_local_paths_for_ranges'`. Doesn't affect this PR since [the method is rewritten here](https://github.com/lhoupert/filesystem_spec/blob/fb94b5935852a484b70af0974b9ad0ebb43de12e/fsspec/implementations/cached.py#L855-L887),  @itzzdev09 you may want to have a look
- I dropped `on_error=on_error` from `self.fs.get()`: `get()` has no such parameter, so it
  ends up in the backend's `get_file()` and `http_sync` raises a TypeError.


My last commit, fb94b59, fixes a regression in my own PR's code.

itzzdev09:
@lhoupert 
Thanks for catching both of those! Fixed in [fsspec/filesystem_spec#2163](https://github.com/fsspec/filesystem_spec/pull/2163): the helper was on the wrong class, and `on_error` shouldn't have been passed to `get()`.

martindurant:
Sorry, this will take a little work now to align.

lhoupert:
No problem, it is done!

martindurant:
Windows error looks real

lhoupert:
Sorry for that, I cannot test windows locally easily. It should be fixed now

itzzdev09:
@lhoupert I checked out b2b321c on Windows 11 (Python 3.12). The full test suite passes (1886 passed, 0 failed), including test_cached.py, so the Windows fix looks good on my side.

> Sorry for that, I cannot test windows locally easily. It should be fixed now

lhoupert:
Thanks all!

PR #2083, opened 2026-07-19
Propagate storage_options to all backends resolved by GenericFileSystem (#1856)

### LINKED ISSUE #1856: rsync accross FS with different protocol or credentials is broken
Hi,

I'm completely new to fsspec and was extremely interested by the `rsync` utility (even if signaled as experimental) and gave it a try (version 2025.5.1).

My first use case is to sync a S3 source to a local directory.

Having a quick look at the source code, I tried something like this:

```python
rsync('s3://bucket-name/key', './path/to/local/directory',
  inst_kwargs = {
    'default_method': 'options',
    'storage_options': {
      's3': { 'endpoint_url': 'xxxxxx', 'key': 'xxxxxxx', 'secret': 'xxxx', },
    },
  }
)
```

The idea being to let the `GenericFileSystem` instance used by rsync resolve both underlying filesystem behind source and destination.

Unfortunately this didn't work, because the `GenericFileSystem`:

- didn't override all methods `rsync` uses, namely `isdir`
- didn't resolve correctly URL

First problem can be solved by adding the following method to `GenericFileSystem` (I have absolutely no idea if this is correct, it simply mimics the way it is done for other methods):

```python
    async def _isdir(
        self,
        url,
        **kwargs,
    ):
        fs = _resolve_fs(url, self.method, storage_options=self.st_opts)
        if fs.async_impl:
            return await fs._isdir(url, **kwargs)
        else:
            return fs.isdir(url, **kwargs)
```

Second problem was addressed by making sure every call to `_resolve_fs` uses the `storage_options=self.st_opts` argument (it is missing in many places).

Then it worked.

This is however not very efficient, because every call to `_resolve_fs` creates a new instance. `rsync` should probably call it only twice (once for the source, once for the destination, if they do not have the same protocol).

Then came the next question: what if I want to synchronise two S3 directories in differents buckets with different credentials? Also a use case I have.

Since `_resolve_fs` only uses the protocol to find a filesystem instance, this will not work.

Among the possible solutions I figured out:
- let `_resolve_fs` handle URL as URI and discriminate both on protocol and authority. At least the bucket name could be used to select different credentials in `inst_kwargs`. This would also probably work with other protocols
- make `rsync` take a `source_fs` and `destination_fs` argument instead of a unique `fs`

Hope this helps.

Best regards,

Antoine

balramthewarrior:
Fixes #1856.

### Root cause

`GenericFileSystem` stores `storage_options` as `self.st_opts` and resolves a backend per call via `_resolve_fs`. With `default_method="options"`, `_resolve_fs` looks up `storage_options[protocol]` to construct the backend — so omitting the argument silently builds an unconfigured filesystem.

Of the 14 `_resolve_fs` call sites, only 5 passed `storage_options=self.st_opts`. The other 9 dropped it: `_info`, `_ls`, `_cat_file`, `_rm`, `_cp_file` (both sides), `_make_many_dirs`, and `_copy` (both sides).

Any credentialed backend reached through those paths was constructed bare, which is what breaks `rsync` across filesystems with different protocols or credentials.

### On the reported `isdir` failure

The issue proposes adding an explicit `GenericFileSystem._isdir`. I don't think that's needed, and it matches your reply on the thread:

> This is surprising, since it ought to fall back to info() which in turn falls back to ls()[0].

That fallback works fine. The reason `isdir` failed is that it reaches the backend through `_info`, which was one of the 9 sites dropping `storage_options`. Fixing the propagation makes `isdir` work through the existing fallback — no new method required. There's a regression test covering exactly this.

### Change

Pass `storage_options=self.st_opts` at the 9 remaining `_resolve_fs` call sites. No behaviour change when `storage_options` is unset (`_resolve_fs` already coerces `None` to `{}`), and no change for the `default`/`generic`/`current` methods, which ignore the argument.

### Tests

Added to `fsspec/tests/test_generic.py`:

- `_OptionsRequiredFS`, a `MemoryFileSystem` subclass that raises unless its `token` storage option arrives — standing in for a filesystem needing real credentials, so dropped options fail loudly instead of silently falling back to an unauthenticated instance.
- `test_storage_options_propagated_to_backend` — parametrized over `info`, `ls`, `cat_file`, `isdir`, `find`.
- `test_storage_options_propagated_cross_protocol` — the reporter's original call shape: `rsync` from a credentialed remote to local, configured via `inst_kwargs`.

Against unpatched master these give 5 failed / 8 passed. `find` passes on both, since `_find` was already among the 5 correct call sites, so the tests track this specific defect rather than failing broadly. Full suite: 1558 passed, 0 failed. `ruff check`, `ruff format` and `codespell` clean.

### DIFF
--- a/fsspec/tests/test_generic.py
+++ b/fsspec/tests/test_generic.py
@@ -1,6 +1,9 @@
 import pytest
 
 import fsspec
+from fsspec.generic import rsync
+from fsspec.implementations.memory import MemoryFileSystem
+from fsspec.registry import _registry, register_implementation
 from fsspec.tests.conftest import data, server  # noqa: F401
 
 
@@ -116,3 +119,57 @@ def test_rsync(tmpdir, m):
         allfiles[f"file://{pos_tmpdir}/deep/path/afile"]
         == allfiles2[f"file://{pos_tmpdir}/deep/path/afile"]
     )
+
+
+class _OptionsRequiredFS(MemoryFileSystem):
+    """Backend that raises unless its storage_options reached the constructor."""
+
+    protocol = "optsrequired"
+
+    def __init__(self, *args, token=None, **kwargs):
+        if token is None:
+            raise ValueError("storage_options were not propagated")
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def _strip_protocol(cls, path):
+        if isinstance(path, str):
+            path = path.removeprefix(f"{cls.protocol}://")
+        return MemoryFileSystem._strip_protocol(path)
+
+
+@pytest.fixture()
+def opts_fs(m):
+    register_implementation(
+        _OptionsRequiredFS.protocol, _OptionsRequiredFS, clobber=True
+    )
+    m.pipe_file("/afile", b"hello")
+    m.makedirs("/adir", exist_ok=True)
+    m.pipe_file("/adir/nested", b"world")
+    yield fsspec.filesystem(
+        "generic",
+        default_method="options",
+        storage_options={_OptionsRequiredFS.protocol: {"token": "set"}},
+    )
+    _registry.pop(_OptionsRequiredFS.protocol, None)
+    _OptionsRequiredFS.clear_instance_cache()
+
+
+def test_storage_options_propagated_to_backend(opts_fs):
+    opts_fs.info("optsrequired:///afile")
+    opts_fs.ls("optsrequired:///adir")
+    opts_fs.cat_file("optsrequired:///afile")
+    opts_fs.isdir("optsrequired:///adir")
+    opts_fs.find("optsrequired:///adir")
+
+
+def test_storage_options_propagated_cross_protocol(opts_fs, tmpdir):
+    rsync(
+        "optsrequired:///adir",
+        f"file://{tmpdir}",
+        inst_kwargs={
+            "default_method": "options",
+            "storage_options": {_OptionsRequiredFS.protocol: {"token": "set"}},
+        },
+    )
+    assert (tmpdir / "nested").read_binary() == b"world"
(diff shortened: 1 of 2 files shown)


fsspec/tests/test_generic.py
```diff
@@ -116,3 +119,80 @@ def test_rsync(tmpdir, m):
...
+    Stands in for a real filesystem needing credentials (e.g. s3 key/secret or
+    endpoint_url), so that dropped storage_options fail loudly instead of
+    silently falling back to an unauthenticated instance.
+    """
+
+    protocol = "optsrequired"
+    _token = "s3kr3t"
+
+    def __init__(self, *args, token=None, **kwargs):
+        if token != self._token:
```
martindurant: Same as `if token is not None`? Then you would not need the class attribute `_token`.

fsspec/tests/test_generic.py
```diff
@@ -116,3 +119,80 @@ def test_rsync(tmpdir, m):
...
+        lambda fs: fs.cat_file("optsrequired:///afile"),
+        lambda fs: fs.isdir("optsrequired:///adir"),
+        lambda fs: fs.find("optsrequired:///adir"),
+    ],
+)
+def test_storage_options_propagated_to_backend(opts_fs, op):
+    # GH#1856: several GenericFileSystem methods resolved the backend without
+    # passing storage_options, so credentialed filesystems were constructed
+    # bare. isdir() is included because it reaches the backend via _info().
+    op(opts_fs)
```
martindurant: This is unnecessarily verbose: please just move those operations inline rather than parametrising, and remove the comment here and in the other tests.

balramthewarrior:
Thanks for the review. Done both:
- Dropped the `_token` attribute; the backend now just checks `token is None`.
- De-parametrised `test_storage_options_propagated_to_backend` into inline calls and removed the explanatory comments.

Pushed. Confirmed the two tests still fail against master without the `generic.py` change and pass with it.

PR #2085, opened 2026-07-21
Permit composite URL protocol schemes in `infer_storage_options`

amotl:
Hi there. In [omniload](https://omniload.readthedocs.io/), we are [using](https://github.com/panodata/omniload/blob/v0.7.0/omniload/core/registry.py) composite URL protocol schemes like `https+webdav://`, `mongodb+srv://`, or `mysql+pymysql://`. While the code used custom auxiliary/helper functions before, we are now aiming to modernize and refactor a bit, for example towards using `fsspec.utils.infer_storage_options` across the board. On this occasion, we found that a specific regex pattern needed a minor adjustment to permit such URL protocol schemes that include a `+` character.

- Reference: https://github.com/panodata/omniload/pull/236

### DIFF
--- a/fsspec/utils.py
+++ b/fsspec/utils.py
@@ -56,11 +56,18 @@ def infer_storage_options(
     "host": "node", "port": 123, "path": "/mnt/datasets/test.csv",
     "url_query": "q=1", "extra": "value"}
     """
-    # Handle Windows paths including disk name in this special case
-    if (
-        re.match(r"^[a-zA-Z]:[\\/]", urlpath)
-        or re.match(r"^[a-zA-Z0-9]+://", urlpath) is None
-    ):
+
+    # Discover Windows paths including disk name in this special case.
+    is_filesystem = re.match(r"^[a-zA-Z]:[\\/]", urlpath)
+
+    # Discover URI according to RFC 3986: Scheme names consist of a
+    # sequence of characters beginning with a letter and followed by
+    # any combination of letters, digits, plus ("+"), period ("."),
+    # or hyphen ("-").
+    # https://datatracker.ietf.org/doc/html/rfc3986#section-3.1
+    is_uri = re.match(r"^[a-zA-Z0-9+.-]+://", urlpath)
+
+    if is_filesystem or is_uri is None:
         return {"protocol": "file", "path": urlpath}
 
     parsed_path = urlsplit(urlpath)
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -223,6 +223,13 @@ def test_infer_simple():
     assert out.get("host", None) is None
 
 
+def test_infer_composite_protocol():
+    out = infer_storage_options("foo+bar-baz://db.example.org")
+    assert out["protocol"] == "foo+bar-baz"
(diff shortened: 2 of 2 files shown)


martindurant:
What is fsspec supposed to do with there composite protocols, though - do they map onto omniload only and exactly?

Aside: omniload and intake should probably talk. Intake is caring less about writing yaml catalogues and pipelines, and (if anything) more about guessing file types and providing dataset summaries. Enumerating in detail all the things duckdb knows how to do, and providing a clear way to determine dataset invocation and summary all sound like good ideas.

amotl:
Hi Martin,

thanks for your swift reply.

> What is fsspec supposed to do with there composite protocols, though - do they map onto omniload only and exactly?

In our case, omniload is responsible for routing and dispatching of protocol schemes to **dlt-based components** that wrap the fsspec filesystem implementations. In this spirit, fsspec will probably not do anything specific to them today other than treating them opaquely, which seems to work well.

We just recently discovered `infer_storage_options` as a useful utility to streamline the downstream auxiliary logic around fsspec details in omniload, with that minor drawback. If it is not applicable or possible to adjust this utility function so it can be reused in omniload, we will either need to work around it or vendorize it.

> Aside: omniload and intake should probably talk.

Thank you for quickly demonstrating ingredients of Intake. Indeed, that sounds nice. I've started a discussion thread so the conversation can grow beyond this specific PR discussion. 🙇 

- https://github.com/fsspec/filesystem_spec/discussions/2086

With kind regards,
Andreas.

fsspec/utils.py
```diff
@@ -57,9 +57,10 @@ def infer_storage_options(
     "url_query": "q=1", "extra": "value"}
     """
     # Handle Windows paths including disk name in this special case
+    # Also permit composite URL protocol schemes like `mysql+pymysql://`.
     if (
         re.match(r"^[a-zA-Z]:[\\/]", urlpath)
-        or re.match(r"^[a-zA-Z0-9]+://", urlpath) is None
+        or re.match(r"^[a-zA-Z0-9+]+://", urlpath) is None
```
Sanjays2402: This still classifies valid schemes containing `-` or `.` as local paths, for example `foo+bar-baz://host`. Since `urlsplit` already accepts the RFC 3986 scheme grammar, should this guard use `[A-Za-z][A-Za-z0-9+.-]*` too?

amotl: Thank you for looking up RFC 3986 for me. Indeed, it [says](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1):

> Scheme names consist of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").

I've adjusted the patch accordingly.

martindurant:
Please check the test failures: I think they are all cases where the protocol is omitted completely, which should result in the default "file", but the URL otherwise unchanged. These protocol-less URLs can contain some variants (not just "/...") due to windows.

amotl:
Apologies. I made a mistake when refactoring the code block a bit to make it more readable. Should be fixed now.

amotl:
Thank you for merging!

PR #1982, opened 2026-01-28
fixes to parquet and known cache

martindurant:
Fixes #1973

PR #2091, opened 2026-07-29
Fix for `merge_offset_ranges` becoming quadratic

BrianMichell:
#1982 introduced a performance regression that was raised as an issue here: https://github.com/TGSAI/segy/issues/343

This PR seeks to restore the original `O(n log n)` algorithm while retaining the fixes for Parquet.

### DIFF
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -472,9 +472,8 @@ def test_merge_offset_ranges_drops_nested():
     [
         # Exact duplicates: keep one
         ([0, 0], [50, 50], [(0, 50)]),
-        # Nested in either order
+        # Nested range
         ([0, 10], [80, 20], [(0, 80)]),
-        ([10, 0], [20, 80], [(0, 80)]),
         # None end covers to EOF
         ([0, 0], [None, 50], [(0, None)]),
         ([0, 0], [50, None], [(0, None)]),
@@ -493,6 +492,19 @@ def test_merge_offset_ranges_edges(starts, ends, expected, sort):
     )
 
     assert list(zip(*result)) == [("f", *rng) for rng in expected]
+
+
+def test_merge_offset_ranges_sorts_unsorted_nested_ranges():
+    result = merge_offset_ranges(
+        ["f", "f"],
+        [10, 0],
+        [20, 80],
+        max_gap=100,
+        max_block=1000,
+        sort=True,
+    )
+
+    assert list(zip(*result)) == [("f", 0, 80)]
 
 
 @pytest.mark.parametrize("max_block", [None, 4, 128])
@@ -537,21 +549,24 @@ def test_merge_offset_ranges_covers_every_input_range():
 
 def test_merge_offset_ranges_many_sequential_is_fast():
     # Regression: O(n²) nested-range filter added in #1982
-    n = 20_000
-    paths = ["file"] * n
-    starts = list(range(0, n * 240, 240))
-    ends = [s + 240 for s in starts]
+    def run(n):
+        paths = ["file"] * n
+        starts = list(range(0, n * 240, 240))
+        ends = [s + 240 for s in starts]
+        t0 = time.perf_counter()
+        result = merge_offset_ranges(
+            paths, starts, ends, max_block=8_388_608, sort=True
+        )
+        return time.perf_counter() - t0, result, ends[-1]
 
-    t0 = time.perf_counter()
-    result_paths, result_starts, result_ends = merge_offset_ranges(
-        paths, starts, ends, max_block=8_388_608, sort=True
-    )
-    elapsed = time.perf_counter() - t0
+    small_elapsed, _, _ = run(5_000)
+    large_elapsed, result, expected_end = run(20_000)
+    result_paths, result_starts, result_ends = result
 
-    assert elapsed < 1.0
+    assert large_elapsed < small_elapsed * 8
     assert result_paths == ["file"]
     assert result_starts == [0]
-    assert result_ends == [ends[-1]]
+    assert result_ends == [expected_end]
 
 
 def test_size():
--- a/fsspec/tests/test_utils.py
+++ b/fsspec/tests/test_utils.py
@@ -507,6 +507,19 @@ def test_merge_offset_ranges_sorts_unsorted_nested_ranges():
     assert list(zip(*result)) == [("f", 0, 80)]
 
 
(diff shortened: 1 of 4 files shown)


fsspec/utils.py
```diff
@@ -537,17 +537,19 @@ def nullcontext(obj: T) -> Iterator[T]:
...
     sort: bool = True,
-) -> tuple[list[str], list[int], list[int]]:
+) -> tuple[list[str], list[int], list[int | None]]:
     """Merge adjacent byte-offset ranges when the inter-range
     gap is <= `max_gap`, and when the merged byte range does not
-    exceed `max_block` (if specified). By default, this function
-    will re-order the input paths and byte ranges to ensure sorted
-    order. If the user can guarantee that the inputs are already
+    exceed `max_block` (if specified). Overlapping ranges are always
+    merged, even past `max_block`, so returned ranges never overlap.
```
Sanjays2402: the docstring now promises returned ranges never overlap, but that only holds when `sort=True`. the old code dropped nested ranges with a global pairwise scan, so unsorted interleaved input was fine; the new single-pass only compares against the previous block, e.g. `merge_offset_ranges(["f","g","f"], [0,0,10], [100,10,20], sort=False)` used to give `(["f","g"],[0,0],[100,10])` and now returns the nested `f` 10-20 as its own block, overlapping 0-100. `parquet.py` passes `sort=False` on the metadata path with a "Should be sorted" comment, which is an assumption rather than a guarantee. worth either scoping the docstring claim to sort=True or grouping by path before the merge loop.

BrianMichell: Good catch, thanks!

fsspec/tests/test_utils.py
```diff
@@ -450,6 +452,123 @@ def test_merge_offset_ranges(max_gap, max_block):
...
+        result = merge_offset_ranges(
+            paths, starts, ends, max_block=8_388_608, sort=True
+        )
+        return time.perf_counter() - t0, result, ends[-1]
+
+    small_elapsed, _, _ = run(5_000)
+    large_elapsed, result, expected_end = run(20_000)
+    result_paths, result_starts, result_ends = result
+
+    assert large_elapsed < small_elapsed * 8
```
Sanjays2402: the ratio guard in test_merge_offset_ranges_many_sequential_is_fast is tight enough to flake. 4x the input on an n log n sort is already ~4x the time, so the 8x bound is under 2x headroom — I ran the new implementation standalone and got ratios of 3.3-4.9 across six trials on an idle machine, and the 5k baseline is only ~1.6ms, small enough that one GC pause or a noisy CI runner inverts it. comparing 20k against a linear extrapolation of a much larger baseline, or just asserting large_elapsed stays under some absolute ceiling, would still catch the quadratic without the timing sensitivity.

Sanjays2402: the wall-clock ratio assert is going to flake on a shared runner: run(5_000) is a couple of ms, so `small_elapsed * 8` is a tiny absolute bound and one GC pause or a noisy neighbour in the 20k run blows it. n log n only buys ~4.2x over 4x the input too, so the headroom is thinner than it looks. counting merges (or just capping n and asserting the result) would test the same regression without the timer.

fsspec/utils.py
```diff
@@ -560,59 +563,57 @@ def merge_offset_ranges(
...
+            # Cannot merge with previous block
+            new_paths.append(path)
+            new_starts.append(start)
+            new_ends.append(end)
+        elif prev_end is None:
+            # Previous block already covers the rest of the file
+            continue
+        elif start < prev_end:
+            # Overlap / nested: merge into the union even past max_block
+            new_starts[-1] = min(new_starts[-1], start)
```
Sanjays2402: the `min(new_starts[-1], start)` here is unreachable under `sort=True` (already ordered by start), and under `sort=False` it can walk the current block's start backwards behind the block that was already emitted: `merge_offset_ranges(["f"]*3, [0,100,5], [10,110,7], max_gap=0, sort=False)` returns starts `[0,5]` / ends `[10,110]`, so the two output ranges overlap and the second one starts before the first ends. parquet.py calls this with `sort=False` on the "should already be sorted" assumption, so it is not obviously unreachable in practice. worth either dropping the min and documenting sorted-input-only, or raising instead of silently emitting overlapping blocks?

fsspec/utils.py
```diff
@@ -560,59 +568,64 @@ def merge_offset_ranges(
...
+        if path != new_paths[-1]:
+            # Cannot merge with previous block
+            new_paths.append(path)
+            new_starts.append(start)
+            new_ends.append(end)
+        elif start < new_starts[-1]:
+            # Starts must be non-decreasing within a path; walking the
+            # current block start backwards can overlap a prior emitted
+            # block when sort=False.
+            raise ValueError(
```
Sanjays2402: the new ValueError is reachable from fsspec's own sort=False caller: `_get_parquet_byte_ranges_from_metadata` passes `sort=False  # Should be sorted`, but when `row_groups` is a list of row-group metadata objects rather than ints, `FastparquetEngine._parquet_byte_ranges` sets `row_group_indices = None` and iterates the caller's list order, so `open_parquet_file(..., metadata=md, row_groups=[rg2, rg1])` yields descending starts and now raises instead of merging. worth either sorting there or dropping to sort=True for that call?

fsspec/utils.py
```diff
@@ -560,59 +568,64 @@ def merge_offset_ranges(
...
+                "starts must be sorted ascending within each path; "
+                f"got start={start} before current block start={new_starts[-1]}"
+            )
+        elif prev_end is None:
+            # Previous block already covers the rest of the file
+            continue
+        elif start < prev_end:
+            # Overlap / nested: merge into the union even past max_block
```
Sanjays2402: the "overlap merges even past max_block" rule chains: each merge extends new_ends[-1], so the next range only has to overlap the *extended* block to keep going. 1000 ranges of 150 bytes at stride 100 with max_block=8_388_608 collapse to a single 100_050-byte block, and with a real-world overlapping request set that grows without bound. parquet.py passes max_block=256_000_000 straight into cat_ranges, so the cap silently stops being a cap on the fetch size. worth capping the chain at max_block and just letting consecutive blocks overlap there instead?

BrianMichell:
If there are no more concerns regarding this PR could we please enable the pipelines @martindurant ?

martindurant:
CI is good - I'll look at it tomorrow.

martindurant:
I can't fault it at all!

PR #2097, opened 2026-08-11
fix: use real LRU eviction in DirCache

### LINKED ISSUE #2095: [Bug] DirCache LRU implementation fails to evict items on write and wipes cache during iteration
### **Describe the Bug**
`DirCache` uses `functools.lru_cache` to limit the number of cached paths to `max_paths`:
```python
if max_paths:
    self._q = lru_cache(max_paths + 1)(lambda key: self._cache.pop(key, None))
```

However, `functools.lru_cache` only calls the wrapped function (`self._cache.pop`) when a key is a **cache miss** (computing a value for a new key), **NOT** when an old key is evicted from `lru_cache`'s internal index.

This causes three major issues in `DirCache`:

1. **Memory Leak (No Eviction on Write):** Adding entries beyond `max_paths` never deletes old entries from `self._cache`. `self._cache` grows without bound.
2. **Deferred Deletion on Read:** When an old key that fell out of `lru_cache` is read via `__getitem__`, `_q` treats it as a miss and calls `self._cache.pop(key)`, deleting `key` from `_cache` at read time and raising a `KeyError`.
3. **Cache Destruction during Iteration:** `__iter__` filters entries using `k in self` (`__getitem__`), triggering cascading `lru_cache` misses. As a result, calling `list(dircache)` deletes **every single item** in `_cache` and returns `[]`.

---

### **Minimal Reproducible Example**

```python
from fsspec.dircache import DirCache

dc = DirCache(max_paths=2)

# 1. Add 4 items (exceeding max_paths=2)
dc["a"] = 1
dc["b"] = 2
dc["c"] = 3
dc["d"] = 4

# Bug 1: _cache size is 4 instead of 2
print("Actual size:", len(dc._cache))  # Output: 4 (Expected: 2)

# Bug 2: Iterating wipes out the cache and returns []
print("Keys yielded by iter:", list(dc))  # Output: []
print("Cache after iter:", dc._cache)     # Output: {}
```

---

### **Expected Behavior**
* Adding keys beyond `max_paths` should immediately evict the least recently set keys from `self._cache`. `len(dc._cache)` should never exceed `max_paths`.
* Iterating over `dircache` (`list(dc)`) should yield valid, unexpired keys without mutating or clearing the cache.

---

### **Environment Information**
* **Library:** `fsspec`
* **Python Version:** 3.8+

---

### **Suggested Fix**
Replace `functools.lru_cache` with `collections.OrderedDict` (or standard dict popping) in `DirCache` to handle LRU ordering directly upon `__setitem__` and `__getitem__`.

aryansk:
Fixes #2095

## Problem

`DirCache` uses a `functools.lru_cache` as an eviction helper, but `lru_cache` calls its wrapped function only on cache misses — never when an old key is evicted from its recency index. As a result:

1. Adding entries beyond `max_paths` never removes old entries from `_cache` (unbounded growth).
2. Reading an evicted key pops it from `_cache` lazily and raises `KeyError`.
3. `__iter__` filters through `__getitem__`, so iterating destroys the entire cache and returns `[]`.

## Change

Track recency with an `OrderedDict` instead:

- on insert, evict the least-recently-used entry (from both `_cache` and `_times`) when over `max_paths`;
- on access, refresh the recency order without popping the listing;
- `__delitem__` and `clear()` keep the recency index consistent.

## Testing

Added `fsspec/tests/test_dircache.py` covering eviction, iteration stability, recency refresh, delete, and expiry. With the fix: `160 passed` in the cache test modules; the remaining failures/errors are pre-existing missing-optional-dependency (`aiohttp`) import issues.

## Notes

- `max_paths` is currently undocumented in the class docstring (only in the `__init__` docstring); I left that unchanged to keep the diff focused.

### DIFF
--- a/fsspec/dircache.py
+++ b/fsspec/dircache.py
@@ -1,6 +1,6 @@
 import time
+from collections import OrderedDict
 from collections.abc import MutableMapping
-from functools import lru_cache
 
 
 class DirCache(MutableMapping):
@@ -48,7 +48,7 @@ class DirCache(MutableMapping):
         self._cache = {}
         self._times = {}
         if max_paths:
-            self._q = lru_cache(max_paths + 1)(lambda key: self._cache.pop(key, None))
+            self._q = OrderedDict()
         self.use_listings_cache = use_listings_cache
         self.listings_expiry_time = listings_expiry_time
         self.max_paths = max_paths
@@ -58,11 +58,15 @@ class DirCache(MutableMapping):
             if self._times.get(item, 0) - time.time() < -self.listings_expiry_time:
                 del self._cache[item]
         if self.max_paths:
-            self._q(item)
+            # Refresh the recency order; the item may already be evicted.
+            self._q.pop(item, None)
+            self._q[item] = None
         return self._cache[item]  # maybe raises KeyError
 
     def clear(self):
         self._cache.clear()
+        if self.max_paths:
+            self._q.clear()
 
     def __len__(self):
         return len(self._cache)
@@ -78,13 +82,20 @@ class DirCache(MutableMapping):
         if not self.use_listings_cache:
             return
         if self.max_paths:
-            self._q(key)
+            self._q.pop(key, None)
+            self._q[key] = None
+            while len(self._q) > self.max_paths:
+                oldest, _ = self._q.popitem(last=False)
+                self._cache.pop(oldest, None)
+                self._times.pop(oldest, None)
         self._cache[key] = value
         if self.listings_expiry_time is not None:
             self._times[key] = time.time()
 
     def __delitem__(self, key):
         del self._cache[key]
+        if self.max_paths:
+            self._q.pop(key, None)
 
     def __iter__(self):
         entries = list(self._cache)
--- a/fsspec/tests/test_dircache.py
+++ b/fsspec/tests/test_dircache.py
@@ -1,0 +1,64 @@
+import time
+
+from fsspec.dircache import DirCache
+
+
+def test_max_paths_evicts_oldest_on_write():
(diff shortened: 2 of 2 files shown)


fallenmi (changes requested):
The new recency queue is mutated before `self._cache[item]` establishes that the key exists. `MutableMapping.__contains__` calls `__getitem__`, so an ordinary miss corrupts the eviction order:

```python
d = DirCache(max_paths=2)
d["a"] = 1
d["b"] = 2
assert "missing" not in d
d["c"] = 3
assert dict(d._cache) == {"b": 2, "c": 3}
```

On exact head `4544763`, the final cache is only `{"c": 3}` and `_q` is `OrderedDict({"missing": None, "c": None})`: the ghost key makes the next insertion evict both valid entries. The exact base fails the normal LRU oracle while this head fixes it, but this absent-key oracle still fails. Please read the cached value, allowing `KeyError` to happen, before refreshing `_q`, and cover missing and expired lookups. I also ran the changed target plus `test_core.py`: 62 passed, 6 skipped; the diff check and compile check are clean.

Reviewed with Codex (GPT-5); the exact base/head oracles and listed tests were run locally.

itzzdev09:
For reference, #2165 is a second fix for #2095. I ran both against the issue's repro and some extra cases (eviction beyond `max_paths`, a read protecting an entry from eviction, iteration, expiry, unbounded mode), and they behave the same.

This one is the narrower change: it keeps the existing meaning of `max_paths=0` (no limit, as on master) and the existing expiry behaviour. #2165 changes both (details on that PR). `test_memory.py` and `test_http.py` pass with this branch.

martindurant:
(closed in favour of #2165)

PR #2099, opened 2026-08-12
Use mv_file for single-file moves

### LINKED ISSUE #2017: Use mv_file instead of copy + rm to implement mv
As discussed in https://github.com/fsspec/gcsfs/pull/800#discussion_r3041091240, we can consider using mv_file to implement mv method.  We might see performance gains if some fsspec implementations include atomic mv_file like we see in case of GCSFS.

aryansk:
Fixes #2017

### Summary
- add a default `mv_file` implementation that preserves the existing copy/delete behavior
- route file-only moves through `mv_file` so filesystems can provide an atomic override
- preserve recursive directory move behavior and add MemoryFileSystem regression coverage

### Validation
- `ruff format fsspec/spec.py fsspec/implementations/tests/test_memory.py`
- `ruff check fsspec/spec.py fsspec/implementations/tests/test_memory.py`
- `pytest fsspec/implementations/tests/test_memory.py -q`
- `pytest fsspec/tests/abstract/mv.py -q`
- `git diff --check`

Prepared with AI assistance; I reviewed the implementation and validation output.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1280,12 +1280,71 @@ class AbstractFileSystem(metaclass=_Cached):
         """Move file(s) from one location to another"""
         if path1 == path2:
             logger.debug("%s mv: The paths are the same, so no files were moved.", self)
+            return
+
+        if isinstance(path1, list) and isinstance(path2, list):
+            # No need to expand paths when both source and destination are provided as lists
+            paths1 = path1
+            paths2 = path2
         else:
-            # explicitly raise exception to prevent data corruption
+            from .implementations.local import trailing_sep
+
+            source_is_str = isinstance(path1, str)
+            paths1 = self.expand_path(
+                path1, recursive=recursive, maxdepth=maxdepth, **kwargs
+            )
+            if source_is_str and (not recursive or maxdepth is not None):
+                # Non-recursive glob does not move directories
+                paths1 = [p for p in paths1 if not (trailing_sep(p) or self.isdir(p))]
+                if not paths1:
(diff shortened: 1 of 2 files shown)


martindurant:
I feel this might have been done - perhaps only for async? Or perhaps in a PR that hasn't been merged? Please check.

aryansk:
Thanks for flagging this. I checked current `master`: `AsyncFileSystem._mv_file` already exists, and GCSFS overrides it for its atomic move path, but synchronous `AbstractFileSystem.mv` still calls `copy` followed by `rm` and there is no corresponding sync `mv_file` hook.

This draft is the sync counterpart: file-only moves go through the new default `mv_file` (which preserves copy/delete behavior and can be overridden by a backend), while directory moves retain the existing recursive copy/remove path. The regression test verifies that `mv` calls an override. If you were referring to a pending sync implementation elsewhere, please point me to it and I'll compare.

fsspec/spec.py
```diff
@@ -1280,12 +1280,71 @@ def mv(self, path1, path2, recursive=False, maxdepth=None, **kwargs):
...
+            exists = source_is_str and (
+                (has_magic(path1) and source_is_file)
+                or (not has_magic(path1) and dest_is_dir and not trailing_sep(path1))
+            )
+            paths2 = other_paths(
+                paths1,
+                path2,
+                exists=exists,
+                flatten=not source_is_str,
+            )
```
martindurant, line 1326: Is this identical code to what it seen in the cp method?

PR #2101, opened 2026-08-15
[bug]  Fix range reads on whole-file caches and crossed ranges in cat_file

nvbkdw:
`filecache` and `simplecache` serve range reads out of the local copy of a file rather than from the remote. Several defects in that path make range reads on a cached file fail, and chasing them down surfaced one more in the shared base-class implementation.

## `WholeFileCacheFileSystem` / `SimpleCacheFileSystem`

**1. Negative `start`/`end` raise `OSError(EINVAL)`**

`AbstractFileSystem.cat_file` documents negative offsets as counting backwards from the end of the file ("like usual python slices"), and remote backends such as `s3fs` implement them as an HTTP suffix range. `WholeFileCacheFileSystem._cat_file` passed the value straight to `f.seek(start)`, which asks for a negative *absolute* position:

```python
fs = fsspec.filesystem("simplecache", target_protocol="s3")
fs.cat_file(path, start=-260)   # OSError: [Errno 22] Invalid argument
```

The failure happens *after* the whole object has been downloaded into the cache, so the local copy that could trivially serve the read is fetched and then discarded.

**2. `AttributeError: '_io.BufferedReader' object has no attribute 'size'`**

The sync path goes through `AbstractFileSystem.cat_file`, which needs `f.size` to resolve a negative offset. Both `_open` implementations here return a bare `io.BufferedReader`, so it raises. `LocalFileOpener` already attaches `size` to the raw file object (`local.py`), so this follows the existing idiom.

**3. `_check_file` return type is not handled consistently**

`WholeFileCacheFileSystem._check_file` returns a `(detail, path)` tuple, while the `SimpleCacheFileSystem` override returns the path alone. `_cat_file` and `_cat_ranges` assume the latter, so on `filecache`:

```python
fs.cat_file(path)          # populate cache
asyncio.run(fs._cat_file(path))    # TypeError: expected str, bytes or os.PathLike object, not tuple
asyncio.run(fs._cat_ranges([path], [2], [5]))  # AttributeError: 'tuple' object has no attribute 'startswith'
```

`_cat_ranges` additionally never downloaded an uncached path, because it tests `fn is None` while the cache-miss sentinel here is `False`.

**4. A range crossed by one byte reads the rest of the file**

`_cat_file` computed the read length as `end - f.tell()`. `read()` treats a negative length as "to the end of the file", so a range crossed by exactly one byte hit that sentinel by accident. Resolving negative offsets to absolute ones (defect 1) makes this reachable from ordinary input:

```python
# 10-byte cached file
asyncio.run(fs._cat_file(path, start=6, end=-5))   # b"6789", expected b""
asyncio.run(fs._cat_file(path, start=8, end=-5))   # ValueError: read length must be non-negative or -1
```

Note the two different failures: a one-byte crossing computes `-1` and silently returns data, anything wider computes `-2` or less and raises.

**5. Repeated paths in `_cat_ranges` resolve to the cache-miss sentinel**

`_cat_ranges` used `rset` purely as a membership set, so only the *first* occurrence of an uncached path was assigned a local path. Every later occurrence appended the falsy sentinel to `lpaths`, and `LocalFileSystem.cat_ranges` returned an exception object in that slot — silently, since `on_error` defaults to `"return"`:

```python
asyncio.run(fs._cat_ranges(["/d1", "/d1", "/d1"], [2, 0, 7], [5, 3, 9]))
# [b'234', AttributeError("'bool' object has no attribute 'startswith'"), AttributeError(...)]
```

Asking for several ranges of one object is the normal access pattern for sharded formats, which is exactly what this PR is about.

## `AbstractFileSystem.cat_file`

Defect 4 is not specific to the caches — the base implementation computes the length the same way, so it is reachable on *every* backend that does not override `cat_file`:

```python
fsspec.filesystem("file").cat_file(fn, start=6, end=-5)   # b"6789", expected b""
```

The offsets were also normalized inconsistently: a negative `start` clamped at 0, a negative `end` did not, so a suffix longer than the file raised instead of yielding empty (`data[:-100]` is `b""`, but `cat_file(end=-100)` raised `ValueError`). And `None` was handled by control flow rather than as a value, giving three different mechanisms for one idea.

`slice.indices` already performs exactly this normalization — clamping included — so both offsets now resolve through it, which is literally what the docstring promises. It needs the size up front and not every file object has one (`TarFileSystem` yields a `tarfile.ExFileObject`, which would then break even a plain full-file read), so the size is consulted only when an offset is actually negative — the same condition under which the old code read `f.size`. No backend loses a case that previously worked.

### How this came up

zarr v3 sharded arrays store the inner-chunk index at the *end* of each shard object, so every read begins with a suffix-range request. Reading such an array over `simplecache::s3://` downloads the full shard and use negative index to read files.

### Changes

`fsspec/implementations/cached.py`

- `_cat_file`: resolve negative `start`/`end` against the file size before seeking, mirroring `AbstractFileSystem.cat_file`, and clamp the read length so a crossed range cannot collide with `read()`'s to-EOF sentinel.
- `_open` (both classes): expose `size` on the returned file object.
- `_cat_file` / `_cat_ranges`: accept either `_check_file` return shape, and treat any falsy value as a cache miss.
- `_cat_ranges`: track the resolved local path per unique path, so repeated paths all resolve.

`fsspec/spec.py`

- `cat_file`: resolve `start`/`end` through `slice.indices` against the file size, consulted only when an offset is negative.

Tests

- Negative offsets, crossed ranges, and cached `cat_ranges` on both `filecache` and `simplecache`, over sync `cat_file` and the awaited `_cat_file` that async consumers (e.g. zarr's `FsspecStore`) use.
- Repeated uncached paths through `_cat_ranges`, against an async target so the download branch is actually exercised.
- `cat_file` start/end combinations checked against `data[start:end]` directly, since python slice semantics are what the docstring promises.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -853,19 +853,24 @@ class AbstractFileSystem(metaclass=_Cached):
         """
         # explicitly set buffering off?
         with self.open(path, "rb", **kwargs) as f:
-            if start is not None:
-                if start >= 0:
-                    f.seek(start)
-                else:
-                    f.seek(max(0, f.size + start))
-            if end is not None:
-                if end < 0:
-                    end = f.size + end
-                # a crossed range reads nothing, like a python slice. Without
-                # the clamp, a range crossed by exactly one byte computes -1,
-                # which ``read`` takes as "to the end of the file"
-                return f.read(max(0, end - f.tell()))
-            return f.read()
+            if (start is not None and start < 0) or (end is not None and end < 0):
+                # A negative offset counts backwards from the end of the file,
+                # so resolve both to absolute positions -- which is precisely
+                # what a python slice does, clamping included. Only this case
+                # needs the file size, and not every file object carries one
+                # (``tarfile.ExFileObject``, for instance), so nothing else
+                # may touch it.
+                start, end, _ = slice(start, end).indices(f.size)
+
+            if start:
+                f.seek(start)
+            if end is None:
+                # no upper bound: read to the end of the file
+                return f.read()
+            # a crossed range reads nothing. Without the clamp, a range
+            # crossed by exactly one byte gives a length of -1, which ``read``
+            # takes as "to the end of the file"
+            return f.read(max(0, end - f.tell()))
 
     def pipe_file(self, path, value, mode="overwrite", **kwargs):
         """Set the bytes of given file"""
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -861,7 +861,10 @@ class AbstractFileSystem(metaclass=_Cached):
             if end is not None:
                 if end < 0:
                     end = f.size + end
-                return f.read(end - f.tell())
+                # a crossed range reads nothing, like a python slice. Without
+                # the clamp, a range crossed by exactly one byte computes -1,
+                # which ``read`` takes as "to the end of the file"
+                return f.read(max(0, end - f.tell()))
             return f.read()
 
     def pipe_file(self, path, value, mode="overwrite", **kwargs):
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -748,9 +748,12 @@ class WholeFileCacheFileSystem(CachingFileSystem):
             # infer the compression from the original filename, like
             # the `TarFileSystem`, let's extend the `io.BufferedReader`
             # fileobject protocol by adding a dedicated attribute
-            # `original`.
+            # `original`. `size` is added for the same reason: callers such
+            # as `AbstractFileSystem.cat_file` need it to resolve negative
+            # offsets, and `LocalFileOpener` exposes it the same way.
             f = open(fn, mode)
             f.original = detail.get("original")
+            f.size = os.fstat(f.fileno()).st_size
             return f
 
         hash = self._mapper(path)
@@ -768,12 +771,26 @@ class WholeFileCacheFileSystem(CachingFileSystem):
         path = self._strip_protocol(path)
         sha = self._mapper(path)
         fn = self._check_file(path)
+        if isinstance(fn, tuple):
+            # `WholeFileCacheFileSystem._check_file` returns `(detail, fn)`;
+            # the `SimpleCacheFileSystem` override returns just the path.
+            _, fn = fn
 
         if not fn:
             fn = os.path.join(self.storage[-1], sha)
             await self.fs._get_file(path, fn, **kwargs)
 
         with open(fn, "rb") as f:  # noqa ASYNC230
+            if (start is not None and start < 0) or (end is not None and end < 0):
+                # Negative offsets count backwards from the end of the file, as
+                # documented on ``AbstractFileSystem.cat_file``. They must be
+                # resolved against the file size first: seeking to a negative
+                # absolute position raises ``OSError(EINVAL)``.
+                file_size = os.fstat(f.fileno()).st_size
+                if start is not None and start < 0:
+                    start = max(0, file_size + start)
+                if end is not None and end < 0:
+                    end = max(0, file_size + end)
             if start:
                 f.seek(start)
             size = -1 if end is None else end - f.tell()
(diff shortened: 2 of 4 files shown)


fallenmi (changes requested):
`filecache` still does not record files downloaded through the async range paths, so the main async/Zarr use case repeatedly fetches the full remote object instead of serving later ranges from the cache.

On exact head `95df12a`, I wrapped `MemoryFileSystem.get` / `get_file` with counters and awaited the same read twice. For `WholeFileCacheFileSystem`, both `_cat_file` and `_cat_ranges` made one backend download per invocation (counts `1` then `2`), and `_check_file("/afile")` remained `False` after each successful download. The same probes on `SimpleCacheFileSystem` made one download total and then returned the cached local path.

The miss branches currently write directly to the hash path but never create/save the metadata that `WholeFileCacheFileSystem._check_file()` requires. This also means the new repeated-range test's comment “again now that the file is cached” is false for its `filecache` parameter: the bytes pass because the second invocation silently downloads the object again.

Please preserve the `filecache` metadata contract after a successful async download and assert the backend download count across a second `_cat_file` and `_cat_ranges` call. That is important here because sharded readers repeatedly request suffix/range data from the same full object.

For context, I independently confirmed the intended functional fixes: exact base returns errors/wrong bytes for the added cases, while this head returns the expected ranges. Both touched test files pass (`151 passed, 17 skipped`), as do pinned Ruff 0.14.3 check/format, compileall, and `git diff --check`.

Disclosure: OpenAI Codex assisted with this independent review and the deterministic probes above.

fsspec/spec.py
```diff
@@ -853,16 +853,24 @@ def cat_file(self, path, start=None, end=None, **kwargs):
...
+
+            if start:
+                f.seek(start)
+            if end is None:
+                # no upper bound: read to the end of the file
+                return f.read()
+            # a crossed range reads nothing. Without the clamp, a range
+            # crossed by exactly one byte gives a length of -1, which ``read``
+            # takes as "to the end of the file"
+            return f.read(max(0, end - f.tell()))
```
martindurant, line 873: We know that read(0) should return b"", but I'm not convinced that all backends will honour this.

You are principally guarding against `end < start`? If that's essentially the task here, it seems to me this could be simpler.

itzzdev09:
Reviewed this carefully, including tracing the offset math by hand against real Python slice semantics (crossed ranges, deeply-negative starts, out-of-range suffixes, mixed-sign start/end) — it holds up in every case I checked. The `_check_file` tuple/non-tuple normalization is correct for both `WholeFileCacheFileSystem` and `SimpleCacheFileSystem`, verified against `CacheMetadata.check_file`'s actual return contract (`False | tuple[Detail, str]`). The `_cat_ranges` repeated-path fix is a genuine correctness improvement — the old `rset`-based dedup left repeated occurrences of an uncached path pointing at the pre-download placeholder instead of the resolved local file.

One small thing worth considering: `_cat_file`'s negative-offset handling in `cached.py` hand-rolls the same clamping (`max(0, file_size + start)` / `max(0, file_size + end)`) that `spec.py::cat_file` now does more robustly by delegating to `slice(start, end).indices(f.size)`. Both are correct today, but it's now two independently-maintained implementations of the same "resolve like a Python slice" rule — a future edge-case fix to one is easy to forget to apply to the other. Might be worth having the async path build a `slice(...).indices(...)` call too, or extracting a small shared helper, just to remove the duplication risk. Not blocking either way — nice fix overall.

martindurant:
> worth having the async path build a slice(...).indices(...) call too

Yes indeed, and it's faster too

martindurant:
(ping me when you have done the slice thing and there are now conflicts too)

itzzdev09 (approved):
Confirmed each of the three failures on `master`, with both `filecache` and `simplecache`:

- `await fs._cat_file("/afile", start=-3)` on a cached file raises `OSError(EINVAL)` from the negative seek.
- A crossed range such as `start=2, end=-2`, or `start=6, end=5`, raises `ValueError: read length must be non-negative or -1`, since `end - f.tell()` lands on exactly `-1`.
- The cached read path exposes no `size`, so `AbstractFileSystem.cat_file` cannot resolve a negative offset through it.

All of them behave correctly with this change, and the full `test_cached.py` suite passes.

One thing worth keeping in mind for later: the `isinstance(fn, tuple)` unpacking is a trap for the next caller, since the base `_check_file` returns `(detail, fn)` while the `SimpleCacheFileSystem` override returns a path, and this PR now has to handle that in two more places. A follow-up making the two agree would let those checks go away.

PR #2104, opened 2026-08-17
Implement created and modified methods for SFTPFileSystem

hallhagen:
This PR is based on (and extremely similary to) [PR 1971](https://github.com/fsspec/filesystem_spec/pull/1971).

In the same way as that PR, this implementation returns the modified timestamp when `created` is called as well, as there is no ctime returned by `info`. 

The only difference to the WebHDFS implementation is that the values returned by `info` already are datetime objects, so no conversion is necessary.

### DIFF
--- a/fsspec/implementations/tests/test_sftp.py
+++ b/fsspec/implementations/tests/test_sftp.py
@@ -2,6 +2,7 @@ import os
 import shlex
 import subprocess
 import time
+from datetime import datetime, timezone
 from tarfile import TarFile
 
 import pytest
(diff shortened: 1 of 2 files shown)


fsspec/implementations/tests/test_sftp.py
```diff
@@ -246,3 +247,69 @@ def test_makedirs_exist_ok(ssh, path):
...
+    with f.open(file_path, "wb") as wf:
+        wf.write(b"test content")
+
+    modified_file_date: datetime = f.modified(file_path)
+    assert modified_file_date >= modified_dir_date
+    assert modified_file_date <= datetime.now(timezone.utc)
+
+
+# NOTE: These following two tests are a copy of the modified ones, as
+# WebHDFS does not have a created time API, we are using modified as a proxy.
```
martindurant: Should not be the case for sftp, right?

hallhagen: It unfortunately is, paramiko only provides `st_atime` and `st_mtime`, and not `st_ctime`. (Paramiko actually states that it might not even return those values, as _"An SFTP server may return as much or as little info as it wants, so the results may vary from server to server"_)

Sorry about leaving in this comment referring to WebHDFS, I have updated it to clarify the values returned from paramikos stat call.

Issue #1851, opened 2025-05-30
Wrong behaviour of fs.get in "Directory to existing directory" case


nsundalov:
**Problem**:
fs.get copies source dir inside target dir but expected to copy content of source dir into target dir.
[case 1e of the doc](https://filesystem-spec.readthedocs.io/en/latest/copying.html#copying-files-and-directories)

Script to reproduce error:
```python
import os

from fsspec import filesystem
from fsspec.implementations.dirfs import DirFileSystem

DATA_PATH = "dir"
REMOTE_PATH = "data/remote"
LOCAL_PATH = "data/local"


def init_test():
    # remote
    os.makedirs(REMOTE_PATH, exist_ok=True)
    data_dir = os.path.join(REMOTE_PATH, DATA_PATH)
    os.makedirs(data_dir, exist_ok=True)
    with open(os.path.join(data_dir, "f.txt"), "w") as f:
        f.write("hello world")

    # local
    os.makedirs(LOCAL_PATH, exist_ok=True)


def test():
    fs = filesystem("file")
    fs = DirFileSystem(REMOTE_PATH, fs)

    # copy from data/remote/dir to data/local/dir
    # first leads to data/local/dir/dir/f.txt if called 2 times
    fs.get(DATA_PATH, os.path.join(LOCAL_PATH, DATA_PATH), recursive=True)


init_test()
test()
```

After calling script 2 times I have following dir structure:
```
data
├── local
│   └── dir
│       └── dir
│           └── f.txt
└── remote
    └── dir
        └── f.txt
```

Expected to have following:
```
data
├── local
│   └── dir
│       └── f.txt
└── remote
    └── dir
        └── f.txt
```

PR #2110, opened 2026-08-31
Preserve the source trailing slash in `DirFileSystem._join`

PhuocOng:
Addresses #1851.

The nesting reported in that issue is not specific to `DirFileSystem` — local and memory filesystems do the same, matching the documented semantics for a source without a trailing slash (copying.rst, 1e). But it did surface a dirfs-specific bug: `_join` runs paths through `_strip_protocol`, which strips a trailing slash, so the documented way to ask for "copy the contents" is silently lost. On a second `get` of the same directory, dirfs nests a fresh copy inside the target where every other filesystem copies the contents:

```python
>>> dirfs.get("dir/", "local/dir", recursive=True)  # local/dir/f.txt
>>> dirfs.get("dir/", "local/dir", recursive=True)  # local/dir/dir/f.txt without the fix
```

The fix keeps the caller's trailing slash across the join. Single-path operations are unaffected since the wrapped filesystem's own `_strip_protocol` normalizes them anyway.

Added tests `test_join_preserves_trailing_slash` and `test_get_directory_contents_to_existing_directory` in `test_dirfs.py`; both fail without the fix.

### DIFF
--- a/fsspec/implementations/dirfs.py
+++ b/fsspec/implementations/dirfs.py
@@ -68,6 +68,10 @@ class DirFileSystem(AsyncFileSystem, ChainedFileSystem):
                 return path
             if not path:
                 return self.path
+            # a trailing slash requests "copy the contents" in bulk
+            # operations (docs/source/copying.rst), so preserve it across the
+            # join even though _strip_protocol removes it
+            suffix = "/" if path.endswith("/") else ""
             path = self._strip_protocol(path)
             # ".." only navigates above the root on filesystems that resolve it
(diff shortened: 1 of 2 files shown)


itzzdev09 (approved):
Checked the join against both memory and local: `_join("a/")` keeps the slash, `_join("/")`, `_join("")` and `..` are unchanged, and a trailing slash reaching the wrapped filesystem on a file path is harmless (`cat`, `info` and `open` on `a/b/` behave the same before and after on local). `_relpath` round-trips through the join, and the full dirfs suite passes.

PR #2117, opened 2026-09-06
Allow per-instance filesystem protocols

### LINKED ISSUE #1800: Don't type `protocol` as `ClassVar`
Protocol is defined as a `ClassVar`:
https://github.com/fsspec/filesystem_spec/blob/ffdfed146f070cd05bce78767fc9873d165c59fd/fsspec/spec.py#L109

This means that pylance gives me an error when I try to set it in a constructor:

Image

Typing protocol as a `ClassVar` makes sense if you assume that you have separate, concrete classes for each implementation, but this is not always true. In [Obstore](https://github.com/developmentseed/obstore), we have **one class** (`FsspecStore`) which dynamically supports different protocols based on the type of `obstore.store` used inside it.

be-student:
Closes #1800

`AbstractFileSystem.protocol` is currently declared as a `ClassVar`, so type checkers reject filesystem adapters that choose their protocol from constructor state. This removes the `ClassVar` restriction while preserving the existing string-or-tuple type and class default.

Concrete implementations with a fixed protocol can continue declaring their own `ClassVar`. The changelog now documents support for per-instance protocol selection.

Validation:

- The issue-shaped Pyright fixture fails before the change with `Attribute "protocol" cannot be assigned through a class instance because it is a ClassVar` and passes afterward with 0 errors.
- `fsspec/tests/test_spec.py` portable tests: 186 passed, 10 skipped, 1 xfailed. The 52 excluded Bash-reference cases require GNU `stat -c` and fail unchanged on macOS.
- Ruff check, Ruff format check, and `git diff --check` pass.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,8 @@ Dev
 
 Fixes
 
+- Allow filesystem implementations to assign ``protocol`` per instance
+
 - Avoid mutating live ``BlockCache`` and ``BackgroundBlockCache`` instances
   when pickling (#2102)
 
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -10,7 +10,7 @@ import weakref
 from errno import ESPIPE
 from glob import has_magic
 from hashlib import sha256
-from typing import Any, ClassVar
+from typing import Any
 
 from .callbacks import DEFAULT_CALLBACK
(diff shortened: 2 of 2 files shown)


martindurant:
OK, I suppose so; this is typing getting things wrong. It it a class variable *usually*....

PR #2118, opened 2026-09-06
Add age-based cleanup for simplecache

### LINKED ISSUE #2105: simplecache retains written files indefinitely
Hi, we ran into a problem where processes would ran out of disk space because `SimpleCacheFileSystem` keeps the `LocalTempFile`s forever. 

It is also not quite obvious how to clear the cache because `clear_cache()` has the potential to remove currently open files and also might need a call to `_mkcache()` afterwards.

There is this comment in `LocalTempFile` that says 
```
# we do not delete the local copy, it's still in the cache.
```
but it does not really provide a solution. 

Happy to provide a PR if there is a clear direction how the project would like to solve this (if it wants to).

be-student:
Fixes #2105.

`SimpleCacheFileSystem` does not persist cache metadata, so its inherited `clear_expired_cache()` had no records to inspect and silently retained every cached file. This override uses each cached file's local modification time, requires an explicit non-negative maximum age, and only modifies the writable cache location.

The regression tests cover files retained after writes, selective age-based removal, and invalid expiry arguments. Cache lookups in the regression test are normalized through the wrapped filesystem before `os.utime`, matching the cache-key contract on Windows. The feature documentation and changelog now describe manual age-based cleanup.

Validation at `b6f8f8054c97139249fa932098ad7a6241835c96`:

- focused expiry tests: 2 passed, 118 deselected
- complete cached-filesystem module: 89 passed, 31 skipped
- all pre-commit hooks on changed files passed, including Ruff and codespell
- `git diff --check` passed
- full suite: 1788 passed, 170 skipped, 2 xfailed, 52 failed

All 52 full-suite failures are the same macOS baseline failures in the unchanged `test_posix_tests_bash_stat` parameterizations. Those tests invoke GNU `stat -c %N`; macOS BSD `stat` does not support `-c`, so the command produces no matching output. No Windows manual test was performed.

Hosted CI:

- Windows passed, including the regression that previously failed before cache-path normalization
- Python 3.10, 3.11, 3.12, and 3.14 passed
- Python 3.13 ran 2192 passing tests and failed only six unchanged `GitHubFileSystem` tests because unauthenticated GitHub API requests returned HTTP 403 rate-limit responses; these failures are unrelated to this branch
- Python 3.14 free-threaded passed

AI assistance was used for implementation and testing.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,8 @@ Dev
 
 Fixes
 
+- Allow ``simplecache`` files to be removed by age using their modification times
+
 - Avoid mutating live ``BlockCache`` and ``BackgroundBlockCache`` instances
   when pickling (#2102)
 
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -289,6 +289,9 @@ should not mix block- and file-caches in the same directory. "simplecache" is th
 except without the options for cache expiry and to check the original source - it can be used where the
 target can be considered static, and particularly where a large number of target files are expected
 (because no metadata is written to disc). Only "simplecache" is guaranteed thread/process-safe.
+Cached files can still be removed by age using their local modification times, for example
+``fs.clear_expired_cache(expiry_time=7 * 24 * 60 * 60)``. This only changes the writable cache
+location when multiple cache directories are configured.
 
 Remote Write Caching
 --------------------
--- a/fsspec/implementations/cached.py
+++ b/fsspec/implementations/cached.py
@@ -847,6 +847,38 @@ class SimpleCacheFileSystem(WholeFileCacheFileSystem):
     def load_cache(self):
         pass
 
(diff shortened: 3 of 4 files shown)


martindurant:
Runs into the following issue on windows (seen in the test):
```
>       os.utime(old_cached, (1, 1))
E       TypeError: utime: path should be string, bytes or os.PathLike, not NoneType
```

be-student:
Rebased onto current `master` and revalidated the Windows cache-path fix. The hosted Windows job now passes, as do Python 3.10, 3.11, 3.12, 3.14, and free-threaded Python 3.14. Python 3.13 reached 2192 passing tests; its only six failures are unchanged `GitHubFileSystem` tests where unauthenticated GitHub API requests returned HTTP 403 rate-limit responses, so they are unrelated to this branch.

martindurant:
> GitHub API requests returned HTTP 403 rate-limit responses

We've been hitting CI pretty hard...

PR #2120, opened 2026-09-07
Close underlying files when using compression

### LINKED ISSUE #1672: Context not preserve when open file with compression
## Context 
- When providing the `compression` parameter to `fsspec.open`, the underlying Filesystem object is not properly cleaned up (even with context).
- This causes any errors related to the Filesystem object cleanup to be suppressed and show up during garbage collection instead.

### Reproducible code
```python
import fsspec

def foo():
  with fsspec.open('s3://non-access-bucket/non-access-data.gz', mode = 'wb', compression = 'gzip') as f:
      f.write(b'hello world')

foo()
``` 
The above code snippet will happily run (assuming s3 credentials are setup properly etc) and function exits successfully, but later when the file object destructor get called, we will see the following exception

### Exception
```
Exception ignored in: <funciton AbstractBufferredFile.__del__ at 0x7bc52f060040>
...
self.close
...
self.flush(force=True)
... snip ...
PermissionError: Access Denied
```

I have not tried with any other FS yet but I reckon as long as we can trigger the error during flush/close, we should be able to reproduce the error.

### Temporary work around
- Separate fsspec open and compression library opener.

### Relevant library code
```python
            f = self._open(
                path,
                mode=mode,
                block_size=block_size,
                autocommit=ac,
                cache_options=cache_options,
                **kwargs,
            )
            if compression is not None:
                from fsspec.compression import compr
                from fsspec.core import get_compression

                compression = get_compression(path, compression)
                compress = compr[compression]
                f = compress(f, mode=mode[0]) # Here
```

Jokasa7:
Closing a file returned by `AbstractFileSystem.open(..., compression=...)` can leave the underlying filesystem file open. For codecs such as gzip, bz2 and lzma, closing the compression stream does not close a caller-provided file. This delays resource release and can defer a write/upload error until garbage collection.

Keep ownership of both streams in a private wrapper. Close the compression stream first to finish its trailer, then close the underlying file in `finally`. Preserve the unwrapped object when compression inference selects no codec, and keep transaction queues pointing to the raw file.

The returned object retains the file-like interface but is a wrapper, so its concrete type is no longer the codec class. This avoids replacing methods on codec objects, including C extension streams that do not allow setting `close`.

Fixes #1672.

Validation on Python 3.12.13 / Windows:

- The original 30 new regression cases fail against unmodified `master`: 24 leave the raw file open and six fail to propagate its close error.
- The final compression module tests pass: 68 passed, 1 skipped (optional snappy dependency unavailable). Coverage includes binary/text reads and writes, direct/context-manager close, compression/raw close failures, transaction commit/discard, iteration/seek/readinto, no-codec identity, and codecs that already close their input.
- The affected test set reports 491 passed, 135 skipped, 10 xfailed and 2 failed. Both failures are existing `test_glob_weird_characters` cases that try to create a directory containing `|`, which Windows rejects with `WinError 123`; the same failures were reproduced on unmodified `master`.
- Ruff check/format, `git diff --check`, and a full Sphinx HTML build with `-W --keep-going` pass. The optional full cloud/backend and downstream integration suites were not run locally.

AI assistance: this patch and its regression tests were developed and validated with OpenAI Codex, including a separate code review. The test results above are from commands executed by Codex.

### DIFF
--- a/fsspec/spec.py
+++ b/fsspec/spec.py
@@ -1417,12 +1417,14 @@ class AbstractFileSystem(metaclass=_Cached):
             if not ac and "r" not in mode:
                 self.transaction.files.append(f)
             if compression is not None:
-                from fsspec.compression import compr
+                from fsspec.compression import _ClosingFile, compr
                 from fsspec.core import get_compression
 
                 compression = get_compression(path, compression)
                 compress = compr[compression]
-                f = compress(f, mode=mode[0])
+                compressed = compress(f, mode=mode[0])
+                if compressed is not f:
+                    f = _ClosingFile(compressed, f)
             return f
 
     def touch(self, path, truncate=True, **kwargs):
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,9 @@ Dev
 
 Fixes
 
+- Close the underlying file when closing a stream opened with compression
+  through ``AbstractFileSystem.open``, and propagate errors raised on close (#1672)
+
 - Avoid mutating live ``BlockCache`` and ``BackgroundBlockCache`` instances
   when pickling (#2102)
 
--- a/fsspec/compression.py
+++ b/fsspec/compression.py
@@ -11,6 +11,43 @@ def noop_file(file, mode, **kwargs):
     return file
 
 
+class _ClosingFile:
+    """A compression stream that owns its underlying file."""
+
+    def __init__(self, file, raw):
+        self._file = file
+        self._raw = raw
+
+    def __getattr__(self, name):
+        return getattr(self._file, name)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        return next(self._file)
+
+    def __enter__(self):
+        self._file.__enter__()
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def close(self):
+        try:
+            # Finalize the compression stream before flushing the raw file.
+            if not self._file.closed:
+                self._file.close()
+        finally:
+            # Some codecs already close their input, while others leave it open.
+            if not self._raw.closed:
+                self._raw.close()
+
+    def __del__(self):
+        self.close()
+
+
 # TODO: files should also be available as contexts
 # should be functions of the form func(infile, mode=, **kwargs) -> file-like
 compr = {None: noop_file}
--- a/fsspec/tests/test_compression.py
+++ b/fsspec/tests/test_compression.py
@@ -1,3 +1,4 @@
+import io
 import pathlib
 import sys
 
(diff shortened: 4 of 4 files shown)


itzzdev09:
Went through this in detail — no issues found. Specifically checked:

- **Close ordering**: `_ClosingFile.close()` finalizes the compression stream first (so the trailer/CRC gets flushed) then closes the raw file in a `finally`, with independent `if not X.closed` guards on each, so a failure finalizing the compression stream (tested by `test_fs_open_closes_raw_file_on_compression_error`) still guarantees the raw file gets closed rather than leaking it.
- **Proxy correctness**: `__getattr__` forwards to the wrapped stream, but `__iter__`/`__next__`/`__enter__`/`__exit__` are all explicitly defined rather than relying on `__getattr__` — correctly sidesteps the usual gotcha where Python's special-method lookup bypasses `__getattr__` for dunders, so `for line in f` and `with f:` actually work through the wrapper.
- **Transaction interaction**: `self.transaction.files.append(f)` in `spec.py` runs *before* the compression wrap, so the transaction ends up tracking the raw pre-compression file object, not the `_ClosingFile` the caller writes through. Confirmed this still works correctly, though: `_ClosingFile.close()` closes that same raw object via `self._raw`, so `transaction.files[0].closed` is genuinely `True` after the wrapper closes, matching what `test_fs_open_compression_transaction` checks.
- **Tests are real**: reverted just the `spec.py` change (kept `_ClosingFile` itself) and reran `test_compression.py` — 39 of the new tests fail immediately without the fix, so this is genuine regression coverage, not decoration.

Good fix, nice test coverage (including the zstandard third-party-codec case, which has a different close-ownership shape than the stdlib gzip/bz2/lzma wrappers).

fsspec/spec.py
```diff
@@ -1417,12 +1417,14 @@ def open(
...
-                from fsspec.compression import compr
+                from fsspec.compression import _ClosingFile, compr
                 from fsspec.core import get_compression
 
                 compression = get_compression(path, compression)
                 compress = compr[compression]
-                f = compress(f, mode=mode[0])
+                compressed = compress(f, mode=mode[0])
+                if compressed is not f:
+                    f = _ClosingFile(compressed, f)
```
martindurant: Wouldn't it make sense to put the call to _ClosingFile inside compress() ?

itzzdev09:
Worth flagging one snag with moving it into `compress()`: `compress` is whatever is registered in the `compr` registry, including third-party codecs added through `register_compression`, so each of those would have to do the wrapping itself to get the behaviour.

Putting it in the registry lookup instead would reach the other call sites that read `compr[...]` directly and already manage the raw file themselves: `core.py:120`, where `OpenFile` appends both objects to `self.fobjects` and closes them in reverse on exit; `cached.py:386/715/941`; `local.py:397`, where `LocalFileOpener` keeps `self.f`; and `tar.py:81`, where `fo` belongs to the caller. `OpenFile` would end up closing the raw file twice, and `tar.py` would close a file it doesn't own.

Keeping the wrapping at the `AbstractFileSystem.open` call site is what confines the ownership change to the one place that creates the raw file, which seems right to me.

martindurant:
You make a good point, but then we should also consider other filesystem-specific open() implementations that accept compression= ? Maybe there are none (because _open is usually what is used).

itzzdev09:
You're right that there are none — `DirFileSystem.open` is the only override, and it forwards to the wrapped filesystem, so it inherits this. Everything else, here and in s3fs/gcsfs/adlfs, only implements `_open`.

So the base-class call site covers everything today, and I'd keep it there. One improvement worth making: give `compression.py` a small helper, say `open_compressed(f, compression, mode)`, holding the `compress(...)` call and the `_ClosingFile` wrapping, and have `AbstractFileSystem.open` call that. Then the ownership rule lives next to `_ClosingFile` rather than being spelled out at the call site, and anything that does end up needing it later is a one-liner rather than a copy of the same three lines. The other `compr[...]` users stay untouched.

martindurant:
Yes, I like moving it to compression.py

itzzdev09:
😃😃

martindurant:
(ping me when you've done that. I can handle the conflict if you like - it's just the changelog)

itzzdev09:
shall i go for it? happy to do it myself if @Jokasa7 doesnt, nd if u need it quick
@martindurant

martindurant:
no rush

itzzdev09:
Sounds good — I'll leave it with @Jokasa7 then, it's their patch.

@Jokasa7 if you'd rather not pick up the refactor, just say and I'll open a follow-up. Otherwise it's a small move: lift the compress(...) call and the _ClosingFile wrap into a helper in compression.py and have AbstractFileSystem.open call that — the other compr[...] callers stay as they are.

PR #2121, opened 2026-09-08
Add Internet Archive filesystem implementation

lfoppiano:
Adds `InternetArchiveFileSystem` (`fsspec/implementations/ia.py`, protocol ia): `ia://<identifier>/<filename>` for files in archive.org items, plus registry entry, API docs, changelog and tests.

It remap to HTTP (HTTPFileSystem) with a path mapping and credentials, from https://archive.org/download/<identifier>/<filename>, which redirects to a data node that support HTTP Range queries. That is the same URL the `internetarchive` package downloads from.

**S3-like API** (s3.us.archive.org, which an S3FileSystem + endpoint_url would use): Internet Archive data nodes ignore Range and answer a ranged GET with 200 and the whole file, so every block read would download the entire object. Measured on a 1.6 GB public item (Range: bytes=0-99 on every request):

| Endpoint                               | URL                                                | Status                                       | Bytes sent    |
|--- |--- |--- | ---|
| S3-like front end                      | https://s3.us.archive.org/<item>/<file>            | 307 → http://dnNNN.s3dns.us.archive.org:80/… | 410 (XML)     |
| S3-like data node (following the 307)  | http://dnNNN.s3dns.us.archive.org/<item>/<file>    | 200, Accept-Ranges: bytes, no Content-Range  | 1,689,855,303 |
| Download endpoint                      | https://archive.org/download/<item>/<file>         | 302 → https://dnNNN.ca.archive.org/0/items/… | 0             |
| Download data node (following the 302) | https://dnNNN.ca.archive.org/0/items/<item>/<file> | 206, Content-Range: bytes 0-99/1689855303    | 100           |

IA documents this: "HTTP 1.1 Range headers are ignored" (https://archive.org/developers/ias3.html#how-this-is-different-from-normal-s3). botocore also misreads the 307 as an AWS region redirect and recurses on HeadBucket until RecursionError.

Credentials. Public items need none. For restricted items the class reads the ia.ini written by ia configure, found the way the internetarchive package finds it ($IA_CONFIG_FILE, $XDG_CONFIG_HOME/internetarchive/ia.ini, ~/.config/ia.ini, ~/.ia), with IA_ACCESS_KEY_ID/IA_SECRET_ACCESS_KEY overriding the file. 
Cookies go into the session's cookie jar scoped to .archive.org rather than a Cookie header, because every download is a cross-origin redirect and aiohttp drops Cookie/Authorization headers there; the jar re-attaches them to the data node. 401/403 raise PermissionError (otherwise HTTPFileSystem._info reports a 403 as FileNotFoundError).

Tested with `fsspec/implementations/tests/test_ia.py`.

### DIFF
--- a/fsspec/implementations/ia.py
+++ b/fsspec/implementations/ia.py
@@ -1,0 +1,226 @@
+"""Files in Internet Archive items, ``ia://<identifier>/<filename>``."""
+
+import configparser
+import functools
+import os
+from dataclasses import dataclass, field
+from http.cookies import SimpleCookie
+
+import aiohttp
+import yarl
+
+from ..utils import stringify_path
+from .http import HTTPFileSystem, get_client
+
+DOWNLOAD_URL = "https://archive.org/download/"
+COOKIE_DOMAIN = ".archive.org"
+# The environment variable names the ``internetarchive`` package uses.
+ENV_ACCESS_KEY = "IA_ACCESS_KEY_ID"
+ENV_SECRET_KEY = "IA_SECRET_ACCESS_KEY"
+ENV_CONFIG_FILE = "IA_CONFIG_FILE"
+
+
+def ia_config_path():
+    """The ``ia.ini`` written by ``ia configure``, or None.
+
+    Searched the way the ``internetarchive`` package searches: ``$IA_CONFIG_FILE``,
+    ``$XDG_CONFIG_HOME/internetarchive/ia.ini`` (``XDG_CONFIG_HOME`` defaulting to
+    ``~/.config``), ``~/.config/ia.ini``, ``~/.ia``. The first existing file wins.
+    """
+    home = os.path.expanduser("~")
+    xdg = os.environ.get("XDG_CONFIG_HOME") or os.path.join(home, ".config")
+    candidates = [
+        os.environ.get(ENV_CONFIG_FILE),
+        os.path.join(xdg, "internetarchive", "ia.ini"),
+        os.path.join(home, ".config", "ia.ini"),
+        os.path.join(home, ".ia"),
+    ]
+    return next((p for p in candidates if p and os.path.isfile(p)), None)
+
+
+@dataclass(frozen=True)
+class IACredentials:
+    """What logs a request in at archive.org; all-empty means anonymous."""
+
+    access_key: "str | None" = None
+    secret_key: "str | None" = None
+    cookies: dict = field(default_factory=dict)  # logged-in-user / logged-in-sig
+    config_file: "str | None" = None  # where they were read from
+
+    @property
+    def anonymous(self):
+        return not (self.access_key or self.cookies)
+
+
+def load_ia_credentials(config_file=None):
+    """Credentials from ``ia.ini`` and the environment; anonymous when there are none.
+
+    ``[s3] access/secret`` and ``[cookies] logged-in-user/logged-in-sig`` are read
+    from ``config_file`` (default: :func:`ia_config_path`). Cookie values in the file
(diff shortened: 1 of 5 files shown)


martindurant:
This looks like fun - please let me know when you are ready. Are you affiliated with IA in any way? If not, should we approach someone for their opinion?

lfoppiano:
@martindurant thank you for reaching out. I'm from the Common Crawl Foundation. 
We've started discussing with the IA about it and I still have to do some testing, I let you know as soon as it's ready 🙂

wumpus:
I asked Mark Graham @ IA a week ago and he did not reply. (I used to work there.)

We want to stress test this PR and also PR #2124 -- I have a 500 TByte dataset of ~10 GByte files that I want to read from IA that ought to be a good stress test.

fsspec/implementations/ia.py
```diff
@@ -0,0 +1,226 @@
+"""Files in Internet Archive items, ``ia://<identifier>/<filename>``."""
```
wumpus, line 1: The name for this in IA jargon is "item", not "identifier".

fsspec/implementations/ia.py
```diff
@@ -0,0 +1,226 @@
...
+    if env_access:
+        access_key, secret_key = env_access, env_secret
+    if not (access_key and secret_key):
+        access_key = secret_key = None
+    found = path if path and os.path.isfile(path) else None
+    return IACredentials(access_key, secret_key, cookies, found)
+
+
+class InternetArchiveFileSystem(HTTPFileSystem):
+    """Files in Internet Archive items, addressed as ``ia://<identifier>/<filename>``.
```
wumpus, line 95: ditto

fsspec/implementations/ia.py
```diff
@@ -0,0 +1,226 @@
...
+                morsels[name]["domain"] = domain
+                morsels[name]["path"] = "/"
+            session.cookie_jar.update_cookies(
+                morsels, yarl.URL(f"https://{domain.lstrip('.')}/")
+            )
+        return session
+
+    @classmethod
+    def _strip_protocol(cls, path):
+        """``ia://item/file`` (or bare ``item/file``) becomes the download URL.
```
wumpus, line 181: Ahah! 😹

wumpus:
@martindurant I have asked jjjake to weigh in.

jjjake:
@wumpus Thanks for the heads up on this.

My only suggestion is to use IA-S3 keys via Authorization header for everything, and not use cookies at all. You'll just need to make sure that the Authorization header survives the redirect to the data node. Everything else looks good to me.

I'll pass this along to the rest of our team in case anyone else has feedback, but the rest looks good to me! Thanks for submitting this, lfoppiano!

lfoppiano:
@jjjake thanks for your feedback. I've replaced the cookies auth with the authorization header.  

@wumpus thanks for the corrections, I've fixed the naming.

martindurant:
The failures seem to be real and persistent - but I don't see what they have to do with changes here. 

Perhaps update the branch to see if these were already fixed?

wumpus:
`assert ia_config_path() == str(xdg_default)` looks like it was caused by this PR.

That socket thing, I don't know.

@lfoppiano do you want to try to see if you can get Windoze and MacOS working in CI for `warc2zip`?  You can use `warcio` CI as an example.

wumpus:
... although @lfoppiano on 3rd thought perhaps you can directly debug this `ia_config_path()` thing and see if it's causing the other failures.

lfoppiano:
@martindurant @wumpus there was a small glitch on the userprofile mocking setup in the test.

Issue #1895, opened 2025-07-23
`HTTPFileSystem` reads an incorrect amount of data


bveeramani:
Here's a minimal-ish repro:
```python
from fsspec.implementations.http import HTTPFileSystem

remote_path = "https://huggingface.co/api/datasets/abisee/cnn_dailymail/parquet/3.0.0/train/0.parquet"

expected_data_size = 256540614

filesystem = HTTPFileSystem()
with filesystem.open(remote_path) as file:
    total_read = 0
    while data := file.read(256 * 1024):
        total_read += len(data)

assert (
    total_read == expected_data_size
), f"Data mismatch: {total_read} != {expected_data_size}"
# AssertionError: Data mismatch: 5767168 != 256540614
```

This issue causes Ray Data's `from_huggingface` API to break. See https://github.com/ray-project/ray/issues/54101

martindurant:
Thanks for reporting. I am looking into it. 

I notice that the URL redirects twice before getting to the actual download (cloudflare "intelligent tiering", apparently), so my first guess is that the final URL expires between starting the first read and the read that returns no bytes. A workaround is `fs.open(remote_path, block_size=0)`, which gets the bytes in streaming mode, with a connection that stays alive - but would not give you random-access to pull out the parquet columns.

Issue #550, opened 2021-03-01
ServerDisconnectedError and Cannot seek streaming HTTP file ValueError in version 0.8.7  and  0.8.4


jkellndorfer:
I wonder if @rsignell-usgs  or @martindurant have insights into this issue:

I have the following code snippet that worked in fsspec version 0.8.4 (for TEST 2):

```
import netrc
import fsspec
import aiohttp

(username, account, password) = netrc.netrc(file='/home/ubuntu/.netrc').authenticators("urs.earthdata.nasa.gov")
cloud_fs = fsspec.filesystem('http', client_kwargs={'auth': aiohttp.BasicAuth(username, password)})

# TEST 1
url='https://e4ftl01.cr.usgs.gov/MEASURES/NASADEM_HGT.001/2000.02.11/NASADEM_HGT_n00e006.zip'
ftmp=cloud_fs.open(url)
z=fsspec.filesystem('zip', fo=ftmp)
print('\n'.join(z.find('/')))

# TEST 2
url='https://datapool.asf.alaska.edu/SLC/SB/S1B_IW_SLC__1SDH_20201106T085415_20201106T085442_024142_02DE43_E9A2.zip'
ftmp=cloud_fs.open(url)
z=fsspec.filesystem('zip', fo=ftmp)
print('\n'.join(z.find('/')))
```

There are two issues:
TEST 1 produces a `ServerDisconnectedError`, but on a *second run* returns the expected output. This happens in both tested versions, 0.8.4 and 0.8.7
TEST 2 produces the correct output in version 0.8.4 (on the first run) but fails with e `ValueError: Cannot seek streaming HTTP file` in version 0.8.7

Should I approach this differently? 
I attach the tracebacks below. Thanks!

# ServerDisconnectedError
```
---------------------------------------------------------------------------
ServerDisconnectedError                   Traceback (most recent call last)
<ipython-input-3-1bac1fbcb8ec> in <module>
      1 url='https://e4ftl01.cr.usgs.gov/MEASURES/NASADEM_HGT.001/2000.02.11/NASADEM_HGT_n00e006.zip'
      2 ftmp=cloud_fs.open(url)
----> 3 z=fsspec.filesystem('zip', fo=ftmp)
      4 print('\n'.join(z.find('/')))

/s/anaconda/envs/seppo/lib/python3.8/site-packages/fsspec/registry.py in filesystem(protocol, **storage_options)
    233     """
    234     cls = get_filesystem_class(protocol)
[...]

PR #2124, opened 2026-09-09
HTTP: retry transient failures of cat_file and HTTPFile block reads

lfoppiano:
`HTTPFileSystem` does not seems to have a mechanism for retry: every `cat_file` and every `HTTPFile` block read was one `session.get` plus one body read, so a single 503, a throttling 429 or a connection dropped mid-body killed the whole read.

With this PR we added a small loop around those two paths that retries 408/425/429/5xx, dropped or reset connections, timeouts and truncated bodies, honouring Retry-After and otherwise waiting retry_wait * 2**n (capped, jittered). 
404, 401/403 and other 4xx are never retried, so subclasses that map them in _raise_not_found_for_status keep working. The loop wraps request and body read together on purpose: a mid-body ClientPayloadError surfaces from r.read(), which a request-level retry client never sees.

The HTTP error handling the retry can be altered using ``retry_statuses`` option to HTTPFileSystem (default: 408/425/429/500/502/503/504) and an overridable ``_is_retryable(exc)`` method, so a caller can opt extra codes in (a CDN answering 403 under load) or opt default codes out without patching the module. 

Two silent truncations become (retried) errors: a 206 body shorter than the requested range while the file size says more exists, and a 416 for a range inside the file, the CloudFront behaviour reported in #1895.

New HTTPFileSystem options retries=3 and retry_wait=1.0, overridable per open(); retries=0 restores the previous behaviour. The shared test server gains header-driven fault injection (fail_status, fail_times, retry_after, truncate_body, short_body) and a reset_faults fixture.

Refs #550, #1895.

This PR was derived from one issue raised and fixed [here](https://github.com/commoncrawl/warc2zip/pull/21).

### DIFF
--- a/fsspec/implementations/http.py
+++ b/fsspec/implementations/http.py
@@ -34,16 +34,16 @@ _RETRY_MAX_WAIT = 60.0
 _retry_sleep = asyncio.sleep  # module attribute so tests can substitute a fake
 
 
-def _is_retryable(exc):
+def _is_retryable(exc, statuses=_RETRYABLE_STATUSES):
     """Whether a failed HTTP read is worth repeating.
 
-    Throttling and server-side errors, dropped or reset connections, timeouts
-    and truncated bodies are transient; anything else (including the
-    ``FileNotFoundError``/``PermissionError`` a subclass may map 4xx codes to)
-    is deterministic and propagates at once.
+    A response error is retried when its status is in ``statuses``; dropped
+    or reset connections, timeouts and truncated bodies are always transient;
+    anything else (including the ``FileNotFoundError``/``PermissionError`` a
+    subclass may map 4xx codes to) is deterministic and propagates at once.
     """
     if isinstance(exc, aiohttp.ClientResponseError):
-        return exc.status in _RETRYABLE_STATUSES
+        return exc.status in statuses
     if isinstance(exc, aiohttp.ClientSSLError):
         return False
     return isinstance(
@@ -67,12 +67,15 @@ def _retry_after(exc):
         return None
 
 
-async def _with_retries(attempt, retries, retry_wait, label):
+async def _with_retries(
+    attempt, retries, retry_wait, label, is_retryable=_is_retryable
+):
     """Await ``attempt()``, repeating it on transient failures.
 
     ``attempt`` is a zero-argument coroutine function covering the whole
     request *and* body read, so a connection dropped mid-body counts as a
-    failure too. A ``Retry-After`` header is honoured when present; otherwise
+    failure too. ``is_retryable(exc)`` decides whether a failure is worth
+    repeating. A ``Retry-After`` header is honoured when present; otherwise
     the wait doubles from ``retry_wait`` on each retry, capped at
     ``_RETRY_MAX_WAIT`` seconds and jittered. Non-retryable errors and the
     final failure propagate unchanged.
@@ -81,7 +84,7 @@ async def _with_retries(attempt, retries, retry_wait, label):
         try:
             return await attempt()
         except Exception as exc:
-            if n >= retries or not _is_retryable(exc):
+            if n >= retries or not is_retryable(exc):
                 raise
             delay = _retry_after(exc)
             if delay is None:
@@ -134,6 +137,7 @@ class HTTPFileSystem(AsyncFileSystem):
         encoded=False,
         retries=3,
         retry_wait=1.0,
+        retry_statuses=None,
         **storage_options,
     ):
         """
@@ -168,6 +172,11 @@ class HTTPFileSystem(AsyncFileSystem):
             Seconds to wait before the first retry; doubled on each further
             retry (capped at 60 s, with jitter). A ``Retry-After`` header sent
             by the server takes precedence.
+        retry_statuses: iterable of int or None
+            HTTP status codes of a failed read that are retried; default
+            (None) is 408, 425, 429, 500, 502, 503 and 504. Dropped
+            connections, timeouts and truncated bodies are retried regardless.
+            Override ``_is_retryable`` in a subclass for finer control.
         storage_options: key-value
             Any other parameters passed on to requests
         cache_type, cache_options: defaults used in open()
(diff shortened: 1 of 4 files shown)


martindurant:
We already have various retries in various places (including in gcsfs and s3fs - is there anything that can be reused or factored out? Or are these all sufficiently different?

lfoppiano:
@martindurant Thank you, it's a good point. 

I did a short analysis and there are common grounds but all other components have custom policites, often in contrast with each other. E.g. GCS: retries **401 when the message says "Invalid Credentials"** (see [here](https://github.com/fsspec/gcsfs/blob/3d003b65b4bd9730e722e12c7e36a3f299190da7/gcsfs/retry.py#L90)), S3FS has other custom logic because of botocore interface hiding the error code (see [here](https://github.com/fsspec/s3fs/blob/62f45e6f667e1fe745b2f613d0217b296b2de0db/s3fs/core.py#L64) and [here](https://github.com/fsspec/s3fs/blob/62f45e6f667e1fe745b2f613d0217b296b2de0db/s3fs/core.py#L170)). 

Ideally we could share the skeleton of a retry call, but we might end up having so much custom approaches that it won't be worth the effort. 

What do you think?

martindurant:
There is also `retrying` and `tenacity`, but I would rather not have `fsspec` depend on anything we can avoid.

fsspec/implementations/http.py
```diff
@@ -28,6 +29,78 @@
...
+    if isinstance(exc, aiohttp.ClientResponseError):
+        return exc.status in statuses
+    if isinstance(exc, aiohttp.ClientSSLError):
+        return False
+    return isinstance(
+        exc,
+        (
+            aiohttp.ClientConnectionError,
+            aiohttp.ClientPayloadError,
+            asyncio.TimeoutError,
```
martindurant, line 54: We imply these should all be retriable? It might be nice to be able to define what set of exceptions are allowed as well as the statuses above.

fsspec/implementations/http.py
```diff
@@ -28,6 +29,78 @@
 ex2 = re.compile(r"""(?P<url>http[s]?://[-a-zA-Z0-9@:%_+.~#?&/=]+)""")
 logger = logging.getLogger("fsspec.http")
 
+_RETRYABLE_STATUSES = frozenset({408, 425, 429, 500, 502, 503, 504})
+_RETRY_MAX_WAIT = 60.0
+_retry_sleep = asyncio.sleep  # module attribute so tests can substitute a fake
```
martindurant, line 34: Can't you use monkeypatch alone?

fsspec/implementations/http.py
```diff
@@ -28,6 +29,78 @@
...
+    )
+
+
+def _retry_after(exc):
+    """Delta-seconds ``Retry-After`` carried by a response error, else None."""
+    headers = getattr(exc, "headers", None)
+    if not headers:
+        return None
+    try:
+        return max(0.0, float(headers.get("Retry-After")))
```
martindurant, line 65:
```suggestion
        return float(headers.get("Retry-After", "0"))
```
?

fsspec/implementations/http.py
```diff
@@ -87,6 +163,20 @@ def __init__(
             A callable, which takes keyword arguments and constructs
             an aiohttp.ClientSession. Its state will be managed by
             the HTTPFileSystem class.
+        retries: int
+            How many times a transient failure of a read (HTTP 408/425/429/5xx,
```
martindurant, line 167: Actually, the set of statuses is configurable just below, so don't list here.

fsspec/implementations/http.py
```diff
@@ -87,6 +163,20 @@ def __init__(
...
+            a dropped or reset connection, a timeout, a truncated range body)
+            is retried in ``cat_file`` and ``HTTPFile`` block reads; 0 disables
```
martindurant, line 169: *Not* used for PUT/POST, right? What about ls() and other routes?

fsspec/implementations/http.py
```diff
@@ -100,6 +190,26 @@ def __init__(
...
+                # "503" would otherwise iterate to {5, 0, 3}
+                raise ValueError(
+                    f"retry_statuses must be an iterable of ints, got {retry_statuses!r}"
+                )
+            try:
+                self.retry_statuses = frozenset(int(s) for s in retry_statuses)
+            except (TypeError, ValueError):
+                raise ValueError(
+                    f"retry_statuses must be an iterable of ints, got {retry_statuses!r}"
+                ) from None
```
martindurant, line 212: This type checking is over-cautious. 
```suggestion
        self.retry_statuses = retry_statuses if retry_statuses is not None else _RETRYABLE_STATUSES
```

fsspec/implementations/http.py
```diff
@@ -232,6 +342,16 @@ def _raise_not_found_for_status(self, response, url):
             raise FileNotFoundError(url)
         response.raise_for_status()
 
+    def _is_retryable(self, exc):
```
martindurant, line 345: Should inline this

fsspec/implementations/http.py
```diff
@@ -379,6 +505,8 @@ def _open(
         if mode != "rb":
             raise NotImplementedError
         block_size = block_size if block_size is not None else self.block_size
+        # per-open retry overrides are for HTTPFile only, never request options
+        retry_kw = {k: kwargs.pop(k) for k in ("retries", "retry_wait") if k in kwargs}
```
martindurant, line 509: This isn't documented; do we want to allow per-file overrides at all?

fsspec/implementations/http.py
```diff
@@ -591,6 +720,12 @@ class HTTPFile(AbstractBufferedFile):
     size: None or int
         If given, this is the size of the file in bytes, and we don't attempt
         to call the server to find the value.
+    retries: int or None
+        Retries for a transient failure of a block read; None (default) uses
+        the value configured on the filesystem.
+    retry_wait: float or None
+        Wait before the first retry, doubled on each further retry; None
+        (default) uses the value configured on the filesystem.
```
martindurant, line 728: Suggest removing the arguments on a file for this PR and defer to the attributes stored by the filesystem

fsspec/implementations/http.py
```diff
@@ -694,51 +836,80 @@ async def async_fetch_range(self, start, end):
...
+        async def _once():
+            r = await self.session.get(
+                self.fs.encode_url(self.url), headers=headers, **kwargs
             )
+            async with r:
+                if r.status == 416:
+                    if self.size is not None and start < self.size:
+                        # Some servers (CloudFront under load, see #1895)
+                        # answer 416 to a range that lies inside the file;
+                        # treating it as EOF would silently truncate the read.
```
martindurant, line 849: So there is no content here? 

This doesn't have much to do with error retries, though. Or are you suggesting that this unique set of circumstances is indeed retriable? A full URL link to doc/issue would be nice.

fsspec/implementations/http.py
```diff
@@ -694,51 +836,80 @@ async def async_fetch_range(self, start, end):
...
+                                f"{headers['Range']} of {self.url} returned "
+                                f"{len(out)} bytes, expected {expected}"
+                            )
+                elif start > 0:
+                    raise ValueError(
+                        "The HTTP server doesn't appear to support range "
+                        "requests. Only reading this file from the beginning "
+                        "is supported. Open with block_size=0 for a streaming "
+                        "file interface."
+                    )
```
martindurant, line 887: I wonder if we should support this. We can discard leading bytes of course; but users might not appreciate the unnecessary bandwidth.

PR #2126, opened 2026-09-10
Make the SFTP host key policy configurable

### LINKED ISSUE #2119: Insecure default SSH host key policy (AutoAddPolicy) in SFTPFileSystem
### Description

`SFTPFileSystem._connect()` in `fsspec/implementations/sftp.py` hardcodes
`paramiko.AutoAddPolicy()` as the missing-host-key policy, with no way for
callers to override it:

https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/sftp.py#L53

```python
def _connect(self):
    logger.debug("Connecting to SFTP server %s", self.host)
    self.client = paramiko.SSHClient()
    self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
    self.client.connect(self.host, **self.ssh_kwargs)
    self.ftp = self.client.open_sftp()
```

### Why this is a problem

`AutoAddPolicy` silently trusts and caches *any* host key presented on first
connection, with no verification against `known_hosts` or any other source.
This is textbook CWE-295 (Improper Certificate/Host Key Validation) — an
attacker in the network path (rogue AP, ARP/DNS spoofing, compromised
router, etc.) on that first connection can present their own key and it
will be accepted without warning, enabling a man-in-the-middle attack.

ump45nose:
`SFTPFileSystem._connect()` hardcoded `paramiko.AutoAddPolicy()` as the missing host key
policy, so a caller had no way to ask for a stricter one. As reported in #2119, that
policy accepts and remembers any key presented on a first connection to a host, which
allows a man-in-the-middle to impersonate the server.

This adds a `host_key_policy` argument to `SFTPFileSystem`:

```python
fs = fsspec.filesystem("sftp", host="example.com", host_key_policy="reject")
```

It accepts either one of the strings `"auto_add"` (default), `"warning"` or `"reject"`,
or a `paramiko.MissingHostKeyPolicy` subclass or instance:

| value | policy |
| --- | --- |
| `"auto_add"` (default) | `paramiko.AutoAddPolicy` — accept and remember unknown keys, as before |
| `"warning"` | `paramiko.WarningPolicy` — log a warning and accept |
| `"reject"` | `paramiko.RejectPolicy` — refuse to connect to unknown hosts |

The value is popped from the ssh kwargs, so it is not forwarded to
`SSHClient.connect()`, and it is applied with
`set_missing_host_key_policy()` before connecting, as before. Unknown strings raise
`ValueError` and anything else that is not a policy raises `TypeError`, at construction
time rather than at connect time.

**The default is deliberately unchanged** (`"auto_add"`), since switching it to
`"reject"` would break every existing user whose known hosts file does not already
contain the server key; the docstring and changelog entry document that the default
trusts any key and that `"reject"` is available for verification. If you would rather
change the default in a future release, that is a one-line change and I am happy to
send it separately — the parameter has to exist first either way.

### Tests

The docker-based SFTP tests exercise a live server, so the new tests use a stand-in for
`paramiko.SSHClient` and check how the client is configured before connecting:

- the default is `AutoAddPolicy`
- each of the three names selects the matching paramiko policy
- a policy class and a policy instance are both used as given
- `host_key_policy` is not passed to `connect()`, while the other ssh kwargs still are
- an unknown name raises `ValueError`, a non-policy raises `TypeError`

```
$ pytest fsspec/implementations/tests/test_sftp.py -q
9 passed, 17 skipped
```

Without the change to `sftp.py`, 7 of the 9 new tests fail. Running the whole suite
(`pytest fsspec/ -q`) gives the same set of failures and errors before and after this
change (170 in this environment, all from optional dependencies that are not installed);
`ruff check` and `ruff format --check` are clean on the changed files.

Closes #2119

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,6 +14,11 @@ Fixes
 - End the transaction even when a commit or discard raises, so the filesystem
   is not left in transaction mode and deferred temporary files are cleaned up
 
+- The SFTP host key policy is configurable with the new ``host_key_policy``
+  argument of ``SFTPFileSystem``, which accepts ``"auto_add"`` (default,
+  unchanged behaviour), ``"warning"`` or ``"reject"``, or any
+  ``paramiko.MissingHostKeyPolicy`` (#2119)
+
 2026.7.0
 --------
 
--- a/fsspec/implementations/sftp.py
+++ b/fsspec/implementations/sftp.py
@@ -12,6 +12,37 @@ from ..utils import infer_storage_options
 
 logger = logging.getLogger("fsspec.sftp")
 
+#: Short names for the missing host key policies built into paramiko, accepted
+#: as the ``host_key_policy`` argument of ``SFTPFileSystem``.
+HOST_KEY_POLICIES = {
+    "auto_add": paramiko.AutoAddPolicy,
+    "reject": paramiko.RejectPolicy,
+    "warning": paramiko.WarningPolicy,
+}
+
+
+def _resolve_host_key_policy(policy):
+    """Turn a ``host_key_policy`` argument into a paramiko policy instance."""
+    if policy is None:
+        return paramiko.AutoAddPolicy()
+    if isinstance(policy, str):
+        try:
+            return HOST_KEY_POLICIES[policy]()
+        except KeyError:
+            raise ValueError(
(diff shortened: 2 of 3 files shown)


martindurant:
This is fine by me.

I wonder, should there be a "confirm" option (i.e., equivalent to what the `ssh` CLI does)?

PR #2128, opened 2026-09-10
Add resumable downloads to get_file

### LINKED ISSUE #1617: Suggestion: continue download
When downloading large files, it can commonly happen that the transfer aborts part way. Rather than having to restart the process, it would be nice to be able to read only the remaining bytes on a subsequent attempts. `cat_file()` already allows for providing a start byte (and end byte), but `get_file()` does not. I am proposing to both implement start/end byte and, more importantly, to allow "continue" whereby start=size-of-local-file.

nickita-khylkouski:
Closes #1617.

Adds `resume=True` to `get_file` and `get` for local path destinations. The generic path opens an existing destination in append mode and seeks the remote file to the same offset. HTTP resumes with an internal range request, restarts safely if the server ignores it, and checks the remote size before accepting a completed local file. There is no public byte-range API.

Tests:
- 8 focused resume/get-file tests passed
- full suite: 1,917 passed, 70 skipped, 2 xfailed
- 52 existing GNU `stat -c` comparison cases fail on macOS with BSD `stat`
- Ruff check/format passed on all changed files

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,6 +4,11 @@ Changelog
 Dev
 ---
 
+Enhancements
+
+- Add byte-range and resumable ``get_file`` / ``get`` downloads (#1617),
+  with validated HTTP range responses and progress reporting.
+
 Fixes
 
 - Allow filesystem implementations to assign ``protocol`` per instance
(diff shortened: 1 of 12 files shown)


martindurant:
This is a very big PR and hard to review.

I don't mind the idea of resumable downloads (open local file with "ab" and seek in remote file) not byte ranges, although the two conflict with each other. I think the implementation should be significantly simpler than this!

PR #2131, opened 2026-09-11
Close cached readers when exiting an open_files context

### LINKED ISSUE #2130: Cached open_files readers remain open after context exit
Exiting `with fsspec.open_files(...)` leaves the returned readers open when using `simplecache` or `filecache`.

Reproduced on master `be32e2b2bb775104bc3f2bf291a87bbfedd8a273`, Linux, Python 3.10, with only an in-memory source and temporary cache:

```python
import tempfile
import fsspec

memory = fsspec.filesystem("memory")
memory.pipe("/example", b"data")
with tempfile.TemporaryDirectory() as cache:
    with fsspec.open_files("simplecache::memory:///example", cache_storage=cache) as files:
        assert files[0].read() == b"data"
    print(files[0].closed)  # False; expected True
    files[0].close()
```

This also happens with `filecache`, a warm cache, and an exception inside the context. `OpenFiles.__enter__` returns the handles from `open_many`, but `__exit__` only closes the unpopulated `OpenFile` wrappers. Keeping the returned list alive therefore retains open file descriptors after the context ends.

AI assistance was used to investigate and run this reproducer.

betacatsling:
Exiting an `open_files` context leaves `simplecache` and `filecache` readers open because `open_many()` returns handles without populating the `OpenFile` wrappers. Close those handles on exit, including when the consumer raises an exception.

Regression coverage exercises both cache implementations, normal and exceptional exits, and cold and warm cache reads. The four regressions fail on the unchanged base and pass with the fix. Local validation: cached-filesystem and core suites **137 passed, 37 skipped**; Ruff lint/format and `git diff --check` passed. Remote backends and downstream suites were not run.

AI assistance was used for implementation and local verification.

Fixes #2130.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,8 @@ Dev
 
 Fixes
 
+- Close cached readers when exiting an ``open_files`` context
+
 - Allow filesystem implementations to assign ``protocol`` per instance
 
 - Avoid mutating live ``BlockCache`` and ``BackgroundBlockCache`` instances
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -192,7 +192,11 @@ class OpenFiles(list):
     def __exit__(self, *args):
         fs = self.fs
         [s.__exit__(*args) for s in self]
-        if "r" not in self.mode:
+        if "r" in self.mode:
+            # open_many() returns files without populating the OpenFile objects.
+            for f in self.files:
+                f.close()
+        else:
             while True:
                 if hasattr(fs, "open_many"):
                     # check for concurrent cache upload
--- a/fsspec/implementations/tests/test_cached.py
+++ b/fsspec/implementations/tests/test_cached.py
@@ -955,6 +955,36 @@ def test_again(protocol):
 
 
 @pytest.mark.parametrize("protocol", ["simplecache", "filecache"])
+@pytest.mark.parametrize("raise_in_context", [False, True])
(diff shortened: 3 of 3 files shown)


martindurant (commented):
It seems to me, that `[s.__exit__(*args) for s in self]` was supposed to do this, and the real problem is not correctly populating the internal list, no?

betacatsling:
Thanks — yes, the individual `OpenFile.fobjects` lists remain empty on this path, which is why the existing `__exit__` calls do nothing. `open_many()` returns the actual handles into `OpenFiles.files` without entering the individual `OpenFile`s.

I kept cleanup at the `OpenFiles` level because that is already where the bulk-opened handles are tracked. Populating `fobjects` for both modes would also make the existing loop close `LocalTempFile`s before `commit_many()`, triggering their per-file autocommit and changing write-path ordering. Closing `self.files` only on reads leaves that write path unchanged.

There is also an ownership wrinkle: slicing creates a new `OpenFiles` container but shares its `OpenFile` instances, so assigning handles to those shared descriptors can overwrite another active container's tracking, whereas each container has its own `files` list.

Populating `fobjects` only for reads would fix the ordinary case too, so that is a viable alternative — I preferred the narrower change here: close the read handles returned by `open_many()` without changing descriptor ownership or write behavior.

martindurant:
OK, your reasoning makes sense to me.

PR #2132, opened 2026-09-11
Add resumable downloads to get_file

### LINKED ISSUE #1617: Suggestion: continue download

nickita-khylkouski:
Closes #1617.

Supersedes #2128 with the smaller approach requested in review.

Adds `resume=True` to `get_file` and `get` for local path destinations. The generic path opens an existing destination in append mode and seeks the remote file to the same offset. HTTP resumes with an internal range request, restarts safely if the server ignores it, and checks the remote size before accepting a completed local file. There is no public byte-range API.

Tests:
- 8 focused resume/get-file tests passed
- full suite: 1,917 passed, 70 skipped, 2 xfailed
- 52 existing GNU `stat -c` comparison cases fail on macOS with BSD `stat`
- Ruff check/format passed on all changed files


fsspec/spec.py
```diff
@@ -984,17 +999,31 @@ def get_file(
             os.makedirs(lpath, exist_ok=True)
             return None
 
+        offset = 0
+        if resume:
+            if outfile is not None:
+                raise ValueError("resume requires a local path, not a file-like")
+            if os.path.isfile(lpath):
+                offset = os.path.getsize(lpath)
```
martindurant: Else 0 - we can resume even if it never started. That enables users to pass resume for everything if they like.

fsspec/spec.py
```diff
@@ -984,17 +999,31 @@ def get_file(
...
+
         if outfile is None:
             fs = LocalFileSystem(auto_mkdir=True)
             fs.makedirs(fs._parent(lpath), exist_ok=True)
 
         with self.open(rpath, "rb", **kwargs) as f1:
+            size = getattr(f1, "size", None)
+            if offset:
+                if size is not None and offset > size:
+                    raise ValueError(f"Local file {lpath} is larger than {rpath}")
```
martindurant: If they are equal, we just skip, since the file is complete

Issue #1906, opened 2025-08-13
memory ls/walk/find algorithms


smorken:
I was wondering if anyone had considered using a tree-based structure rather than flat structure for the in-memory implementation.  For my current experimentation, use case I'm ending up with A LOT of calls to `ls` via `walk` and `find` calls, and that calls `startswith` very many times when many paths are stored in the flat structuring.  

For me the in-memory implementation as it stands is not useable for when you have a significant number of paths.  I guess maybe the in-memory implementation is intended more for testing, and less for production?  

Any value in pursing an update to the in-memory implementation? Any workarounds or better approaches?  I have considered fuse but it's not a cross platform dependency.

smorken:
I ended up drafting an implementation using combined prefix trie and simple path caching.

Its here:

https://github.com/smorken/triememfs

I don't have a lot of time for the needed refinements at the moment, but it does pass all of the [fsspec abstract test cases](https://github.com/smorken/triememfs/tree/main/trie_memfs/test) and from basic informal benchmarking so far it's dramatically faster for walk/ls/find routines.  

Will be using this script for something though and may keep pushing refinements for a while.

martindurant:
Thanks for posting. You can of course publish this and might want to include it in fsspec.registry.known_implementations if you want others to benefit too - when you're ready, of course.

smorken:
@skshetry thanks for pointing this out, I would have probably just used that instead of going to the effort of writing my own, if I had known it was there.  That said, after working out a couple of minor issues (fixes all pushed at my link a little further up this thread) I am very happy with how my little memory fs package is working. 

I polished and added a couple of [benchmarks](https://github.com/smorken/triememfs/blob/main/trie_memfs/benchmarks.ipynb)  and they are pretty much comparable and similar in performance.  

Could it make sense to eventually replace the fsspec packaged memory implementation using a combination or one or the other or both of these approaches? That would probably save folks time in the long run.

martindurant:
#1905 is also relevant, although not concerning performance - if someone wants to pick it up.

skshetry:
> [#1905](https://github.com/fsspec/filesystem_spec/pull/1905) is also relevant, although not concerning performance - if someone wants to pick it up.

In DVC, we don't like __global__ store from `MemoryFileSystem`, so we added `global_store=True|False` kwarg to our implementation. In DVC, we create fsspec filesystems with `skip_instance_cache=True` so instance cache is not a problem for us.

[...]

PR #2137, opened 2026-09-12
Allow independent stores in MemoryFileSystem

### LINKED ISSUE #1904: create multiple separate memory filesystems

aprylewu:
Memory filesystem instances always share their files and directories, even when callers bypass the instance cache. Add `global_store=False` for instance-local storage, so callers can create in-memory workspaces with overlapping paths.

Use `global_store=False, skip_instance_cache=True` to create a fresh independent workspace on each call. Normal instance caching still applies when the cache option is omitted. Each private instance owns its files and empty directories and has a distinct filesystem token. Both options also work through fsspec configuration.

Pickling a private filesystem copies its contents into a fresh instance, including when the source was cached. File objects point back to the restored filesystem through pickle's memo. Default construction preserves the global store, instance reuse, and transaction behavior, including writes through `fsspec.open()` and `fsspec.get_mapper()`. This uses the existing cache machinery without a new metaclass.

Closes #1904; follows the abandoned #1905. The discussion in #1906 suggested disabling MemoryFS caching altogether; keeping normal caching preserves the transaction association of the top-level helpers.

Validation on macOS / Python 3.13.5:

- Eight related test modules: **563 passed, 118 skipped, 2 xfailed**, including **51 memory tests**.
- An independent agent reviewed the revision and ran 31 additional checks covering cache options, configuration/environment overrides, default and private transactions, pickle protocols 2–5, deepcopy, and restored file back-references.
- Ruff 0.14.3, formatting, codespell 2.4.0, and `git diff --check` passed.

The full Docker/FUSE/downstream suite and documentation build were not run. JSON serialization still records constructor options; it does not copy memory contents.

AI assistance: OpenAI Codex assisted with the implementation, tests, and this description. The validation results above are from local execution.

### DIFF
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -4,7 +4,9 @@ from pathlib import PurePosixPath, PureWindowsPath
 
 import pytest
 
+import fsspec
 from fsspec import filesystem
+from fsspec.config import conf
 from fsspec.implementations.local import LocalFileSystem, make_path_posix
 from fsspec.implementations.memory import MemoryFileSystem
 
@@ -48,6 +50,42 @@ def test_default_store_is_shared(m):
     assert other.cat("file") == b"shared"
     assert other.isdir("empty")
     assert other == m
+    assert other is m
+
+
+@pytest.mark.parametrize("use_mapper", [False, True])
+@pytest.mark.parametrize("rollback", [False, True])
+def test_default_store_transaction_helpers(m, use_mapper, rollback):
+    def write():
+        if use_mapper:
+            fsspec.get_mapper("memory://")["file"] = b"data"
+        else:
+            with fsspec.open("memory://file", "wb") as f:
+                f.write(b"data")
+        assert not m.exists("file")
+
+    if rollback:
+        with pytest.raises(RuntimeError), m.transaction:
+            write()
+            raise RuntimeError("discard transaction")
+        assert not m.exists("file")
+    else:
+        with m.transaction:
+            write()
+        assert m.cat("file") == b"data"
+
+
+def test_independent_store_from_config(m, monkeypatch):
+    monkeypatch.setitem(conf, "memory", {"global_store": False})
+    first = filesystem("memory")
+    second = filesystem("memory", skip_instance_cache=False)
+    first.pipe("file", b"private")
+    assert not second.exists("file")
+    assert not m.exists("file")
+
+    shared = filesystem("memory", global_store=True)
+    assert shared is filesystem("memory", global_store=True)
+    assert shared.store is m.store
 
 
 def test_independent_store_transaction(m):
--- a/fsspec/implementations/tests/test_memory.py
+++ b/fsspec/implementations/tests/test_memory.py
@@ -1,10 +1,97 @@
 import os
+import pickle
 from pathlib import PurePosixPath, PureWindowsPath
 
 import pytest
 
+from fsspec import filesystem
 from fsspec.implementations.local import LocalFileSystem, make_path_posix
 from fsspec.implementations.memory import MemoryFileSystem
+
+
+def test_independent_stores(m):
+    first = filesystem("memory", global_store=False)
+    second = filesystem("memory", global_store=False)
+    for fs, data in [(m, b"global"), (first, b"first"), (second, b"second")]:
+        fs.pipe("same/path", data)
+        fs.mkdir("empty")
+
+    assert m.cat("same/path") == b"global"
+    assert first.cat("same/path") == b"first"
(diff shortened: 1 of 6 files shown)


fsspec/implementations/tests/test_memory.py
```diff
@@ -1,12 +1,137 @@
...
+    assert not m.exists("discarded")
+
+
+def test_independent_store_pickle(m):
+    fs = filesystem("memory", global_store=False)
+    fs.pipe("file", b"original")
+    fs.mkdir("empty")
+
+    restored = pickle.loads(pickle.dumps(fs))
+    assert restored != fs
```
martindurant:
```suggestion
    assert restored is not fs
```
(might argue that they should evaluate as equal)

fsspec/implementations/tests/test_memory.py
```diff
@@ -1,12 +1,137 @@
...
+    assert restored.cat("file") == b"original appended"
+    assert fs.cat("file") == b"original"
+    assert fs.isdir("empty")
+    assert m.ls("") == []
+
+
+def test_default_store_pickle(m):
+    m.pipe("file", b"shared")
+    restored = pickle.loads(pickle.dumps(m))
+    assert restored == m
```
martindurant:
```suggestion
    assert restored is m
```

fsspec/implementations/tests/test_memory.py
```diff
@@ -1,12 +1,137 @@
...
+def test_independent_store_pickle(m):
+    fs = filesystem("memory", global_store=False)
+    fs.pipe("file", b"original")
+    fs.mkdir("empty")
+
+    restored = pickle.loads(pickle.dumps(fs))
+    assert restored != fs
+    assert restored.cat("file") == b"original"
+    assert restored.isdir("empty")
+    assert restored.store["/file"].fs is restored
```
martindurant, line 135: I agree this is as it should be - but where in the code is it set?

martindurant (commented):
Because MemoryFS doesn't take any arguments, users were previously getting not just the same underlying storage, but the same instance back every time (unless they knew what they were doing). 

The code here introduces the only subclass of `_Cached` so far, and I think we need to be really careful with metaclasses - is there no simpler way? Users can already pass skip_instance_cache= themselves after all.

fsspec/implementations/tests/test_memory.py
aprylewu: Changed this to `assert restored is not fs`, so the pickle test checks identity separately from the restored contents. The content and mutation-isolation assertions remain.

fsspec/implementations/tests/test_memory.py
aprylewu: Done — the default-store pickle test now asserts `restored is m`.

fsspec/implementations/tests/test_memory.py
aprylewu: Pickle restores that reference through its memo: the state returned by `__reduce__` contains the store, whose `MemoryFile` objects refer back to the filesystem. When pickle reaches those references, it reuses the filesystem being reconstructed. There is no separate assignment during restoration. I added a comment beside the state and kept the back-reference assertion.

aprylewu:
Thanks, the metaclass was unnecessary. Removed it in de740c9; `global_store=False` now selects instance-local storage while callers use the existing `skip_instance_cache=True` option when they need a fresh instance. The docs and config tests follow that behavior.

One detail needed care after removing it: unpickling a private filesystem must bypass the cache so it cannot overwrite the original cached instance. The reducer now does that explicitly. A regression test takes a snapshot, modifies the source, then checks that loading the snapshot leaves the source untouched.

Also applied the two identity assertions and explained the file back-reference inline. The eight related test modules passed locally (563 passed, 118 skipped, 2 xfailed), as did Ruff and codespell. An independent agent reviewed the revision and passed 31 additional cache/serialization/transaction checks.

PR #2144, opened 2026-09-14
Delegate invalidate_cache, ukey and checksum from DirFileSystem

itzzdev09:
`DirFileSystem` doesn't override `invalidate_cache`, `ukey` or `checksum`, so it gets the base class versions, and none of them reach the wrapped filesystem:

- **`invalidate_cache`** is a no-op in the base class. On a wrapped filesystem with a listings cache (s3fs, gcsfs, ...), `dirfs.invalidate_cache(path)` never clears anything, so stale listings survive even after the caller asks for them to be dropped.
- **`ukey` and `checksum`** hash the result of `dirfs.info()` instead of calling the wrapped filesystem's own implementation. For example, s3fs bases `ukey` on the ETag.

```python
calls = []

class Recording(MemoryFileSystem):
    def invalidate_cache(self, path=None):
        calls.append(path)

dirfs = DirFileSystem("/base", Recording())
dirfs.invalidate_cache("x")
calls  # [] -- expected ["/base/x"]
```

## Change

All three now delegate to `self.fs` with the joined path, the same way `info`, `sizes` and the rest already do.

`invalidate_cache()` with no path clears what is cached under the dirfs root, not the whole wrapped filesystem. A dirfs with an empty root passes `None` through, which means "everything". The base implementation is still called, so invalidation inside a transaction is recorded as before.

## Tests

Added `test_invalidate_cache`, `test_invalidate_cache_without_path`, `test_invalidate_cache_without_path_or_root`, `test_ukey` and `test_checksum` to `test_dirfs.py`, following the mock-based pattern of the neighbouring tests. All of them fail against `master`'s `dirfs.py` (10 failures across the sync/async params) and pass with the change. `test_dirfs.py`: 180 passed. `ruff` 0.14.3 check and format are clean.


fsspec/implementations/dirfs.py
```diff
@@ -285,6 +285,18 @@ async def _sizes(self, paths):
     def sizes(self, paths):
         return self.fs.sizes(self._join(paths))
 
+    def invalidate_cache(self, path=None):
+        super().invalidate_cache(path)
```
martindurant: This is not necessary: dirFS doesn't use dircache.

itzzdev09: Good point, thanks. Removed in 7cc443a, so it now only delegates to the wrapped filesystem.

PR #2145, opened 2026-09-14
Raise FileNotFoundError from archive ls for a missing path

itzzdev09:
`AbstractArchiveFileSystem.ls` (used by zip, tar and libarchive) returns `[]` in two cases where other filesystems don't:

```python
fs = fsspec.filesystem("zip", fo="archive.zip")   # contains "b" and "deeply/nested/path"

fs.ls("i-do-not-exist")   # [] -- expected FileNotFoundError
fs.ls("b")                # [] -- expected [info of "b"]
```

`info("i-do-not-exist")` already raises `FileNotFoundError` on the same filesystem, and `LocalFileSystem`, `MemoryFileSystem` and the object stores raise from `ls` too. Returning an empty list makes a typo look like an empty directory.

## Change

When nothing is listed under a non-root path, `ls` now falls back to `info`:
- a missing path raises `FileNotFoundError`;
- a file is listed as itself;
- an empty directory still returns `[]`.

Listing the root of an empty archive still returns `[]` (`test_empty`), and listings that find entries are unchanged.

## Tests

Added `TestAnyArchive.test_ls_file_or_missing`, which runs for zip, tar, tar.gz, tar.bz2, tar.xz and libarchive. It fails against `master`'s `archive.py` for every scenario that runs here (libarchive isn't installed locally, so it's skipped) and passes with the change. `test_archive.py`, `test_zip.py`, `test_tar.py`, `test_libarchive.py` and `fsspec/tests/test_api.py`: 193 passed, 21 skipped. `ruff` 0.14.3 check and format are clean.


itzzdev09:
CI caught a test I hadn't run: `test_dirfs.py::test_find_detail_single_tar_file` (from #2127) asserted that `find` on a single tar member returns the `{}` placeholder the base `find` falls back to when `ls` is empty. Now that `ls` lists a file as itself, `find` returns the member's real info, the same as on other filesystems, so 4118cce updates that test to check the name, type and size instead. The no-name fallback in dirfs is still covered by `test_detail_without_name`. I compared the full suite against `master` locally, and that test was the only difference.

PR #2147, opened 2026-09-14
Keep reference cat_file offsets inside the referenced part

itzzdev09:
For a reference to part of a file (`[url, offset, size]`), `ReferenceFileSystem.cat_file` adds `start`/`end` to the part's offset without keeping the result inside the part. When they go past the part, the read returns bytes of the target file that aren't part of the reference:

```python
m.pipe("/b", b"other test data")
fs = fsspec.filesystem("reference", fo={"d": ["memory://b", 4, 6]})   # d == b"r test"

fs.cat_file("d", start=2, end=100)   # b"test data" -- expected b"test"
fs.cat_file("d", start=-100)         # reads from the end of /b, not from the start of d
```

This breaks slicing semantics for the key. For kerchunk-style references, where many chunks sit in one target file, it silently returns bytes that belong to neighbouring chunks.

## Change

In `_cat_common`, offsets are resolved like slicing the part: `start` and `end` are clamped to `[offset, offset + size]`, whether given as positive or negative. In-range reads compute exactly the same offsets as before. Whole-file references and inline data are untouched, and so is the sync/async `cat` path, which never passes `start`/`end`.

## Tests

Extended `test_cat_file_ranges` with offsets past each end of the part (`end=100`, `start=-100`, `start=100`, `end=-100`), each compared to slicing the part's bytes. The test fails against `master`'s `reference.py` and passes with the change. `test_reference.py` has the same 10 failures before and after this change: tests that need the HTTP test server or network, which isn't available on this machine. `ruff` 0.14.3 check and format are clean.


martindurant:
(failure caused by github rate limit - will run again later)

Issue #1960, opened 2025-12-11
Block Cache requests the data twice for reads greater than 160MB


googlyrahman:
The code for block cache is located [https://github.com/fsspec/filesystem_spec/blob/master/fsspec/caching.py#L331](here)

The relevant code pulled from the above file

```
    def _fetch(self, start: int | None, end: int | None) -> bytes:
        ....

        # these are cached, so safe to do multiple calls for the same start and end.
        for block_number in range(start_block_number, end_block_number + 1):
            self._fetch_block_cached(block_number)

        return self._read_cache(
            start,
            end,
            start_block_number=start_block_number,
            end_block_number=end_block_number,
        )
```

Problem
- BlockCache is an LRU-based cache with a default size of 32. Given the default block size of 5MB, the maximum cache size is 160MB at any given time.

- In the code above, we first iterate through the required blocks to fetch them. If the requested data exceeds 160MB (more than 32 blocks), the earlier blocks fetched at the start of the loop are evicted to make room for the later blocks before the loop even finishes.

- When _read_cache is called immediately after the loop, it again iterates from block 1 assuming data is already present. However, because the first few blocks were evicted during the fetch loop, _read_cache forces a new network request to retrieve them again. This causes a cascading eviction cycle where we end up requesting each data twice.

Example Scenario
- If we request 33 blocks (165MB of data):
- The loop fetches blocks 1 through 33, By the time it fetches block 33, the cache has evicted block 1 to maintain the limit of 32 items. The cache now holds [2, 3, ..., 33].
- We then enter self._read_cache. It attempts to read block 1 first. Since block 1 is missing, it triggers a new network call, Fetching block 1 evicts block 2. The cache now holds [1, 3, 4, ..., 33].
- _read_cache then tries to read block 2. Since it was just evicted, another network call is made, evicting block 3. The cache becomes [1, 2, 4, ..., 33].

This cycle continues for the entire read. Consequently, for any request where size > block_size * max_blocks (currently 160MB), we effectively download the data twice.

martindurant:
I think you have found a real problem. Aside from making the blocksize and/or the LRU count bigger, do you have a specific solution you'd like to propose? One might suppose that when we know the read is big (>>block size), we should probably be doing a single big read and populate the blocks after the fact?

TLCFEM:
Is the pre-fetch loop necessary?
`_read_cache` fetches and stores the target block anyway, isn't it?

```py
[...]

PR #2150, opened 2026-09-15
Stop BackgroundBlockCache refetching blocks it already read

itzzdev09:
Follow-up to #1984 for `BackgroundBlockCache` (`cache_type="background"`), which still has both of the problems that PR fixed in `BlockCache`. This is the remaining part of #1960.

**1. Reads larger than the LRU fetch blocks again.** `_fetch` first loops over every block in the range calling `_fetch_block_cached`, then `_read_cache` reads them all again. If the range spans more than `maxblocks` blocks, the loop evicts the first blocks before it finishes, so `_read_cache` downloads them a second time, and each of those re-fetches evicts another block. #1960 describes this for `BlockCache`: with the defaults, any read over 160 MB.

**2. A read ending on a block boundary fetches an extra block.** `end` is exclusive, but `end_block_number = end // blocksize` treats it as inclusive. `_fetch(0, blocksize)` fetches block 1 synchronously, then slices zero bytes from it. It also starts the background prefetch at block 2 instead of block 1.

```python
cache = BackgroundBlockCache(4, fetcher, 52, maxblocks=2)
cache._fetch(0, 12)
cache.cache_info().misses   # 7 on master -- blocks 0-2 are needed, so expected 3

cache = BackgroundBlockCache(4, fetcher, 52)
cache._fetch(0, 4)
cache.cache_info().misses   # 2 on master -- expected 1
```

## Change

This is the same change as #1984:
- remove the prefetch loop, since `_read_cache` already fetches each block it needs through the LRU;
- compute the last block as `(end - 1) // blocksize`;
- treat `end_pos == 0` as a full final block in `_read_cache`.

The background-fetch logic is unchanged. It still joins a pending fetch for a block inside the range and prefetches the block after the read.

## Tests

Added `test_background_block_cache_no_redundant_reads`, which mirrors #1984's `BlockCache` test, and `test_background_block_cache_read_ending_on_block_boundary`. Both count LRU misses, which background fetches don't add to, so the counts are deterministic. Both fail against `master`'s `caching.py` and pass with the change. In `test_caches.py`, the only other failure is `test_background`, which needs the HTTP test server (not available here) and fails the same way on `master`: 158 passed. `ruff` 0.14.3 check and format are clean.


martindurant:
We still need this one, right?

itzzdev09:
Yes, I believe we still need this one. #1984 fixed the redundant-fetch issue for "BlockCache", but "BackgroundBlockCache" retained the same prefetch loop and still exhibits the redundant reads when a request spans more than "maxblocks". The block-boundary issue is also still present there.

I checked the current "master": "BackgroundBlockCache._fetch()" still prefetches the entire requested range before "_read_cache()", so #2150's changes don't appear to have been covered by a subsequent PR.

itzzdev09:
Idk, something is wrong. On my GitHub mobile it's showing I commented once and on desktop it's 2 same comments. So if it's two, please ignore the spam and redundancy! 🙃🙂

PR #2152, opened 2026-09-15
Add MountFileSystem to present several filesystems as one tree

### LINKED ISSUE #1994: Equivalent to pyfilesystem2's MountFS?
Hi,

just wanted to make sure I'm not completely missing something before I have a go at this myself. Is there an equivalent to https://pyfilesystem2.readthedocs.io/en/latest/reference/mountfs.html?

Thanks 🙏

itzzdev09:
Closes #1994.

This adds `MountFileSystem` (protocol `mount`): several filesystems presented as a single tree, each mounted at its own directory. It's the counterpart to `DirFileSystem`. dirfs exposes one directory *of* a filesystem, while this places whole filesystems (or directories of them) *into* a shared namespace, like pyfilesystem2's `MountFS`, which #1994 asked about.

```python
memory = fsspec.filesystem("memory")
fs = MountFileSystem({
    "/data": "s3://bucket/dataset",              # URL; its path becomes the root
    "/scratch": "/tmp/scratch",                  # local path
    "/mem": (memory, "/workspace"),              # (filesystem, root) pair
})
fs.mount("/public", "gcs://public-bucket", token="anon")  # storage options for a URL
fs.mount("/cached", "simplecache::s3://bucket/lookup")    # chained URLs work too

fs.ls("/", detail=False)          # ['/cached', '/data', '/mem', '/public', '/scratch']
fs.find("/data", maxdepth=2)
fs.cp("/data/part-0.parquet", "/scratch/")   # streams S3 -> local
fs.get("/", "backup/", recursive=True)
```

## Behaviour

- **Paths** are absolute and "/"-separated. `mount://`, duplicate slashes, and `.`/`..` parts are normalized.
- **Mounting.** `mount(path, fs, root=None, **storage_options)` accepts a filesystem instance with an optional root, or a URL resolved with `url_to_fs`; chained URLs work. `unmount(path)` returns the filesystem, and `mounts` lists what's mounted.
- **Mount points** can't be the root and can't contain one another. An overlapping mount raises `ValueError`, so every path belongs to at most one filesystem.
- **Virtual directories.** The directories above mount points exist only virtually. `ls`, `info`, `find`, `walk`, `glob` and `du` include them. Writing to them, or to a mount point itself, or removing either, raises `PermissionError`, so `rm("/data", recursive=True)` can't wipe a mounted root. Paths outside every mount raise `FileNotFoundError` when read and `PermissionError` when written. Reading a directory as a file raises `IsADirectoryError`.
- **Delegation.** Anything that stays on one filesystem instance is passed straight to it, with paths translated on the way in and names translated on the way out. That includes `cp_file` and `mv` between two mounts of the same instance, which therefore use that filesystem's server-side copy or rename. Copies between different filesystems stream from one to the other. `get`/`put` delegate to `get_file`/`put_file`.
- **Batching.** `cat_ranges` and `rm` send one call per mounted filesystem, preserving result order, so a backend like s3fs can still run its requests concurrently. Paths that fail to resolve are reported per item according to `on_error`.
- **`find`** on a virtual directory only descends into the mounts under it. It passes `maxdepth` on relative to each mount point and doesn't walk directories one `ls` at a time. The results match what the same tree gives on a single filesystem, and a test checks that exhaustively (see below).
- **Transactions.** Files opened for writing inside `fs.transaction` are committed by their own filesystem when it ends, for backends that honour `autocommit`.
- **Serialization.** `cachable = False`, because mounts can change after creation. Mounts added later are kept in `storage_options`, so pickling and `to_json`/`from_json` rebuild the same tree.
- **Sync only.** Martin raised the sync/async question in #1994. I went with sync: asynchronous filesystems can be mounted and are used through their blocking methods, and batching per filesystem keeps their concurrency for ranged reads. An async variant could follow if there's demand, but it would roughly double the surface of this first version.

## Tests

- **`fsspec/implementations/tests/test_mount.py` (79 tests)** covers:
  - listing and `info` for virtual directories, mount points and mounted paths;
  - path normalization;
  - reads, writes and the errors for each kind of path;
  - `mkdir`/`makedirs`, `find`, `glob`, `walk` and `du`;
  - copies and moves both within one filesystem (asserting the backend's own `cp_file`/`mv` is used) and between memory and local mounts;
  - `get`/`put`, including getting the whole virtual tree;
  - `cat_ranges` ordering, error handling and per-filesystem batching;
  - mount validation, `unmount`, and URLs with storage options;
  - a mount filesystem mounted inside another;
  - pickle and JSON round-trips, `invalidate_cache`, and transactions.
- **`test_find_matches_a_single_filesystem`** lays the same tree out on one memory filesystem and compares `find` for every combination of start path (`/`, a virtual directory, a mount point, a mount), `maxdepth` (None, 1–4) and `withdirs`.
- **`fsspec/implementations/tests/mount/`** runs the shared `AbstractCopyTests`, `AbstractGetTests`, `AbstractPutTests`, `AbstractPipeTests` and `AbstractOpenTests` against a memory filesystem mounted at `/mnt`, as `tests/memory` does for memory itself: 137 passed.

The full suite shows the same failures and errors as `master` locally (tests needing network or optional dependencies that aren't installed here), and nothing new. `ruff` 0.14.3 check and format are clean.

Also registered in `registry.py`, listed in `api.rst`, and noted in the changelog.


martindurant:
This being rather big, it will take me some time to review (help appreciated).

itzzdev09:
Understood, thanks. 
To make it easier: it's purely additive. All the logic is in implementations/mount.py (~600 lines), and the rest is tests plus a registry line. The core is _resolve (~L177), which decides whether a path is in a mount, a virtual directory, or neither; most other methods just delegate through it. The parts worth a careful look are _find_in_virtual_dir (~L325) and cp_file/mv/cat_ranges (~L407–578), where operations span filesystems. 

If it helps, I can split it into a first PR with mounting, reads and listing, and a follow-up with writes, copies and batching.

martindurant:
> cp_file/mv/cat_ranges (~L407–578), where operations span filesystems

Do you reuse code in fsspec.generic?

itzzdev09:
No, not at the moment. It follows the same idea but doesn't call into it:

- **Same filesystem instance:** `cp_file` does the same check as `GenericFileSystem._cp_file`. When `fs1 is fs2`, it hands the copy to that filesystem, so something like S3 → S3 stays a server-side copy.
- **Different filesystems:** it streams the file with `shutil.copyfileobj` instead of going through `copy_file_op`.
- **Several files:** they're copied one at a time through `AbstractFileSystem.copy`.

I could switch the cross-filesystem path to `fsspec.generic.copy_file_op`. Copies between mounts would then be batched and concurrent in the same way generic's are, and the extra code in `cp_file` would go away. The catch is that `copy_file_op` saves each file to a local temp file first (`get_file`, then `put_file`), while the current code streams straight from one filesystem to the other with no temp file. For large objects I think that difference matters, but I'm happy to change it if you'd rather reuse generic. I'd pass `on_error="raise"` so a failed copy isn't silently skipped.

The other two:

- `cat_ranges` has no equivalent in generic. It groups the ranges by mounted filesystem and makes one `cat_ranges` call per filesystem, so each backend can still batch or run its own requests concurrently.
- `mv` only adds the same-instance case, where the move is passed straight to that filesystem's `mv`. Anything else goes to the base implementation

PR #2155, opened 2026-09-16
Fix tail on simplecache and filecache

itzzdev09:
`AbstractFileSystem.tail` reads `f.size`, which only fsspec's own file classes have. `simplecache` and `filecache` hand back a plain local file object from `open`, so `tail` always failed on them:

```python
import fsspec
fsspec.filesystem("memory").pipe("/a/b.txt", b"12345")
fsspec.filesystem("simplecache", target_protocol="memory").tail("memory://a/b.txt", 2)
# AttributeError: '_io.BufferedReader' object has no attribute 'size'
```

`tail` now seeks to the end and uses `tell()` to find the length, which every seekable file supports. Added a test for both cache types; it fails on master.


fsspec/spec.py
```diff
@@ -1175,7 +1175,8 @@ def head(self, path, size=1024):
     def tail(self, path, size=1024):
         """Get the last ``size`` bytes from file"""
         with self.open(path, "rb") as f:
-            f.seek(max(-size, -f.size), 2)
+            f.seek(0, 2)
+            f.seek(max(f.tell() - size, 0))
```
martindurant, line 1179: Same as `f.seek(-size, 2)` ?

itzzdev09: lmost. Local files raise `OSError` when seeking before 0, so `tail(path, size)` fails if `size` exceeds the file length. Seeking to the end first gives us the length without `.size`. I can add a test for this case if you'd like.

martindurant: Good point, and yes please (add it into an existing test if possile)

itzzdev09:
Added in 369abcf, inside test_whole_file_cache_tail: sizes equal to, larger than, and zero relative to the file length. With f.seek(-size, 2) the test fails with OSError: [Errno 22] Invalid argument.

PR #2093, opened 2026-08-03
feat(AdaptiveReadaheadCache): introducing a new cache-type adaptive

raj-prince:
**Descriprtion**
This PR introduces the adaptive prefetching functionality as a first-class fsspec cache implementation, with prefetch logic and tests consolidated under fsspec.

To save the review time - providing the diff of already reviewed moved code across the repo.
```sh
#!/bin/sh

prefetcher_gcsfs="https://raw.githubusercontent.com/fsspec/gcsfs/refs/heads/main/gcsfs/prefetcher.py"
prefetcher_fsspec="https://raw.githubusercontent.com/raj-prince/filesystem_spec/refs/heads/adaptive-readahead/fsspec/prefetcher.py"
prefetcher_test_gcsfs="https://raw.githubusercontent.com/fsspec/gcsfs/refs/heads/main/gcsfs/tests/test_prefetcher.py"
prefetcher_test_fsspec="https://raw.githubusercontent.com/raj-prince/filesystem_spec/refs/heads/adaptive-readahead/fsspec/tests/test_prefetcher.py"

tmp1=$(mktemp "${TMPDIR:-/tmp}/prefetcher.gcsfs.XXXXXX.py")
tmp2=$(mktemp "${TMPDIR:-/tmp}/prefetcher.fsspec.XXXXXX.py")
tmp3=$(mktemp "${TMPDIR:-/tmp}/test_prefetcher.gcsfs.XXXXXX.py")
tmp4=$(mktemp "${TMPDIR:-/tmp}/test_prefetcher.fsspec.XXXXXX.py")

cleanup() {
	rm -f "$tmp1" "$tmp2" "$tmp3" "$tmp4"
}
trap cleanup EXIT INT TERM

curl -s "$prefetcher_gcsfs" > "$tmp1"
curl -s "$prefetcher_fsspec" > "$tmp2"
diff -u "$tmp1" "$tmp2"

curl -s "$prefetcher_test_gcsfs" > "$tmp3"
curl -s "$prefetcher_test_fsspec" > "$tmp4"
diff -u "$tmp3" "$tmp4"
```

Diff:
```diff
--- /tmp/prefetcher.gcsfs.5KiKLH.py     2026-09-13 00:00:46.052236250 +0000
+++ /tmp/prefetcher.fsspec.e88GLQ.py    2026-09-13 00:00:46.243236208 +0000
@@ -1,41 +1,15 @@
 import asyncio
-import ctypes
 import logging
 import weakref
 from collections import deque
 
[...]

PR #2163, opened 2026-09-17
Fix cat_ranges on filecache and on targets without on_error (follow-up to #2154)

itzzdev09:
Follow-up to #2154, fixing the two problems @lhoupert found while merging main into #2075.

**filecache `_cat_ranges` raised `AttributeError`.** #2154 put the new `_local_paths_for_ranges` helper on `SimpleCacheFileSystem`, but the async `_cat_ranges` that calls it is defined on `WholeFileCacheFileSystem`, so filecache could not find it:

```python
await filecache_fs._cat_ranges(["memory://f"], [0], [4])
# AttributeError: ... has no attribute '_local_paths_for_ranges'
```

The helper now lives on `WholeFileCacheFileSystem`. It also accepts what filecache's `_check_file` returns for a cached file (`(detail, fn)`, or `False` when missing), as well as simplecache's plain path.

**`on_error` was passed to `get()`/`_get()`.** Neither takes it, so it ended up in the target's `get_file`, and implementations without `**kwargs` (e.g. `http_sync`) raise `TypeError`. It is no longer passed there. It is still used for the local `cat_ranges` that reads the ranges.

Tests, all in `test_cached.py`:
- `test_async_cat_ranges_downloads_uncached` for filecache and simplecache, including a second call that reads the cached files.
- `test_simplecache_cat_ranges_target_get_file_without_on_error`, with a target whose `get_file` has no `**kwargs`.

The filecache and strict `get_file` cases fail on master, and `test_cached.py` passes.

This overlaps with #2101 and #2075, which also touch these methods. Happy to rebase if either lands first.


itzzdev09:
The CI failures are from master: test_fast_slice_pypy_fallback has been failing since #2093. Fixed in [fsspec/filesystem_spec#2164](https://github.com/fsspec/filesystem_spec/pull/2164)

martindurant:
(I'll merge that shortly ,and you can update here)

martindurant:
Out of interest: how many implementations lack on_error?

itzzdev09:
None of them, as far as I can tell: every get_file/_get_file in fsspec/implementations takes **kwargs. The problem is the ones that pass those kwargs on to their client. http_sync does session.get(url, **kw), so requests raises TypeError: unexpected keyword argument 'on_error', and async http does the same via aiohttp. Everywhere else it's silently swallowed, which isn't much better, since get() never had an on_error to begin with. I'd expect s3fs/gcsfs to behave like the HTTP ones since they forward kwargs to the client call too.

martindurant:
I should have said: are there implementations that don't *use* on_error; but we do also have "special" implementations like caches and the asyns wrapper to make life more complicated.

itzzdev09:
Good question, and my PR description was sloppy: http_sync does take **kwargs, it just passes them straight to requests.get, which is where the TypeError comes from. Same with async http and aiohttp. I'll fix the wording.

On the actual question: none of them use on_error. No get/_get declares it anywhere, base classes included, so it's always just a passthrough kwarg. It lives on cat/cat_ranges/copy/walk instead.

And yeah, the special ones don't help. Caches hand get to the target so it ends up in its get_file, and the async wrapper passes whatever it gets to the wrapped sync get. The caches do implement on_error on their own cat/cat_ranges, which is probably how it ended up here in the first place.

PR #2164, opened 2026-09-17
Fix test_fast_slice_pypy_fallback failing on master

itzzdev09:
Since #2093, every pytest job on master fails `test_fast_slice_pypy_fallback`:

```
AttributeError: <module 'fsspec.prefetcher' ...> does not have the attribute 'HAS_CPYTHON_API'
```

`_fast_slice` and `HAS_CPYTHON_API` live in `fsspec.utils` (`prefetcher` only imports `_fast_slice`), so the test's `mock.patch("fsspec.prefetcher.HAS_CPYTHON_API", False)` has nothing to patch. This patches `fsspec.utils.HAS_CPYTHON_API` instead, which is the flag `_fast_slice` actually reads.

`test_prefetcher.py` passes (60 tests). This failure also shows up on open PRs based on the current master, e.g. #2163.


martindurant:
Thanks for the quick fix.

PR #2165, opened 2026-09-18
Fix DirCache LRU eviction and iteration using OrderedDict

### LINKED ISSUE #2095: [Bug] DirCache LRU implementation fails to evict items on write and wipes cache during iteration

raj-prince:
### Summary

Fixes #2095

`DirCache` previously attempted to limit the number of cached paths via `functools.lru_cache`:
```python
if max_paths:
    self._q = lru_cache(max_paths + 1)(lambda key: self._cache.pop(key, None))
```

Because `functools.lru_cache` only invokes its wrapped function on a **cache miss** rather than upon eviction, this led to three major issues:
1. **Unbounded Memory Growth**: Setting items beyond `max_paths` never evicted old entries from the cache.
2. **Deferred Deletion on Read**: Reading a key that fell out of `lru_cache` was treated as a miss, popping the key and raising an unexpected `KeyError`.
3. **Iteration Destruction**: Iterating over the cache (`list(dircache)`) triggered cascading misses, wiping out all entries in the cache.

---

### Changes

- **`OrderedDict`-based LRU**: Replaced `functools.lru_cache` with `collections.OrderedDict` to manage LRU ordering directly.
  - Setting items (`dc[key] = value`) immediately evicts the least recently used entry once the cache exceeds `max_paths`.
  - Getting items (`dc[key]`) marks them as recently used, preserving true LRU ordering.
  - Unlimited caches (`max_paths=None`) use a standard dictionary for minimal overhead.
- **Safe Expiry Handling (`listings_expiry_time`)**:
  - Membership checks (`key in dircache`) and lookups check whether the key is in the cache before evaluating expiry timestamps, ensuring membership checks never raise a `KeyError` on non-existent or expired keys.
  - Expired items are purged on lookup and during iteration without affecting active entries.
- **Support for `max_paths=0`**: Correctly recognizes `max_paths=0` as a zero-capacity cache rather than disabling eviction.
- **Unit Tests**: Added focused unit tests in `fsspec/tests/test_dircache.py` validating LRU eviction, access recency, time expiry, non-existent key lookups, disabled caching (`use_listings_cache=False`), mapping methods, and serialization via `pickle`.


itzzdev09:
I compared this with #2097, which fixes the same issue (#2095). Both give the same results on the repro from the issue and on the cases I tried: eviction beyond `max_paths`, a read protecting an entry from eviction, iteration no longer emptying the cache, expiry, and unbounded mode.

Two places where this one changes behaviour beyond the fix:

- **`max_paths=0`**: on master (and #2097) this means no limit, because the old code checked `if max_paths:`. Here it caches nothing (`d = DirCache(max_paths=0); d["x"] = 1` then `"x" in d` is `False`). Since `max_paths` is a documented storage option passed through filesystem constructors, that could change behaviour for anyone passing 0.
- **The docstring** now says expiry "refers to when the entry was set or accessed", but a read doesn't reset the timer. With `listings_expiry_time=0.3`, an entry read at 0.2 s is still gone at 0.4 s. Either the wording or the behaviour needs to change.

The `ValueError` for a negative `max_paths` is new too. That seems reasonable, but it's another behaviour change.

martindurant:
@raj-prince , would appreciate your comparison of the two PRs.

raj-prince:
Thanks @martindurant and @itzzdev09 for looking into this!

To be clear, I am completely open to whichever PR you prefer to move forward with. I originally opened #2165 only because there hadn't been updates on #2097 for a while and I already had this branch ready.

At a high level, the main differences are:

1. Single vs. Redundant Storage -  PR#2097 uses two separate data structures—one for data (`_cache`) and one for LRU tracking (`_q`). Maintaining two data-structures adds unnecessary memory/CPU overhead and extra complexity to keep them synchronized. Using a single `OrderedDict` directly for `_cache` in PR#2165 avoids this redundancy and keeps the implementation clean.
2. Proper Expiry Cleanup: #2165 ensures that expired entries and their timestamps (`_times`) are properly cleaned up across evictions, deletions, and `clear()`.

Regarding the concern,
1. non-negative max-path scenario: I agree with the concern regarding behavior changes. I have updated it so that any non-positive value (`0`, negative, or `None`) is treated as an unlimited cache, fully preserving `master`'s historical behavior without errors.
2. listing_expiry_time docstring clarification: Updated the docstring to clearly separate `listings_expiry_time` (TTL/item validity measured from set time) from `max_paths` (LRU capacity ordering), removing the wording that caused the confusion.

martindurant:
@itzzdev09 , your concern are answered? If so, I suggest you close your PR and we go with this one.

itzzdev09:
@martindurant 
Yes, both resolved — I checked: max_paths=0 and negative values are now treated as unlimited, matching master's behavior, and the docstring now correctly says expiry is measured from when the entry was set. One correction: I don't have a PR open on this — #2097 is @aryansk, I only left the comparison comment. #2165 looks good to merge from my side.

PR #2168, opened 2026-09-20
Find file length without .size in cat_file and read_block

itzzdev09:
Follow-up to the `tail` fix (#2155). `AbstractFileSystem.cat_file` (for a negative `start`/`end`) and `read_block` both read `f.size` to find the file length:

```python
# cat_file
f.seek(max(0, f.size + start))
...
end = f.size + end

# read_block
size = f.size
```

Only fsspec's own file classes expose `.size`. `simplecache` and `filecache` return a plain local file from `open`, so both methods raise `AttributeError: '_io.BufferedReader' object has no attribute 'size'` on them:

```python
import fsspec
fsspec.filesystem("memory").pipe("/f", b"0123456789")
fs = fsspec.filesystem("simplecache", target_protocol="memory")
fs.cat_file("/f", start=-3)     # AttributeError
fs.read_block("/f", 0, None)    # AttributeError
```

`cat_file` now measures the length with a seek to the end (only when a bound is negative, so the common path is unchanged), and `read_block` falls back to a seek when `.size` is absent. This mirrors what `tail` already does.

Added `test_whole_file_cache_cat_file_negative` and `test_whole_file_cache_read_block` for both cache types; both fail on `master`. The api/spec/utils/memory/local suites pass (475).


fsspec/spec.py
```diff
@@ -856,14 +856,29 @@ def cat_file(self, path, start=None, end=None, **kwargs):
...
         with self.open(path, "rb", **kwargs) as f:
+            # A negative start or end counts back from the file length. Only
+            # fsspec's own file classes expose `.size`; a plain file (e.g. from
+            # the caching filesystems) does not, so measure with a seek instead.
+            measured = (start is not None and start < 0) or (
+                end is not None and end < 0
+            )
+            if measured:
+                f.seek(0, 2)
+                size = f.tell()
```
martindurant:
```suggestion
                size = f.seek(0, 2)
```

fsspec/spec.py
```diff
@@ -856,14 +856,29 @@ def cat_file(self, path, start=None, end=None, **kwargs):
         """
         # explicitly set buffering off?
         with self.open(path, "rb", **kwargs) as f:
+            # A negative start or end counts back from the file length. Only
+            # fsspec's own file classes expose `.size`; a plain file (e.g. from
+            # the caching filesystems) does not, so measure with a seek instead.
+            measured = (start is not None and start < 0) or (
+                end is not None and end < 0
+            )
```
martindurant: Call it `needs_size` ?

fsspec/spec.py
```diff
@@ -856,14 +856,29 @@ def cat_file(self, path, start=None, end=None, **kwargs):
...
+            if measured:
+                f.seek(0, 2)
```
martindurant: Still use `.size` if it exists

fsspec/spec.py
```diff
@@ -856,14 +856,29 @@ def cat_file(self, path, start=None, end=None, **kwargs):
...
+                size = f.tell()
             if start is not None:
                 if start >= 0:
                     f.seek(start)
                 else:
-                    f.seek(max(0, f.size + start))
+                    f.seek(max(0, size + start))
+            elif measured:
```
martindurant: Only if we needed the size, but it was not present as an attribute.

itzzdev09:
Done — using .size first now, only measuring with seek(0, 2) when it's absent, and only resetting the cursor in that case. Renamed to needs_size. All tests still pass

PR #2169, opened 2026-09-20
Treat a leading dot as a hidden file, not an extension

Rodrigo-Palma:
`get_file_extension` splits the file name on its last dot, so a dotfile whose name carries no extension reports its whole name as one:

```python
>>> from fsspec.utils import get_file_extension
>>> get_file_extension("/path/to/.bashrc")
'bashrc'
>>> get_file_extension(".gitignore")
'gitignore'
>>> get_file_extension("s3://bucket/.env")
'env'
```

A leading dot marks the file as hidden, it does not introduce an extension. `pathlib` draws the same line:

```python
>>> PurePosixPath(".bashrc").suffix
''
```

This ignores the leading dots when looking for the extension separator. A hidden file that does carry an extension is unaffected, so `".hidden.txt"` still reports `"txt"`.

Everything the existing parametrization already covers is unchanged, including the parent-directory-with-a-dot case added in #2071 and a trailing dot. I extended that parametrization rather than adding a separate test, and also checked the result against `pathlib.PurePosixPath(...).suffix` across the cases above plus `"file.tar.gz"`, `"..file.txt"` and `"a..b"`: they agree on all of them.

`fsspec/tests/test_utils.py` is green (113 passed) and `ruff check` is clean.


Rodrigo-Palma:
The runs here are still waiting for maintainer approval, so I ran this repository's CI on my fork at the same commit as this PR (`9fc82dc`):

https://github.com/Rodrigo-Palma/filesystem_spec/actions/runs/35598923905

11 jobs, all green: `lint`, `pytest` on 3.10 through 3.14 (including 3.14t), `pytest-win`, and the downstream jobs `s3fs-pytest`, `gcsfs-pytest` and `downstream`.

itzzdev09:
Looks good to me. I ran a few more names through it: `..`, `...` and `.x.` give `""`, `file..txt` gives `txt`, and `/d/.data.tar.gz` still gives `gz`.

I also checked who else is affected. Nothing outside the tests calls `get_file_extension`, and `infer_compression` does its own suffix check. A bare `.gz` was already `None` there on master, so compression inference doesn't change.

PR #2170, opened 2026-09-20
Keep the whole IPv6 address when parsing the host

Rodrigo-Palma:
`infer_storage_options` takes the host from the netloc by splitting on the last colon. That colon is the port separator for a named host, but it lands inside the address when the host is an IPv6 literal without a port:

```python
>>> from fsspec.utils import infer_storage_options
>>> infer_storage_options("hdfs://[::1]/path/f.csv")["host"]
'[:'
>>> infer_storage_options("hdfs://[2001:db8::1]/f.csv")["host"]
'[2001:db8:'
```

With a port the same expression happens to be right, because the last colon is then the port separator:

```python
>>> infer_storage_options("hdfs://[::1]:8020/f.csv")["host"]
'[::1]'
```

which is probably why it went unnoticed.

`parsed_path.hostname` is not used here on purpose, as the comment right above the line explains: it lowercases the host, which matters for S3 buckets (dask#1417). So this keeps parsing the netloc by hand and only stops at the closing bracket when the host is bracketed.

Named hosts take the same path as before, and case is still preserved: `hdfs://Node-Name.COM/f.csv` keeps `Node-Name.COM`, and `s3://Bucket-Name.com/test.csv` still folds the bucket into the path unchanged.

I checked the result against `urllib.parse.urlsplit(...).hostname` for bracketed and named hosts, with and without a port, and with a user and password present: they agree on all of them (modulo the brackets, which this function keeps, as it already did for the working port case).

There was no IPv6 coverage in `test_utils.py`, so I added `test_infer_ipv6_host`. The file is green (109 passed) and `ruff check` is clean.


Rodrigo-Palma:
Same as on #2169: the runs here are awaiting approval, so I ran this repository's CI on my fork at the same commit as this PR (`b32be72`):

https://github.com/Rodrigo-Palma/filesystem_spec/actions/runs/35598926539

11 jobs, all green: `lint`, `pytest` on 3.10 through 3.14 (including 3.14t), `pytest-win`, and the downstream jobs `s3fs-pytest`, `gcsfs-pytest` and `downstream`.

itzzdev09:
Checked this out and tried a few more IPv6 shapes on top of the ones in the test. All give the right host and port:

```
hdfs://[fe80::1%25eth0]:9000/p      host='[fe80::1%25eth0]' port=9000
webhdfs://[::ffff:1.2.3.4]:50070/p  host='[::ffff:1.2.3.4]' port=50070
sftp://u@[::1]/p                    host='[::1]' user='u'
hdfs://[::1]:/p                     host='[::1]' (empty port ignored)
hdfs://Host.Name/p                  host='Host.Name' (case kept, as before)
```

An unclosed bracket (`hdfs://[::1`) already raises `ValueError: Invalid IPv6 URL` from `urlsplit` before this code runs, so the `"]" in host` check is only a safety net, which is fine. `test_utils.py` passes for me. LGTM.

PR #2171, opened 2026-09-20
Reject opening implied memory directories as files

agammann:
Paused pending agreement on MemoryFileSystem semantics and performance. Thanks for identifying both concerns.

This proposal would reject opening an implied directory as a file. It assumes local-filesystem behavior, but the maintainer has noted that object-store behavior may be preferable. The intended behavior is unresolved.

The proposed `isdir()` call scans the store on every open. A reviewer reported about 8.1 seconds versus 0.01 seconds for 2,000 new writes with 20,000 existing files. Those are the reviewer’s measurements; I have not independently reproduced them. The current implementation is not ready to merge.

Original local validation on Python 3.12 / Windows: 12 regression cases failed before the change; the memory, mapping and filesystem specification suites then reported 231 passed, 123 skipped and 1 xfailed, with one socket teardown warning. Ruff checks passed for the changed files. These correctness checks do not establish acceptable performance or settle the design question.

See the [design feedback](https://github.com/fsspec/filesystem_spec/pull/2171#issuecomment-5766461261) and [performance review](https://github.com/fsspec/filesystem_spec/pull/2171#issuecomment-5766348855).


itzzdev09:
Tested this locally and the fix is right: on master, `m.open("/d0", "wb")` succeeds when `/d0/f` exists, and on this branch it raises `IsADirectoryError` for `/d0`, `d0` and `/d0/`.

One thing worth weighing though: `isdir()` goes through `info()` and `ls()`, which walk the whole store, and `_open` now does that on every open, including every new write. With 20,000 files already in the store, writing 2,000 new files took about 8.1 s on this branch vs 0.01 s on master (Windows, Python 3.12):

```python
m = MemoryFileSystem(skip_instance_cache=True); m.store = {}; m.pseudo_dirs = [""]
for i in range(20000):
    m.pipe_file(f"/d{i % 100}/f{i}", b"x")
for i in range(2000):  # timed
    with m.open(f"/new/f{i}", "wb") as f:
        f.write(b"x")
```

I tried an early-exit prefix scan over `store` instead of `isdir()`, and it was still ~6.5 s. So any check for implied directories here looks linear in the store size unless parent directories are indexed somewhere. That may well be acceptable for how MemoryFileSystem is usually used, but I thought it was worth flagging.

martindurant:
I'm not sure about the premise here. It depends on whether we think memFS should mimic a local FS really, or behave more like an object store - I would probably prefer the latter. In S3, for instance, it is permissible to write to a key that happens to a prefix (directory) to other keys. Having said that, if you write to "key", then writing to "key/file" *is* currently disallowed (FileExistsError).

itzzdev09 (commented):
The behaviour change looks right: an implied directory should not open as a file.

The cost is the concern. `self.isdir(path)` routes through `info`, which builds `list(self.store) + self.pseudo_dirs` and scans every key, on every `_open`. The old check was a membership test on a short list. With 20k files in the store, 200 opens go from 0.001s to 0.77s here, and 200 `pipe_file` calls to a new path cost 0.45s, so reads and writes both become O(files) per call.

Doing the cheap test first keeps the fix and most of the speed for the common case of an existing file:

```python
if path not in self.store and self.isdir(path):
    raise IsADirectoryError(path)
```

That still scans when writing a new file, so if the write path matters too, the alternative is testing for a child directly rather than going through `info`.

Issue #1568, opened 2024-04-10
Sanitize some common TAR path occurrences such as leading dots ./


mxmlnkn:
Hi there,

we already wrote a bit back and forth in the issue in the smart_open repository. I wanted to give fsspec.fuse a quick try using the tar/libarchive backend. Unluckily, my very first test failed. I created the test tar with:

```bash
echo foo > large
tar -cf large{.tar,}
tar tvlf large.tar  # -rwx------ user/user   4 2024-04-10 23:04 ./large
```

Then, I tried to mount it with:

```python
from fsspec.implementations.tar import TarFileSystem as tafs
fs = tafs("large.tar")
import fsspec.fuse
fsspec.fuse.run(fs, "", "mounted")'
```

and access it with:

```bash
ls -la mounted/
# total 4
# drwxrwxrwx 0 user user    0 Apr 10 23:07 .
# drwxrwxrwx 0 user user    0 Apr 10 23:07 .
# drwx------ 1 user user 4096 Apr 10 23:04 ..
ls -la mounted/./
# total 4
# drwxrwxrwx 0 user user    0 Apr 10 23:07 .
# drwxrwxrwx 0 user user    0 Apr 10 23:07 .
# drwx------ 1 user user 4096 Apr 10 23:04 ..
```

As you can see, the leading dot is interpreted as a valid folder even though it isn't. And even though it is shown because of the FUSE-specifics, which already normalizes paths before the implementation is called, it is not possible to access the `large` file.

I think, the TAR and libarchive backends should normalize paths to some degree. At least leading dots. Maybe also `.` and `..` inside the path. Funnily enough, I had the exact same issue with fuse-archive: https://github.com/google/fuse-archive/issues/2 . There are also more complex cases, e.g., try this:

```
tar -cf large.tar ./././large
tar tvlf large.tar 
#-rwx------ user/user   4 2024-04-10 23:04 ./././large
```

I was not able to create a path with `..` in it. Gnu tar strips it and even the leading dots when a .. occurs. But it might be possible to create such TARs with Python's tarfile and/or with other tools.

Specifying `path = "./"` to `fsspec.fuse.run` kinda works around this issue and `large` will be visible in the mount point, but:

 - I would have to know whether there are leading dots first, which complicates usage.
 - It doesn't fix working with TARs with files both with leading dot and without, e.g.:

```bash
echo foo > bar
echo foo > large
tar -cf large.tar bar ./large
tar tvlf large.tar 
[...]

PR #2174, opened 2026-09-22
Normalize dot components in TarFileSystem paths

pentaoa:
TAR members such as `./file.txt` and `././dir/file.txt` appear beneath a literal `.` directory and cannot be read using the paths a filesystem caller expects. Normalize empty and single-dot components consistently in the TAR index and in listing/lookup paths, while retaining the original member names for `tarfile` reads and link resolution. Explicit `./` root entries are not listed as their own children; `..` components remain literal.

Related to #1568. This addresses the TAR path handling independently of FUSE; it does not change the libarchive backend. Regression coverage includes mixed prefixed/plain names, hidden files, explicit/implicit directories, duplicate members, symlinks and hardlinks.

Validation on Python 3.12: all nine new regressions fail against the original implementation and pass with the patch; TAR/shared archive/ZIP tests report 190 passed and 21 skipped (optional dependencies), and core/spec/mapping tests report 214 passed, 129 skipped and one expected failure. All pre-commit checks pass. The local Windows HTTP fixture emitted one shutdown-thread warning.

[CI](https://github.com/fsspec/filesystem_spec/actions/runs/35696135709) now passes all 11 checks: Python 3.10–3.14, free-threaded 3.14, Windows, lint, downstream, s3fs and gcsfs. The free-threaded run reports 2,224 passed, 39 skipped and two expected failures; all nine new TAR regressions execute and pass in CI.

AI assistance: OpenAI Codex implemented the change, wrote the regression tests, and ran the reported local checks.


martindurant:
This is an improvement, I will merge.

Question: is it possible to have the *same* path duplicated because of normlising out the "/." pieces?

pentaoa:
Yes. `./file.txt` and `file.txt` can normalize to the same path. The index keeps one entry, with the last archive member winning, so `ls()` has no duplicate and reads use that member's original name. `test_dot_normalization_duplicate_members` covers both orderings and checks the listing, size, and contents.

martindurant:
I suppose that's reasonable behaviour. If archives do indeed have this kind of duplication, we could call that a manner of corruption and it's not clear there's anything else we can do. I assume (gnu) `tar` would also overwrite on extract and so indeed the latter wins.

